PR #5, opened 2021-08-23
Use mock model for HorizontalNav to limit tabs from being exposed in Operator DetailsPage

bipuladh:


src/models/odf.ts
```diff
@@ -12,3 +12,15 @@ export const ODFStorageSystem: K8sKind = {
   id: 'storagesystem',
   crd: true,
 };
+
```
GowthamShanmugam, line 15: Please add a few lines about why this mock kind and where it is used

bipuladh: added

PR #6, opened 2021-08-24
fix: Stop storage system name from overflowing to a different line. 

bipuladh:
Screenshot from 2021-08-24 13-50-35


bipuladh:
Screenshot from 2021-08-24 13-48-45

cloudbehl (approved):

PR #30, opened 2021-09-09
Update Cypress webpack and TSC configs

rexagod:
2021-09-09-133731_2560x1440_scrot

/cc @bipuladh

### DIFF
--- a/cypress/plugin.js
+++ b/cypress/plugin.js
@@ -14,10 +14,15 @@ module.exports = (on, config) => {
         rules: [
           {
             test: /\.ts$/,
-            loader: 'ts-loader',
+            use: 'ts-loader',
+            exclude: /node_modules/,
           },
         ],
       },
+      output: {
+        filename: 'bundle.js',
+        path: path.resolve(__dirname, 'cypress-dist'),
+  },
     },
   };
   // `on` is used to hook into various events Cypress emits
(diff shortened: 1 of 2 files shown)


cypress/plugin.js
```diff
@@ -8,16 +8,21 @@ module.exports = (on, config) => {
   const options = {
     webpackOptions: {
       resolve: {
-        extensions: ['.ts', '.js'],
+        extensions: ['.tsx', '.ts', '.js'],
```
bipuladh, line 11: Why tsx? All our files are going to `spec.ts`.

cypress/plugin.js
```diff
@@ -8,16 +8,21 @@ module.exports = (on, config) => {
...
+            test: /\.tsx?$/,
+            use: 'ts-loader',
+            exclude: /node_modules/,
           },
         ],
       },
+      output: {
+        filename: 'bundle.js',
+        path: path.resolve(__dirname, 'dist'),
+  },
```
bipuladh, line 25: Why are we handling the output file? Do we want create one file which contains all tests?

bipuladh (changes requested):

cypress/plugin.js
rexagod: Fixed this and the regex below, PTAL.

cypress/plugin.js
rexagod: Cypress folks [recommend bundling](https://github.com/cypress-io/cypress/blob/master/npm/webpack-preprocessor/examples/use-ts-loader/webpack.config.js#L18) to reduce request times for Cypress GET, POST, etc. requests. Nonetheless, this will still show up as different tests in the console same as we see in console.
2021-09-09-143805_2560x1440_scrot

bipuladh: ah okay!

bipuladh (approved):
/lgtm

PR #31, opened 2021-09-09
Bug 1999050: Fix links to Vendor Dashboard from all cards

bipuladh:
Fix status of ODF Operator in Status Card
Fix CSS to align cards
Refactor constants and types


src/components/odf-dashboard/status-card/status-card.tsx
```diff
@@ -88,7 +88,7 @@ export const StatusCard: React.FC = () => {
   );
 
   return (
-    <DashboardCard>
+    <DashboardCard className="odfDashboard-card--height">
```
SanjalKatiyar, line 91: Is this a Global class ??
I can't see any css file imported above.

SanjalKatiyar: same in `src/components/odf-dashboard/object-storage-card/capacity-card.tsx` & `src/components/odf-dashboard/system-capacity-card/capacity-card.tsx` files.

bipuladh: It's part of `capacity-card.scss`. I am doing this to not have to have same class declaration across 2 files.

src/components/odf-dashboard/performance-card/performance-card.tsx
```diff
@@ -150,9 +156,9 @@ const PerformanceCard: React.FC = () => {
...
         />
       )}
-      {loading && <PerformanceCardLoading />}
+      {loading && !error && <PerformanceCardLoading />}
       {(error && !loading) ||
-        (_.isEmpty(rawRows) && (
+        (!error && !loading && _.isEmpty(rawRows) && (
           <div className="performanceCard--error">
             <DataUnavailableError />{' '}
           </div>
```
SanjalKatiyar, line 164: Right now, if condition on L160 is true (`(error && !loading)`)--> It will not display <DataUnavailableError /> FC because it is `&&`ed with condition on L161 only.
Shouldn't it be ??
```
      { ( (error && !loading) ||
        (!error && !loading && _.isEmpty(rawRows)) ) && (
          <div className="performanceCard--error">
            <DataUnavailableError />{' '}
          </div>
      )}
```

PR #34, opened 2021-09-16
Updates dockerfile for the CI

vbnrh:

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -1,6 +1,9 @@
-FROM quay.io/coreos/tectonic-console-builder:v22 AS build
+FROM quay.io/coreos/tectonic-console-builder:v22
 
-RUN mkdir -p /app
-ADD . /app
-WORKDIR /app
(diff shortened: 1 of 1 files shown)


Dockerfile.ci
```diff
@@ -1,6 +1,9 @@
...
-RUN mkdir -p /app
-ADD . /app
-WORKDIR /app
+WORKDIR /usr/src/app
+COPY . /usr/src/app
 RUN yarn install
+RUN yarn build
+
+EXPOSE 9001
+ENTRYPOINT [ "./http-server.sh", "./dist" ]
```
bipuladh, line 9: Why do we need to do this?

vbnrh: @bipuladh  added it with assumption that the container port needs to be exposed and server to be running for the dynamic plugin to work ? Correct me if i am wrong

vbnrh: Or is it specifically introduced for local builds ?

PR #48, opened 2021-10-08
Fix for expand PVC test

vbnrh:


cypress/support/index.ts
```diff
@@ -0,0 +1,6 @@
+
+Cypress.Cookies.debug(true);
+
+Cypress.Cookies.defaults({
+  preserve: ['openshift-session-token', 'csrf-token'],
+});
```
bipuladh, line 6: Add new line

cypress/support/index.ts
```diff
@@ -0,0 +1,6 @@
+
+Cypress.Cookies.debug(true);
```
bipuladh, line 2: Why the debug part?

bipuladh (commented):
/lgtm cancel

cypress/support/index.ts
vbnrh: By turning on debugging, Cypress will automatically generate logs to the console when it sets or clears cookie values. This is useful to help you understand how Cypress clears cookies before each test, and is useful to visualize how to handle preserving cookies in between tests.

PR #54, opened 2021-10-18
Bug 2013086: Fix IBM dashboard route issuse

bipuladh:
`/odf` route(non-exact) is getting registered before IBM routes
Configuring routes in a manner where order doesn't matter

### DIFF
--- a/src/components/odf-dashboard/dashboard.tsx
+++ b/src/components/odf-dashboard/dashboard.tsx
@@ -3,6 +3,7 @@ import { HorizontalNav } from '@openshift-console/dynamic-plugin-sdk';
 import { Helmet } from 'react-helmet';
 import { useTranslation } from 'react-i18next';
 import { RouteComponentProps } from 'react-router';
+import { useLocation } from 'react-router-dom';
 import { Grid, GridItem } from '@patternfly/react-core';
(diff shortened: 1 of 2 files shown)


src/components/odf-dashboard/dashboard.tsx
```diff
@@ -57,6 +58,15 @@ const ODFDashboardPage: React.FC = () => {
       component: ODFDashboard,
     },
   ];
+  const { history } = props;
+  const location = useLocation();
+
+  React.useEffect(() => {
+    if (location.pathname.endsWith('/odf/systems')) {
```
rexagod, line 65: Just a suggestion. Can we do this for `/odf/system` in general (covering both cases)? Currently the OCS SS directs to `http://localhost:9000/odf/**system**/ocs.openshift.io~v1~storagecluster/ocs-storagecluster/overview/block-file` so if  we remove the `ocs.openshift.io~v1~storagecluster/ocs-storagecluster/overview/block-file`, then this change will ensure nothing breaks.

bipuladh: okay

bipuladh: I tried doing this but that would end up breaking the whole thing.

PR #55, opened 2021-10-18
Add dockerignore file to the project

bipuladh:
/assign @rexagod

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+node_modules
+dist
+.cache-loader
+.husky/_/


.dockerignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+dist
+.cache-loader
```
rexagod, line 3:
```suggestion
.cache-loader
.husky/_/
```
We can safely remove `.husky/_/` as well.
```
❯ g clean -fxd                                                                                                                                                              odf-console/git/master 
Removing .cache-loader/
Removing .husky/_/
Removing dist/
Removing node_modules/
```

bipuladh: ack

PR #60, opened 2021-10-20
Add dockerfile and nginx configuration file

bipuladh:
/assign @rexagod

### DIFF
--- a/Dockerfile.prod
+++ b/Dockerfile.prod
@@ -1,0 +1,30 @@
+FROM registry-proxy.engineering.redhat.com/rh-osbs/rhacm2-yarn-builder:v2.3.0_14-1.20210810163126 AS builder
+ARG TARGET_BRANCH=master
+
+RUN dnf install -y git
+RUN git clone https://github.com/red-hat-storage/odf-console.git
+WORKDIR /odf-console
+RUN git fetch origin ${TARGET_BRANCH} && git checkout origin/${TARGET_BRANCH}
+
+RUN yarn install --prod=false
+RUN yarn build
+RUN mv ./dist ../app
+
+RUN git fetch origin ${TARGET_BRANCH}-compatibility && git checkout origin/${TARGET_BRANCH}-compatibility
+RUN yarn install
+RUN yarn build
+RUN mv ./dist ../compatibility
+
+FROM registry.access.redhat.com/ubi8/nginx-118
+ADD default.conf "${NGINX_CONFIGURATION_PATH}"
+COPY --from=builder /app .
+COPY --from=builder /compatibility ./compatibility
+CMD /usr/libexec/s2i/run
+
+LABEL maintainer="Bipul Adhikari <badhikar@redhat.com>"
+LABEL name="odf-console"
+LABEL version="4.9"
+LABEL description="OpenShift Data Foundation Console container"
+LABEL summary="Provides the latest console for OpenShift Data Foundation."
+LABEL io.k8s.display-name="ODF Console"
+LABEL io.openshift.tags="odf"
--- a/default.conf
+++ b/default.conf
@@ -1,0 +1,16 @@
+server {
+    listen       9001 ssl;
+    ssl_certificate /var/serving-cert/tls.crt;
+    ssl_certificate_key /var/serving-cert/tls.key;
+    location / {
+        root   /opt/app-root/src;
+    }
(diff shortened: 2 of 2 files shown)


Dockerfile.prod
```diff
@@ -0,0 +1,29 @@
...
+RUN dnf install -y git
+RUN git clone https://github.com/red-hat-storage/odf-console.git
+WORKDIR /odf-console
+RUN git fetch origin release-4.9 && git checkout origin/release-4.9
+
+RUN yarn install --prod=false
+RUN yarn build
+RUN mv ./dist ../app
+
+RUN git fetch origin release-4.9-compatibility && git checkout origin/release-4.9-compatibility
```
rexagod, line 12:
```suggestion
RUN git fetch origin "${TARGET_BRANCH}-compatibility" && git checkout "origin/${TARGET_BRANCH}-compatibility"
```

Dockerfile.prod
```diff
@@ -0,0 +1,29 @@
...
+
+FROM registry.access.redhat.com/ubi8/nginx-118
+ADD default.conf "${NGINX_CONFIGURATION_PATH}"
+COPY --from=builder /app .
+COPY --from=builder /compatibility ./compatibility
+CMD /usr/libexec/s2i/run
+
+LABEL maintainer="Bipul Adhikari <badhikar@redhat.com>"
+LABEL name="odf-console"
+LABEL version="4.19"
```
rexagod, line 25: Should this be 4.9?

Dockerfile.prod
```diff
@@ -0,0 +1,29 @@
...
+RUN git fetch origin release-4.9-compatibility && git checkout origin/release-4.9-compatibility
+RUN yarn install --prod=false
+RUN yarn build
+RUN mv ./dist ../compatibility
+
+FROM registry.access.redhat.com/ubi8/nginx-118
+ADD default.conf "${NGINX_CONFIGURATION_PATH}"
+COPY --from=builder /app .
+COPY --from=builder /compatibility ./compatibility
+CMD /usr/libexec/s2i/run
```
rexagod, line 21: I think perhaps we need to run this command as root, or a user with the same perms as the host of the environment that started the docker daemon. I think OCS Operator does something of the sort and we can explore that if we want to go that way.

2021-10-20-230946_1920x1080_scrot

default.conf
```diff
@@ -0,0 +1,17 @@
+server {
+    listen       9001;
+    ssl 	     on;
+    ssl_certificate /var/serving-cert/tls.crt;
+    ssl_certificate_key /var/serving-cert/tls.key;
+    location / {
+        root   /opt/app-root/src;
```
rexagod, line 7: Not able to verify the `/opt` paths at the moment due to the cert and key permission failures above, will come back to this once we patch that.

Dockerfile.prod
bipuladh: you need to found those certificates into those directories docker can't pull from your local system

default.conf
bipuladh: You are testing it incorrectly.

Dockerfile.prod
```diff
@@ -0,0 +1,29 @@
+FROM registry-proxy.engineering.redhat.com/rh-osbs/rhacm2-yarn-builder:v2.3.0_14-1.20210810163126 AS builder
+
```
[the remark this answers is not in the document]
bipuladh: since we don't need this to persist inside the container we can go for `arg` instead.

Dockerfile.prod
```diff
@@ -0,0 +1,30 @@
+FROM registry-proxy.engineering.redhat.com/rh-osbs/rhacm2-yarn-builder:v2.3.0_14-1.20210810163126 AS builder
+ARG TARGET_BRANCH=master
```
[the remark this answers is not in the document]
rexagod: We're assuming that `TARGET_BRANCH` has a `-compatibility` counterpart, so we might need to change this (https://github.com/red-hat-storage/odf-console/pull/60/files#diff-a811e274b83c4976b7b48589ed3273262e4adb67534b49208d484fe7f0bc06a0R13) up a bit if this variable can point to branches that do not have that counterpart.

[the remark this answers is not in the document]
rexagod: Oh okay, sounds good.

PR #62, opened 2021-10-21
Fix routes for breadcrumbs in Noobaa resource pages

bipuladh:
Before:
route-bug

After:
route-bug-fixed

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -36,9 +36,20 @@
   {
     "type": "console.page/route",
     "properties": {
-      "path": "/odf",
+      "path": ["/odf", "/odf/overview"],
       "exact": true,
       "component": { "$codeRef": "dashboard.Reroute" }
+    },
+    "flags": {
+      "required": ["ODF_MODEL"]
+    }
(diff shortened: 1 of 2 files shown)


console-extensions.json
```diff
@@ -43,5 +43,16 @@
...
+  {
+    "type": "console.page/route",
+    "properties": {
+      "path": "/odf/resource/:kind",
+      "exact": true,
+      "component": { "$codeRef": "dashboard.RerouteResource" }
+    },
+    "flags": {
+      "required": ["ODF_MODEL"]
+    }
```
SanjalKatiyar, line 56: Still need Route for this as well, right ??
https://github.com/openshift/console/blob/54ff88a061c57e426f49a82287476644c238923d/frontend/packages/ceph-storage-plugin/src/components/odf-resources/resource-details-page.tsx#L83

SanjalKatiyar: Or, why are we even redirecting to `odf/overview` above. Isn't this should be just `/odf` ?

PR #66, opened 2021-11-03
Bug 2019654: url link for system systems

shdn-ibm:
fix https://bugzilla.redhat.com/show_bug.cgi?id=2019654


SanjalKatiyar:
small FYI, as per https://bugzilla.redhat.com/show_bug.cgi?id=2005282, now, system-list page also points to the URL with systemName (just like dashboard) instead of pointing to storageCluster name.

SanjalKatiyar:
/retitle Bug 2019654: url link for system systems

afreen23:
Yes, with fix went in https://github.com/openshift/console/pull/10262
We dont want to do this anymore.
All routes will incorporate StorageSystem name

shdn-ibm:
This fix is for the overview page. I think it is different from the list page.

PR #67, opened 2021-11-11
Use KBps for Throughput chart

rexagod:
Use KBps for Throughput chart in the Performance card inside the ODF
Dashboard.
***
#### Earlier
image
***
#### Now
image
***
This change was made in order to use the same unit ("KBps") as defined in the SS List page (see below) in the ODF Dashboard for Throughput.
***
image


bipuladh (approved):
/lgtm

PR #74, opened 2021-11-29
Improve support for Systems that cannot provide total capacity

bipuladh:
Screenshot from 2021-11-29 12-06-11
Note: It's reporting 1B since the metrics exporter of Noobaa is broken.

### DIFF
--- a/packages/odf/components/odf-dashboard/object-storage-card/capacity-card.tsx
+++ b/packages/odf/components/odf-dashboard/object-storage-card/capacity-card.tsx
@@ -2,9 +2,7 @@ import * as React from 'react';
 import CapacityCard from '@odf/shared/dashboards/capacity-card/capacity-card';
 import { humanizeBinaryBytes } from '@odf/shared/utils/humanize';
 import { PrometheusResponse } from '@openshift-console/dynamic-plugin-sdk';
-import {
-  usePrometheusPoll,
-} from '@openshift-console/dynamic-plugin-sdk-internal';
+import { usePrometheusPoll } from '@openshift-console/dynamic-plugin-sdk-internal';
 import { useTranslation } from 'react-i18next';
 import { Card, CardBody, CardHeader, CardTitle } from '@patternfly/react-core';
 import { ODFStorageSystem } from '../../../models';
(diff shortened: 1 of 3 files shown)


src/components/odf-dashboard/system-capacity-card/capacity-card.tsx
```diff
@@ -65,7 +65,9 @@ const SystemCapacityCard: React.FC = () => {
             managedSystemName: system.spec.name,
             managedSystemKind: referenceFor(apiGroup)(apiVersion)(kind),
             usedValue: humanizeBinaryBytes(usedMetric?.value?.[1]),
-            totalValue: humanizeBinaryBytes(totalMetric?.value?.[1]),
+            totalValue: _.isNumber(totalMetric?.value?.[1])
+              ? humanizeBinaryBytes(totalMetric?.value?.[1])
+              : undefined,
```
SanjalKatiyar, line 70: Can it return totalCapacity as "0" ??
Will it work in that case ?

src/components/common/capacity-card/capacity-card.tsx
```diff
@@ -244,16 +256,22 @@ const CapacityCard: React.FC<CapacityCardProps> = ({
...
-              isPercentage={isPercentage}
-              isRelative={relative}
-              largestValue={sortedMetrics[0].usedValue}
-            />
-          ))}
+          <CapacityCardHeader showPercentage={securePercentage} />
+          {sortedMetrics.map((item, index) => {
+            if (relative === undefined) {
+              secureRelative = sortedMetrics[index]?.totalValue === undefined;
+              securePercentage = !secureRelative;
```
SanjalKatiyar, line 263: Just need to confirm once, we are changing the value of `securePercentage` on L263, but it is not a state variable. And, sending the same value to L259 which we got as a props (irrespective of any change on L263). 
This is intentional ??

SanjalKatiyar: I assume it is intentional, plz ignore above comment.

src/components/odf-dashboard/system-capacity-card/capacity-card.tsx
SanjalKatiyar: IIRC what was happening earlier was that reported usedCapacity was 3B and reported totalCapacity was 0B... for such case it will still show infinity%

SanjalKatiyar: Need to add one a safety check in `getPercentage` function for cases when denominator if zero.
`isNaN(0/0) === true` but `isNaN(2/0) === false`

bipuladh:
Screenshot from 2022-02-02 22-54-16
Working as expected with Noobaa.

PR #75, opened 2021-11-30
Migration: Moves Storage System List Page and Add Capacity Modal

bipuladh:
Screenshot(New list page is StorageSystem List)

Screenshot from 2021-12-01 11-39-53


bipuladh:
/hold 
Get more eyes on it

vbnrh:
Screenshot from 2021-12-10 16-21-33

Working ! 

Will take some time for code reviews

package.json
```diff
@@ -41,14 +40,16 @@
...
     "react-helmet": "^6.1.0",
     "react-i18next": "^11.11.4",
+    "react-redux": "^7.2.6",
     "react-router": "5.2.0",
     "react-router-dom": "5.2.0"
   },
   "devDependencies": {
     "@cypress/webpack-preprocessor": "^5.9.1",
-    "@openshift-console/dynamic-plugin-sdk": "0.0.2",
+    "@openshift-console/dynamic-plugin-sdk": "0.0.3",
```
afreen23, line 49: is this required in this PR ?
Its a good practice to bump the  versions in separate PR /commit.

packages/odf/modals/add-capacity/add-capacity-modal.scss
```diff
@@ -0,0 +1,51 @@
+@import '~@patternfly/patternfly/sass-utilities/colors';
+
+.ceph-add-capacity__current-capacity {
```
afreen23, line 3: We should change these css prefixes now.

afreen23 (changes requested):

package.json
bipuladh: We need to add kubevirt package and kubevirt has dependency on this version.

packages/odf/modals/add-capacity/add-capacity-modal.scss
bipuladh: yeah we should definitely do it. It would be easy to do in a follow up PR though.

bipuladh: I will create an issue in the repo.

packages/shared/status/StatusIconAndText.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react';
+import classNames from 'classnames';
+import { CamelCaseWrap } from '../utils/CamelCaseWrap';
+import { StatusComponentProps } from './types';
+
+const DASH = '-';
```
SanjalKatiyar, line 6: very small things...
since we are using this variable in multiple files we can import it from some common file exporting constants.

bipuladh:
Currently I have not added AddCapacity extension for CSV pages. Need some more code for it. So will do it in a different PR. Will create an issue to track it.

bipuladh:
> Currently I have not added AddCapacity extension for CSV pages. Need some more code for it. So will do it in a different PR. Will create an issue to track it.

https://github.com/red-hat-storage/odf-console/issues/85

packages/odf/components/odf-dashboard/dashboard.tsx
```diff
@@ -57,6 +58,11 @@ const ODFDashboardPage: React.FC<ODFDashboardPageProps> = (props) => {
       name: t('Overview'),
       component: ODFDashboard,
     },
+    {
+      href: 'systems',
+      name: t('Storage Systems'),
+      component: StorageSystemListPage,
+    },
```
SanjalKatiyar, line 65: just a doubt... can't we add it as a extension here in `odf-console` as well ?
like we were doing in `console` repo...

bipuladh: We can, but why would we do it? The reason we did that is because we couldn't directly put it here.

SanjalKatiyar: ack...

SanjalKatiyar:
> Screenshot from 2021-12-10 16-21-33
> 
> Working !
> 
> Will take some time for code reviews

/lgtm

PR #84, opened 2021-12-14
use PF components for the cards on the dashboard

SanjalKatiyar:
Based on https://github.com/openshift/console/pull/10410 we need to remove the usage of console dashboard components and use PF components instead.

**Before:**
Screenshot from 2021-12-14 17-07-42

**After:**
Screenshot from 2021-12-14 17-07-01


rexagod:
/lgtm

Fixes the ODF dashboard cards.

PR #92, opened 2022-01-13
Add Common Resource Dropdown Component

bipuladh:
Example:
Screenshot from 2022-01-18 11-40-06

### DIFF
--- a/packages/shared/selectors/k8s.ts
+++ b/packages/shared/selectors/k8s.ts
@@ -1,0 +1,7 @@
+import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
+
+type GetStringProperty<T = K8sResourceCommon> = (resource: T) => string;
+
+export const getName: GetStringProperty = (resource) => resource?.metadata?.name;
+
+export const getUID: GetStringProperty = (resource) => resource?.metadata?.uid;
--- a/yarn.lock
+++ b/yarn.lock
@@ -4631,6 +4631,11 @@ functional-red-black-tree@^1.0.1:
   resolved "https://registry.yarnpkg.com/functional-red-black-tree/-/functional-red-black-tree-1.0.1.tgz#1b0ab3bd553b2a0d6399d29c0e3ea0b252078327"
   integrity sha1-GwqzvVU7Kg1jmdKcDj6gslIHgyc=
 
+fuzzysearch@^1.0.3:
+  version "1.0.3"
+  resolved "https://registry.yarnpkg.com/fuzzysearch/-/fuzzysearch-1.0.3.tgz#dffc80f6d6b04223f2226aa79dd194231096d008"
+  integrity sha1-3/yA9tawQiPyImqnndGUIxCW0Ag=
+
 gauge@~2.7.3:
   version "2.7.4"
   resolved "https://registry.yarnpkg.com/gauge/-/gauge-2.7.4.tgz#2c03405c7538c39d7eb37b317022e325fb018bf7"
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "@patternfly/react-icons": "^4.11.0",
     "@patternfly/react-table": "^4.29.0",
     "classnames": "^2.3.1",
+    "fuzzysearch": "^1.0.3",
     "lodash-es": "^4.17.21",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
--- a/packages/shared/selectors/index.ts
+++ b/packages/shared/selectors/index.ts
@@ -1,0 +1,1 @@
+export * from './k8s';
--- a/packages/shared/utils/ResourceDropdown.tsx
+++ b/packages/shared/utils/ResourceDropdown.tsx
@@ -1,0 +1,242 @@
+import * as React from 'react';
+import {
+  useK8sWatchResource,
+  WatchK8sResource,
+} from '@openshift-console/dynamic-plugin-sdk';
+import { K8sModel } from '@openshift-console/dynamic-plugin-sdk-internal-kubevirt/lib/api/common-types';
+import classNames from 'classnames';
+import * as fuzzy from 'fuzzysearch';
+import { useTranslation } from 'react-i18next';
+import {
+  Dropdown,
+  DropdownItem,
+  DropdownToggle,
+  TextInput,
+} from '@patternfly/react-core';
+import { CaretDownIcon } from '@patternfly/react-icons';
+import { LoadingInline } from '../generic/Loading';
+import { getName, getUID } from '../selectors';
+import './resourceDropdown.scss';
+
+const filterName = (searchText: string, resourceName: string) =>
+  fuzzy(searchText, resourceName);
+
+type ResourceBadgeProps = {
+  model: K8sModel;
+};
+
+const ResourceBadge: React.FC<ResourceBadgeProps> = ({
+  model: resourceModel,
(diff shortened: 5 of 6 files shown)


packages/shared/selectors/k8s.ts
```diff
@@ -0,0 +1,7 @@
+import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
+
+type GetStringProperty<T = K8sResourceCommon> = (resource: T) => string;
+
+export const getName: GetStringProperty = (resource) => resource.metadata?.name;
+
+export const getUID: GetStringProperty = (resource) => resource.metadata?.uid;
```
afreen23, line 7: if this is generic utility , then in usage resource can be undefined as well.

```suggestion
export const getName: GetStringProperty = (resource) => resource?.metadata?.name;

export const getUID: GetStringProperty = (resource) => resource?.metadata?.uid;
```

packages/shared/dropdown/resourceDropdown.scss
```diff
@@ -0,0 +1,7 @@
+.odf-resourceDropdown__container {
+  min-width: 20em;
+}
+
+.odf-resourceDropdown__container > button {
+  width: 100%;
+}
```
afreen23, line 7: can you use pf properties respectively

packages/shared/dropdown/ResourceDropdown.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+type ResourceBadgeProps = {
+  model: K8sModel;
+};
+
+const ResourceBadge: React.FC<ResourceBadgeProps> = ({
+  model: resourceModel,
+}) => (
+  <>
+    <span className="sr-only">{resourceModel.abbr.toLocaleUpperCase()}</span>
```
afreen23, line 22: what is "sr-only" ?

afreen23 (changes requested):

packages/shared/dropdown/resourceDropdown.scss
bipuladh: Which PF property?

packages/shared/dropdown/ResourceDropdown.tsx
bipuladh: This style is defined on console project.

SanjalKatiyar:
**Few functionalities that can be added:**
1) LoadingInline component untill resources are loaded and error in case of loaderror.
2) A filtering prop, which will filter data before creating dropdown items.
3) If we are using this component as a StorageClass dropdown as well, then it should have one more functionality of pre-selecting `default` StorageClass if any initial `selection` is not provided by the parent component (if this functionality is not needed we can skip it and can make a new FC for SC dropdown instead).

bipuladh:
> **Few functionalities that can be added:**
> 
> 1. LoadingInline component untill resources are loaded and error in case of loaderror.
> 2. A filtering prop, which will filter data before creating dropdown items.
> 3. An initial `selection` prop, if we want any dropdown item pre-selected.
This and the 4th one is the same. You can implement 4 using the above prop.
> 4. If we are using this component as a StorageClass dropdown as well, then it should have one more functionality of pre-selecting `default` StorageClass if any initial `selection` is not provided by the parent component (if this functionality is not needed we can skip it or can make a new FC for SC dropdown as well).

packages/shared/dropdown/ResourceDropdown.tsx
```diff
@@ -0,0 +1,241 @@
...
+  const [selectedItem, setSelectedItem] = React.useState(null);
+  const [searchText, setSearchText] = React.useState('');
+
+  const [resources, loaded, loadError] =
+    useK8sWatchResource<T[]>(watchResource);
+
+  React.useEffect(() => {
+    if (initialSelection && selectedItem === null && loaded && !loadError) {
+      const item = initialSelection(resources);
+      setSelectedItem(item);
```
SanjalKatiyar, line 132:
```suggestion
      setSelectedItem(item);
      onSelect(item);
```

packages/shared/dropdown/ResourceDropdown.tsx
```diff
@@ -0,0 +1,241 @@
...
+          {loaded && !loadError && (
+            <ResourceDropdownText
+              text={selectedItem ? propertySelector(selectedItem) : ''}
+              resourceModel={resourceModel}
+              showBadge={showBadge}
+            />
+          )}
+          {loaded && loadError && (
+            <span className="odf-resourceDropdownContainer__toggle--error">
+              Error Loading
```
SanjalKatiyar, line 221:
```suggestion
              t('Error Loading')
```

packages/shared/dropdown/ResourceDropdown.tsx
```diff
@@ -0,0 +1,241 @@
...
+        </DropdownToggle>
+      }
+      isOpen={isOpen}
+    >
+      <TextInput
+        id="search-bar"
+        iconVariant="search"
+        value={searchText}
+        onChange={(val) => setSearchText(val)}
+        onClick={onClick}
```
SanjalKatiyar, line 233:
```suggestion
```
it's not doing anything, right ?

packages/shared/dropdown/ResourceDropdown.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+  const dropdownItems = React.useMemo(() => {
+    if (!loaded && loadError) {
+      return [];
+    } else {
+      return resources
+        .filter((res) =>
+          filterName(
+            searchText,
+            getName(res) && (filterResource ? filterResource(res) : true)
```
SanjalKatiyar, line 166:
```suggestion
            getName(res)) && (filterResource ? filterResource(res) : true)
```

SanjalKatiyar: missing `)` else we are passing boolean value to `filterName` function.

packages/shared/selectors/k8s.ts
```diff
@@ -0,0 +1,7 @@
+import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
```
SanjalKatiyar, line 1: instead of creating a new directory, can we add these functions under `@odf/shared/utils` ??
(that's where I added them in migration PRs as well along with lot more different utilities)

yarn.lock
```diff
@@ -4631,6 +4631,11 @@ functional-red-black-tree@^1.0.1:
   resolved "https://registry.yarnpkg.com/functional-red-black-tree/-/functional-red-black-tree-1.0.1.tgz#1b0ab3bd553b2a0d6399d29c0e3ea0b252078327"
   integrity sha1-GwqzvVU7Kg1jmdKcDj6gslIHgyc=
 
+fuzzysearch@^1.0.3:
+  version "1.0.3"
+  resolved "https://registry.yarnpkg.com/fuzzysearch/-/fuzzysearch-1.0.3.tgz#dffc80f6d6b04223f2226aa79dd194231096d008"
+  integrity sha1-3/yA9tawQiPyImqnndGUIxCW0Ag=
+
```
SanjalKatiyar, line 4638: can we have one commit for `ResourceDropdown` FC & other for `package.json` (so 2 commits total for this PR) ??

packages/shared/selectors/k8s.ts
bipuladh: I think selectors are good here.

PR #93, opened 2022-01-13
Add dropdown with filter

SanjalKatiyar:


SanjalKatiyar:
just for review purpose right now... 
/hold

SanjalKatiyar:
Screenshot from 2022-01-17 21-01-59

There is different component for StorageClass Dropdown (https://github.com/red-hat-storage/odf-console/blob/032abf40d90894d51613d9e37452afce332f6f9a/packages/shared/dropdown/storageclassdropdown.tsx) but pasting its screenshot as well because it looks similar (and multus dropdown don't have any data right now)
Screenshot from 2022-01-17 21-02-06

PR #101, opened 2022-01-19
Bug 2041507: Fix issues in AddCapacity Modal

bipuladh:
StorageClassDropdown now references the component.
Screenshot from 2022-01-19 15-36-51

### DIFF
--- a/packages/odf/modals/add-capacity/add-capacity-modal.tsx
+++ b/packages/odf/modals/add-capacity/add-capacity-modal.tsx
@@ -22,7 +22,7 @@ import {
 import { usePrometheusPoll } from '@openshift-console/dynamic-plugin-sdk-internal';
 import {
   useK8sGet,
-  StorageClassDropdown,
+  StorageClassDropdown as ScModule,
   createModalLauncher,
 } from '@openshift-console/dynamic-plugin-sdk-internal-kubevirt';
 import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk-internal/lib/extensions/console-types';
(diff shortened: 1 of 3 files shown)


packages/odf/modals/add-capacity/add-capacity-modal.tsx
```diff
@@ -57,6 +57,8 @@ import {
 } from '../../utils/ocs';
 import './add-capacity-modal.scss';
 
+const StorageClassDropdown = (ScModule as any).StorageClassDropdown;
```
afreen23, line 60: Why this ?

afreen23 (changes requested):

bipuladh:
Changes in the kubevirt SDK, Previously it was exporting the component now it is exporting the whole module but under the same name. 
Reference: 
https://github.com/openshift/console/commit/e94be3265f2ced9c4aca1587ee3ac011a665d534#diff-5e83ab3d6fc56a1651d0621bac9b3d901cadaa18c868b99c4582262fbbb1e453L175-L176

packages/odf/modals/add-capacity/add-capacity-modal.tsx
bipuladh: See https://github.com/red-hat-storage/odf-console/pull/101#issuecomment-1016277766

afreen23 (approved):
/lgtm

packages/odf/modals/add-capacity/add-capacity-modal.tsx
afreen23: Makes sense now, thanks.

PR #105, opened 2022-01-24
Wait for created resources to be deleted

vbnrh:
This is to avoid issue with hitting an error while a namespace store is being deleted.
Screenshot from 2022-01-24 13-31-50


bipuladh (changes requested):
Wait until the resource is removed than changing names.

bipuladh (approved):
/lgtm

PR #113, opened 2022-02-09
(fix) Show OCS subcomponent health issue in UI

bipuladh:
Screenshot: 
Screenshot from 2022-02-09 17-50-02
Screenshot from 2022-02-09 17-42-13

### DIFF
--- a/locales/en/plugin__odf-console.json
+++ b/locales/en/plugin__odf-console.json
@@ -1,4 +1,12 @@
 {
+  "Block and File service is unhealthy": "Block and File service is unhealthy",
+  "Object service is unhealthy": "Object service is unhealthy",
+  "Warning": "Warning",
+  "Error": "Error",
+  "All resources are unhealthy": "All resources are unhealthy",
+  "Object Bucket has an issue": "Object Bucket has an issue",
+  "Many buckets have issues": "Many buckets have issues",
+  "Some buckets have issues": "Some buckets have issues",
   "Activity": "Activity",
   "OpenShift Data Foundation": "OpenShift Data Foundation",
   "Overview": "Overview",
--- a/packages/ocs/hooks/useOcsHealth.ts
+++ b/packages/ocs/hooks/useOcsHealth.ts
@@ -1,0 +1,138 @@
+import * as React from 'react';
+import { K8sResourceKind } from '@odf/shared/types';
+import { referenceForModel } from '@odf/shared/utils';
+import {
+  HealthState,
+  useK8sWatchResource,
+  WatchK8sResource,
+} from '@openshift-console/dynamic-plugin-sdk';
+import { usePrometheusPoll } from '@openshift-console/dynamic-plugin-sdk-internal';
+import { useTranslation } from 'react-i18next';
+import {
+  CephClusterModel,
+  CephObjectStoreModel,
+  NooBaaSystemModel,
+} from '../models/core';
+import { Health, HEALTH_QUERY } from '../queries';
+import {
+  getCephHealthState,
+  getNooBaaState,
+  getRGWHealthState,
+} from '../utils';
+
+const cephClusterResource: WatchK8sResource = {
+  kind: referenceForModel(CephClusterModel),
+  isList: true,
+};
+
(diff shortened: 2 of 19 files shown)


locales/en/plugin__odf-console.json
```diff
@@ -1,4 +1,13 @@
 {
+  "Block and File service is unhealthy": "Block and File service is unhealthy",
+  "Object Service is Unhealthy": "Object Service is Unhealthy",
+  "Object Service is unhealthy": "Object Service is unhealthy",
```
SanjalKatiyar, line 4:
```suggestion
  "Object Service is Unhealthy": "Object service is Unhealthy",
  "Object Service is unhealthy": "Object service is unhealthy",
```
Dashboard name is `Object`. So same as in case of `Block and File`, "service" should be lowercase ?

packages/ocs/utils/ceph-health.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export enum Phase {
+  CONNECTED = 'Connected',
+  PROGRESSING = 'Progressing',
+  FAILURE = 'Failure',
+  READY = 'Ready',
+}
+
+export const getRGWHealthState = (cr: K8sResourceKind): SubsystemHealth => {
+  const health = cr?.status?.phase;
```
SanjalKatiyar, line 64: Isn't phase will tell us status of the CR not the health ?
This might be discussed already, just confirming once.

packages/ocs/hooks/useOcsHealth.ts
```diff
@@ -0,0 +1,143 @@
...
+  ).state;
+
+  const mcgState = AcceptableHealthStates.includes(interimMCGState)
+    ? HealthState.OK
+    : HealthState.ERROR;
+
+  const rgwState =
+    interimRGWState === NA
+      ? HealthState.OK
+      : AcceptableHealthStates.includes(interimRGWState)
```
SanjalKatiyar, line 97: just nits: we can add NA to `AcceptableHealthStates` as well, it will save one extra condition.

packages/ocs/hooks/useOcsHealth.ts
```diff
@@ -0,0 +1,143 @@
...
+  kind: referenceForModel(CephObjectStoreModel),
+  prop: 'rgw',
+  isList: true,
+};
+
+const noobaaResource: FirehoseResource = {
+  kind: referenceForModel(NooBaaSystemModel),
+  isList: true,
+  prop: 'noobaa',
+};
```
SanjalKatiyar, line 39: nits: type is not `FirehoseResource`, we don't need `prop`.

packages/odf/components/odf-dashboard/system-capacity-card/capacity-card.tsx
```diff
@@ -66,7 +66,7 @@ const SystemCapacityCard: React.FC = () => {
             managedSystemName: system.spec.name,
             managedSystemKind: referenceFor(apiGroup)(apiVersion)(kind),
             usedValue: humanizeBinaryBytes(usedMetric?.value?.[1]),
-            totalValue: _.isNumber(totalMetric?.value?.[1])
+            totalValue: _.isNumber(Number(totalMetric?.value?.[1]))
```
SanjalKatiyar, line 69: won't it be always true ?

bipuladh: it will be NaN if the value number is trying to parse is `undefined`.

packages/ocs/utils/ceph-health.ts
bipuladh: we determine health from the phase.

afreen23:
> Screenshot: Screenshot from 2022-02-09 17-50-02 Screenshot from 2022-02-09 17-42-13

The unhealthy icon can be before the texts - "Block and file unhealthy"
The separate column for health is a bit odd since its not reflecting any value but just plain icon

If we need to show Health column then lets keep it like:
Block and File  <icon>Warning/Error
Object               <icon>Ok/Healthy

Otherwise:
<icon> Block and File services are unhleathy

PR #134, opened 2022-03-04
fix entry point for /compatibility basePath

SanjalKatiyar:
For OCP 4.11 + ODF 4.10 --> basePath: `/compatibility/`
But console is fetching files from `/app-root/src/compatibility/compatibility/plugin-manifest.json` which is incorrect because there is no `compatibility` directory inside parent `compatibility` directory.
Console should fetch files from `/app-root/src/compatibility/plugin-manifest.json` instead, thus, need to change the entry-point for such requests.

### DIFF
--- a/default.conf
+++ b/default.conf
@@ -6,7 +6,7 @@ server {
         root   /opt/app-root/src;
     }
     location /compatibility/ {
-        root   /opt/app-root/src/compatibility;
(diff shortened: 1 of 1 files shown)


bipuladh:
/approve
/lgtm 
Thank you @SanjalKatiyar for debugging this.

PR #137, opened 2022-03-07
remove usage of setFlag from wizard footer.tsx

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-2805

As per the discussion: https://github.com/openshift/console/pull/10872#issuecomment-1017732666

`setFlag` will not be exposed. Hence, removing its usage from the `footer.tsx` file.

Values of flags after creating internal mode SS from the UI (without any re-load):
```
MCG_STANDALONE
OCS
OCS_CONVERGED
OCS_INDEPENDENT
```
Screenshot from 2022-02-28 19-26-32

### DIFF
--- a/packages/odf/features.ts
+++ b/packages/odf/features.ts
@@ -1,4 +1,8 @@
-import { SetFeatureFlag } from "@openshift-console/dynamic-plugin-sdk";
+import { OCSStorageClusterModel } from '@odf/core/models';
+import { StorageClusterKind } from '@odf/shared/types';
+import { SetFeatureFlag, k8sList } from "@openshift-console/dynamic-plugin-sdk";
+import * as _ from 'lodash';
+import { CEPH_STORAGE_NAMESPACE, SECOND } from './constants';
 
 export const ODF_MODEL_FLAG = 'ODF_MODEL';
 export const OCS_INDEPENDENT_FLAG = 'OCS_INDEPENDENT';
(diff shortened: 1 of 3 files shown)


packages/odf/features.ts
```diff
@@ -47,3 +51,35 @@ export const OCS_FEATURE_FLAGS = {
...
+      setFlag(OCS_CONVERGED_FLAG, false);
+      setFlag(OCS_INDEPENDENT_FLAG, false);
+    }
+  };
+
+  // calling first time instantaneously
+  // else it will wait for 15s before start polling
+  ocsDetector();
+  ocsIntervalId = setInterval(ocsDetector, 15 * SECOND);
+};
```
SanjalKatiyar, line 85: Still need to add an extension `"type": "console.flag"`

packages/odf/features.ts
```diff
@@ -47,3 +51,35 @@ export const OCS_FEATURE_FLAGS = {
 };
 
 export const setODFFlag = (setFlag: SetFeatureFlag) => setFlag(ODF_MODEL_FLAG, true);
+
+export const detectOCS = async (setFlag: SetFeatureFlag) => {
+  let ocsIntervalId = null;
+  const ocsDetector = async () => {
+    try {
+      const storageClusters = await k8sList({model: OCSStorageClusterModel, queryParams: { CEPH_STORAGE_NAMESPACE }});
+      if (storageClusters?.length > 0) {
```
bipuladh, line 60: else let's set set these all flags to false.

bipuladh (changes requested):

packages/odf/features.ts
SanjalKatiyar: won't it trigger re-render unnecessarily every 15 sec in this case until storageClusters is found ??

SanjalKatiyar: I can use a flag to set all flags to false only once...

packages/odf/features.ts
```diff
@@ -47,3 +51,43 @@ export const OCS_FEATURE_FLAGS = {
...
+        setFlag(
+          MCG_STANDALONE,
+          storageCluster?.spec?.multiCloudGateway?.reconcileStrategy === 'standalone',
+        );
+        clearInterval(ocsIntervalId);
+      } else if(setFlagFalse) {
+        setFlagFalse = false;
+        setFlag(OCS_FLAG, false);
+        setFlag(OCS_CONVERGED_FLAG, false);
+        setFlag(OCS_INDEPENDENT_FLAG, false);
```
bipuladh, line 80: Let's setup MCG standalone as false as well( also in the catch statement)

bipuladh (changes requested):

packages/odf/features.ts
```diff
@@ -46,4 +50,47 @@ export const OCS_FEATURE_FLAGS = {
...
+  setFlag(OCS_INDEPENDENT_FLAG, false);
+  setFlag(MCG_STANDALONE, false);
+};
+
 export const setODFFlag = (setFlag: SetFeatureFlag) => setFlag(ODF_MODEL_FLAG, true);
+
+export const detectOCS = async (setFlag: SetFeatureFlag) => {
+  let ocsIntervalId = null;
+  // to prevent unnecessary re-render every 15 sec
+  // untill storageClusters are found
```
afreen23, line 65:
```suggestion
  // until storageClusters are found
```

afreen23 (changes requested):

packages/odf/features.ts
```diff
@@ -46,4 +50,47 @@ export const OCS_FEATURE_FLAGS = {
   [FEATURES.ODF_VAULT_SA_KMS]: 'vault-sa-kms',
 };
 
+const setFlagsFalse = (setFlag: SetFeatureFlag) => {
```
afreen23, line 53:
```suggestion
const setOCSFlagsFalse = (setFlag: SetFeatureFlag) => {
```

afreen23 (changes requested):

afreen23 (approved):

afreen23:
/lgtm cancel

@bipuladh review pending

PR #140, opened 2022-03-08
Fix issue with breadcrumb header

SanjalKatiyar:
**Why to remove custom breadcrumbs from header.tsx file ?** Console's extension `console.resource/create` by default adds a breadcrumb to the create resource page. No more need to add custom breadcrumb any more.

**Why to poll for subscriptions resource in odf-system-list.tsx ?** https://github.com/openshift/console/blob/d2b18efe2c60c4973436e1b25b66e73120ced077/frontend/packages/operator-lifecycle-manager/src/components/operand/create-operand.tsx#L176 `match.params.csvName` should be having correct CSV name in order of poll for the corresponding CSV, in case of any error in polling for the CSV resource breadcrumb will get hidden.

**Before:**
Screenshot from 2022-03-08 17-43-19

**After:**
CSS (alignment) of the breadcrumb will be fixed on the console side. Need to create a BZ for that.
Screenshot from 2022-03-08 16-51-44


packages/shared/utils/common.ts
```diff
@@ -94,3 +94,8 @@ export const referenceForOwnerRef = (
   referenceForGroupVersionKind(groupVersionFor(ownerRef.apiVersion).group)(
     groupVersionFor(ownerRef.apiVersion).version
   )(ownerRef.kind);
+
+export const getCSVFromSubscription = (
+  subscriptions: SubscriptionKind[],
+  operator: String,
+): String => _.find(subscriptions, (sub) => sub?.spec?.name === operator)?.status?.currentCSV;
```
afreen23, line 101: you can do filter ?

afreen23 (changes requested):

packages/shared/utils/common.ts
SanjalKatiyar: even if I add a filter still I need to do, right ??
`_.filter(subscriptions, (sub) => sub?.spec?.name === operator)[0]?.status?.currentCSV`

afreen23: Not really.
```suggestion
): String => subscriptions.filter((sub) => sub?.spec?.name === operator)
```

packages/shared/utils/common.ts
```diff
@@ -94,3 +94,7 @@ export const referenceForOwnerRef = (
   referenceForGroupVersionKind(groupVersionFor(ownerRef.apiVersion).group)(
     groupVersionFor(ownerRef.apiVersion).version
   )(ownerRef.kind);
+
+export const getCSVFromSubscription = (
+  subscription: SubscriptionKind,
+): String => subscription?.status?.currentCSV;
```
afreen23, line 100:
```suggestion
```

packages/odf/components/system-list/odf-system-list.tsx
```diff
@@ -271,21 +272,36 @@ const StorageSystemList: React.FC<StorageSystemListProps> = (props) => {
 };
 
 const StorageSystemListPage: React.FC<RouteComponentProps> = (props) => {
+  const [subs, subsLoaded, subsLoadError] = useK8sGet<SubscriptionKind>(
+    SubscriptionModel,
+    ODF_OPERATOR,
+    CEPH_STORAGE_NAMESPACE,
+  );
+
+  const csvName = getCSVFromSubscription(subs) || ODF_OPERATOR;
```
afreen23, line 281:
```suggestion
  const csvName = subscription?.status?.currentCSV || ODF_OPERATOR;
```

packages/odf/components/system-list/odf-system-list.tsx
```diff
@@ -271,21 +272,36 @@ const StorageSystemList: React.FC<StorageSystemListProps> = (props) => {
...
-      kind={referenceForModel(ODFStorageSystem)}
-      namespace={CEPH_STORAGE_NAMESPACE}
-      canCreate
-      createProps={createProps}
-    />
+    <StatusBox
+      skeleton={<div className="loading-skeleton--table" />}
+      data={subs}
+      loaded={subsLoaded}
+      loadError={subsLoadError}
```
afreen23, line 293: done we have any mechanism in ListPage to handle this ?

afreen23 (changes requested):

packages/odf/components/system-list/odf-system-list.tsx
SanjalKatiyar: yeah there was... thanks for suggesting it...

But one thing I would like to point out is that, `ListPage` in this `odf-system-list.tsx` file is still been imported from kubevirt package.
Every file which is part of `create-storage-system` directory, I have already fixed the imports and all are now using correct/new table components. This PR is to fix any bug that was left undone related to that wizard flow.

For `odf-system-list.tsx` and any other files in `odf-console` that are not under `create-storage-system`, they are still using kubevirt currently, we need to fix those files as well in different PR.

afreen23 (approved):
/lgtm

packages/odf/components/system-list/odf-system-list.tsx
afreen23: Ok, thats not related to this PR. Feel free to do later

PR #147, opened 2022-03-15
Add the dr-list page

RishabhKodes:
Initialized the Disaster recovery list page and created the headers for the list.

PR #155, opened 2022-03-24
DRPolicy Delete action

GowthamShanmugam:
On top of: https://github.com/red-hat-storage/odf-console/pull/147


GowthamShanmugam:
image

GowthamShanmugam:
@bipuladh please review

packages/shared/kebab/kebab.tsx
```diff
@@ -13,21 +13,33 @@ type KebabProps = {
...
+      {t('plugin__odf-console~Edit Annotations')}
+    </DropdownItem>),
+   [ModalKeys.DELETE]:
+    (<DropdownItem key={ModalKeys.DELETE} id={ModalKeys.DELETE}>
+      {t('plugin__odf-console~Delete')}
+    </DropdownItem>),
+});
+
+const overrideDefaultKebabItems = (t: TFunction, cutomKeys: string[]) => {
+  const kebabItems = defaultKebabItems(t);
```
GowthamShanmugam, line 39: If customkeys has the same key as default then custom then default key will be overridden by customkeys props and its modal.

packages/mco/components/data-policies/drpolicy-list-page/drpolicy-list-page.tsx
```diff
@@ -37,20 +48,51 @@ const tableColumnInfo = [
...
+  React.useEffect(() => {
+    if(drPlacementsControlLoaded && !drPlacementsControlLoadError){
+      setFilteredDRPlacementControl(
+        drPlacementControls?.filter((drPlacementControl) => drPlacementControl?.spec?.drPolicyRef?.name === obj?.metadata?.name)
+      );
+    };
+  }, [drPlacementControls, drPlacementsControlLoaded, drPlacementsControlLoadError,  obj?.metadata?.name]);
+
+  const kebabActionItems = () => ({
+    [ModalKeys.DELETE]: {
```
GowthamShanmugam, line 79: Here i am overriding DELETE, but it will still use the default delete modal. If you want to override the default modal also for delete then pass your modal here: https://github.com/red-hat-storage/odf-console/pull/155/commits/6a1e8db428d172a92583b11fcbe921481879088d#diff-1b32ef98d898b5f4a22e51d9aa49e14d03849959c2d1836f34042b0044acef49R202

With this change, if you want to only override the default action dropdown and its props then it is possible. if you want to override default modal as well, that is also possible.

packages/shared/kebab/kebab.tsx
bipuladh: Could we use the same modal keys and use Object.assign instead?

bipuladh (changes requested):

packages/mco/components/data-policies/drpolicy-actions/policy-actions.tsx
```diff
@@ -0,0 +1,3 @@
+export enum Actions {
+    DELETE_DR_POLICY = 'Delete DRPolicy',
+};
```
SanjalKatiyar, line 3: shouldn't this be under constants ??

SanjalKatiyar (approved):

locales/en/plugin__odf-console.json
```diff
@@ -19,6 +19,7 @@
   "Checkbox to select cluster": "Checkbox to select cluster",
   "Data policies": "Data policies",
   "Disaster recovery": "Disaster recovery",
+  "Cannot delete while connected to an Application": "Cannot delete while connected to an Application",
```
afreen23, line 22:
```suggestion
  "Cannot delete while connected to an application": "Cannot delete while connected to an Application",
```

packages/shared/kebab/kebab.tsx
```diff
@@ -56,16 +75,16 @@ export const Kebab: React.FC<KebabProps> = ({
...
 
   const drpodownItems = React.useMemo(() => {
-    const customItems = Object.entries(
+    const customItems: KebabItems = Object.entries(
       customKebabItems ? customKebabItems(t) : {}
-    ).map(([k, v]) => {
-      return (
-        <DropdownItem key={k} id={k}>
-          {v}
+    ).reduce((obj, [k, v]) => (
```
afreen23, line 80: in case of  {}
```suggestion
    ).?reduce((obj, [k, v]) => (
```

afreen23 (changes requested):

locales/en/plugin__odf-console.json
GowthamShanmugam: Application is a CR in OCP, its ok to start as capital

afreen23: Alright then if its a CR.

packages/mco/models/models.ts
```diff
@@ -12,3 +12,16 @@ export const DRPolicyModel: K8sModel = {
...
+export const DRPlacementControlModel: K8sModel = {
+    label: 'DR Placement Control',
+    labelPlural: 'DR Placement Controls',
+    apiVersion: 'v1alpha1',
+    apiGroup: 'ramendr.openshift.io',
+    plural: 'drplacementcontrols',
+    abbr: 'DRPC',
+    namespaced: true,
+    kind: 'DRPlacementControl',
+    id: 'drplacementcontrol',
```
afreen23, line 25:
```suggestion
```
No need to maintain id, they are no  longer used in console

packages/mco/constants/dr-policy.ts
```diff
@@ -6,3 +6,7 @@ export const REPLICATION_TYPE = (t: TFunction) => ({
   async: t('plugin__odf-console~Asynchronous'),
   sync: t('plugin__odf-console~Synchronous'),
 });
+
+export enum Actions {
+  DELETE_DR_POLICY = 'Delete DRPolicy',
```
afreen23, line 11: externalize

packages/mco/types/types.ts
```diff
@@ -29,3 +29,24 @@ export type DRPolicyKind = K8sResourceCommon & {
...
+      };
+      preferredCluster?: string;
+      pvcSelector: {
+          matchLabels: {
+              [key in string]: string;
+          }
+      }
+  };
+  status?: {
+      Phase: string;
```
afreen23, line 50: is it `phase` or `Phase` only ??

locales/en/plugin__odf-console.json
```diff
@@ -28,6 +29,15 @@
...
   "Synchronous": "Synchronous",
+  "StorageClasses": "StorageClasses",
+  "Replicas": "Replicas",
+  "UsedCapacity": "UsedCapacity",
+  "Mirroring Status": "Mirroring Status",
+  "Overall Image Health": "Overall Image Health",
+  "Compression Status": "Compression Status",
+  "Compression Savings": "Compression Savings",
+  "BlockPools": "BlockPools",
+  "Create BlockPool": "Create BlockPool",
```
afreen23, line 40: I dont see these strings externalized in this PR ? You might need to run yarn i18n to update these.

afreen23 (changes requested):

packages/mco/models/models.ts
GowthamShanmugam: Ah! looks like all PR got affected by these change

afreen23 (approved):
/lgtm

packages/shared/kebab/kebab.tsx
```diff
@@ -63,23 +67,25 @@ export const Kebab: React.FC<KebabProps> = ({
 
   const dropdownItems = React.useMemo(() => {
     const defaultResolved = defaultKebabItems(t);
-    const customResolved = customKebabItems(t);
+    const customResolved = customKebabItems ? customKebabItems(t): {};
```
GowthamShanmugam, line 70: UI is breaking when no customkebab item is passed

bipuladh: ack

PR #156, opened 2022-03-29
ACM overview dashboard

SanjalKatiyar:
Route: `${cluster_url}/multicloud/data-services/storagesystem`

Screenshot from 2022-03-29 18-34-47
Screenshot from 2022-03-29 18-34-42
Screenshot from 2022-03-29 19-26-40
Screenshot from 2022-03-31 19-14-46

PR #159, opened 2022-03-29
Enabling Multicluster dashboard when ACM observability is enabled - WIP

GowthamShanmugam:
on top of: https://github.com/red-hat-storage/odf-console/pull/156


packages/mco/components/mco-dashboard/queries.ts
```diff
@@ -0,0 +1,17 @@
...
+    [StorageDashboard.TOTAL_CAP_FILE_BLOCK]:
+      '(label_replace(odf_system_map{target_namespace="openshift-storage"} , "managedBy", "$1", "target_name", "(.*)"))  * on (namespace, managedBy) group_right(storage_system) odf_system_raw_capacity_total_bytes',
+    [StorageDashboard.USED_CAPACITY_FILE_BLOCK]:
+      '(label_replace(odf_system_map{target_namespace="openshift-storage"} , "managedBy", "$1", "target_name", "(.*)"))  * on (namespace, managedBy) group_right(storage_system) odf_system_raw_capacity_used_bytes',
+  };
+  
+  export const STATUS_QUERIES = {
+    [StorageDashboard.HEALTH]: '(label_replace(odf_system_map{target_namespace="openshift-storage"} , "managedBy", "$1", "target_name", "(.*)"))  * on (namespace, managedBy) group_right(storage_system) odf_system_health_status'
+  };
+
```
bipuladh, line 17: new line missing

packages/mco/components/mco-dashboard/status-card/status-card.tsx
```diff
@@ -0,0 +1,93 @@
...
+    case '0':
+      return HealthState.OK;
+    case '1':
+      return HealthState.WARNING;
+    case '2':
+      return HealthState.ERROR;
+    default:
+      return HealthState.LOADING;
+  }
+};
```
bipuladh, line 33: Refactor this from ODF package and put it in shared.

packages/mco/components/mco-dashboard/status-card/status-card.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  const [worstHealth, setWorstHealth] = React.useState<string>('0');
+  const [healthData, healthError, healthLoading] =
+    useURLPoll<PrometheusResponse>(`${ACM_ENDPOINT}=${STATUS_QUERIES[StorageDashboard.HEALTH]}`);
+
+  const parsedHealthData = React.useMemo(() => {
+      let health = '0';
+      let data = !healthError && !healthLoading
+        ? healthData?.data?.result.reduce((acc, statusMetric) => {
+            health = health < statusMetric?.value[1] ? statusMetric?.value[1] : health;
```
bipuladh, line 46: Ideally reduce should not have side effects to variables outside it's scope. So I don't think reduce or map should be used in this case.

bipuladh, line 46: health is a string in the outer scope, I am not sure even if intrinsic conversion happens it's a right way to do the comparisons. Use  explicit conversions is required?

packages/mco/components/mco-dashboard/status-card/status-card.tsx
```diff
@@ -0,0 +1,93 @@
...
+          <Flex
+            justifyContent={{ default: 'justifyContentSpaceBetween' }}
+            className="odf-StatusCard__items--width"
+          >
+            <FlexItem>
+              <HealthItem
+                title={t('Openshift Data Foundation')}
+                state={HealthState.OK}
+              >
+                {/*<StorageSystemPopup systemHealthMap={healthySystems} />*/}
```
bipuladh, line 74: Why?

packages/mco/components/mco-dashboard/status-card/storage-system-popup.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+const getSystemHealthCount = (systemHealthMap: SystemHealthMap[]) => {
+    let critical, warning, normal;
+    critical = warning = normal = 0;
+    systemHealthMap.forEach((item) => {
+        if (item?.rawHealthData === '0') normal++;
+        else if (item?.rawHealthData === '1') warning++;
+        else if (item?.rawHealthData === '2') critical++;
+    });
+    return {critical, warning, normal};
```
bipuladh, line 47: Use reduce here.

packages/mco/components/mco-dashboard/status-card/storage-system-popup.tsx
```diff
@@ -0,0 +1,99 @@
...
+  systemHealthMap: SystemHealthMap[];
+};
+
+type SystemHealthCountProps = {
+    value: string;
+    health: HealthState;
+    count: number;
+}
+
+const healthStateToIcon = {
```
bipuladh, line 27: How about using https://github.com/red-hat-storage/odf-console/blob/master/packages/shared/dashboards/status-card/states.tsx#L42 ?

packages/odf/features.ts
```diff
@@ -94,3 +101,49 @@ export const setOCSFlags = async (setFlag: SetFeatureFlag) => {
...
+    k8sList({model: mcokind, queryParams: {}})
+      .then((data: ACMMultiClusterObservability[]) => {
+        const isEnabled = data?.[0].status?.conditions?.some((condition) => condition.status == "True"  && condition.type == "Ready");
+        if (isEnabled) {
+          k8sGet({model: configMapkind, name: ACM_WHITELISTING_METRICS_CONFIG_MAP, ns: ACM_OBSERVABILITY_NS}).then((config: ConfigMapKind) => {
+            if (config?.metadata?.labels?.[ODF_METRICS_LABEL] === "true"){
+              setFeatureFlag(ACM_OBSERVABILITY_FLAG, true);
+            };
+          }).catch((error) => {
+
```
bipuladh, line 122: console.error or do something here.

packages/odf/features.ts
```diff
@@ -94,3 +101,49 @@ export const setOCSFlags = async (setFlag: SetFeatureFlag) => {
   ocsDetector();
   ocsIntervalId = setInterval(ocsDetector, 15 * SECOND);
 };
+
+
+export const detectACMObservability = async (setFeatureFlag: SetFeatureFlag) => {
+  let id = null;
+  let isInitial = true;
+  let mcokind: K8sModel = {...AcmMultiClusterObservabilityModel}
```
bipuladh, line 109:
```suggestion
  let mcokind: K8sModel = AcmMultiClusterObservabilityModel
```
Wouldn't this work?
Same for configMapKind

bipuladh (changes requested):

packages/odf/features.ts
```diff
@@ -1,6 +1,11 @@
 import { OCSStorageClusterModel } from '@odf/core/models';
-import { StorageClusterKind } from '@odf/shared/types';
-import { SetFeatureFlag, k8sList } from "@openshift-console/dynamic-plugin-sdk";
+import { ACM_OBSERVABILITY_NS, ACM_WHITELISTING_METRICS_CONFIG_MAP, ODF_METRICS_LABEL  } from '@odf/mco/constants';
+import { AcmMultiClusterObservabilityModel } from '@odf/mco/models';
+import { ACMMultiClusterObservability } from '@odf/mco/types';
```
SanjalKatiyar, line 4: https://github.com/red-hat-storage/odf-console/blob/7150378e539894dcb672ef7079487faa1beb7258/packages/odf/package.json#L6
now won't `/odf` be having a dependency on `/mco` as well ??

packages/mco/components/mco-dashboard/status-card/status-card.tsx
SanjalKatiyar: ack

packages/mco/components/mco-dashboard/status-card/status-card.tsx
SanjalKatiyar: there is no type conversion here (implicit or explicit)... I am doing string comparison only...
`health` & `statusMetric?.value[1]` both are strings which can be `0 or 1 or 2` that's all, so even if we do comparison between these strings result will be same like doing comparison between numbers. So, I don't think there is need to do type conversion.

packages/mco/components/mco-dashboard/queries.ts
GowthamShanmugam: @SanjalKatiyar

packages/mco/components/mco-dashboard/status-card/status-card.tsx
SanjalKatiyar: It is WIP, we had a discussion regarding this yesterday only... will update this...

packages/mco/components/mco-dashboard/status-card/storage-system-popup.tsx
SanjalKatiyar: ack...

packages/odf/features.ts
GowthamShanmugam: yes, mco is seperate packages now, which has only multicluster dashbaord related modals, i.e ACM and ramen models

packages/odf/features.ts
GowthamShanmugam: ack

packages/odf/features.ts
GowthamShanmugam: mmmm. it makesense. i will move this to shared

GowthamShanmugam:
@SanjalKatiyar @bipuladh changes done on enabling/disabling dashboard part.

packages/shared/models/acm.ts
```diff
@@ -0,0 +1,15 @@
+import { K8sModel } from '@openshift-console/dynamic-plugin-sdk/lib/api/common-types';
```
GowthamShanmugam, line 1: moved to shared because, it might be required by MCO package also in future

PR #160, opened 2022-03-30
Set MCO flag at compile time

bipuladh:
- Creates a build command for MCO
- Sets MCO flag based on build parameter.
- Updates the http-server to remove dependency on certs.
MCO_flag_set

### DIFF
--- a/http-server.sh
+++ b/http-server.sh
@@ -6,4 +6,4 @@ public_path="$1"
 shift
 server_opts="$@"
 
-http-server $public_path -p 9001 -S true -C /var/serving-cert/tls.crt -K /var/serving-cert/tls.key -c-1 --cors $server_opts 
+http-server $public_path -p 9001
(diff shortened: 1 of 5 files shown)


http-server.sh
```diff
@@ -6,4 +6,4 @@ public_path="$1"
 shift
 server_opts="$@"
 
-http-server $public_path -p 9001 -S true -C /var/serving-cert/tls.crt -K /var/serving-cert/tls.key -c-1 --cors $server_opts 
+http-server $public_path -p 9001
```
GowthamShanmugam, line 9: is it wont run as http instead of https?

GowthamShanmugam (approved):

http-server.sh
SanjalKatiyar: I think this file is for running local only... that is why...

PR #164, opened 2022-04-05
Add Creation of BackingStore Page

bipuladh:
Screenshot from 2022-04-17 13-15-49


weirdwiz:
LGTM
image

packages/shared/dropdown/StaticDropdown.tsx
```diff
@@ -0,0 +1,70 @@
...
+          {dropdownText}
+        </DropdownToggle>
+      }
+      autoFocus={false}
+      onSelect={onChange}
+    />
+  );
+};
+
+export default StaticDropdown;
```
SanjalKatiyar, line 70: There is a `SingleSelectDropdown` which can be used to display static dropdownitems... ptal once...

bipuladh: I checked it out I still prefer StaticDropdown. 
Reasons: 
1) Static Dropdown handles the generation of DropdownItem.
2) Select has check mark next to it. Dropdown doesn't show this. 

However we should align these two components together and also ask UXD to help us determine which to use where.

SanjalKatiyar: ack...

1. In case of  `SingleSelectDropdown` I intentionally passed DropdownItem as a prop. My thinking at that time was that it will be helpful in case we need any custom changes to dropdownitems (may our own dropdownitem component etc).
For example: https://github.com/red-hat-storage/odf-console/blob/976ffc34c617d852104fe23dbf15866d847ddbd6/packages/odf/components/utils/osd-size-dropdown.tsx#L27
here we needed `description` as well, so parent component had full control to add any custom prop to the dropdownitem instead of always changing generic SingleSelectDropdown code. 
Maybe we have some other use case sometime, in that case we can directly make changes to dropdownitem and pass it down as a prop. 
This was just in my mind at that time, not saying that it is a good approach or not.

2. I agree on this check mark part.

PR #170, opened 2022-04-05
Handle null values in the dashboard when connectivity is lost

bipuladh:
Screenshot from 2022-04-05 17-41-28

### DIFF
--- a/packages/odf/components/odf-dashboard/object-storage-card/capacity-card.tsx
+++ b/packages/odf/components/odf-dashboard/object-storage-card/capacity-card.tsx
@@ -9,19 +9,19 @@ import { ODFStorageSystem } from '../../../models';
 import { CAPACITY_QUERIES, StorageDashboard } from '../queries';
 
 const parseMetricData = (metric: PrometheusResponse) =>
-  metric.data.result.map((datum) => ({
-    name: datum.metric.type,
-    usedValue: humanizeBinaryBytes(datum.value[1]),
-  }));
+  metric?.data?.result?.map((datum) => ({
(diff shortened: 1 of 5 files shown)


packages/odf/components/odf-dashboard/object-storage-card/capacity-card.tsx
```diff
@@ -21,7 +21,7 @@ const ObjectCapacityCard: React.FC = () => {
     endpoint: 'api/v1/query' as any,
   });
 
-  const dataFrames = !loaded && !error ? parseMetricData(data) : [];
+  const dataFrames = loaded && !error ? parseMetricData(data) : [];
```
SanjalKatiyar, line 24:
```suggestion
  const dataFrames = !loaded && !error ? parseMetricData(data) : [];
```
Earlier it was correct, it is bad naming convention. It should be renamed to `loading` because it is Prometheus polling response...

packages/odf/components/odf-dashboard/status-card/status-card.tsx
```diff
@@ -2,12 +2,10 @@ import * as React from 'react';
...
 import HealthItem from '@odf/shared/dashboards/status-card/HealthItem';
 import { ClusterServiceVersionKind } from '@odf/shared/types';
-import {
-  getGVK,
-  referenceForModel,
-} from '@odf/shared/utils';
+import { getGVK, referenceForModel } from '@odf/shared/utils';
 import {
   HealthState,
+  k8sCreate,
```
SanjalKatiyar, line 8: it is not used anywhere in this page ??

PR #176, opened 2022-04-12
Migrate ListPage for BlockPools

bipuladh:
Screenshot from 2022-04-12 12-07-08


packages/ocs/block-pool/utils.ts
```diff
@@ -0,0 +1,49 @@
+import { StorageClassResourceKind } from '@odf/shared/types';
```
GowthamShanmugam, line 1: Rename this file to blockPoolUtils.ts and move this under some common utils folder

GowthamShanmugam (commented):
One minor comment,

GowthamShanmugam (approved):

PR #180, opened 2022-04-14
Generic localization between odf and odfmco

GowthamShanmugam:


locales/en/plugin__odf-console.json
```diff
@@ -28,6 +28,15 @@
   "DRPolicies": "DRPolicies",
   "Asynchronous": "Asynchronous",
   "Synchronous": "Synchronous",
+  "StorageClasses": "StorageClasses",
+  "Replicas": "Replicas",
+  "UsedCapacity": "UsedCapacity",
+  "Mirroring Status": "Mirroring Status",
+  "Overall Image Health": "Overall Image Health",
+  "Compression Status": "Compression Status",
+  "Compression Savings": "Compression Savings",
```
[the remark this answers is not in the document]
SanjalKatiyar: okay, this PR is not specifically adding these texts...

packages/shared/status/statuses.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
-import { useTranslation } from 'react-i18next';
 import { InProgressIcon } from '@patternfly/react-icons';
+import { useSharedTranslation } from '../useTranslationHook';
```
SanjalKatiyar, line 3: nits: useTranslation is already provided by i18n... `useCustomTranslationHook` or `useSharedTranslationHook` as a name for the file would be better IMO.
```suggestion
import { useSharedTranslation } from '../useSharedTranslationHook';
```

locales/en/plugin__odf-console.json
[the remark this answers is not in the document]
GowthamShanmugam: Yep, looks like this was missed in already merged PR. i will change this.

packages/shared/status/statuses.tsx
GowthamShanmugam: ack

GowthamShanmugam:
@SanjalKatiyar @bipuladh Changes done

packages/shared/useCustomTranslationHook.ts
```diff
@@ -0,0 +1,8 @@
+import { useTranslation } from 'react-i18next';
+
+
+export const NS = process.env.NS;
+
+export const useCustomTranslation = () => {
+    return useTranslation(NS);
+};
```
afreen23, line 8:
```suggestion
export const useCustomTranslation = () => useTranslation(NS);
```

packages/shared/useCustomTranslationHook.ts
```diff
@@ -0,0 +1,8 @@
+import { useTranslation } from 'react-i18next';
+
+
+export const NS = process.env.NS;
```
afreen23, line 4:
```suggestion
export const NS = process.env.I8N_NS;
```

afreen23: It will give more clarity about NS

webpack.config.ts
```diff
@@ -5,6 +5,11 @@ import { ConsoleRemotePlugin } from '@openshift-console/dynamic-plugin-sdk-webpa
 import * as CopyWebpackPlugin from 'copy-webpack-plugin';
 import * as webpack from 'webpack';
 
+const resolveLocale = (dirName: string, ns: string) =>
+  ['en', 'ja', 'ko', 'zh'].reduce((acc, lang) => ([...acc, { from: path.resolve(dirName, `locales/${lang}/plugin__*.json`), to: `locales/${lang}/${ns}.[ext]`}]), [])
```
afreen23, line 9: Create a constant for the languages used:
export const LANGUAGES =  ['en', 'ja', 'ko', 'zh'];
```suggestion
 LANGUAGES.reduce((acc, lang) => ([...acc, { from: path.resolve(dirName, `locales/${lang}/plugin__*.json`), to: `locales/${lang}/${ns}.[ext]`}]), [])
```

afreen23 (changes requested):

PR #182, opened 2022-04-18
Migrate Namespace Store Creation Page

bipuladh:


packages/odf/components/namespace-store/create-namespace-store.tsx
```diff
@@ -0,0 +1,66 @@
...
+        redirectHandler={(resources) => {
+          const lastIndex = resources.length - 1;
+          const resourcePath = `${referenceForModel(
+            NooBaaNamespaceStoreModel
+          )}/${getName(resources[lastIndex])}`;
+          isODF
+            ? history.push(`/odf/resource/${resourcePath}`)
+            : history.push(
+                `/k8s/ns/${ns}/clusterserviceversions/${appName}/${resourcePath}`
+              );
```
SanjalKatiyar, line 52: don't think we need it anymore, but we can remove later as well ! whichever is more alright.

packages/odf/components/namespace-store/namespace-store-modal.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <ModalBody>
+          <p>
+            {t(
+              'Represents an underlying storage to be used as read or write target for the data in the namespace buckets.'
+            )}
+          </p>
+          <NamespaceStoreForm
+            namespace={namespace}
+            onCancel={() => close()}
+            redirectHandler={() => close()}
```
SanjalKatiyar, line 29:
```suggestion
            onCancel={closeModal}
            redirectHandler={closeModal}
```
shouldn't this be `closeModal` instead ??

weirdwiz:
LGTM
image

packages/odf/components/namespace-store/create-namespace-store.tsx
bipuladh: Will do it in the bucketclass PR.

packages/odf/components/namespace-store/namespace-store-modal.tsx
bipuladh: Yes they should be there are other problems in this component as well. I am fixing them as part of create-bc PR as it's easier to work with them when I can actually see them. :)

PR #183, opened 2022-04-19
Avoid UI conflict between odf-console and odf-multicluster-console using flags

GowthamShanmugam:

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -16,6 +16,17 @@
     }
   },
   {
+    "type": "console.flag/model",
+    "properties": {
+      "model": {
+        "group": "ramendr.openshift.io",
+        "version": "v1alpha1",
+        "kind": "DRPolicy"
+      },
+      "flag": "ODR_MODEL"
+    }
+  },
+  {
     "type": "console.navigation/href",
     "properties": {
       "id": "odfdashboard",
@@ -25,7 +36,8 @@
       "href": "/odf/cluster"
     },
     "flags": {
-      "required": ["ODF_MODEL"]
+      "required": ["ODF_MODEL"],
+      "disallowed": ["MCO"]
     }
   },
   {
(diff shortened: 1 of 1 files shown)


console-extensions.json
```diff
@@ -15,6 +15,17 @@
...
   },
+  {
+    "type": "console.flag/model",
+    "properties": {
+      "model": {
+        "group": "ramendr.openshift.io",
+        "version": "v1alpha1",
+        "kind": "DRPolicy"
+      },
+      "flag": "ODR_MODEL"
```
bipuladh, line 26: Where is this used?

GowthamShanmugam: For DataService navigation on the ACM page. https://github.com/red-hat-storage/odf-console/pull/183/files#diff-2d79046ce9b0f8fda82dd631ce959a36f6c350960d54cf8fdb1579f3c718e83aR88

GowthamShanmugam: right now we have only one DataPolicies, i.e DisasterRecovery, if the ODR Hub operator is not installed then we dont need to show DataService options.

In future, once we started supporting multiple data services and data policies we will need to hide only the DisasterRecovery tab

afreen23: Is this flag currently defined anywhere ? I cant find that.

afreen23 (changes requested):

console-extensions.json
GowthamShanmugam: it is used: https://github.com/red-hat-storage/odf-console/pull/183/files#diff-2d79046ce9b0f8fda82dd631ce959a36f6c350960d54cf8fdb1579f3c718e83aR88

This extension will create this flag based on CRD present

GowthamShanmugam: @afreen23 ^

PR #190, opened 2022-04-21
Add support for Generic Resource List Pages

bipuladh:
Screenshot from 2022-04-21 11-19-18
Screenshot from 2022-04-21 11-19-13


packages/odf/components/resource-pages/label-list.tsx
```diff
@@ -0,0 +1,74 @@
...
+  name: string;
+  value: string;
+  expand: boolean;
+};
+
+export type LabelListProps = WithTranslation & {
+  labels: { [key: string]: string };
+  kind: K8sResourceKindReference;
+  expand?: boolean;
+};
```
SanjalKatiyar, line 74: can we move this to shared ?
looks like can be used by many components...

SanjalKatiyar: okay, we already have this component in `shared/details-page/label-list.tsx` let's use from there...

packages/odf/components/resource-pages/link.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const path = `/odf/resource/${referenceForModel(resourceModel)}/${name}`;
+  return (
+    <span>
+      <ResourceIcon resourceModel={resourceModel} />
+      <Link to={path}>{name}</Link>
+    </span>
+  );
+};
+
+export default ODFResourceLink;
```
SanjalKatiyar, line 22: we have `ResourceLink` in shared, can we use that directly ?

bipuladh: +1

packages/odf/components/resource-pages/label-list.tsx
bipuladh: ack

packages/odf/components/resource-pages/link.tsx
```diff
@@ -0,0 +1,24 @@
...
+      <ResourceLink
+        resourceModel={resourceModel}
+        resourceName={name}
+        link={path}
+      />
+    </span>
+  );
+};
+
+export default ODFResourceLink;
```
SanjalKatiyar, line 24: actually we can use ResourceLink directly on the list page component, right ?
why do we need this wrapper ??

PR #191, opened 2022-04-21
Migrate BackingStore details page

bipuladh:
Screenshot: 
Screenshot from 2022-04-22 13-04-35


packages/odf/components/resource-pages/BackingStoreDetailsPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+  }, [launchModal, resource]);
+
+  const Details = React.useMemo(
+    () => BSDetails(launchModal, t),
+    [launchModal, t]
+  );
+
+  const YAMLEditor = React.useCallback(
+    () => <ResourceYAMLEditor initialResource={resource} />,
+    [JSON.stringify(resource)]
```
SanjalKatiyar, line 118: let's use `useDeepCompareMemoize` instead. I can see multiple warnings here !!

SanjalKatiyar: or if ResourceYAMLEditor is expecting prop to be a string, let's stringify it inside the useCallBack !!

packages/odf/components/resource-pages/BackingStoreDetailsPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+              <ResourceSummary
+                resource={obj}
+                launchModal={launchModal}
+                resourceModel={NooBaaBackingStoreModel}
+              />
+            </div>
+          </div>
+        </div>
+        <div className="co-m-pane__body">
+          <SectionHeading text={t('Provider Details')} />
```
SanjalKatiyar, line 51:
```suggestion
          <SectionHeading text={t('Provider details')} />
```

packages/odf/components/resource-pages/BackingStoreDetailsPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+      <Kebab
+        toggleType="Dropdown"
+        launchModal={launchModal}
+        extraProps={{
+          resource,
+          resourceModel: NooBaaBackingStoreModel,
+        }}
+      />
+    );
+  }, [launchModal, resource]);
```
SanjalKatiyar, line 109: I think we can skip useCallBack here, wdyt ??
actions function is not passed to any React hook, PagHeading is directly calling this fucntion.

bipuladh: yes but we don't want the function to be recreated every time. I will pass it memoized resource to keep it stable.

packages/odf/components/resource-pages/BackingStoreDetailsPage.tsx
bipuladh: ack

packages/odf/components/resource-pages/BackingStoreDetailsPage.tsx
bipuladh: ack

packages/odf/components/resource-pages/BackingStoreDetailsPage.tsx
```diff
@@ -0,0 +1,153 @@
...
+              href: '',
+              name: 'Details',
+              component: Details as any,
+            },
+            {
+              href: 'yaml',
+              name: 'YAML',
+              component: YAMLEditor,
+            },
+          ]}
```
SanjalKatiyar, line 144: I think there is no need to add Details & YAMLEditor inside a React hooks.  Instead we can use useMemo hook on this "pages" array itself ??
cause reference of this array can still change on re-render

packages/odf/components/resource-pages/Providers.tsx
```diff
@@ -0,0 +1,220 @@
...
+    </FlexItem>
+    <FlexItem>{children}</FlexItem>
+  </Flex>
+);
+
+const AWSDetails: React.FC<ProviderDetailsProps> = ({ resource }) => {
+  const { t } = useTranslation();
+  const region = getRegion(resource);
+  const secret = resource.spec.awsS3.secret;
+  const targetBucket = resource.spec.awsS3.targetBucket;
```
SanjalKatiyar, line 33:
```suggestion
  const secret = resource?.spec?.awsS3?.secret;
  const targetBucket = resource?.spec?.awsS3?.targetBucket;
```

packages/odf/components/resource-pages/Providers.tsx
```diff
@@ -0,0 +1,220 @@
...
+        <DetailsItem field={t('Target Bucket')}>{targetBucket}</DetailsItem>
+      </FlexItem>
+    </Flex>
+  );
+};
+
+const AzureBlobDetails: React.FC<ProviderDetailsProps> = ({ resource }) => {
+  const { t } = useTranslation();
+  const secret = resource.spec.azureBlob.secret;
+  const targetBucket = resource.spec.azureBlob.targetBlobContainer;
```
SanjalKatiyar, line 62:
```suggestion
  const secret = resource?.spec?.azureBlob?.secret;
  const targetBucket = resource?.spec?.azureBlob?.targetBlobContainer;
```

packages/odf/components/resource-pages/Providers.tsx
```diff
@@ -0,0 +1,220 @@
...
+      </FlexItem>
+    </Flex>
+  );
+};
+
+const S3CompatibleDetails: React.FC<ProviderDetailsProps> = ({ resource }) => {
+  const { t } = useTranslation();
+  const secret = resource.spec.s3Compatible.secret;
+  const endpoint = resource.spec.s3Compatible.endpoint;
+  const targetBucket = resource.spec.s3Compatible.targetBucket;
```
SanjalKatiyar, line 91:
```suggestion
  const secret = resource?.spec?.s3Compatible?.secret;
  const endpoint = resource.spec.s3Compatible.endpoint;
  const targetBucket = resource.spec.s3Compatible.targetBucket;
```

packages/odf/components/resource-pages/Providers.tsx
```diff
@@ -0,0 +1,220 @@
...
+      </FlexItem>
+    </Flex>
+  );
+};
+
+const IBMDetails: React.FC<ProviderDetailsProps> = ({ resource }) => {
+  const { t } = useTranslation();
+  const secret = resource.spec.ibmCos.secret;
+  const endpoint = resource.spec.ibmCos.endpoint;
+  const targetBucket = resource.spec.ibmCos.targetBucket;
```
SanjalKatiyar, line 121: `?.` in case of `null`

packages/odf/components/resource-pages/Providers.tsx
```diff
@@ -0,0 +1,220 @@
...
+        <DetailsItem field={t('Target Bucket')}>{targetBucket}</DetailsItem>
+      </FlexItem>
+    </Flex>
+  );
+};
+
+const GCPDetails: React.FC<ProviderDetailsProps> = ({ resource }) => {
+  const { t } = useTranslation();
+  const secret = resource.spec.googleCloudStorage.secret;
+  const targetBucket = resource.spec.googleCloudStorage.targetBucket;
```
SanjalKatiyar, line 150: handling null/undefined values.

packages/odf/components/resource-pages/Providers.tsx
```diff
@@ -0,0 +1,220 @@
...
+        <DetailsItem field={t('Target Bucket')}>{targetBucket}</DetailsItem>
+      </FlexItem>
+    </Flex>
+  );
+};
+
+const PVCDetails: React.FC<ProviderDetailsProps> = ({ resource }) => {
+  const { t } = useTranslation();
+  const numVolumes = resource.spec.pvPool.numVolumes;
+  const storageClass = resource.spec.pvPool.storageClass;
```
SanjalKatiyar, line 176: same here...

packages/odf/components/resource-pages/Providers.tsx
```diff
@@ -0,0 +1,220 @@
...
+        <DetailsItem field={t('StorageClass')}>
+          <ResourceLink kind={StorageClassModel.kind} name={storageClass} />
+        </DetailsItem>
+      </FlexItem>
+    </Flex>
+  );
+};
+
+const ProviderDetails: React.FC<ProviderDetailsProps> = ({ resource }) => {
+  const type = resource.spec.type;
```
SanjalKatiyar, line 196: null values

packages/odf/components/resource-pages/Providers.tsx
```diff
@@ -0,0 +1,220 @@
...
+      case NOOBAA_TYPE_MAP[BC_PROVIDERS.IBM]:
+        return IBMDetails;
+      case NOOBAA_TYPE_MAP[BC_PROVIDERS.PVC]:
+        return PVCDetails;
+      case NOOBAA_TYPE_MAP[BC_PROVIDERS.GCP]:
+        return GCPDetails;
+      default:
+        return AWSDetails;
+    }
+  }, [type]);
```
SanjalKatiyar, line 215: we can move it outside React FC.

packages/shared/utils/Conditions.tsx
```diff
@@ -0,0 +1,138 @@
...
+        return status;
+    }
+  };
+
+  const rows = (
+    conditions as Array<K8sResourceCondition | ClusterServiceVersionCondition>
+  )?.map?.(
+    (
+      condition: K8sResourceCondition & ClusterServiceVersionCondition,
+      i: number
```
SanjalKatiyar, line 45: nits:
```suggestion
      index: number
```

SanjalKatiyar: not anything important though

packages/odf/components/resource-pages/BackingStoreDetailsPage.tsx
SanjalKatiyar: similar to how we do for dashboard pages...

bipuladh: It would be the same. If resource, launchModal changes then the memoized value would also change and the same with the callback.

packages/odf/components/resource-pages/Providers.tsx
bipuladh: Rather than doing this I will handle the loading and Error before we render this component. Make sure null values are not passed down.

packages/odf/components/resource-pages/Providers.tsx
bipuladh: We will not be able to memoize. The dependency array value is a string so even the resource might get updated the ProviderComponent will not be updated. Putting it outside we would return a new component every time.

packages/shared/utils/Conditions.tsx
bipuladh: `i`, `j`, `k` are well know imo.

packages/odf/components/resource-pages/BackingStoreDetailsPage.tsx
SanjalKatiyar: ack... but React will atleast now have to keep track of one less hook this way... maybe it will not any significant affect on performance, but we can reduce number of hooks if it is possible IMO :(
earlier two hooks, each for Details & YAMLEditor
now one hook for pages only

packages/odf/components/resource-pages/Providers.tsx
SanjalKatiyar: ack... so u mean it is already handled in parent component ?

packages/odf/components/resource-pages/BackingStoreDetailsPage.tsx
SanjalKatiyar: ack...

packages/odf/components/resource-pages/Providers.tsx
SanjalKatiyar: perfect thanks... yeah did not think of that !!

SanjalKatiyar: ack... thanks... make sense...

packages/odf/components/resource-pages/BackingStoreDetailsPage.tsx
bipuladh: makes sense.

packages/odf/components/resource-pages/Providers.tsx
bipuladh: yeah i am passing loading and loadError to details page to handle it there itself.

packages/odf/components/resource-pages/Providers.tsx
SanjalKatiyar: if we move this out of React FC, then the fucntion reference will remain constant... so if function is same, won't it return the reference of same component as well ??

bipuladh: Function would called everytime `type` is updated. That happens everytime useK8sWatchResource updates it's values. Here the dependency array when compared will be the same so it won't even bother to run the above switch statement.

packages/shared/details-page/DetailsPage.tsx
```diff
@@ -60,17 +65,29 @@ const DetailsPage: React.FC<DetailsPageProps> = ({
...
-      title={
-        <DetailsPageTitle resource={resource} resourceModel={resourceModel} />
-      }
-      actions={actions}
-      resource={resource}
-    />
-    <HorizontalNav pages={pages} resource={resource} />
+    {!loaded && <LoadingBox />}
+    {loaded && loadError && <ErrorPage message={loadError?.message} />}
+    {loaded && !loadError && (
```
SanjalKatiyar, line 74: there could be case when `loaded = true, error = false and data = [] or null`... but not blocking the PR cause it rarely happens...

PR #192, opened 2022-04-22
Migrate BucketClass Details Page

bipuladh:
Screenshot from 2022-04-22 18-13-16


packages/odf/components/resource-pages/BucketClassDetailsPage.tsx
```diff
@@ -0,0 +1,193 @@
...
+      ))}
+    </>
+  );
+};
+
+const NamespacePolicyDetails: React.FC<PolicyDetailsProps> = ({ resource }) => {
+  const { t } = useTranslation();
+  const type = resource.spec.namespacePolicy.type;
+  return <DetailsItem field={t('Policy type')}>{type}</DetailsItem>;
+};
```
SanjalKatiyar, line 60: in current NamespacePolicy section we have other info as well... 
1. like read/write resource etc...
2. And we have info popups as well when we click on the title...
are we skipping them from the designs ?

packages/odf/components/resource-pages/BucketClassDetailsPage.tsx
```diff
@@ -0,0 +1,193 @@
...
+  resource: BucketClassKind;
+};
+
+const PlacementPolicyDetails: React.FC<PolicyDetailsProps> = ({ resource }) => {
+  const { t } = useTranslation();
+  const tiers = resource.spec.placementPolicy.tiers;
+  return (
+    <>
+      {tiers.map((tier) => (
+        <DetailsItem field={t('Tier')} key={tier.placement} padChildElement>
```
SanjalKatiyar, line 43: as per current design shouldn't this be outside the `.map` and "Tier" should be "Tiers" instead ??

packages/odf/components/resource-pages/BucketClassDetailsPage.tsx
bipuladh: Those values are in reality are not being rendered by the parser as well. So I am avoiding them for now.

packages/odf/components/resource-pages/BucketClassDetailsPage.tsx
bipuladh: ack

PR #193, opened 2022-04-22
Migrate NamespaceStore Details Page

bipuladh:


packages/odf/utils/mcg.ts
```diff
@@ -248,8 +248,10 @@ export const getBucketName = (bs: BackingStoreKind): string => {
   return bs.spec?.[PROVIDERS_NOOBAA_MAP[type]]?.[BUCKET_LABEL_NOOBAA_MAP[type]];
 };
 
-export const getRegion = (bs: BackingStoreKind): string => {
-  const type = getBackingStoreType(bs);
+export const getRegion = (
+  bs: BackingStoreKind | NamespaceStoreKind
+): string => {
+  const type = getBackingStoreType(bs as BackingStoreKind);
```
SanjalKatiyar, line 254: this is not specific to BackingStore, right. Can we change to better naming.

PR #194, opened 2022-04-25
Adds generic resource editing page and route

bipuladh:
Adds Edit button to Kebab menu

Screenshot from 2022-04-25 14-47-03
Screenshot from 2022-04-25 14-44-56


packages/odf/components/ResourceEditor/GenericResourceEditor.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
```
GowthamShanmugam, line 1: This component is common for all, move this to share

bipuladh: It is used at a specific Route so I don't want to put it there.

GowthamShanmugam: ok

GowthamShanmugam (approved):

PR #196, opened 2022-04-25
Fix MirrorPeer creation with OCS instead of StorageSystem

GowthamShanmugam:

### DIFF
--- a/packages/shared/utils/storage.ts
+++ b/packages/shared/utils/storage.ts
@@ -1,8 +1,24 @@
-import { ClusterServiceVersionKind } from '@odf/shared/types';
+import { OCSStorageClusterModel } from '@odf/shared/models';
+import {
+  ClusterServiceVersionKind,
+  StorageSystemKind,
+} from '@odf/shared/types';
 import { ODF_OPERATOR } from '../constants';
+import { getGVKLabel } from '../utils/common';
 
 export const getODFCsv = (csvList: ClusterServiceVersionKind[] = []) =>
   csvList.find(
     (csv) =>
       csv?.metadata.name?.substring(0, ODF_OPERATOR.length) === ODF_OPERATOR
+  );
+
+export const getOCSStorageSystem = (ssList: StorageSystemKind[] = []) =>
+  ssList.find(
+    (ss) =>
+      ss?.spec?.kind ===
+      getGVKLabel({
+        kind: OCSStorageClusterModel.kind,
+        apiVersion: OCSStorageClusterModel.apiVersion,
+        apiGroup: OCSStorageClusterModel.apiGroup,
+      })
   );
--- a/packages/mco/components/data-policies/create-dr-policy/create-dr-policy.tsx
+++ b/packages/mco/components/data-policies/create-dr-policy/create-dr-policy.tsx
@@ -10,14 +10,15 @@ import {
 import {
   K8sResourceKind,
   ClusterServiceVersionKind,
-  ListKind,
   StorageSystemKind,
+  ListKind,
 } from '@odf/shared/types';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
 import {
   referenceForGroupVersionKind,
   referenceForModel,
   getODFCsv,
+  getOCSStorageSystem,
 } from '@odf/shared/utils';
 import {
   getAPIVersionForModel,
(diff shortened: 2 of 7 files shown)


packages/shared/utils/storage.ts
```diff
@@ -1,8 +1,26 @@
...
+      ss?.spec?.kind ===
+      getStorageSystemKind({
+        kind: OCSStorageClusterModel.kind,
+        apiVersion: OCSStorageClusterModel.apiVersion,
+        apiGroup: OCSStorageClusterModel.apiGroup,
+      })
+  );
+
+export const getStorageSystemKind = ({ kind, apiVersion, apiGroup }) =>
+  `${kind.toLowerCase()}.${apiGroup}/${apiVersion}`;
```
SanjalKatiyar, line 26: this should go under `common` and function name should generic not specific to StorageSystem.

SanjalKatiyar:
```suggestion
export const getGVKLabel = ({ kind, apiVersion, apiGroup }) =>
  `${kind.toLowerCase()}.${apiGroup}/${apiVersion}`;
```

SanjalKatiyar: and instead of three arguments we can pass `Model` itself to the fucntion.

GowthamShanmugam: If this function need only a three-argument, so no need to use the model as an argument. It is always better to use exact arg instead of passing the whole object.


Also, these three values are non-optional here, so sometimes we need to pass Model and sometime we need to pass K8sModel.  For K8sModel,   apiGroup is optional so we can't pass an entire object to this func args.

packages/shared/utils/storage.ts
```diff
@@ -1,8 +1,26 @@
...
+      ss?.spec?.kind ===
+      getGVKLabel({
+        kind: OCSStorageClusterModel.kind,
+        apiVersion: OCSStorageClusterModel.apiVersion,
+        apiGroup: OCSStorageClusterModel.apiGroup,
+      })
+  );
+
+export const getGVKLabel = ({ kind, apiVersion, apiGroup }) =>
+  `${kind.toLowerCase()}.${apiGroup}/${apiVersion}`;
```
SanjalKatiyar, line 26: can we move it under `utils/common.ts` ??

packages/shared/utils/storage.ts
```diff
@@ -1,8 +1,26 @@
...
+
+export const getOCSStorageSystem = (ssList: StorageSystemKind[] = []) =>
+  ssList.find(
+    (ss) =>
+      ss?.spec?.kind ===
+      getGVKLabel({
+        kind: OCSStorageClusterModel.kind,
+        apiVersion: OCSStorageClusterModel.apiVersion,
+        apiGroup: OCSStorageClusterModel.apiGroup,
+      })
```
SanjalKatiyar, line 22:
```suggestion
      getGVKLabel(OCSStorageClusterModel);
```
this should work as well, right ?
similar to how we have done above...

packages/shared/utils/storage.ts
SanjalKatiyar: ack !!

packages/shared/utils/storage.ts
GowthamShanmugam: ack

packages/shared/utils/storage.ts
GowthamShanmugam: no, it won't, OCSStorageClusterModel has apiGroup optional. but getGVKLabel has apiGroup non-optional.  we need to make apiGroup args optional on getGVKLabel to work but it can't be optional here

GowthamShanmugam:
@SanjalKatiyar Changes done

packages/shared/utils/storage.ts
SanjalKatiyar: OCSStorageClusterModel is a static object, we know that apiGroup will always be present in it...
we can pass like this: getGVKLabel(`OCSStorageClusterModel as any`) instead of adding three extra lines everytime we are calling getGVKLabel fucntion.
wdyt ??

SanjalKatiyar: or we can handle type in a better way...

GowthamShanmugam: I feel its ok, this might be the safe way. If you see here model also created on same way: https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/create-storage-system/external-storage/index.ts#L18

PR #197, opened 2022-04-26
Fix edit action issue

GowthamShanmugam:


packages/shared/kebab/kebab.tsx
```diff
@@ -83,9 +83,9 @@ export const Kebab: React.FC<KebabProps> = ({
       customActionMap?.[actionKey]
         ? customActionMap[actionKey]?.()
         : history.push(
-            `/odf/edit/${referenceForModel(resourceModel)}/${
+            `/k8s/cluster/${referenceForModel(resourceModel)}/${
```
bipuladh, line 86: This only works for non-namespaced resource you need to check if it's namespaced and proceed accordingly.

bipuladh (changes requested):

bipuladh (approved):
/lgtm

PR #203, opened 2022-05-06
Disable ODF BackingStore, BucketClass, and NamespaceStore tabs and OCS BlockPool tab

weirdwiz:
Disables ODF BackingStore, BucketClass, and NamespaceStore tabs (based on MCG_FLAG) and OCS BlockPool tab (based on CEPH_FLAG).
output

### DIFF
--- a/packages/ocs/dashboards/odf-system-dashboard.tsx
+++ b/packages/ocs/dashboards/odf-system-dashboard.tsx
@@ -3,9 +3,11 @@ import { LoadingBox } from '@odf/shared/generic/status-box';
 import PageHeading from '@odf/shared/heading/page-heading';
 import { referenceForModel } from '@odf/shared/utils';
 import Tabs, { TabPage } from '@odf/shared/utils/Tabs';
+import { useFlag } from '@openshift-console/dynamic-plugin-sdk';
 import { useTranslation } from 'react-i18next';
 import { RouteComponentProps } from 'react-router';
 import { match as Match } from 'react-router-dom';
+import { CEPH_FLAG } from '../../odf/features';
 import { BlockPoolListPage } from '../block-pool/BlockPoolListPage';
 import { CephBlockPoolModel } from '../models';
 import OCSSystemDashboard from './ocs-system-dashboard';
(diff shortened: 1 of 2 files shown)


packages/ocs/dashboards/odf-system-dashboard.tsx
```diff
@@ -34,26 +36,24 @@ const ODFSystemDashboard: React.FC<ODFSystemDashboardPageProps> = ({
...
-    setPages(overviewPage);
-  }, [overviewPage]);
+      isCephAvailable
+        ? {
+            title: t('BlockPools'),
+            href: blockPoolHref,
+            component: BlockPoolListPage,
+          }
+        : null,
+    ]);
```
bipuladh, line 55: This will result in a array with value something like `[{overview}, null]`. Most likely the app will break. Instead use a temp variable to store the value. 
Something like: 
```
const pages = {
title: 'Overview,
...
};

if (isCephAvailable) { 
 pages.push(blockPool);
}
setPages(pages)
```

bipuladh, line 55: This will result in a array with value something like `[{overview}, null]`. Most likely the app will break. Instead use a temp variable to store the value. 
Something like: 
```
const pages = {
title: 'Overview,
...
};

if (isCephAvailable) { 
 pages.push(blockPool);
}
setPages(pages)
```

packages/odf/components/odf-dashboard/dashboard.tsx
```diff
@@ -56,6 +57,7 @@ export const ODFDashboard: React.FC = () => {
 
 const ODFDashboardPage: React.FC<ODFDashboardPageProps> = (props) => {
   const { t } = useTranslation('plugin__odf-console');
+  const hasMCG = useFlag(MCG_FLAG);
```
bipuladh, line 60: This value will be undefined for some time. We could put our app on loading while the value is `undefined` same for all values that we are pulling from flags.

bipuladh, line 60: This value will be undefined for some time. We could put our app on loading while the value is `undefined` same for all values that we are pulling from flags.

bipuladh (changes requested):

packages/odf/components/odf-dashboard/dashboard.tsx
```diff
@@ -68,22 +74,34 @@ const ODFDashboardPage: React.FC<ODFDashboardPageProps> = (props) => {
...
-      name: t('Namespace Store'),
-      component: NamespaceStoreListPage,
-    },
-  ];
+  ]);
+
+  React.useEffect(() => {
+    const hasAlreadyAdded = pages.find(
+      (page) => page.name === t('Backing Store')
+    );
```
bipuladh, line 82: You need to test for each of them `Bucket Class` and `Namespace Store`. You could create a temp array and call setPages in the end.

bipuladh (changes requested):

bipuladh (approved):
/lgtm

PR #207, opened 2022-05-11
Bug 2084041: Fix DetailsCard

bipuladh:
Updates the SDK version.
Updates the Internal SDK version.

`DetailItem` was recently deprecated hence it's usage needed to be replaced with the new `OverviewDetailsItem`.
Adds `@openshift-console/plugin-shared` to project deps.
Update `webpack.config.ts` to support compiling `scss` from `plugin-shared`
Update files based on the changes introduced in the SDKs.

Screenshot from 2022-05-11 15-01-18

### DIFF
--- a/packages/ocs/block-pool/BlockPoolDetailsPage.tsx
+++ b/packages/ocs/block-pool/BlockPoolDetailsPage.tsx
@@ -48,7 +48,7 @@ export const BlockPoolDetailsPage: React.FC<BlockPoolDetailsPageProps> = ({
     },
     {
       name: t('StorageSystem details'),
-      path: location.pathname.split(`/${poolName}`)[0],
+      path: location.pathname.split(`/${kind}`)[0],
     },
     {
       name: poolName,
--- a/packages/ocs/dashboards/object-service/details-card/details-card.tsx
+++ b/packages/ocs/dashboards/object-service/details-card/details-card.tsx
@@ -25,10 +25,8 @@ import {
   useK8sWatchResource,
   useFlag,
 } from '@openshift-console/dynamic-plugin-sdk';
-import {
-  DetailsBody,
-  DetailItem,
-} from '@openshift-console/dynamic-plugin-sdk-internal';
+import { DetailsBody } from '@openshift-console/dynamic-plugin-sdk-internal';
+import { OverviewDetailItem as DetailItem } from '@openshift-console/plugin-shared';
 import * as _ from 'lodash';
 import { useTranslation } from 'react-i18next';
 import { Link } from 'react-router-dom';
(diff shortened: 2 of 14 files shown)


webpack.config.ts
```diff
@@ -53,7 +53,10 @@ const config: webpack.Configuration = {
       },
       {
         test: /\.scss$/,
-        exclude: /node_modules/,
+        include: [
+          /node_modules\/@openshift-console\/plugin-shared/,
+          /packages/,
+        ],
```
SanjalKatiyar, line 59: just a doubt: for other dependencies like `patternfly` etc we never compile the scss files... why do we have to do it for `plugin-shared` package ??

bipuladh: Plugin shared does not compile these scss files at build time.

packages/ocs/dashboards/object-service/details-card/details-card.tsx
```diff
@@ -130,10 +127,7 @@ export const ObjectServiceDetailsCard: React.FC<{}> = () => {
           <DetailItem
             key="provider"
             title={t('Provider')}
-            error={
-              !!infrastructureError ||
-              (infrastructure && !infrastructurePlatform)
-            }
+            error={!!infrastructureError || infrastructureError}
```
SanjalKatiyar, line 130: why not only `error={infrastructureError}`

packages/ocs/block-pool/BlockPoolDetailsPage.tsx
```diff
@@ -48,7 +48,7 @@ export const BlockPoolDetailsPage: React.FC<BlockPoolDetailsPageProps> = ({
     },
     {
       name: t('StorageSystem details'),
-      path: location.pathname.split(`/${poolName}`)[0],
+      path: location.pathname.split(`/${kind}`)[0],
```
SanjalKatiyar, line 51: kind is `CephBlockPool` right ?
`.../odf/system/ocs.openshift.io~v1~storagecluster/ocs-storagecluster/ceph.rook.io~v1~CephBlockPool/ocs-storagecluster-cephblockpool` is the dashboard URL... if we split on `kind` it will give wrong path...

SanjalKatiyar, line 51: `.../odf/system/ocs.openshift.io~v1~storagecluster/ocs-storagecluster/ceph.rook.io~v1~CephBlockPool/ocs-storagecluster-cephblockpool`
if we split over `kind` it will give path as `.../odf/system/ocs.openshift.io~v1~storagecluster/ocs-storagecluster/` which is incorrect for the BlockPool list page...

webpack.config.ts
SanjalKatiyar: ack...

SanjalKatiyar: do we compile css for other dependencies while build time ? isn't it already in compiled form and is stored under nodes_modules ?
sorry if it sounds illogical... I am not sure about the process right now... will look it up a bit online as well...

PR #214, opened 2022-05-18
After clicking on Create StorageSystem, it navigates to Storage Systems tab but shows an error message

SanjalKatiyar:


bipuladh:
Does this happen all the time?

packages/odf/components/system-list/odf-system-list.tsx
```diff
@@ -363,17 +363,20 @@ export const StorageSystemListPage: React.FC<StorageSystemListPageProps> = ({
...
     selector,
   });
 
-  const odfCsvName: string = csvLoaded
-    ? csv?.find((item) => item.metadata.name.includes('odf-operator'))?.metadata
-        ?.name
-    : null;
+  const odfCsvName: string =
+    csvLoaded && !csvError
+      ? csv?.find((item) => item?.metadata?.name.includes('odf-operator'))
```
GowthamShanmugam, line 377:
```suggestion
      ? csv?.find((item) => item?.metadata?.name?.includes('odf-operator'))
```

GowthamShanmugam: For extra safety :)

SanjalKatiyar: it should not be an issue IMO, if we are getting metadata then we will get the name as well (as name is required field)...
but I will add it !!

GowthamShanmugam: what if metadata is undefined,  up to item?.metadata? is safe .name.includes will break

SanjalKatiyar: if metadata is undefined this: `metadata?.name` will take care of it...

GowthamShanmugam: if metadata is undefined this: metadata?.name will take care of it, yes, but include will fail. undefined metadata will save the name by ?, now undefined name will break the include

SanjalKatiyar: no it will not fail... if metadata is undefined, `?.` that we have in between `name` and `metadata` will take care of it !
but I have updated the PR anyways...

SanjalKatiyar: it will return undefined exactly at the point it will encounter `undefined` for the first time...
```
var a = undefined;
console.log(a?.name.includes('a'));
```
will work correctly...

GowthamShanmugam: try:
var a = {};
console.log(a?.name.includes('a'));

GowthamShanmugam: yes, that right

SanjalKatiyar: got it... thanks !

SanjalKatiyar:
@bipuladh I think this can be merged as well !

PR #224, opened 2022-05-22
Fix Multiple instances of Object Bucket, Object Bucket Claims under storage section

GowthamShanmugam:

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -72,6 +72,7 @@
       "id": "objectbuckets",
       "section": "storage",
       "name": "Object Buckets",
+      "insertAfter": "volumesnapshotcontents",
       "model": {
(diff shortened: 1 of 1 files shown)


bipuladh:
Let's fix Data Foundation to before PV.

GowthamShanmugam:
image

GowthamShanmugam:
> Let's fix Data Foundation to before PV.

Data Foundation is already has inserBefore pv check
https://github.com/red-hat-storage/odf-console/blob/886140ac6b8b3c4fe652f349f033d96dc4736b7a/console-extensions.json#L145

bipuladh (changes requested):
Please fix order of data foundation nav item as well

PR #237, opened 2022-05-25
Add yaml and events tab for ObjectBucketClaim page

vbnrh:
This commit fixes a regression where the YAML and Events tabs are not
migrated to the ObjectBucketClaim page
Before: 

Screenshot from 2022-05-25 18-38-18

After:

Screenshot from 2022-05-25 18-44-37


packages/odf/components/mcg/ObjectBucketClaim.tsx
```diff
@@ -33,6 +33,10 @@ import { sortable } from '@patternfly/react-table';
 import { MCG_FLAG, RGW_FLAG } from '../../features';
 import { NooBaaObjectBucketModel } from '../../models';
 import { getPhase, isBound, obcStatusFilter } from '../../utils';
+import {
+  YAMLEditorWrapped,
+  EventStreamWrapped,
+} from '../resource-pages/CommonDetails';
```
SanjalKatiyar, line 39: ideally as Gowtham suggested, we should import these from shared...
but let's do it next time we will raise a similar PR...

PR #258, opened 2022-06-06
Bug 2091638: add yaml tab to blockpool

weirdwiz:
Screenshot from 2022-06-06 11-57-34
Screenshot from 2022-06-06 12-57-49


packages/ocs/block-pool/BlockPoolDetailsPage.tsx
```diff
@@ -7,6 +7,7 @@ import { referenceForModel } from '@odf/shared/utils';
 import { useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
 import { useTranslation } from 'react-i18next';
 import { RouteComponentProps, useLocation } from 'react-router-dom';
+import { YAMLEditorWrapped } from '../../odf/components/resource-pages/CommonDetails';
```
SanjalKatiyar, line 10: don't import from odf package

SanjalKatiyar: shift wrapped components to shared... then import from there...

bipuladh (approved):
/lgtm

PR #259, opened 2022-06-06
prevent duplicate blockpool tabs

weirdwiz:
#203 introduces a bug, where duplicate blockpool tabs get added


weirdwiz:
ptal @bipuladh

weirdwiz:
ptal @bipuladh

PR #274, opened 2022-06-10
Fix uploding JSON data for external cluster connection strict checking issue

GowthamShanmugam:
For more info: https://bugzilla.redhat.com/show_bug.cgi?id=2092217

### DIFF
--- a/packages/odf/components/utils/common.ts
+++ b/packages/odf/components/utils/common.ts
@@ -211,9 +211,11 @@ export const checkError = (
   });
 
   // Check for missing keys
-  const missingKeys = _.difference(
-    _.concat(requiredKeys, requiresEncodingKeys),
-    providedKeys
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
image

packages/odf/components/utils/common.ts
```diff
@@ -211,9 +211,11 @@ export const checkError = (
   });
 
   // Check for missing keys
-  const missingKeys = _.difference(
-    _.concat(requiredKeys, requiresEncodingKeys),
-    providedKeys
```
bipuladh, line 216: Could `_.deepFlatten(providedKeys)` work?

GowthamShanmugam: Do you mean _.flattenDeep? 

it is for converting the multidimensional list to a single multidimensional. problem is providedKey keys have some postfix. we can't do direct differences, instead, we need to compare the substring.

bipuladh (approved):

PR #282, opened 2022-06-14
Fix Dashboard crashes in ODF

bipuladh:
`pages` was being used in conjunction with `setPages` inside the same `useEffect` hook which was causing an infinite loop. 

### DIFF
--- a/packages/ocs/dashboards/odf-system-dashboard.tsx
+++ b/packages/ocs/dashboards/odf-system-dashboard.tsx
@@ -47,8 +47,8 @@ const ODFSystemDashboard: React.FC<ODFSystemDashboardPageProps> = ({
   React.useEffect(() => {
     const isBlockPoolAdded = pages.find((page) => page.href === blockPoolHref);
     if (isCephAvailable && !isBlockPoolAdded) {
-      setPages([
-        ...pages,
+      setPages((p) => [
+        ...p,
         {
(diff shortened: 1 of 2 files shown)


packages/ocs/dashboards/odf-system-dashboard.tsx
```diff
@@ -47,16 +47,16 @@ const ODFSystemDashboard: React.FC<ODFSystemDashboardPageProps> = ({
...
+        ...p,
         {
           title: t('BlockPools'),
           href: blockPoolHref,
           component: BlockPoolListPage,
         },
       ]);
     }
-  }, [isCephAvailable, pages, t]);
+  }, [isCephAvailable, JSON.stringify(pages), setPages, t]);
```
SanjalKatiyar, line 59: what if we move `const isBlockPoolAdded = pages.find((page) => page.href === blockPoolHref);` in another useEffect with `pages` as only dependency...
and in this useEffect we pass `isBlockPoolAdded` (boolean) to the dependency array... that way we don't have to use `JSON.stringify` ??

bipuladh: It would be the same thing.

bipuladh: updated as per our discussions.

PR #294, opened 2022-06-16
added the description for StorageClass dropdown in Add Capacity modal

RishabhKodes:
Added the `SecondaryTextGenerator` field in the StorageClassDropdown which provides the description for the storage class headings.

image

Signed-off-by: Rishabh Bhandari <rishabhbhandari6@gmail.com>

### DIFF
--- a/packages/odf/modals/add-capacity/add-capacity-modal.tsx
+++ b/packages/odf/modals/add-capacity/add-capacity-modal.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { getStorageClassDescription } from '@odf/core/utils';
 import ResourceDropdown from '@odf/shared/dropdown/ResourceDropdown';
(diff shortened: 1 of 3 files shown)


packages/odf/modals/add-capacity/add-capacity-modal.tsx
```diff
@@ -73,6 +74,17 @@ type StorageClassDropdownProps = {
...
+  const descriptionList = [sc.provisioner, sc.parameters?.type];
+  return descriptionList.reduce(
+    (description, dl) => {
+      if (dl) return description ? `${description} | ${dl}` : dl;
+      return description;
+    },
+    isDefaultClass(sc) ? '(default)' : ''
+  );
+};
+
```
SanjalKatiyar, line 87: move this to common place as this function is used in other files as well...
and import from there at all place...
if this is common to `/odf` move it to utils for odf...
if common between different packages... them move to shared...

PR #316, opened 2022-06-21
wizard page going blank for KMS

SanjalKatiyar:
Screenshot from 2022-06-21 15-06-44

### DIFF
--- a/packages/odf/components/kms-config/vault-config.tsx
+++ b/packages/odf/components/kms-config/vault-config.tsx
@@ -16,6 +16,7 @@ import {
   ValidatedOptions,
 } from '@patternfly/react-core';
 import { PencilAltIcon } from '@patternfly/react-icons';
+import { VaultEmptyState } from '../../constants';
 import { FEATURES } from '../../features';
 import {
   VaultConfig,
@@ -86,11 +87,13 @@ export const VaultConfigure: React.FC<KMSConfigureProps> = ({
 
   const setAuthMethod = React.useCallback(
     (authMethod: VaultAuthMethods) => {
-      !!vaultStateClone.authMethod && setAuthValue('');
+      if (!!vaultStateClone.authMethod) {
+        vaultStateClone.authValue = VaultEmptyState.authValue;
+      }
       vaultStateClone.authMethod = authMethod;
       updateVaultState(vaultStateClone);
     },
-    [setAuthValue, updateVaultState, vaultStateClone]
+    [updateVaultState, vaultStateClone]
   );
 
   const filteredVaultAuthMethodMapping = React.useMemo(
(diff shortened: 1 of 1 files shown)


packages/odf/components/kms-config/vault-config.tsx
```diff
@@ -86,11 +87,13 @@ export const VaultConfigure: React.FC<KMSConfigureProps> = ({
 
   const setAuthMethod = React.useCallback(
     (authMethod: VaultAuthMethods) => {
-      !!vaultStateClone.authMethod && setAuthValue('');
```
GowthamShanmugam, line 89: Are we sure this is the causing issue?

SanjalKatiyar: no it is extra...

SanjalKatiyar: `encryption` in `ValutConnectionForm` is causing the issue

SanjalKatiyar:
```
no it is extra...
when we change the authmethod we want to reset the authvalue...
so basically go back to initial state...
```
earlier we were just resetting the value to `''` but we want to reset valid as `true` as well (initial state)...

GowthamShanmugam: true? no authValue is string only, the default value is ''

 authValue: {
    value: '',
    valid: true,
  },
  
 you are doing the same.

GowthamShanmugam: and it is missing the authValue validation

SanjalKatiyar: yeah... when we reset we want  to go back to initial state which was, right ??
That is what I am doing...
```
authValue: {
value: '',
valid: true,
},
```

SanjalKatiyar: no authValue is string only, the default value is  --> yes... but `valid` is true by default...

SanjalKatiyar: not required... we are re-setting the state right... so we don't need validation...
we need validation when we type/input the authvalue... we have setAuthValue function for that...
here we are explicitly setting it to `''` to if we add validation it will store authValue.valid as `false` (which is not initial state)

GowthamShanmugam: the issue is when you change auth method, the auth value becomes empty but it still allows you to create the KMS config. Please check.

GowthamShanmugam: mmm no, it will work

GowthamShanmugam (approved):

PR #337, opened 2022-07-01
Bug 2043572: Deleting namespace store in UI secrets remains

TimothyAsirJeyasing:
Random characters added to the secret at the end

### DIFF
--- a/packages/odf/components/namespace-store/namespace-store-form.tsx
+++ b/packages/odf/components/namespace-store/namespace-store-form.tsx
@@ -78,9 +78,10 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
     setProgress(true);
     /** Create a secret if secret ==='' */
     let { secretName } = providerDataState;
+    let m4 = () => Math.floor((1 + Math.random()) * 0x10000).toString(16);
     const promises = [];
     if (!secretName) {
-      secretName = nsName.concat('-secret');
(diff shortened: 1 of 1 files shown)


TimothyAsirJeyasing:
@bipuladh , @GowthamShanmugam , @SanjalKatiyar Please review

packages/shared/modals/DeleteModal.tsx
```diff
@@ -55,6 +61,29 @@ const DeleteModal: React.FC<CommonModalProps<DeleteModalExtraProps>> = ({
       });
   });
 
+  const getSecName = (resource: K8sResourceCommon) => {
+    const getType = (resource) => {
+      switch (resource['spec']['type']) {
```
GowthamShanmugam, line 66: i deleting namespace store should deleted secret using owner ref.

TimothyAsirJeyasing: Secret named with unique

bipuladh (changes requested):
Lets do this only if a secret of that particular name exists. We can handle this if we get an error around secret not being present. So the flow would be. 
1) Try and create a secret without the random characters.
2) If the secret is present then we would get an error. If we get such an error then only append these random characters to the name. We could avoid that by adding more intelligence. We could try checking the content of these two secrets and if it matches we could use the same secret instead of creating but this could be later tackled (the matching of secret content).

PR #340, opened 2022-07-07
DRPolicy related actions are only restricted to authorized users

GowthamShanmugam:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2105262

### DIFF
--- a/locales/en/plugin__odf-console.json
+++ b/locales/en/plugin__odf-console.json
@@ -48,6 +48,7 @@
   "No disaster recovery policies yet": "No disaster recovery policies yet",
   "Configure recovery to your failover cluster with a disaster recovery policy.<1></1>Click the <3>Create DRPolicy</3> button to get started.": "Configure recovery to your failover cluster with a disaster recovery policy.<1></1>Click the <3>Create DRPolicy</3> button to get started.",
   "Empty page logo": "Empty page logo",
+  "You are not authorized to complete this action. See your cluster administrator for role-based access control information.": "You are not authorized to complete this action. See your cluster administrator for role-based access control information.",
   "Storage System": "Storage System",
   "Overview": "Overview",
   "Systems": "Systems",
--- a/packages/mco/components/data-policies/drpolicy-list-page/drpolicy-list-page.tsx
+++ b/packages/mco/components/data-policies/drpolicy-list-page/drpolicy-list-page.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { useAccessReview } from '@odf/shared/hooks/rbac-hook';
 import { Kebab } from '@odf/shared/kebab/kebab';
 import {
   LaunchModal,
@@ -252,51 +253,62 @@ export const DRPolicyListPage: React.FC = () => {
       cluster: HUB_CLUSTER_NAME,
     });
 
+  const [canCreateDRPolicy, isLoading] = useAccessReview(
+    {
+      group: DRPolicyModel?.apiGroup,
+      resource: DRPolicyModel?.plural,
+      namespace: null,
+      verb: 'create',
+    },
+    HUB_CLUSTER_NAME
+  );
+
   const [data, filteredData, onFilterChange] = useListPageFilter(drPolicies);
 
   return (
     <>
-      {drPoliciesLoaded &&
-        (drPolicies.length === 0 || drPoliciesLoadError ? (
-          <EmptyPage
-            title={t('No disaster recovery policies yet')}
-            buttonText={t('Create DRPolicy')}
-            onClick={() => history.push(createProps)}
-          >
-            <Trans t={t}>
-              Configure recovery to your failover cluster with a disaster
-              recovery policy.
-              <br />
-              Click the <strong>Create DRPolicy</strong> button to get started.
-            </Trans>
-          </EmptyPage>
-        ) : (
-          <>
-            <ModalComponent {...props} />
-            <ListPageBody>
-              <div className="mco-drpolicy-list__header">
-                <ListPageFilter
-                  data={data}
-                  loaded={drPoliciesLoaded}
-                  onFilterChange={onFilterChange}
-                  hideColumnManagement={true}
-                />
-                <div className="mco-drpolicy-list__createlink">
-                  <ListPageCreateLink to={createProps}>
-                    {t('Create DRPolicy')}
-                  </ListPageCreateLink>
-                </div>
-              </div>
-              <DRPolicyList
-                data={filteredData as DRPolicyKind[]}
-                unfilteredData={drPolicies}
+      {drPolicies?.length === 0 || (!canCreateDRPolicy && !isLoading) ? (
+        <EmptyPage
+          title={t('No disaster recovery policies yet')}
+          buttonText={t('Create DRPolicy')}
+          canCreate={canCreateDRPolicy}
+          t={t}
+          onClick={() => history.push(createProps)}
+        >
+          <Trans t={t}>
+            Configure recovery to your failover cluster with a disaster recovery
+            policy.
+            <br />
+            Click the <strong>Create DRPolicy</strong> button to get started.
+          </Trans>
+        </EmptyPage>
+      ) : (
+        <>
+          <ModalComponent {...props} />
+          <ListPageBody>
+            <div className="mco-drpolicy-list__header">
+              <ListPageFilter
+                data={data}
                 loaded={drPoliciesLoaded}
-                loadError={drPoliciesLoadError}
-                rowData={{ launchModal }}
+                onFilterChange={onFilterChange}
+                hideColumnManagement={true}
               />
-            </ListPageBody>
-          </>
-        ))}
+              <div className="mco-drpolicy-list__createlink">
+                <ListPageCreateLink to={createProps}>
+                  {t('Create DRPolicy')}
+                </ListPageCreateLink>
+              </div>
+            </div>
+            <DRPolicyList
+              data={filteredData as DRPolicyKind[]}
+              unfilteredData={drPolicies}
+              loaded={drPoliciesLoaded && !isLoading}
+              loadError={drPoliciesLoadError}
+              rowData={{ launchModal }}
+            />
+          </ListPageBody>
+        </>
+      )}
     </>
   );
 };
--- a/packages/mco/components/empty-state-page/empty-page.tsx
+++ b/packages/mco/components/empty-state-page/empty-page.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
+import { TFunction } from 'i18next';
 import {
   Button,
   Title,
@@ -7,6 +8,7 @@ import {
   EmptyStateVariant,
   EmptyStateIcon,
   EmptyStateBody,
+  Tooltip,
 } from '@patternfly/react-core';
 import './empty-page.scss';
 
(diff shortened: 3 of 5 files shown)


GowthamShanmugam:
Screenshot from 2022-07-07 23-04-37

locales/en/plugin__odf-console.json
```diff
@@ -48,6 +48,7 @@
   "No disaster recovery policies yet": "No disaster recovery policies yet",
   "Configure recovery to your failover cluster with a disaster recovery policy.<1></1>Click the <3>Create DRPolicy</3> button to get started.": "Configure recovery to your failover cluster with a disaster recovery policy.<1></1>Click the <3>Create DRPolicy</3> button to get started.",
   "Empty page logo": "Empty page logo",
+  "You are not authorized to complete this action. See your cluster administrator for role-based access control information.": "You are not authorized to complete this action. See your cluster administrator for role-based access control information.",
```
GowthamShanmugam, line 51: Erorr message is copied from ACM

packages/mco/components/empty-state-page/empty-page.tsx
```diff
@@ -23,21 +25,37 @@ const EmptyPageIcon: React.FC = () => {
...
-          data-test="create-button"
+        <EmptyStateBody>{children}</EmptyStateBody>
+        <Tooltip
+          content={
+            !canCreate
+              ? t(
+                  'You are not authorized to complete this action. See your cluster administrator for role-based access control information.'
+                )
+              : ''
+          }
```
SanjalKatiyar, line 44: nits:
```suggestion
          content={
            !canCreate && t(
                  'You are not authorized to complete this action. See your cluster administrator for role-based access control information.'
                )
          }
```

packages/mco/components/data-policies/drpolicy-list-page/drpolicy-list-page.tsx
```diff
@@ -252,51 +253,59 @@ export const DRPolicyListPage: React.FC = () => {
...
-                <div className="mco-drpolicy-list__createlink">
-                  <ListPageCreateLink to={createProps}>
-                    {t('Create DRPolicy')}
-                  </ListPageCreateLink>
-                </div>
-              </div>
-              <DRPolicyList
-                data={filteredData as DRPolicyKind[]}
-                unfilteredData={drPolicies}
+      {drPolicies.length === 0 || !canCreateDRPolicy ? (
```
SanjalKatiyar, line 267:
```suggestion
      {drPolicies?.length === 0 || !canCreateDRPolicy ? (
```

packages/mco/components/data-policies/drpolicy-list-page/drpolicy-list-page.tsx
```diff
@@ -252,51 +253,59 @@ export const DRPolicyListPage: React.FC = () => {
...
+              <div className="mco-drpolicy-list__createlink">
+                <ListPageCreateLink to={createProps}>
+                  {t('Create DRPolicy')}
+                </ListPageCreateLink>
+              </div>
+            </div>
+            <DRPolicyList
+              data={filteredData as DRPolicyKind[]}
+              unfilteredData={drPolicies}
+              loaded={drPoliciesLoaded}
```
SanjalKatiyar, line 302:
```suggestion
              loaded={drPoliciesLoaded && !canCreateDRPolicyLoading}
```

packages/mco/components/data-policies/drpolicy-list-page/drpolicy-list-page.tsx
```diff
@@ -252,51 +253,59 @@ export const DRPolicyListPage: React.FC = () => {
       cluster: HUB_CLUSTER_NAME,
     });
 
+  const canCreateDRPolicy = useAccessReviewAllowed({
```
SanjalKatiyar, line 256: this returns loading state as well... we can pass it in `DRPolicyList`...
```suggestion
  const [canCreateDRPolicy, canCreateDRPolicyLoading] = useAccessReview({
```

SanjalKatiyar, line 256:
```suggestion
  const [canCreateDRPolicy, isLoading] = useAccessReviewAllowed({
```

packages/mco/components/data-policies/drpolicy-list-page/drpolicy-list-page.tsx
SanjalKatiyar, line 302:
```suggestion
              loaded={drPoliciesLoaded && !isLoading}
```

packages/mco/components/data-policies/drpolicy-list-page/drpolicy-list-page.tsx
GowthamShanmugam: useAccessReviewAllowed is not taking any cluster name, so the cluster context switch on the hybrid console will affect this. We need to copy useAccessReviewAllowed to odf-console with target cluster name.

SanjalKatiyar: so we should do that if that is the case !

SanjalKatiyar: we did it here https://github.com/red-hat-storage/mcg-ms-console/blob/master/packages/utils/hooks/rbac.tsx for mcg-ms-console... all u need to pass is cluster name as well...

SanjalKatiyar: does it matter which cluster we are into right now ??
all we are checking if user has given permission or not...

SanjalKatiyar: we copied it as well for mcg-ms-console https://github.com/red-hat-storage/mcg-ms-console/blob/master/packages/utils/hooks/rbac.tsx, but we don't need cluster name in our case...

GowthamShanmugam: We dont need loading, canCreateDRPolicy will be true only when access is loaded successfully

packages/mco/components/data-policies/drpolicy-list-page/drpolicy-list-page.tsx
GowthamShanmugam: drPolicies?.length === 0 then !isLoading dont have any value here, it always displays the custom Empty page with the button disabled.

I feel its ok to skip isLoading from here.

packages/mco/components/data-policies/drpolicy-list-page/drpolicy-list-page.tsx
SanjalKatiyar: yes... but that time it will be false... and `!canCreateDRPolicy` condition can be true which can show `EmptyPage` for a brief moment before showing the list page...
this may result in some flickering effect on the UI...

SanjalKatiyar: edge case !

GowthamShanmugam: We need a cluster name because DR UI is in the Hybrid console. All CRUD operations on OCP will be done using the correct active cluster context.

But we need to makesure all DR CRUD operations are happening on the local-cluster only. So we need to pass the Hub cluster name as a target cluster in all OCP CRUD operation API calls.

GowthamShanmugam: "but that time it will be false... and !canCreateDRPolicy condition can be true"
   drPolicies?.length === 0 || (!isLoading && !canCreateDRPolicy) cornor case is still possible when  drPolicies?.length ===0 and isLoading is false and !canCreateDRPolicy is true.
  
  !isLoading has no effect here.

GowthamShanmugam: drPolicies?.length === 0 && (!isLoading && !canCreateDRPolicy) may be this will avoid but i feel its ok to display unaythorized message when UI unable to figureout the access instead of keep loading.

SanjalKatiyar: drPolicies?.length === 0 means that our data is loaded just it resulted in empty list `[]`
for the case if data is not loaded (undefined), means drPolicies?.length !== 0 and api call for `canCreateDRPolicy` has still not completed, it will show `EmptyState` component unless we use `!isLoading`...
when we use `!isLoading` then at that duration it will show list page component and in list page component we are passing `loaded={drPoliciesLoaded && !isLoading}` so it will display some loading state by default by the time api call for `canCreateDRPolicy` completes...

SanjalKatiyar: we can skip `isloading` as well... it is not anything severe... just good to have thing !!

GowthamShanmugam: In the green case, I am using the empty page as a loading page. Because I feel too much loading one for policies and one for authorization check.

SanjalKatiyar: ok

SanjalKatiyar: got it !!

GowthamShanmugam: i dont see much loading, so i done added your suggestion

packages/mco/components/empty-state-page/empty-page.tsx
```diff
@@ -23,21 +25,36 @@ const EmptyPageIcon: React.FC = () => {
...
-          data-test="create-button"
+        <EmptyStateBody>{children}</EmptyStateBody>
+        <Tooltip
+          content={
+            !canCreate &&
+            t(
+              'You are not authorized to complete this action. See your cluster administrator for role-based access control information.'
+            )
+          }
+          trigger={!canCreate ? 'mouseenter' : 'manual'}
```
GowthamShanmugam, line 44: When canCreate is true tooltip won't be displayed without isVisible.

GowthamShanmugam: isVisible false only work when trigger is manual

GowthamShanmugam:
@SanjalKatiyar changes done

packages/mco/components/data-policies/drpolicy-list-page/drpolicy-list-page.tsx
```diff
@@ -252,51 +252,62 @@ export const DRPolicyListPage: React.FC = () => {
...
-                <div className="mco-drpolicy-list__createlink">
-                  <ListPageCreateLink to={createProps}>
-                    {t('Create DRPolicy')}
-                  </ListPageCreateLink>
-                </div>
-              </div>
-              <DRPolicyList
-                data={filteredData as DRPolicyKind[]}
-                unfilteredData={drPolicies}
+      {drPolicies.length === 0 || (!canCreateDRPolicy && !isLoading) ? (
```
SanjalKatiyar, line 269:
```suggestion
      {drPolicies?.length === 0 || (!canCreateDRPolicy && !isLoading) ? (
```

packages/shared/models/console-models.ts
```diff
@@ -62,3 +62,13 @@ export const ProjectModel: K8sKind = {
...
+
+export const SelfSubjectAccessReviewModel: K8sKind = {
+  abbr: 'SSAR',
+  kind: 'SelfSubjectAccessReview',
+  label: 'SelfSubjectAccessReview',
+  labelPlural: 'SelfSubjectAccessReviews',
+  plural: 'selfsubjectaccessreviews',
+  apiVersion: 'v1',
+  apiGroup: 'authorization.k8s.io',
+};
```
SanjalKatiyar, line 74: it is K8s api... should not be under `console-models`

packages/shared/utils/rbac.ts
```diff
@@ -0,0 +1,135 @@
...
+ * @param resourceAttributes resource attributes for access review
+ * @param impersonate impersonation details
+ * @returns The isAllowed boolean value.
+ */
+export const useAccessReviewAllowed = (
+  resourceAttributes: AccessReviewResourceAttributes,
+  cluster?: string
+): boolean => {
+  return useAccessReview(resourceAttributes, cluster)[0];
+};
```
SanjalKatiyar, line 135: under `shared/hooks`

packages/shared/models/console-models.ts
GowthamShanmugam: ack

GowthamShanmugam:
@SanjalKatiyar Changes done

PR #350, opened 2022-07-28
Bug 2091967: add description to storageclass dropdown

weirdwiz:
Screenshot from 2022-08-01 11-20-01

Bugzilla : https://bugzilla.redhat.com/show_bug.cgi?id=2091967

### DIFF
--- a/packages/odf/components/mcg/CreateObjectBucketClaim.tsx
+++ b/packages/odf/components/mcg/CreateObjectBucketClaim.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { getStorageClassDescription } from '@odf/core/utils';
 import ResourceDropdown from '@odf/shared/dropdown/ResourceDropdown';
 import ResourcesDropdown from '@odf/shared/dropdown/ResourceDropdown';
 import { ButtonBar } from '@odf/shared/generic/ButtonBar';
@@ -134,6 +135,7 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
               id="sc-dropdown"
               data-test="sc-dropdown"
               resource={storageClassResource}
+              secondaryTextGenerator={getStorageClassDescription}
             />
             <p className="help-block">
               {t(
--- a/packages/odf/utils/odf.ts
+++ b/packages/odf/utils/odf.ts
@@ -25,14 +25,15 @@ export const getExternalSubSystemName = (
 export const getStorageClassDescription = (
   sc: StorageClassResourceKind
 ): string => {
-  const descriptionList = [sc.provisioner, sc.parameters?.type];
-  return descriptionList.reduce(
-    (description, dl) => {
-      if (dl) return description ? `${description} | ${dl}` : dl;
-      return description;
-    },
-    isDefaultClass(sc) ? '(default)' : ''
-  );
+  const storageClassProperties = [
+    isDefaultClass(sc) ? '(default)' : '',
+    sc.metadata?.annotations?.['description'],
+    sc.metadata?.annotations?.['storage.alpha.openshift.io/access-mode'],
+    sc.provisioner,
+    sc.parameters?.type,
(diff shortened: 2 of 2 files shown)


packages/shared/dropdown/ResourceDropdown.tsx
```diff
@@ -61,10 +62,16 @@ const ResourceDropdownItem: ResourceDropdownItem = ({
   id,
   secondaryTextGenerator,
   onClick,
+  description,
```
SanjalKatiyar, line 65: can u try `secondaryTextGenerator` ?
most probably that will add description only, we don't need new parameter.

SanjalKatiyar: u just have to pass the function directly...

SanjalKatiyar: no changes in `ResourceDropdown.tsx`.

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -39,6 +39,34 @@ const objectStorageProvisioners = [
...
+      ['metadata', 'annotations', 'storage.alpha.openshift.io/access-mode'],
+      ''
+    ),
+    _.get(sc, 'provisioner'),
+    _.get(sc, 'parameters.type', ''),
+    _.get(sc, 'parameters.zone', ''),
+  ];
+  return _.compact(storageClassProperties).join(' | ');
+};
+
```
SanjalKatiyar, line 69: not needed... already present...
https://github.com/red-hat-storage/odf-console/blob/c055ca23e07159e09c965fac2d0610ed1375ca4a/packages/odf/utils/odf.ts#L25

packages/odf/utils/odf.ts
```diff
@@ -25,14 +25,15 @@ export const getExternalSubSystemName = (
...
   sc: StorageClassResourceKind
 ): string => {
-  const descriptionList = [sc.provisioner, sc.parameters?.type];
-  return descriptionList.reduce(
-    (description, dl) => {
-      if (dl) return description ? `${description} | ${dl}` : dl;
-      return description;
-    },
-    isDefaultClass(sc) ? '(default)' : ''
-  );
```
SanjalKatiyar, line 35: this function is used at other places as well... we can't update it directly...
changes in this PR should not affect other places :)
u can update the function such that, based upon where we are using it... we can append `descriptionList` to add more parameters...

weirdwiz: The changes at other places, was how the description was originally displayed.

Eg. in 4.10: 
Screenshot from 2022-08-16 16-44-55

Eg in the current PR: 
Screenshot from 2022-08-16 16-51-48

SanjalKatiyar: What I am asking is why do we need these extra info at all places ??  Above 2 screenshots are same because right now CR do not have extra fields/info that we have added to `descriptionList`.
So is it a requirement in the BZ or default console's functionality is same as above ??

weirdwiz: The default console has the same implementation, that is why i changed the original function.
https://github.com/openshift/console/blob/84960bbb5c447188730e75437e162b7d8ad56411/frontend/public/components/utils/storage-class-dropdown.tsx#L223

SanjalKatiyar: cool !!

SanjalKatiyar:
/retitle Bug 2091967: add description to storageclass dropdown

packages/odf/utils/odf.ts
```diff
@@ -25,14 +25,15 @@ export const getExternalSubSystemName = (
...
+  const storageClassProperties = [
+    isDefaultClass(sc) ? ' (default)' : '',
```
SanjalKatiyar, line 29:
```suggestion
    isDefaultClass(sc) ? '(default)' : '',
```

PR #353, opened 2022-08-02
Bug 2104492: Fix the capacity queries

aruniiird:
Capacity queries are using 'kubelet_volume_stats_used_bytes', which
gives the result in a per node basis (evenwhen the nodes are sharing
a single PVC).
Corrected the query to give only one result per <namespace>+<PVC>
combination.

Signed-off-by: Arun Kumar Mohan <amohan@redhat.com>


SanjalKatiyar:
/retitle Bug 2104492: Fix the capacity queries

PR #355, opened 2022-08-02
Data Foundation tab should not be blocked on StorageCluster presence

SanjalKatiyar:
**Before:**
Screenshot from 2022-08-02 21-10-38

**After:**
Screenshot from 2022-08-02 21-11-28

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -148,7 +148,7 @@
       "href": "/odf/cluster"
     },
     "flags": {
-      "required": ["ODF_MODEL", "OCS"],
+      "required": ["ODF_MODEL"],
       "disallowed": ["MCO"]
     }
   },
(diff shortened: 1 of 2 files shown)


console-extensions.json
```diff
@@ -148,7 +148,7 @@
       "href": "/odf/cluster"
     },
     "flags": {
-      "required": ["ODF_MODEL", "OCS"],
+      "required": ["ODF_MODEL"],
```
bipuladh, line 151: We need to fix the issue around unprivileged user as well. Apparently this fix is not enough.

SanjalKatiyar: yeah but that is a different bug. Do we want to fix 2 different bugs in same PR ?

SanjalKatiyar: this PR was for solving regression... but if u want I can add fix for other bug in same PR as well...

PR #360, opened 2022-08-03
Bug 2089446 - Handle leftover rook-ceph-external-cluster-details secret

TimothyAsirJeyasing:
Resolves bz#2089446

### DIFF
--- a/packages/odf/components/create-storage-system/footer.tsx
+++ b/packages/odf/components/create-storage-system/footer.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { secretResource } from '@odf/core/resources';
 import { OCSStorageClusterModel } from '@odf/shared/models';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
 import { getGVKLabel } from '@odf/shared/utils';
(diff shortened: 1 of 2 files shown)


TimothyAsirJeyasing:
@bipuladh , @GowthamShanmugam Please review

packages/odf/components/create-storage-system/footer.tsx
```diff
@@ -169,6 +171,7 @@ const handleReviewAndCreateNext = async (
         storageClass.name
       );
 
+      await deleteSecret(secretResource.name);
```
bipuladh, line 174: Why are we deleting secret without checking if the secret actually exists? If the secret exists could we patch it and update it instead of deleting it?

bipuladh (changes requested):

packages/odf/components/create-storage-system/footer.tsx
TimothyAsirJeyasing: deleteSecret functionn is checking the secret availability. Do you mean we will rename the existing one and let the flow creates new one.

SanjalKatiyar:
/close
closing for now as https://bugzilla.redhat.com/show_bug.cgi?id=2089446#c6

PR #379, opened 2022-08-19
Migrate console alert actions

nik-redhat:
Migrated the console alert actions from openshift/console.

Signed-off-by: nik-redhat <nladha@redhat.com>

### DIFF
--- a/packages/odf/constants/common.ts
+++ b/packages/odf/constants/common.ts
@@ -2,6 +2,7 @@ import { Toleration, Taint } from '@odf/shared/types';
 import { TFunction } from 'i18next';
 
 export const CEPH_BRAND_NAME = 'Red Hat Ceph Storage';
+export const CEPH_STORAGE_NAMESPACE = 'openshift-storage';
 export const NO_PROVISIONER = 'kubernetes.io/no-provisioner';
 export const STORAGE_CLUSTER_SYSTEM_KIND = 'storagecluster.ocs.openshift.io/v1';
 export const OCS_EXTERNAL_CR_NAME = 'ocs-external-storagecluster';
--- a/packages/odf/modals/add-capacity/cluster-expansion-modal.tsx
+++ b/packages/odf/modals/add-capacity/cluster-expansion-modal.tsx
@@ -1,0 +1,37 @@
+import * as React from 'react';
+import { useModalLauncher } from '@odf/shared/modals/modalLauncher';
+import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
+import { K8sModel } from '@openshift-console/dynamic-plugin-sdk/lib/api/common-types';
+
+const ALERT_MODAL_KEY = 'ALERT_MODAL';
+
+const extraMap = {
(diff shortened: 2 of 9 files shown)


bipuladh:
Could you please attach a recording of the migrated feature?

nik-redhat:
> Could you please attach a recording of the migrated feature?

Sure

packages/odf/models/ocs.ts
```diff
@@ -109,3 +109,16 @@ export const NooBaaObjectBucketModel: K8sKind = {
...
+  labelPlural: 'Storage Clusters',
+  apiVersion: 'v1',
+  apiGroup: 'ocs.openshift.io',
+  plural: 'storageclusters',
+  abbr: 'OCS',
+  namespaced: true,
+  kind: 'StorageCluster',
+  id: 'ocscluster',
+  crd: true,
+};
```
SanjalKatiyar, line 124: already there: https://github.com/red-hat-storage/odf-console/blob/19d26b1ee2d26bd7d854482f68aeae9e9a94d7d9/packages/shared/models/storage.ts#L15

packages/odf/constants/common.ts
```diff
@@ -2,6 +2,7 @@ import { Toleration, Taint } from '@odf/shared/types';
 import { TFunction } from 'i18next';
 
 export const CEPH_BRAND_NAME = 'Red Hat Ceph Storage';
+export const CEPH_STORAGE_NAMESPACE = 'openshift-storage';
```
SanjalKatiyar, line 5: already there: https://github.com/red-hat-storage/odf-console/blob/19d26b1ee2d26bd7d854482f68aeae9e9a94d7d9/packages/shared/constants/common.ts#L3

packages/odf/modals/add-capacity/cluster-expansion-modal.tsx
```diff
@@ -0,0 +1,37 @@
...
+}) => {
+  const [ModalComponent, props, launcher] = useModalLauncher(extraMap);
+
+  return (
+    <>
+      <ModalComponent {...props} />
+      {launcher(key, extraProps)}
+    </>
+  );
+};
```
SanjalKatiyar, line 37: won't this cause infinite rendering ??
`launcher` is called each time FC is rendering, and each time `launcher` is called it will set some state cause another re render of the FC ??

nik-redhat: I doubt, but will have to check.
Though the issue is I am not able to understand ho test the changes :confused: . I have the cluster up and running with the odf-console added dynamically on 9001, then how to proceed? Can you please help me with that?

packages/odf/models/ocs.ts
nik-redhat: But this is `OCSServiceModel` :thinking:

nik-redhat:
So, I did try running locally and the alert seems to be fired from the odf-console. But, for some reason the modal is not getting triggered, I tried a lot to figure out the issue but couldn't get to the root of it. Can someone please help here?
TIA :)

PR #384, opened 2022-08-22
[release-4.9-compatibility] Bug 2116264: Add admin flag (for hiding Data Foundation navItem from non-admin users)

SanjalKatiyar:


src/features.ts
```diff
@@ -1,5 +1,76 @@
...
+export const detectSSAR = (setFlag: SetFeatureFlag) => {
+    const ssar = {
+        apiVersion: 'authorization.k8s.io/v1',
+        kind: 'SelfSubjectAccessReview',
+    };
+    const ssarDetectors: FeatureDetector[] = ssarChecks.map((ssarObj) => {
+        const fn = async (setFlag: SetFeatureFlag) => {
+        try {
+            ssar['spec'] = { resourceAttributes: ssarObj.resourceAttributes };
+            const result: K8sResourceKind = (await consoleFetchJSON.post(
```
SanjalKatiyar, line 61: `k8sCreate` is not exposed by sdk `0.0.2`, either we can update the sdk or use `consoleFetchJSON`

PR #389, opened 2022-08-23
Integration with Thales enterprise key management

SanjalKatiyar:
Screenshot from 2022-08-25 21-46-00
Screenshot from 2022-08-25 21-47-07


SanjalKatiyar:
Screenshot from 2022-08-25 21-30-43
Screenshot from 2022-08-25 21-30-54
Screenshot from 2022-08-25 21-31-06

SanjalKatiyar:
@GowthamShanmugam @bipuladh PTAL.

packages/odf/components/kms-config/kms-config.tsx
```diff
@@ -58,6 +58,20 @@ export const KMSConfigure: React.FC<KMSConfigureProps> = ({
     (provider) => provider.value === kmsProvider
   );
 
+  const setKMSProvider = React.useCallback(
+    (provider: ProviderNames) => {
+      dispatch({
```
GowthamShanmugam, line 63: Do we really need two different states here? can we make it one

GowthamShanmugam: Also, why you are re-factored again? this might leads to new regressions.  I dont think its a good idea to refactor KMS again.

packages/odf/components/kms-config/name-address-port.tsx
```diff
@@ -0,0 +1,133 @@
+import * as React from 'react';
```
GowthamShanmugam, line 1: This design is not scalable, if new KMS comes with a name and without an address and port it leads to re-factor again, I feel it's ok to keep it redundant between different KMS.

packages/odf/components/kms-config/kms-config.tsx
SanjalKatiyar: we will test it thoroughly, this refactor is required.
currently we have one big state for KMS, irrespective of what provider user is choosing, object will always contain state for all the providers. Now, after refactor object will contain the state of only selected KMS, hence reducing state's size.

packages/odf/components/kms-config/thales-config.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+  return (
+    <>
+      <NameAddrPort
+        className={className}
+        kmsState={thalesState}
+        kmsStateClone={thalesStateClone}
+        updateKmsState={updateThalesState}
+      />
+      {uploadInputItems(t).map((uploadItem: UploadInputItem) => (
```
GowthamShanmugam, line 173: I prefer to move this under an advanced Thales configured file, but keep these as mandatory fields. Design-wise no changes, just moving this to keep a similar file structure to other KMS file structures. If we want to support HTTP way on the future then changes are effortless.

packages/odf/types/kms.ts
```diff
@@ -183,6 +225,6 @@ export const VaultAuthMethodMapping: {
   },
 };
 
-export type KMSConfigMap = VaultConfigMap | HpcsConfigMap;
-
-export type KMSConfiguration = VaultConfig | HpcsConfig;
+export type KMSConfigMap = VaultConfigMap | HpcsConfigMap | ThalesConfigMap;
+export type KMSConfiguration = VaultConfig | HpcsConfig | ThalesConfig;
+export type CommonKMSConfig = VaultConfig | ThalesConfig;
```
GowthamShanmugam, line 230: What is CommonKMSConfig? you can use KMSConfiguration right? why we need this

packages/odf/components/kms-config/name-address-port.tsx
SanjalKatiyar: it is a minor refactor...
we can pass new prop to this component if that happens... or for that new KMS, instead of using this NameAddressPort we can create new component for it depending upon need... 
but NameAddressPort component is shared/common between Thales and Vault both, so IMO make sense not repeat same code twice:)

packages/odf/types/kms.ts
```diff
@@ -53,8 +51,7 @@ export enum KmsEncryptionLevel {
 }
 
 export type KMSConfig = {
-  [ProviderNames.VAULT]: VaultConfig;
-  [ProviderNames.HPCS]: HpcsConfig;
+  providerState: VaultConfig | HpcsConfig | ThalesConfig;
```
GowthamShanmugam, line 54:
```suggestion
  providerState: KMSConfiguration
```

packages/odf/components/kms-config/name-address-port.tsx
SanjalKatiyar: also, that is why I did not add this component under `shared` because this will not be used at many places... but atleast it is used twice, so keeping it under `kms-config` as a common component...

packages/odf/types/kms.ts
```diff
@@ -115,6 +112,42 @@ export type HpcsConfig = {
...
+  caCert: {
+    value: string;
+    fileName: string;
+    error: string;
+  };
+  clientKey: {
+    value: string;
+    fileName: string;
+    error: string;
+  };
```
GowthamShanmugam, line 142: you can group this to KMSTLSConfig? these fields are always common for the KMS provider which are all supporting https.

packages/odf/types/kms.ts
SanjalKatiyar: `CommonKMSConfig` is used in `NameAddrPort` component thus only supporting Vault and Thales... whereas `KMSConfiguration` is for all the providers...

packages/odf/types/kms.ts
```diff
@@ -15,8 +16,11 @@ export enum KmsImplementations {
   VAULT_TOKENS = 'vaulttokens', // used by ceph-csi for token-based vault
   VAULT_TENANT_SA = 'vaulttenantsa', // used by ceph-csi for tenant service account
   IBM_KEY_PROTECT = 'ibmkeyprotect', // used by both rook & ceph-csi
+  KMIP = 'kmip', // used by both rook & ceph-csi
```
GowthamShanmugam, line 19: KMIP?

packages/odf/components/kms-config/utils.tsx
```diff
@@ -443,27 +545,44 @@ export const kmsConfigValidation = (
     case ProviderNames.VAULT: {
       const kmsObj = kms as VaultConfig;
       return (
-        kmsObj.name?.valid &&
-        kmsObj.address?.valid &&
-        kmsObj.port?.valid &&
-        kmsObj.name?.value !== '' &&
-        kmsObj.address?.value !== '' &&
-        kmsObj.port?.value !== ''
+        kmsObj.name.valid &&
```
GowthamShanmugam, line 548: are we sure about removing this optional param? I hope it won't break when QE changes anything directly  on  existing configMap or creating new one.

packages/odf/components/kms-config/utils.tsx
```diff
@@ -362,6 +423,32 @@ const createClusterHpcsResources = (
...
+  kms: ThalesConfig,
+  clusterKmsResources: Promise<K8sResourceKind>[]
+) => {
+  const keySecret: SecretKind = generateThalesSecret(kms);
+
+  const parsedAddress = parseURL(kms.address.value);
+  const configData: ThalesConfigMap = {
+    KMS_PROVIDER: KmsImplementations.KMIP,
+    KMS_SERVICE_NAME: kms.name.value,
+    KMIP_ENDPOINT: `${`${parsedAddress.protocol}//${parsedAddress.hostname}`}:${
```
GowthamShanmugam, line 436: This is used on more than one placed, create a utils function for this

packages/odf/components/kms-config/utils.tsx
```diff
@@ -362,6 +423,32 @@ const createClusterHpcsResources = (
...
+    TLS_SERVER_NAME: kms.tls,
+  };
+  const configMapObj: ConfigMapKind = generateOcsKmsConfigMap(configData);
+
+  clusterKmsResources.push(k8sCreate({ model: SecretModel, data: keySecret }));
+  clusterKmsResources.push(
+    k8sCreate({ model: ConfigMapModel, data: configMapObj })
+  );
+
+  return keySecret.metadata.name;
```
GowthamShanmugam, line 449: getName(keySecret)

packages/odf/components/kms-config/utils.tsx
```diff
@@ -290,6 +308,49 @@ const createCsiHpcsResources = (
...
+  }
+
+  const parsedAddress = parseURL(kms.address.value);
+  const csiConfigData: ThalesConfigMap = {
+    KMS_PROVIDER: KmsImplementations.KMIP,
+    KMS_SERVICE_NAME: kms.name.value,
+    KMIP_ENDPOINT: `${`${parsedAddress.protocol}//${parsedAddress.hostname}`}:${
+      kms.port.value
+    }`,
+    KMIP_SECRET_NAME: secretName || keySecret.metadata.name,
```
GowthamShanmugam, line 332: getName(keySecret)

packages/odf/components/kms-config/utils.tsx
```diff
@@ -290,6 +308,49 @@ const createCsiHpcsResources = (
...
+  update: boolean,
+  secretName = ''
+) => {
+  let keySecret: SecretKind;
+  if (!secretName) {
+    // not required, while setting up storage cluster.
+    // required, while creating new storage class.
+    keySecret = generateThalesSecret(kms);
+    csiKmsResources.push(k8sCreate({ model: SecretModel, data: keySecret }));
+  }
```
GowthamShanmugam, line 323:
```suggestion

```

packages/odf/components/kms-config/utils.tsx
bipuladh, line 436:
```suggestion
    KMIP_ENDPOINT: `${parsedAddress.protocol}//${parsedAddress.hostname}}:${
```
why not?

packages/odf/components/kms-config/utils.tsx
```diff
@@ -290,6 +308,49 @@ const createCsiHpcsResources = (
...
+    const cmPatch = [
+      generateConfigMapPatch(kms.name.value, csiConfigData, false),
+    ];
+    csiKmsResources.push(
+      k8sPatch({ model: ConfigMapModel, resource: csiConfigObj, data: cmPatch })
+    );
+  } else {
+    csiKmsResources.push(
+      k8sCreate({ model: ConfigMapModel, data: csiConfigObj })
+    );
```
GowthamShanmugam, line 350:
```suggestion
    );
     if (!secretName) {
    // not required, while setting up storage cluster.
    // required, while creating new storage class.
    keySecret = generateThalesSecret(kms);
    csiKmsResources.push(k8sCreate({ model: SecretModel, data: keySecret }));
  }
```

packages/odf/components/kms-config/utils.tsx
```diff
@@ -131,6 +134,21 @@ const generateConfigMapPatch = (
   };
 };
 
+const generateThalesSecret = (kms: ThalesConfig) => ({
+  apiVersion: SecretModel.apiVersion,
+  kind: SecretModel.kind,
+  metadata: {
+    name: `thales-kmip-kms-${Math.random().toString(36).substring(7)}`,
```
GowthamShanmugam, line 141: create utils function to generate random name, which will receive prefix as param

bipuladh, line 141:
```suggestion
    name: `thales-kmip-kms-${getRandomChars(7)}`,
```
Can we please refactor this logic out for others as well?

packages/odf/types/kms.ts
GowthamShanmugam: we can't group with NameAddrPort port always when we need to handle which has NameAddrPort differently.

packages/odf/components/kms-config/utils.tsx
```diff
@@ -290,6 +308,49 @@ const createCsiHpcsResources = (
   }
 };
 
+const createCsiThalesResources = (
+  kms: ThalesConfig,
+  csiKmsResources: Promise<K8sResourceKind>[],
```
bipuladh, line 313: Could we keep this a bit more functional. Rather than polluting the argument `csiKmsResources` could we return an array instead. It would be nice to have these helper functions as pure functions.

packages/odf/types/kms.ts
GowthamShanmugam: since we have separate state for each, I dont think we have any necessary to introduce one more grouping relationship

packages/odf/components/kms-config/utils.tsx
```diff
@@ -362,6 +423,32 @@ const createClusterHpcsResources = (
   return keySecret.metadata.name;
 };
 
+const createClusterThalesResources = (
+  kms: ThalesConfig,
+  clusterKmsResources: Promise<K8sResourceKind>[]
+) => {
```
bipuladh, line 429:
```suggestion
): SecretKind['metadata']['name'] => {
```

packages/odf/components/kms-config/utils.tsx
```diff
@@ -362,6 +423,32 @@ const createClusterHpcsResources = (
   return keySecret.metadata.name;
 };
 
+const createClusterThalesResources = (
```
bipuladh, line 426: The return value of the function and the name of the function really do not add up.

packages/odf/components/kms-config/name-address-port.tsx
GowthamShanmugam: I feel it's ok to keep redundant fields between these two. if we do not support anyone  in the future then it is meaning less, also introducing grouping like this makes the KMS structure even more complex.

packages/odf/components/kms-config/utils.tsx
```diff
@@ -384,7 +471,10 @@ export const createCsiKmsResources = (
       createCsiHpcsResources(kms as HpcsConfig, csiKmsResources, update);
       break;
     }
-    default:
+    case ProviderNames.THALES: {
```
bipuladh, line 474: We could instead return the function directly than to update a local variable here and then pass it in the end. Could we please refactor the parent function.

packages/odf/components/kms-config/vault-config.tsx
```diff
@@ -48,10 +46,10 @@ export const VaultConfigure: React.FC<KMSConfigureProps> = ({
   const [Modal, modalProps, launchModal] = useModalLauncher(extraMap);
   const isKmsVaultSASupported = useFlag(FEATURES.ODF_VAULT_SA_KMS);
 
-  const vaultState: VaultConfig = useDeepCompareMemoize(
-    state.kms?.[ProviderNames.VAULT],
+  const vaultState = useDeepCompareMemoize(
+    state.kms.providerState,
     true
-  );
+  ) as VaultConfig;
```
bipuladh, line 52:
```suggestion
  );
```

packages/odf/components/kms-config/vault-config.tsx
```diff
@@ -48,10 +46,10 @@ export const VaultConfigure: React.FC<KMSConfigureProps> = ({
   const [Modal, modalProps, launchModal] = useModalLauncher(extraMap);
   const isKmsVaultSASupported = useFlag(FEATURES.ODF_VAULT_SA_KMS);
 
-  const vaultState: VaultConfig = useDeepCompareMemoize(
-    state.kms?.[ProviderNames.VAULT],
+  const vaultState = useDeepCompareMemoize(
```
bipuladh, line 49:
```suggestion
  const vaultState = useDeepCompareMemoize<VaultConfig>(
```

packages/odf/constants/kms.ts
```diff
@@ -94,8 +99,49 @@ export const HpcsEmptyState: HpcsConfig = Object.seal({
   hasHandled: true,
 });
 
-export const KMSEmptyState: KMSConfig = Object.seal({
+export const ThalesEmptyState: ThalesConfig = Object.seal({
```
bipuladh, line 102:
```suggestion
const ThalesEmptyState: ThalesConfig = Object.seal({
```
Let's make `ProviderStateMap` the only way to access these empty states outside of the file.

bipuladh (changes requested):

packages/odf/components/kms-config/name-address-port.tsx
GowthamShanmugam: Vault and Thals are two different KMS providers. If you are grouping something between the same KMS provider is fine, for example, Vault Token and Vault SA. 

Our design should work like, whatever changes happening on one KMS in the future then another one should work fine.

packages/odf/types/kms.ts
SanjalKatiyar: okay, on re-checking, I remember now... I kept this way intentionally...
here in Thales these are required fields and I have added validations to them as well... whereas in Vault there are no such validations (not sure why, assuming that maybe because there fields are optional in Vault)... so "error" key will be redundant in case of Vault and we will be maintaining extra variables which are not used... so keep it separate from Vault I did this...
let me know if u still think we should combine I can do so ??

packages/odf/components/kms-config/utils.tsx
SanjalKatiyar: because we have `//` between protocol and hostname, so "protocol//hostname" is one string then we need `:` between port and previous string...
may be I am missing something ?

packages/odf/components/kms-config/utils.tsx
SanjalKatiyar: sorry I did not get it... looks to me at correct place...

packages/odf/components/kms-config/utils.tsx
SanjalKatiyar: maybe I am missing something ?
`protocol` + `//` + `hostname` is one string... then this previous string + `:` + `port` is another string... looks okay to me...

SanjalKatiyar: ack !

packages/odf/components/kms-config/utils.tsx
SanjalKatiyar: we can't have it here in else part... during updation as well we can create a new KMS connection, hence new Secret...

packages/odf/components/kms-config/vault-config.tsx
SanjalKatiyar: if I do `useDeepCompareMemoize<VaultConfig>`, still typescript will not be happy about it as `state.kms.providerState` is of type `KMSConfiguration` and `KMSConfiguration` = `VaultConfig | HpcsConfig | ThalesConfig`

SanjalKatiyar: so I had to use `as` explicitly...

packages/odf/constants/kms.ts
SanjalKatiyar: ack...

packages/odf/types/kms.ts
SanjalKatiyar: ack !

packages/odf/components/kms-config/utils.tsx
SanjalKatiyar: updation means adding new connection details to csi configMap...

packages/odf/components/kms-config/utils.tsx
SanjalKatiyar: ack !

packages/odf/components/kms-config/utils.tsx
SanjalKatiyar: ack...

SanjalKatiyar:
@bipuladh @GowthamShanmugam changes done, PTAL.

packages/odf/components/kms-config/utils.tsx
GowthamShanmugam: Ah yes, leave this comment

packages/odf/components/kms-config/name-address-port.tsx
GowthamShanmugam: i agree, it leads to more redundant code

packages/odf/components/kms-config/thales-config.tsx
GowthamShanmugam: Ah! in the future we need to keep each KMS-related components its own directory, only shared one should be in common directly.

packages/odf/components/kms-config/utils.tsx
GowthamShanmugam: ok

packages/odf/types/kms.ts
SanjalKatiyar: also, I forgot to mention... in Vault this `value` is of `SecretKind` where in Thales this `value` is of type `string`... that is one more difference...

humblec:
@GowthamShanmugam @bipuladh can you please review this in priority?

GowthamShanmugam (approved):

PR #393, opened 2022-09-01
Fix cluster claim issue

GowthamShanmugam:


GowthamShanmugam:
@SanjalKatiyar @bipuladh Please review

packages/mco/components/data-policies/create-dr-policy/select-replication-type.tsx
```diff
@@ -186,7 +186,7 @@ export const DRReplicationType: React.FC<DRReplicationTypeProps> = ({
                   { name: c?.name, version: ODF_MINIMUM_SUPPORT }
                 )
               )
-            : !c?.cephFSID &&
+            : c?.cephFSID === '' &&
```
SanjalKatiyar, line 189:
```suggestion
            : c.cephFSID === '' &&
```
`?.` not required... but it is an optional change...

PR #405, opened 2022-09-06
Bug 2061428: add validations to KMS connection name input

SanjalKatiyar:
Screenshot from 2022-09-16 11-09-06
Screenshot from 2022-09-16 11-03-58


packages/odf/components/kms-config/hpcs-config.tsx
```diff
@@ -39,8 +40,10 @@ export const HpcsConfigure: React.FC<KMSConfigureProps> = ({
     (param: string, isRequired: boolean = true) =>
     (value: string) => {
       if (isRequired) {
+        const validParam: boolean =
+          param === HPCSParams.NAME ? isValidName(value) : true;
         kmsObj[param].value = value;
-        kmsObj[param].valid = value !== '';
+        kmsObj[param].valid = validParam && value !== '';
```
GowthamShanmugam, line 46: validParam can check nonempty also,

packages/odf/components/kms-config/utils.tsx
```diff
@@ -50,6 +50,8 @@ export const isValidEndpoint = (url: string) => {
   return validURL || validIP;
 };
 
+export const isValidName = (name: string) => /^[a-zA-Z0-9_.-]*$/.test(name);
```
GowthamShanmugam, line 53: I think this might already checking non empty

packages/odf/components/kms-config/hpcs-config.tsx
SanjalKatiyar: so I only want to check `validParam` if it is `Connection name`, for all other inputs (like address, port etc) I only want to check if they are non-empty, that is why both checks are seperate...

GowthamShanmugam:
```suggestion
        kmsObj[param].valid = HPCSParams.NAME ? isValidName(value) : value !== '';
```

SanjalKatiyar: ack... 
currently `isValidName` is not checking if string is empty or not... I will add that check to it and update...

packages/odf/components/kms-config/utils.tsx
SanjalKatiyar: not yet, but I can add that check to it...

packages/odf/components/kms-config/hpcs-config.tsx
SanjalKatiyar: regex is returning true for empty string, I will add `name !== ''` in `isValidName` function along with regex...

SanjalKatiyar: something like: `name !== '' && /^[a-zA-Z0-9_.-]*$/.test(name)`

packages/odf/components/kms-config/utils.tsx
GowthamShanmugam: leave it as is it now

GowthamShanmugam (approved):

PR #409, opened 2022-09-09
Bug fix on failover and relocate modal

GowthamShanmugam:


GowthamShanmugam:
image

packages/mco/components/disaster-recovery/app-dr-status/dr-status-table.tsx
```diff
@@ -48,13 +48,14 @@ const getTargetClusters = (
...
     const status = DRPC_STATUS[drpc?.status?.phase] || '';
-    return status === currentStatus
+    status === currentStatus
       ? ([DRPC_STATUS.Relocating, DRPC_STATUS.Relocated].includes(status) &&
           acc.add(drpc?.spec?.preferredCluster)) ||
-          ([DRPC_STATUS.FailingOver, DRPC_STATUS.FailedOver].includes(status) &&
-            acc.add(drpc?.spec?.failoverCluster))
+        ([DRPC_STATUS.FailingOver, DRPC_STATUS.FailedOver].includes(status) &&
+          acc.add(drpc?.spec?.failoverCluster))
       : acc;
```
SanjalKatiyar, line 56: this is doing anything, I mean `: acc` part of ternary condition... either we can change it to `: null` or something like
```
if (status === currentStatus) {
    ([DRPC_STATUS.Relocating, DRPC_STATUS.Relocated].includes(status) &&
          acc.add(drpc?.spec?.preferredCluster)) ||
    ([DRPC_STATUS.FailingOver, DRPC_STATUS.FailedOver].includes(status) &&
          acc.add(drpc?.spec?.failoverCluster))
}
return acc;
```

packages/mco/components/disaster-recovery/app-failover-relocate/peer-cluster-status.tsx
```diff
@@ -47,11 +47,11 @@ const getPeerReadiness = (
   obj.peerReadiness.text !== PEER_READINESS.PEER_NOT_READY
```
SanjalKatiyar, line 47: so I have a doubt, `obj.peerReadiness.text` contains translated string... but `PEER_READINESS.PEER_NOT_READY` this is not translated... so in case user changes browser's language, this will always be true... is that okay ?

packages/mco/components/disaster-recovery/app-failover-relocate/peer-cluster-status.tsx
```diff
@@ -47,11 +47,11 @@ const getPeerReadiness = (
...
-          text: t('Peer ready'),
+          text: t('Ready'),
           icon: <GreenCheckCircleIcon />,
         }
       : {
-          text: t('Peer not ready'),
+          text: t('Not ready'),
           icon: <RedExclamationCircleIcon />,
         }
     : obj.peerReadiness;
```
SanjalKatiyar, line 57: also, instead of always looping over entire list (we are using `reduce` currently in `getPeerStatusSummary`), we can return as soon as we find very first instance of DRPC which is not Ready... we can try to use `every`...

SanjalKatiyar: may not be required in this PR, just a suggestion...

GowthamShanmugam: getPeerStatusSummary has a reduced becuase, we need to makesure all ready from the subscription groups are all matching with selected DRPolicy.

So there is some subgroup that can have non-matching DRPolicy name also, we need to find all matching first and makesure all matching are peer ready.

also, we need to find the last lastHealthCheck which is from a subs group that has recent health check time, and also recent lastSnapshotTakenOn time.

All three can come from different subs groups also. so we need to loop through all

GowthamShanmugam: first one loop through all and make sure all peers are ready,
the second  loop through all and fetch which has the least last health check.
third loop through all and fetch which has the least last lastSnapshotTakenOn.

single complete  looping will cover all above three cases.

GowthamShanmugam:
@SanjalKatiyar Changes done

packages/mco/utils/disaster-recovery.ts
```diff
@@ -128,7 +128,7 @@ export const getPlacementKind = (subscription: ACMSubscriptionKind) =>
   subscription?.spec?.placement?.placementRef?.kind;
 
 export const findPeerCondition = (drPolicyControl: DRPlacementControlKind) =>
```
bipuladh, line 130: Shouldn't this function name be `isPeerReady`? `findPeerCondition` should return a condition as per the name.

bipuladh (changes requested):

packages/mco/components/disaster-recovery/app-failover-relocate/subscription-group-selector.tsx
```diff
@@ -110,25 +110,56 @@ export const SubscriptionGroupSelector: React.FC<SubscriptionGroupSelectorProps>
...
+                ? {
+                    validOptions: validPeerStatus
+                      ? [
+                          getOptions(drpcState, isValidTargetCluster, t),
+                          ...acc.validOptions,
+                        ]
+                      : [
+                          ...acc.validOptions,
+                          getOptions(drpcState, isValidTargetCluster, t),
+                        ],
```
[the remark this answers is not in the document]
GowthamShanmugam: Under valid options:
    yes order matters, all options with valid peer status will be checked by default, and invalid will be unchecked. I need all checked should be displayed first and unchecked will be displayed last.

Under invalid options:
   all disabled.

packages/mco/components/disaster-recovery/app-failover-relocate/subscription-group-selector.tsx
```diff
@@ -110,25 +110,56 @@ export const SubscriptionGroupSelector: React.FC<SubscriptionGroupSelectorProps>
...
-        if (!!generatedOptions.length) {
+        );
+        const allOptions = [
+          ...generatedOptions.validOptions,
+          ...generatedOptions.inValidOptions,
+        ];
+        if (!!allOptions.length) {
           setSelected(validState);
-          setOptions(generatedOptions);
+          setOptions(allOptions);
```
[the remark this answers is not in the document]
GowthamShanmugam: I am converting all options object to list, which will have all valid option on top and invalid options at the end

packages/mco/utils/disaster-recovery.ts
GowthamShanmugam: yes

PR #410, opened 2022-09-12
Bug 2043572: adding random characters if secret exists with same name during nss creation

debjyoti-pandit:
fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2043572

**Initially created the NSS then deleted it and then again created it with the same name**
image

### DIFF
--- a/packages/odf/components/namespace-store/namespace-store-form.tsx
+++ b/packages/odf/components/namespace-store/namespace-store-form.tsx
@@ -35,7 +35,12 @@ import {
   secretPayloadCreator,
 } from '../../utils';
 import { S3EndPointType } from '../mcg-endpoints/s3-endpoint-type';
-import { initialState, providerDataReducer } from './reducer';
+import {
+  initialState,
+  providerDataReducer,
+  ProviderDataState,
+  StoreAction,
+} from './reducer';
 import '../mcg-endpoints/noobaa-provider-endpoints.scss';
 
 const PROVIDERS = getProviders(StoreType.NS);
@@ -56,6 +61,48 @@ type NamespaceStoreFormProps = {
   onCancel: () => void;
 };
 
+const createSecret = async (
+  dataSourceName: string,
+  namespace: string,
+  provider: BC_PROVIDERS,
+  providerDataState: ProviderDataState,
+  providerDataDispatch: React.Dispatch<StoreAction>
+) => {
+  const { secretKey, accessKey } = providerDataState;
+  let createdSecret: SecretKind;
+  let secretName = dataSourceName.concat('-secret');
+  const secretPayload = secretPayloadCreator(
+    provider,
+    namespace,
+    secretName,
+    accessKey,
+    secretKey
+  );
+  try {
+    createdSecret = (await k8sCreate({
+      model: SecretModel,
+      data: secretPayload,
+    })) as SecretKind;
+  } catch {
+    secretName = dataSourceName.concat('-');
+    const newSecretPayload = {
+      ...secretPayload,
+      metadata: {
+        generateName: secretName,
+        namespace: secretPayload.metadata.namespace,
+      },
+    };
+    createdSecret = (await k8sCreate({
+      model: SecretModel,
+      data: newSecretPayload,
+    })) as SecretKind;
+  } finally {
+    secretName = createdSecret?.metadata?.name;
+    providerDataDispatch({ type: 'setSecretName', value: secretName });
+  }
+  return secretName;
+};
+
 const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
   const { t } = useCustomTranslation();
   const [nsName, setNsName] = React.useState('');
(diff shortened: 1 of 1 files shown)


packages/odf/components/namespace-store/namespace-store-form.tsx
```diff
@@ -73,14 +74,24 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
   const handleNsNameTextInputChange = (strVal: string) => setNsName(strVal);
   const { onCancel, className, redirectHandler, namespace } = props;
 
-  const onSubmit = (event) => {
+  const onSubmit = async (event) => {
     event.preventDefault();
     setProgress(true);
     /** Create a secret if secret ==='' */
     let { secretName } = providerDataState;
+    let m4 = () => Math.floor((1 + Math.random()) * 0x10000).toString(16);
```
GowthamShanmugam, line 82: you dont need to create random strings, check Kubernetes generate names: https://kubernetes.io/docs/reference/using-api/api-concepts/#generated-values

packages/odf/components/namespace-store/namespace-store-form.tsx
```diff
@@ -73,14 +74,24 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
...
     const promises = [];
     if (!secretName) {
-      secretName = nsName.concat('-secret');
+      try {
+        await k8sGet({
+          name: nsName.concat('-secret'),
+          model: SecretModel,
+          ns: CEPH_STORAGE_NAMESPACE,
+        });
+        secretName = nsName.concat('-' + m4() + m4() + m4());
```
GowthamShanmugam, line 91: Move this under separate function isSecretExits. Based on boolean value create secretName

packages/odf/components/namespace-store/namespace-store-form.tsx
SanjalKatiyar: we do not want to make entire name random, we need to append random chars to the NamespaceStore name...

packages/odf/components/namespace-store/namespace-store-form.tsx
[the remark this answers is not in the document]
SanjalKatiyar: @GowthamShanmugam PTAL: https://github.com/red-hat-storage/odf-console/pull/337#pullrequestreview-1067879951

packages/odf/components/namespace-store/namespace-store-form.tsx
SanjalKatiyar, line 82: use `getRandomChars` fucntion

packages/odf/components/namespace-store/namespace-store-form.tsx
```diff
@@ -73,14 +74,24 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
...
+    let m4 = () => Math.floor((1 + Math.random()) * 0x10000).toString(16);
     const promises = [];
     if (!secretName) {
-      secretName = nsName.concat('-secret');
+      try {
+        await k8sGet({
+          name: nsName.concat('-secret'),
+          model: SecretModel,
+          ns: CEPH_STORAGE_NAMESPACE,
+        });
```
SanjalKatiyar, line 90: I don't like this await here, this is blocking entire `onSubmit` function. Handle Secret in a separate function and use await there. Also, no need to check if Secret is present or not, just create the Secret (without random chars), if it is present it will give `already exists` error, then add random chars and create again.

packages/odf/components/namespace-store/namespace-store-form.tsx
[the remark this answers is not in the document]
SanjalKatiyar: I agree, just create Secret first without random name, if we get "already exists" error then add random chars and create again...

packages/odf/components/namespace-store/namespace-store-form.tsx
```diff
@@ -73,14 +74,24 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
...
     event.preventDefault();
     setProgress(true);
     /** Create a secret if secret ==='' */
     let { secretName } = providerDataState;
+    let m4 = () => Math.floor((1 + Math.random()) * 0x10000).toString(16);
     const promises = [];
     if (!secretName) {
-      secretName = nsName.concat('-secret');
+      try {
+        await k8sGet({
```
SanjalKatiyar, line 86: @bipuladh just FYI, earlier Secret and NamespaceStore were getting created in parallel, but now as per https://github.com/red-hat-storage/odf-console/pull/337#pullrequestreview-1067879951 we will have to wait untill Secret is created then only we can move forward to NamespaceStore creation (because we need to pass Secret name to NamespaceStore CR's spec)...

packages/odf/components/namespace-store/namespace-store-form.tsx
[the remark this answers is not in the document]
debjyoti-pandit: > why do you need to check secret exists or not, create your secret with generated name with some prefix/suffix.

I added this check as @bipuladh asked in this review comment.

https://github.com/red-hat-storage/odf-console/pull/337#pullrequestreview-1067879951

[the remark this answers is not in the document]
debjyoti-pandit: > I agree, just create Secret first without random name, if we get "already exists" error then add random chars and create again...

we can do this, but I felt inside of chaining another promise inside the catch phase of the promise.all. (which is a kind of anti-pattern), we can check first and then decide what can be the name then finally create the nss

packages/odf/components/namespace-store/namespace-store-form.tsx
debjyoti-pandit: > we do not want to make entire name random, we need to append random chars to the NamespaceStore name...

yeah it is not making the entire name as random, it is just creating a random text and appending at the end

bipuladh: @GowthamShanmugam +1

packages/odf/components/namespace-store/namespace-store-form.tsx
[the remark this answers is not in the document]
bipuladh: I prefer @GowthamShanmugam suggestion. Use generateName to deal with this whole situation.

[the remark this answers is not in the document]
debjyoti-pandit: done

packages/odf/components/namespace-store/namespace-store-form.tsx
```diff
@@ -73,89 +73,107 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
...
+            },
+          },
+        };
+      }
+      if (provider === BC_PROVIDERS.S3) {
+        nsPayload.spec.s3Compatible = {
+          ...nsPayload.spec.s3Compatible,
+          endpoint: providerDataState.endpoint,
+        };
+      } else if (provider === BC_PROVIDERS.IBM) {
```
GowthamShanmugam, line 146: why one this need else if, i prefer to use switch here

GowthamShanmugam: `switch(provider) {
	case BC_PROVIDERS.S3:
		nsPayload.spec.s3Compatible = {
          		...nsPayload.spec.s3Compatible,
          		endpoint: providerDataState.endpoint,
        	};
        	break;
        case BC_PROVIDERS.IBM:
        	nsPayload.spec.ibmCos = {
          		...nsPayload.spec.ibmCos,
          		endpoint: providerDataState.endpoint,
        	};
        	break;
        case BC_PROVIDERS.AWS:
		nsPayload.spec.awsS3 = {
          		...nsPayload.spec.awsS3,
          		region: providerDataState.region,
        	};
        	break;
        case BC_PROVIDERS.FILESYSTEM:
        	nsPayload.spec.nsfs = {
          		...nsPayload.spec.nsfs,
          		pvcName: pvc,
          		subPath: folderName,
        	};
        	break;
        default:
        	if (externalProviders.includes(provider)) {
        		nsPayload.spec = {
          			...nsPayload.spec,
          			[PROVIDERS_NOOBAA_MAP[provider]]: {
            				[BUCKET_LABEL_NOOBAA_MAP[provider]]: providerDataState.target,
            				secret: {
              					name: secretName,
              					namespace,
            				},
          			},
        		};
      		};`

GowthamShanmugam: thoughts?

debjyoti-pandit: seems ok, I just missed the default part

debjyoti-pandit: > I think, the code block in default is needed always

we can't keep it there as it will get executed only for non-mentioned cases

packages/odf/components/namespace-store/namespace-store-form.tsx
```diff
@@ -73,89 +73,109 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
...
-    promises.push(
-      k8sCreate({ model: NooBaaNamespaceStoreModel, data: nsPayload })
-    );
+      if (externalProviders.includes(provider)) {
+        nsPayload.spec = {
+          ...nsPayload.spec,
+          [PROVIDERS_NOOBAA_MAP[provider]]: {
+            [BUCKET_LABEL_NOOBAA_MAP[provider]]: providerDataState.target,
+            secret: {
+              name: secretName,
```
GowthamShanmugam, line 135: Here https://github.com/red-hat-storage/odf-console/pull/410/files#diff-932498549506d34852c75aee5147f27f9bdceb13f0ddf87082f1918dbb870196R82,
if the secret name is empty then are we creating this secret with string?

debjyoti-pandit: yes, when the user selects the secret from the secret dropdown then the secret name won't' be empty and we won't have to create any with customized names. But if the user chooses to enter the access key and secret key then the secret name field will be empty, then we need to create a secret name.

packages/odf/components/namespace-store/namespace-store-form.tsx
```diff
@@ -73,89 +73,109 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
...
+          };
+          createdSecret = (await k8sCreate({
+            model: SecretModel,
+            data: newSecretPayload,
+          })) as SecretKind;
+        } finally {
+          secretName = createdSecret?.metadata?.name;
+          providerDataDispatch({ type: 'setSecretName', value: secretName });
+        }
+      }
```
GowthamShanmugam, line 115: Changed looks good to me, but one small optional suggestion is, that I feel a lot is going on the same function, so
  if we split this function into two then it looks even more clean, upto this is one function like generateSecret / createSecrete which will create  a secret and return secret name as result.
  
  and remaining is create namespace.

debjyoti-pandit: Done

GowthamShanmugam (approved):

SanjalKatiyar:
/retitle Bug 2043572: adding random characters if secret exists with same name during nss creation

PR #414, opened 2022-09-15
Bug 2051563: add popover for MCG clusters to depict no limit.

weirdwiz:
Screenshot from 2022-09-15 14-59-03

https://bugzilla.redhat.com/show_bug.cgi?id=2051563

### DIFF
--- a/packages/shared/dashboards/capacity-card/capacity-card.tsx
+++ b/packages/shared/dashboards/capacity-card/capacity-card.tsx
@@ -11,6 +11,8 @@ import {
   ProgressVariant,
   Title,
   Tooltip,
+  Button,
+  Popover,
 } from '@patternfly/react-core';
 import { DataUnavailableError } from '../../generic/Error';
 import ResourceLink from '../../resource-link/resource-link';
@@ -135,6 +137,7 @@ const CapacityCardRow: React.FC<CapacityCardRowProps> = ({
   largestValue,
   resourceModel,
 }) => {
+  const { t } = useCustomTranslation();
   const progress =
     (!isPercentage && isRelative) || (isPercentage && !isRelative)
       ? getProgress(data, isRelative, largestValue)
@@ -203,6 +206,21 @@ const CapacityCardRow: React.FC<CapacityCardRowProps> = ({
       </GridItem>
       <GridItem span={3} key={`${data?.name}~value`}>
         {dataUnavailable ? '-' : value}
+        {!dataUnavailable && !isPercentage && (
+          <span>
+            &nbsp; / &nbsp;
+            <Popover
+              position="right"
+              bodyContent={t(
+                'For standalone MCG clusters, there is no set limit.'
+              )}
+            >
+              <Button variant="plain" className="details-item__popover-button">
+                {t('No Limit')}
+              </Button>
+            </Popover>
+          </span>
+        )}
       </GridItem>
     </>
   );


packages/shared/dashboards/capacity-card/capacity-card.tsx
```diff
@@ -203,6 +205,18 @@ const CapacityCardRow: React.FC<CapacityCardRowProps> = ({
       </GridItem>
       <GridItem span={3} key={`${data?.name}~value`}>
         {dataUnavailable ? '-' : value}
+        {isPercentage ? (
```
weirdwiz, line 208: @bipuladh this is under the assumption that there are only two scenarios. Is this correct?

weirdwiz:
PTAL @GowthamShanmugam @SanjalKatiyar @bipuladh

packages/shared/dashboards/capacity-card/capacity-card.tsx
```diff
@@ -203,6 +205,18 @@ const CapacityCardRow: React.FC<CapacityCardRowProps> = ({
...
+          ''
+        ) : (
+          <Popover
+            position="right"
+            bodyContent="For standalone MCG clusters, there is no set limit."
```
SanjalKatiyar, line 213:
```suggestion
            bodyContent=t('For standalone MCG clusters, there is no set limit.')
```

packages/shared/dashboards/capacity-card/capacity-card.tsx
```diff
@@ -203,6 +205,18 @@ const CapacityCardRow: React.FC<CapacityCardRowProps> = ({
...
+          >
+            <Button variant="plain" className="details-item__popover-button">
+              No Limit
```
SanjalKatiyar, line 216:
```suggestion
              {t('No Limit')}
```

packages/shared/dashboards/capacity-card/capacity-card.tsx
```diff
@@ -143,7 +145,7 @@ const CapacityCardRow: React.FC<CapacityCardRowProps> = ({
       : 0;
   const value = isPercentage
     ? `${data?.usedValue?.string} / ${data?.totalValue?.string}`
-    : data?.usedValue?.string;
+    : data?.usedValue?.string + ' / ';
```
SanjalKatiyar, line 148:
```suggestion
    : data?.usedValue?.string;
```

packages/shared/dashboards/capacity-card/capacity-card.tsx
SanjalKatiyar, line 208:
```suggestion
        {isPercentage && ( <Popover
```

packages/shared/dashboards/capacity-card/capacity-card.tsx
```diff
@@ -203,6 +205,18 @@ const CapacityCardRow: React.FC<CapacityCardRowProps> = ({
       </GridItem>
       <GridItem span={3} key={`${data?.name}~value`}>
         {dataUnavailable ? '-' : value}
+        {isPercentage ? (
+          ''
+        ) : (
+          <Popover
+            position="right"
```
SanjalKatiyar, line 212: can't we have it on auto ? else its position will be right on smaller screen sizes as well...

SanjalKatiyar:
test this PR once on Full Deployment mode as well, just to make sure there is no regression !

packages/shared/dashboards/capacity-card/capacity-card.tsx
SanjalKatiyar: @bipuladh can there be any issue for external mode cluster (RHCS) ? do we have used/total there ?

bipuladh: For External mode we do have used/total information.

packages/shared/dashboards/capacity-card/capacity-card.tsx
bipuladh: Lets ask UX team to review this once.

packages/shared/dashboards/capacity-card/capacity-card.tsx
Chandan141996: @weirdwiz @SanjalKatiyar For larger screen sizes I guess "right" should be the default value, for small screens "auto" should be used to self align the popover with the screen size

weirdwiz: Having the value as "right" will automatically change the position if the screen size goes small. 
Attached screenshot below:
Screenshot from 2022-09-29 16-07-38
Screenshot from 2022-09-29 16-07-02

packages/shared/dashboards/capacity-card/capacity-card.tsx
```diff
@@ -203,6 +206,21 @@ const CapacityCardRow: React.FC<CapacityCardRowProps> = ({
...
+          <span>
+            &nbsp; / &nbsp;
+            <Popover
+              position="right"
+              bodyContent={t(
+                'For standalone MCG clusters, there is no set limit.'
+              )}
+            >
+              <Button variant="plain" className="details-item__popover-button">
+                {t('No Limit')}
```
SanjalKatiyar, line 219: why not directly this ??
```suggestion
                {t(' / No Limit')}
```

weirdwiz: Adding it here adds the underline/hyperlink under the "/" too, which wasn't present in the UX design :sweat_smile: .

This is why I separated the "/"

PR #415, opened 2022-09-15
Bug 2126694: add filter for kms connection details

SanjalKatiyar:


packages/ocs/storage-class/sc-form.tsx
```diff
@@ -392,16 +395,20 @@ const ExistingKMSDropDown: React.FC<ExistingKMSDropDownProps> = ({
           try {
             // removing any object having syntax error
             // or, which are not supported by UI.
-            const kmsData = JSON.parse(connectionDetails);
+            const kmsData: Object = JSON.parse(connectionDetails);
```
GowthamShanmugam, line 398:
```suggestion
            const kmsData: KMSConfigMap = JSON.parse(connectionDetails);
```

packages/odf/constants/kms.ts
```diff
@@ -32,6 +33,15 @@ export const SupportedProviders = {
   [ProviderNames.THALES]: {
     group: 'Thales (using KMIP)',
     supported: [KmsImplementations.KMIP],
+    filter: (kms: Object, secretsList: SecretKind[]) => {
```
GowthamShanmugam, line 36:
```suggestion
    filter: (kms: ThalesConfigMap, secretsList: SecretKind[]) => {
```

packages/ocs/storage-class/sc-form.tsx
SanjalKatiyar: it is not exactly the ConfigMap, but "data" part of the configmap which is a JSON string and we are converting to JS Object...

packages/odf/constants/kms.ts
GowthamShanmugam, line 36: Shall we add the proper type for this object?
something like
export const SupportedProviders : SupportedProvidersProps

SanjalKatiyar: ack... will add that...

packages/ocs/storage-class/sc-form.tsx
SanjalKatiyar: I can check and create a new type for it... let me check...

packages/ocs/storage-class/sc-form.tsx
```diff
@@ -506,19 +527,15 @@ export const StorageClassEncryptionKMSID: React.FC<ProvisionerProps> = ({
...
-    namespace: CEPH_STORAGE_NAMESPACE,
-    name: KMSConfigMapCSIName,
-  };
   const [infra, infraLoaded, infraLoadError] = useK8sGet<any>(
     InfrastructureModel,
     'cluster'
   );
   const [csiConfigMap, csiConfigMapLoaded, csiConfigMapLoadError] =
     useK8sWatchResource<ConfigMapKind>(csiCMWatchResource);
+  const [secrets, secretsLoaded, secretsLoadError] =
```
GowthamShanmugam, line 536: shall we add useK8sWatchResources for this, do you think any performance challenges it there if we use useK8sWatchResources?

SanjalKatiyar: Is there any advantage of using one over another ?? I think both will work similarly...

GowthamShanmugam: no advantage, just for better code maintenance

GowthamShanmugam (approved):

PR #416, opened 2022-09-16
Bug 2126604: change name for Thales provider

SanjalKatiyar:


packages/odf/components/kms-config/kms-config.tsx
```diff
@@ -29,7 +29,7 @@ const KMSProviders: KMSProvidersType = [
     allowedPlatforms: ['IBMCloud'],
   },
   {
-    name: 'Thales (using KMIP)',
+    name: 'Thales CipherTrust Manager (using KMIP)',
```
GowthamShanmugam, line 32: Is this displayed on the UI? dont we need translation?

GowthamShanmugam, line 32: Is this displayed on the UI? dont we need translation?

SanjalKatiyar: it's product name, it is better not to translate IMO.

GowthamShanmugam: ack

GowthamShanmugam (approved):

PR #417, opened 2022-09-16
Bug 2114504: flip menu vertically when it reaches the boundary

SanjalKatiyar:
**Before:**
Screenshot from 2022-09-16 19-38-06
**After:**
Screenshot from 2022-09-16 19-30-12
Screenshot from 2022-09-16 19-30-22

### DIFF
--- a/packages/shared/kebab/kebab.tsx
+++ b/packages/shared/kebab/kebab.tsx
@@ -10,6 +10,7 @@ import {
   KebabToggle,
   DropdownItemProps,
   Tooltip,
+  DropdownDirection,
 } from '@patternfly/react-core';
 import { CaretDownIcon } from '@patternfly/react-icons';
 import { ModalKeys, LaunchModal } from '../modals/modalLauncher';
@@ -76,6 +77,9 @@ export const Kebab: React.FC<KebabProps> = ({
 }) => {
   const { t } = useCustomTranslation();
 
+  const eventRef = React.useRef(undefined);
+  const [toggleDirection, setToggleDirection] =
+    React.useState<DropdownDirection>(DropdownDirection.down);
   const [isOpen, setOpen] = React.useState(false);
 
   const { resourceModel, resource } = extraProps;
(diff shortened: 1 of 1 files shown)


GowthamShanmugam (approved):
LGMT

SanjalKatiyar:
> /lgtm cancel

this PR works totally fine and we can merge this as well if we want.
cancelled lgtm because PF also provides a prop `isFlipEnabled` for a Dropdown, but passing that prop is not working as expected. In sync with PF team as it's a known bug from PF side.

GowthamShanmugam:
> isFlipEnabled

ack

SanjalKatiyar:
> > /lgtm cancel
> 
> this PR works totally fine and we can merge this as well if we want. cancelled lgtm because new version of PF also provides a prop `isFlipEnabled` for a Dropdown, but passing that prop is not working as expected. In sync with PF team as it's a known bug from PF side.

@GowthamShanmugam @bipuladh we can go ahead and merge this PR for now (until it is fixed in newer version of PF).
PF has opened an issue on there end https://github.com/patternfly/patternfly-react/issues/8013 to fix the bug with `isFlipEnabled` prop !

packages/shared/kebab/kebab.tsx
```diff
@@ -159,7 +178,10 @@ export const Kebab: React.FC<KebabProps> = ({
   }, [t, customKebabItemsMap, resourceLabel]);
 
   const toggle = React.useMemo(() => {
-    const onToggle = () => setOpen((open) => !open);
+    const onToggle = (_, e) => {
+      eventRef.current = e;
```
bipuladh, line 182: Could we handle all the logic in `useLayoutEffect` inside here itself?

SanjalKatiyar: that will not work, when this `onToggle` is called at that time event does not has a reference to the popper/dropdown items (cause they are not yet mounted to DOM)... only after the DOM is updated with the re-render which `onToggle` is causing, then only we can access the newly mounted element. Hence handling in a hook.

PR #421, opened 2022-09-20
Bug 2125356: Custom PVC selector option while applying DRPolicy

GowthamShanmugam:


GowthamShanmugam:
Screenshot from 2022-09-20 21-07-48

packages/mco/components/disaster-recovery/drpolicy-actions/apply-policy/apply-policy-modal.tsx
```diff
@@ -389,14 +405,43 @@ const ApplyDRPolicyModal: React.FC<CommonModalProps<ApplyModalExtraProps>> = (
...
         />
-        <Alert
-          className="co-alert mco-apply-policy-modal__alert"
-          variant="info"
-          title={t(
-            'All PersistentVolumeClaims matching the placement rules under an application will be DR protected.'
-          )}
-          isInline
-        />
+        {filteredSelectedApps?.length <= 1 ? (
```
SanjalKatiyar, line 408: one doubt, if all apps are selected then all PVCs will be DR protected else we can select few based on labels.
If filteredSelectedApps === 1, it means we have only one app, and if we select it, it means we have selected all the apps, so in this case also we should show checkbox instead of SelectorInput ??

packages/mco/components/disaster-recovery/drpolicy-actions/apply-policy/apply-policy-modal.tsx
```diff
@@ -389,14 +405,43 @@ const ApplyDRPolicyModal: React.FC<CommonModalProps<ApplyModalExtraProps>> = (
...
+          <Form className="mco-apply-policy-modal__pvcselector">
```
SanjalKatiyar, line 409: everything can be inside one `<Form className="mco-apply-policy-modal__pvcselector">` and inside it we can use conditions to display diff diff components. No need to use same `<Form>` twice.

packages/mco/components/disaster-recovery/drpolicy-actions/apply-policy/apply-policy-modal.tsx
```diff
@@ -389,14 +405,43 @@ const ApplyDRPolicyModal: React.FC<CommonModalProps<ApplyModalExtraProps>> = (
...
+            <Alert
+              className="co-alert mco-apply-policy-modal__alert"
+              variant="warning"
+              title={t(
+                "When multiple applications are selected, DR protection will be applied for all the PVCs under the application's namespace."
+              )}
+              isInline
+            />
+            <Form className="mco-apply-policy-modal__pvcselector">
+              <FormGroup fieldId="all-pvc" isRequired>
```
SanjalKatiyar, line 432: this will not add required astrik (*) sign to this checkbox, user should know that it is a required field, right ?
can we try adding astrik (*) sign to `Protect all PVCs within the application's namespace` text ?

packages/mco/components/disaster-recovery/drpolicy-actions/apply-policy/apply-policy-modal.tsx
```diff
@@ -389,14 +405,43 @@ const ApplyDRPolicyModal: React.FC<CommonModalProps<ApplyModalExtraProps>> = (
...
+                <Checkbox
+                  id="user-agreement"
+                  label={t(
+                    "Protect all PVCs within the application's namespace"
+                  )}
+                  isChecked={isProtectAllPVCChecked}
+                  onChange={() => setProtectAllPVC(!isProtectAllPVCChecked)}
+                />
+              </FormGroup>
```
SanjalKatiyar, line 441: is this a required field as well ? does user have to check the checkbox always ?
If so, can't we check it by default or add a astrik (*) sign to this checkbox ?
if this is not required field, plz skip this comment:)

packages/mco/components/disaster-recovery/drpolicy-actions/apply-policy/apply-policy-modal.tsx
GowthamShanmugam: ack

packages/mco/components/disaster-recovery/drpolicy-actions/apply-policy/apply-policy-modal.tsx
GowthamShanmugam: it is mandatory to select the checkbox, I will add an asterisk(*). The reason we are not auto-enabling is to get the user's attention for this, User might miss this warning message.

packages/mco/components/disaster-recovery/drpolicy-actions/apply-policy/apply-policy-modal.tsx
```diff
@@ -392,14 +408,44 @@ const ApplyDRPolicyModal: React.FC<CommonModalProps<ApplyModalExtraProps>> = (
...
+                )}
+                isInline
+              />
+              <FormGroup fieldId="all-pvc">
+                <Checkbox
+                  id="user-agreement"
+                  label={
+                    <>
+                      {t("Protect all PVCs within the application's namespace")}
+                      <span className="pf-c-form__label-required">*</span>
```
GowthamShanmugam, line 439: Need to upgrade PF to pass isRequired option for checkbox

SanjalKatiyar: ack...

packages/mco/components/disaster-recovery/drpolicy-actions/apply-policy/apply-policy-modal.tsx
SanjalKatiyar: I mean, let's say there is only one app and one placement, and we are selecting it. In this case shouldn't we show checkbox instead ? Just a doubt...

GowthamShanmugam: No, We need to show the PVC selector, it is not about whether the application has only one placement or more than one, it works like how many placements the user is going to select overall from all applications.

At maximum, only one placement rule is allowed from all applications, we won't even show PVC selector if one placement rule is selected per application.

GowthamShanmugam: The reason is, the PVC selector is unique for each placement of PVCs, if more than one got selected then we can't use a common PVC selector for all. So we leave the PVC selector empty and Ramen will protect all PVC for the selected placements. That is why we are warning the user that it will protect all and it is up to the user's risk.

In 4.13, we are working on the design to provide a seperate PVC selector for each placement rule.

SanjalKatiyar: ack...

PR #423, opened 2022-09-21
Bug 2128572: Added application namepace on connected application modal

GowthamShanmugam:


GowthamShanmugam:
After Fix:
image

packages/mco/components/disaster-recovery/drpolicy-list-page/application-status.tsx
```diff
@@ -113,21 +114,18 @@ const subsFilter = (
 const fetchApplications = (
   subscriptionMapping: SubcsriptionMapping,
   app: ApplicationKind
-): string[] => {
-  let appNames: string[] = [];
-  const appName = app?.metadata?.name;
-  const appNamespace = app?.metadata?.namespace;
+): ApplicationKind[] => {
+  let apps = new Set<ApplicationKind>();
```
SanjalKatiyar, line 118: I don't think storing `ApplicationKind` in a set will do anything, set will check by reference only and each object reference will be different (even though contents are same)... am I missing something ?

SanjalKatiyar: why not only add names of the Application ?

SanjalKatiyar: why do we need set ?
application name and namespace combination will be unique anyway...

GowthamShanmugam: we dont need a set here, and we dont need a list also here, if one subscription is matching then we can consider the app

PR #430, opened 2022-09-26
update test cases to run with ODF version 4.12 - CI

SanjalKatiyar:


packages/odf/components/namespace-store/namespace-store-form.tsx
```diff
@@ -242,6 +242,7 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
           onSelect={setProvider as any}
           dropdownItems={PROVIDERS}
           defaultSelection={provider}
+          data-test="namespacestore-provider"
```
vbnrh, line 245: i dont see this id being passed down to the dropdown ?

vbnrh (changes requested):

packages/odf/components/namespace-store/namespace-store-form.tsx
SanjalKatiyar: it will... check https://github.com/red-hat-storage/odf-console/pull/430/files/e7661c08b9183c95a4ff4d94fdbcf20b6c50d7d5#diff-8adfa67001c080b994ad6e8cc465b3263a9dcd9459dcafe0ff68d81ce08fc395 StaticDropdown.tsx

SanjalKatiyar:
/lgtm cancel
updating ODF to 4.12 instead of 4.11

PR #431, opened 2022-09-26
Bug 2022356: Range value is not properly displayed for performance card

debjyoti-pandit:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2022356

Jira: https://issues.redhat.com/browse/OCSBZM-2774

Desc: Added initial value for the ticks when there is only one value on the y-axis. It is failing to scale appropriately.

### DIFF
--- a/packages/shared/dashboards/line-graph/line-graph.tsx
+++ b/packages/shared/dashboards/line-graph/line-graph.tsx
@@ -74,6 +74,9 @@ const LineGraph: React.FC<LineGraphProps> = ({ data }) => {
             tickFormat={(tick, _index, _ticks) => {
               return `${tick} ${unit}`;
             }}
+            tickValues={
+              mappedLineData.length === 1 ? [mappedLineData[0].y] : undefined
+            }
           />
           <ChartGroup>
             <ChartLine data={mappedLineData} />


packages/shared/dashboards/line-graph/line-graph.tsx
```diff
@@ -74,6 +74,9 @@ const LineGraph: React.FC<LineGraphProps> = ({ data }) => {
             tickFormat={(tick, _index, _ticks) => {
               return `${tick} ${unit}`;
             }}
+            tickValues={
+              mappedLineData.length === 1 ? [mappedLineData[0].y] : undefined
```
GowthamShanmugam, line 78: is this always ===1 or <= 1?

debjyoti-pandit: as length can never be less than 0 and when the length is one we need to specify the tick values. that's why I made it === 1

debjyoti-pandit:
Please find the attached video of the changes [here](https://drive.google.com/file/d/1VYR0cfUxWaOUF9iY_JsNcMbVah_mtXAA/view?usp=sharing) (Please start from 7th minute of the video as initial 7 min doesn't contains any important information)

packages/shared/dashboards/line-graph/line-graph.tsx
GowthamShanmugam: Do you mean no tick value is required when mappedLineData more than one value?

debjyoti-pandit: when it has more than one then it will automatically generate the tick values (max 2 as we have tick count 2)

This is because as if we have more points, then ticks will keep on increasing every min.

GowthamShanmugam: ack

GowthamShanmugam:
LGTM

SanjalKatiyar:
@debjyoti-pandit plz remove merge commit from the PR.

SanjalKatiyar:
/hold
commits are not clean

debjyoti-pandit:
> @debjyoti-pandit plz remove merge commit from the PR.

removed :+1:

packages/shared/dashboards/line-graph/line-graph.tsx
debjyoti-pandit: @bipuladh as you can see here in the below snip from the current master 

image

The issue is not about showing small numbers, it is happening because when we are getting the first input/metric. there is no other reference to create the scale. So the Line Chart is creating the scale with the value received +/- a very small delta which is displayed as those values.
Once we have 2 values, the Line Chart creates the proper scale and then removes those values.

Let me know if my understanding is wrong, and if I need to update the implementation.

SanjalKatiyar:
/retitle Range value is not properly displayed for performance card

bipuladh (approved):
/lgtm

PR #434, opened 2022-09-27
Improves build time by 2x

bipuladh:
Brings production build to 25 seconds from 50 seconds.
Brings dev builds to 11 seconds from almost a minute.
Adds ForkTsChecker to save time in type checking.

Screenshot from 2022-09-27 15-16-25


webpack.config.ts
```diff
@@ -41,10 +45,18 @@ const config: webpack.Configuration & DevServerConfiguration = {
...
+            loader: 'thread-loader',
+            options: {
+              ...(NODE_ENV === 'development' ? { poolTimeout: Infinity } : {}),
+            },
+          },
           {
             loader: 'ts-loader',
             options: {
               configFile: path.resolve(__dirname, 'tsconfig.json'),
+              transpileOnly: true,
```
SanjalKatiyar, line 58: its mentioned that `transpileOnly` needed if `ts-loader < 9.3.0`
https://www.npmjs.com/package/fork-ts-checker-webpack-plugin

webpack.config.ts
```diff
@@ -58,7 +70,12 @@ const config: webpack.Configuration & DevServerConfiguration = {
         ],
         use: [
           { loader: 'cache-loader' },
-          { loader: 'thread-loader' },
+          {
+            loader: 'thread-loader',
+            options: {
+              ...(NODE_ENV === 'development' ? { poolTimeout: Infinity } : {}),
+            },
+          },
```
SanjalKatiyar, line 78: shouldn't we add this before `cache-loader` as well ?

webpack.config.ts
SanjalKatiyar: but I guess no harm in adding

webpack.config.ts
```diff
@@ -41,10 +45,18 @@ const config: webpack.Configuration & DevServerConfiguration = {
         test: /(\.jsx?)|(\.tsx?)$/,
         exclude: /node_modules/,
         use: [
+          {
+            loader: 'thread-loader',
+            options: {
+              ...(NODE_ENV === 'development' ? { poolTimeout: Infinity } : {}),
+            },
+          },
```
SanjalKatiyar, line 53: maybe my understanding is little bit wrong... but do we need `thread-loader` here ??
we only have one more loader (`ts-loader`) for this rule that should anyway use separate thread ??
in case of multiple following loaders we can use `thread-loader`...

SanjalKatiyar: maybe I am missing something ?

PR #436, opened 2022-09-27
Bug 2118683: [release-4.11] Fix PVC tests 

vbnrh:
- Increase the waiting for PVC to get bound
- Delete the PVC if it already exists in case cypress retries the tests

Failure on PVC tests previously -
https://user-images.githubusercontent.com/57935785/192537745-9e5530b4-95dc-4686-9175-1a4ef2777c0f.mp4

### DIFF
--- a/cypress/tests/expand-pvc.spec.ts
+++ b/cypress/tests/expand-pvc.spec.ts
@@ -17,13 +17,17 @@ describe('Tests Expansion of a PVC', () => {
   });
 
   it('Test expansion of a CephFS PVC', () => {
-    pvc.createPVC('testpvcfs', '5', 'ocs-storagecluster-cephfs');
+    const cephPvcName = 'testpvcfs';
+    pvc.deleteIfExists(cephPvcName, 'default');
+    pvc.createPVC(cephPvcName, '5', 'ocs-storagecluster-cephfs');
     pvc.expandPVC('10');
     cy.byTestID('pvc-requested-capacity').contains('10 GiB');
   });
 
   it('Test expansion of a RBD PVC', () => {
-    pvc.createPVC('testpvcrbd', '5', 'ocs-storagecluster-ceph-rbd', 'Block');
+    const rbdPvcName = 'testpvcrbd';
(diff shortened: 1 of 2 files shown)


cypress/views/pvc.ts
```diff
@@ -15,7 +15,10 @@ export const pvc = {
       cy.byTestID('Block-radio-input').click();
     }
     cy.byTestID('create-pvc').click();
-    cy.contains('Bound', { timeout: 10000 });
+    cy.contains('Bound', { timeout: 30000 });
+  },
+  deleteIfExists: (pvcName: string, namespace: string) => {
+    cy.exec(`oc delete pvcs ${pvcName} -n ${namespace}`);
```
SanjalKatiyar, line 21: won't it give error if PVC not found ?
```suggestion
    cy.exec(`oc delete pvcs ${pvcName} -n ${namespace}`, {failOnNonZeroExit: false });
```

cypress/tests/expand-pvc.spec.ts
```diff
@@ -17,13 +17,17 @@ describe('Tests Expansion of a PVC', () => {
   });
 
   it('Test expansion of a CephFS PVC', () => {
-    pvc.createPVC('testpvcfs', '5', 'ocs-storagecluster-cephfs');
+    const cephPvcName = 'testpvcfs';
+    pvc.deleteIfExists(cephPvcName, 'default');
+    pvc.createPVC(cephPvcName, '5', 'ocs-storagecluster-cephfs');
```
SanjalKatiyar, line 22: do we need to specify that PVC should get created in `default` NS ?
can it ever happen that it gets created in any other NS and in `deleteIfExists` we are expecting it to be in `default`

cypress/views/pvc.ts
vbnrh: I have made the changes. thanks

cypress/tests/expand-pvc.spec.ts
vbnrh: It gets created in default namespace for our tests. I have not seen any other namespace being used

PR #440, opened 2022-09-28
Adds bundle analyzer

bipuladh:
No asset should be larger than 3 MB
Adds script to verify asset size
Updates build job to parallelize test runs

On Error:
Screenshot from 2022-09-28 17-04-40

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -5,3 +5,4 @@ locales
 .husky
 .vscode
 *.scss
+plugins/**/package.json
--- a/analyzeTest.ts
+++ b/analyzeTest.ts
@@ -1,0 +1,55 @@
+import * as fs from 'fs';
+
+const pluginName = process.env.PLUGIN;
+const MAX_ASSET_SIZE = 3; //3 MiB
+
+const getStatsFilePath = () => `./plugins/${pluginName}/dist/stats.json`;
+
+const toMiB = (value: number) => value / 1024 ** 2;
+
+const stringifyMiB = (value: ReturnType<typeof toMiB>): string =>
+  `${value.toFixed(2)} MB`;
+
+const getParsedStatFile = () => {
+  const filePath = getStatsFilePath();
+  const statsFile = fs.readFileSync(filePath, 'utf-8');
+  return JSON.parse(statsFile);
+};
+
+type BundleDataMap = Record<string, number>;
+
+// [Valid Bundles, Violating Bunldes]
+type GetBundleInformation = () => [BundleDataMap, BundleDataMap];
+
+const getBundleInformation: GetBundleInformation = () => {
+  const statsData = getParsedStatFile();
+  const validAssets = {};
+  const violatingAssets = {};
+
+  statsData.assets.forEach((asset) => {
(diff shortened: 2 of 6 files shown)


webpack-analyze.config.ts
```diff
@@ -0,0 +1,147 @@
...
+          semantic: true,
+          syntactic: true,
+        },
+      },
+    }),
+    new BundleAnalyzerPlugin({
+      analyzerMode: 'static',
+      generateStatsFile: true,
+      openAnalyzer: false,
+    }),
```
SanjalKatiyar, line 138: instead of creating entire new file, can we add this plugin to `webpack.config.ts` file based on some condition ??
like if we are running analyse command then add it to the list else ignore it for normal build command ??

.prettierignore
```diff
@@ -5,3 +5,4 @@ locales
 .husky
 .vscode
 *.scss
+plugins/**/package.json
```
alfonsomthd, line 8: :+1:

analyzeTest.ts
```diff
@@ -0,0 +1,55 @@
...
+const toMB = (value: number) => value / 1024 ** 2;
+
+const stringifyMB = (value: ReturnType<typeof toMB>): string =>
+  `${value.toFixed(2)} MB`;
+
+const getParsedStatFile = () => {
+  const filePath = getStatsFilePath();
+  const file = fs.readFileSync(filePath, 'utf-8');
+  const statsData = JSON.parse(file);
+  return statsData;
```
alfonsomthd, line 16:
```suggestion
  return JSON.parse(file);
```

analyzeTest.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const getBundleInformation: GetBundleInformation = () => {
+  const statsData = getParsedStatFile();
+  const validAssets = {};
+  const violatingAssets = {};
+
+  statsData.assets.forEach((asset) => {
+    const assetSize = toMB(asset.size);
+    const readableSize = stringifyMB(assetSize);
+    if (assetSize > 3) {
```
alfonsomthd, line 32: Just a naming suggestion so the role of the number 3 is more intentional.
```suggestion
    const maxAllowedSize = 3;
    if (assetSize > maxAllowedSize) {
```

analyzeTest.ts
```diff
@@ -0,0 +1,55 @@
+import * as fs from 'fs';
+
+const pluginName = process.env.PLUGIN;
+
+const getStatsFilePath = () => `./plugins/${pluginName}/dist/stats.json`;
+
+const toMB = (value: number) => value / 1024 ** 2;
+
+const stringifyMB = (value: ReturnType<typeof toMB>): string =>
+  `${value.toFixed(2)} MB`;
```
SanjalKatiyar, line 10: nits (for important):
```suggestion
const toMiB = (value: number) => value / 1024 ** 2;

const stringifyMiB = (value: ReturnType<typeof toMiB>): string =>
  `${value.toFixed(2)} MiB`;
```

alfonsomthd (approved):
Tested locally: LGTM

PR #443, opened 2022-09-28
Update staged linter

bipuladh:


bipuladh:
Should go in after https://github.com/red-hat-storage/odf-console/pull/440/

bipuladh:
Overriding as it's changing how staged file linting works.
override ci/prow/images

PR #450, opened 2022-09-29
Build: reduce size: enable production mode and minification for prod.

alfonsomthd:
* Enabled production mode: removed `mode` from webpack config as [Webpack documentation](https://webpack.js.org/configuration/mode/#usage) says: "If mode is not provided via configuration or CLI, CLI will use any valid NODE_ENV value for mode."
  Added fallback 'development' value for NODE_ENV in config.
* Enabled minification for production.
* Added yarn script `serve-local-build` for local testing of a production build without TLS certificates.
* `dist` folder size reduction:
  ODF: 14M ==> 2,6M
  MCO:  6M ==> 1,2M


package.json
```diff
@@ -28,6 +28,8 @@
     "dev": "PLUGIN=odf I8N_NS=plugin__odf-console yarn server:plugin",
     "dev-mco": "PLUGIN=mco I8N_NS=plugin__odf-multicluster-console yarn server:plugin",
     "http-server": "./http-server.sh ./dist",
```
bipuladh, line 30:
```suggestion
```

package.json
```diff
@@ -28,6 +28,8 @@
...
+    "serve-local-build": "http-server plugins/odf/dist -p 9001 -c-1 --cors",
+    "serve-local-build-mco": "http-server plugins/mco/dist -p 9001 -c-1 --cors",
```
bipuladh, line 32:
```suggestion
    "serve-local-build": "./http-server plugins/odf/dist -p 9001 -c-1 --cors",
    "serve-local-build-mco": "./http-server plugins/mco/dist -p 9001 -c-1 --cors",
```
We might have to change the script to remove the requirement for certs.

SanjalKatiyar: can't we have script accept certs (as it is doing now)...
and here directly execute `http-server plugins/odf/dist -p 9001 -c-1 --cors` instead of calling the `http-server.sh` script... like what Alfonso already did ??
we don't have to update script that way

package.json
alfonsomthd: Done.

package.json
alfonsomthd: Updated http-server.sh and adapted Dockerfiles.

PR #451, opened 2022-09-30
Add config file path to eslint package.json

bipuladh:

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
     "node": ">=14.16.0"
   },
   "lint-staged": {
-    "*.{ts,tsx}": "eslint --cache --fix",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -135,7 +135,7 @@
     "node": ">=14.16.0"
   },
   "lint-staged": {
-    "*.{ts,tsx}": "eslint --cache --fix",
+    "*.{ts,tsx}": "eslint --cache --fix -c ./.eslintrc.js",
```
SanjalKatiyar, line 138: we don't have to give path for stylelint and prettier config files as well ??

bipuladh: for some reason eslint is not detecting it. For prettier and stylelint I don't see any problems as of now.

SanjalKatiyar: ack...

PR #466, opened 2022-10-04
Cleanup: enable no-extraneous-dependencies & no-unused-vars

alfonsomthd:
Fix lint issues detected by the following enabled rules:
* **no-extraneous-dependencies**: add explicit dependencies for imports that  previously were relying on dependencies of dependencies (so if a future version of a dependency drops the imported one, our code would break).
  * `import * as _ from 'lodash-es';` is now the only allowed `lodash`import as our current dependency is `lodash-es`.
* **@typescript-eslint/no-unused-vars**:
  *  Removed unnecessary `lodash` imports.
  * Removed unused vars.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -57,7 +57,10 @@ module.exports = {
     'default-case': 'off',
     'global-require': 'off',
     'implicit-arrow-linebreak': 'off',
-    'import/no-extraneous-dependencies': 'off',
+    'import/no-extraneous-dependencies': [
+      'error',
+      { packageDir: ['.', './packages/odf'] },
+    ],
     'jsx-a11y/anchor-is-valid': 'off',
     'jsx-a11y/click-events-have-key-events': 'off',
     'jsx-a11y/label-has-associated-control': 'off',
@@ -92,7 +95,7 @@ module.exports = {
     'react/no-unused-state': 'off',
     'react/sort-comp': 'off',
     'no-unused-vars': 'off',
-    '@typescript-eslint/no-unused-vars': 'off',
+    '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
     'react/function-component-definition': [
       2,
       {
--- a/cypress/helpers/add-capacity.ts
+++ b/cypress/helpers/add-capacity.ts
@@ -1,4 +1,4 @@
-import * as _ from 'lodash';
+import * as _ from 'lodash-es';
 
 export const createOSDTreeMap = (nodes) =>
   nodes.reduce((acc, curr) => Object.assign(acc, { [curr.id]: curr }), {});
--- a/package.json
+++ b/package.json
@@ -51,11 +51,16 @@
     "@patternfly/react-core": "4.175.4",
     "@patternfly/react-icons": "^4.11.0",
     "@patternfly/react-table": "4.104.7",
+    "@patternfly/react-tokens": "^4.12.0",
     "@types/react-dnd-html5-backend": "^3.0.2",
     "classnames": "^2.3.1",
     "eslint-config-airbnb": "18.2.1",
     "fuzzysearch": "^1.0.3",
+    "history": "^4.9.0",
+    "i18next": "^20.2.1",
     "i18next-conv": "^10.1.0",
+    "js-base64": "^2.1.9",
+    "js-yaml": "^3.13.1",
     "lodash-es": "^4.17.21",
     "murmurhash-js": "^1.0.0",
     "react": "^17.0.1",
@@ -69,7 +74,9 @@
     "react-redux": "^7.2.6",
     "react-router": "5.2.0",
     "react-router-dom": "5.2.0",
-    "react-tagsinput": "^3.19.0"
+    "react-tagsinput": "^3.19.0",
+    "redux": "^4.0.0",
+    "victory-core": "^35.4.4"
   },
   "devDependencies": {
     "@cypress/webpack-preprocessor": "^5.9.1",
--- a/cypress/tests/add-capacity.spec.ts
+++ b/cypress/tests/add-capacity.spec.ts
@@ -1,4 +1,4 @@
-import * as _ from 'lodash';
+import * as _ from 'lodash-es';
 import {
   CLUSTER_STATUS,
   STORAGE_SYSTEM_NAME,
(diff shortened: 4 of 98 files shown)


package.json
```diff
@@ -44,18 +44,27 @@
     "memsource-upload": "./i18n-scripts/memsource-upload.sh",
     "memsource-download": "./i18n-scripts/memsource-download.sh"
   },
+  "bundleDependencies": [
+    "@odf/shared"
+  ],
   "dependencies": {
+    "@openshift-console/dynamic-plugin-sdk": "0.0.15",
```
SanjalKatiyar, line 51: it is actually a dev-dependency, it only have types required by the modules, actual components are still served by OCP during runtime...

package.json
```diff
@@ -44,18 +44,27 @@
     "memsource-upload": "./i18n-scripts/memsource-upload.sh",
     "memsource-download": "./i18n-scripts/memsource-download.sh"
   },
+  "bundleDependencies": [
+    "@odf/shared"
+  ],
```
SanjalKatiyar, line 49: instead of here, shouldn't we add `bundleDependencies` to the respective package's  `package.json` file ?? root itself is a mono-repo...

SanjalKatiyar: also, do we even need this ?? I don't think we will ever publish mco/ocs/odf package. We can publish shared package if ever required, but that is independent of any other package. Is there any use case I maybe missing ??

cypress/helpers/add-capacity.ts
```diff
@@ -1,4 +1,4 @@
-import * as _ from 'lodash';
+import * as _ from 'lodash-es';
```
bipuladh, line 1: Rather than doing this. Let's replace all module usage of `lodash` with `lodash-es` in webpack.

.eslintrc.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
     ecmaVersion: 12,
     sourceType: 'module',
   },
-  plugins: ['react', '@typescript-eslint', 'import'],
+  plugins: ['react', '@typescript-eslint', 'import', 'unused-imports'],
```
bipuladh, line 20: `@typescript-eslint` should have `unused-imports` rules.

package.json
```diff
@@ -44,18 +44,27 @@
     "memsource-upload": "./i18n-scripts/memsource-upload.sh",
     "memsource-download": "./i18n-scripts/memsource-download.sh"
   },
+  "bundleDependencies": [
```
bipuladh, line 47: I do not see this required for us as we are not going to publish it to npmjs.

package.json
```diff
@@ -69,11 +78,12 @@
     "react-redux": "^7.2.6",
     "react-router": "5.2.0",
     "react-router-dom": "5.2.0",
-    "react-tagsinput": "^3.19.0"
+    "react-tagsinput": "^3.19.0",
+    "redux": "^4.0.0",
```
bipuladh, line 82: Why `redux`?

package.json
```diff
@@ -69,11 +78,12 @@
...
+    "victory-core": "^35.4.4"
```
bipuladh, line 83: Why `victory core`?

package.json
```diff
@@ -44,18 +44,27 @@
...
+    "@patternfly/react-tokens": "^4.12.0",
     "@types/react-dnd-html5-backend": "^3.0.2",
     "classnames": "^2.3.1",
     "eslint-config-airbnb": "18.2.1",
     "fuzzysearch": "^1.0.3",
+    "history": "^4.9.0",
+    "i18next": "^20.2.1",
     "i18next-conv": "^10.1.0",
+    "js-base64": "^2.1.9",
+    "js-yaml": "^3.13.1",
```
bipuladh, line 67: Why these packages?

package.json
bipuladh, line 51: Lets keep it as `devDependency` it only exposes types.

bipuladh (changes requested):

cypress/helpers/add-capacity.ts
alfonsomthd: Is that not a dirty trick? IMHO is better to have the correct imports than to do a trick in webpack....

package.json
alfonsomthd: This is only to tell the linter that the imports from @odf/shared are inside the project.

cypress/helpers/add-capacity.ts
bipuladh: IMO we will keep on missing this during reviews and people will definitely keep on messing the package name. Doing that would make sure we always use es modules.

package.json
alfonsomthd: It's being imported here:
https://github.com/red-hat-storage/odf-console/blob/master/packages/ocs/dashboards/persistent-internal/raw-capacity-card/raw-capacity-card.tsx#L8

package.json
bipuladh: Doesn't linter also lint code on `@odf/shared`? I am not sure how we are helping linter with this directive.

package.json
alfonsomthd: Because there are explicit imports like in:
packages/shared/utils/SecretValue.tsx
packages/shared/yaml-editor/YAMLEditor.tsx

package.json
alfonsomthd: This is only to tell the linter that the imports from @odf/shared are inside the project.

.eslintrc.js
alfonsomthd: I'll look at it!

package.json
alfonsomthd: It's being imported:
packages/shared/charts/tooltip.tsx

cypress/helpers/add-capacity.ts
alfonsomthd: This approach defeats the purpose of the `no-extraneous-dependencies` which forbids imports not declared...

package.json
alfonsomthd: Addressed.

package.json
alfonsomthd: I updated the linter config so we no longer require this.

package.json
alfonsomthd: Done.

cypress/helpers/add-capacity.ts
alfonsomthd: As per our previous chat conversation, we agree on using the imports from our dependencies.
The linter will throw an error if someone tries to import from a non-existing dependency so the pre-commit and PR checks will fail.

bipuladh (approved):
/lgtm

PR #470, opened 2022-10-07
migrate/enable + fix Vault KMS cypress test cases

SanjalKatiyar:


cypress/helpers/vault.ts
```diff
@@ -0,0 +1,359 @@
+export const roleBindingJSON = {
```
vbnrh, line 1: Can we use these as yaml files instead of using variables ?

cypress/views/common.ts
```diff
@@ -1,8 +1,8 @@
 export const commonFlows = {
   navigateToODF: () => {
     cy.clickNavLink(['Storage', 'Data Foundation']);
-    //cy.byTestOperatorRow('OpenShift Data Foundation').click();
   },
+  // eslint-disable-next-line cypress/require-data-selectors
```
bipuladh, line 5: Add a comment on why we are disabling this rule.

cypress/support/vault-standalone.ts
```diff
@@ -0,0 +1,81 @@
...
+      cy.log('Vault is already deployed');
+    }
+  });
+};
+
+export const isPodRunningWithEncryptedPV = () => {
+  cy.log('Checking pod is up and running with encrypted PV');
+  commandPoll(
+    `oc get Deployment ${testDeploymentJSON.metadata.name} -n default -ojsonpath='{.status.availableReplicas}'`,
+    '1'
```
bipuladh, line 79: What about using oc wait?

cypress/helpers/vault.ts
```diff
@@ -0,0 +1,359 @@
...
+  metadata: {
+    labels: {
+      'app.kubernetes.io/instance': 'vault',
+      'app.kubernetes.io/name': 'vault',
+    },
+    name: 'vault',
+  },
+};
+
+export const getPVCJSON = {
```
bipuladh, line 37:
```suggestion
export const PVCJSON = {
```
Not a function.

cypress/helpers/vault.ts
```diff
@@ -0,0 +1,359 @@
...
+            port: 8200,
+            protocol: 'TCP',
+          },
+        ],
+      },
+    ],
+  },
+};
+
+export const deploymentJson = {
```
bipuladh, line 160:
```suggestion
export const deploymentJSON = {
```
Lets keep the case same. Do we really need to mention these are `JSONs` in the variable name though?

cypress/tests/kms-encryption-sc.ts
```diff
@@ -0,0 +1,22 @@
...
+import { pvc } from '../views/pvc';
+import { createStorageClass } from '../views/storage-class';
+
+describe('Test Ceph pool creation', () => {
+  before(() => {
+    configureVault();
+    cy.login();
+    cy.visit('/');
+    cy.install();
+  });
```
bipuladh, line 14: Missing after block.

cypress/support/vault-standalone.ts
SanjalKatiyar: > oc wait Deployment test1 -n default --for=jsonpath='{.status.availableReplicas}'='1' --timeout=300s
will give error if all the replicas are in not available state, because status.availableReplicas field will not be existing on the CR.
whereas,
> commandPoll( oc get... )
this will not return any error in such case and will keep on retrying (for a specified period) until it finds the correct "availableReplicas" field.

cypress/tests/kms-encryption-sc.ts
SanjalKatiyar: done...

bipuladh (approved):
/lgtm

PR #479, opened 2022-10-10
[wip] Write e2e tests for Compression

bipuladh:


cypress/support.ts
```diff
@@ -87,3 +88,15 @@ Cypress.Commands.add('install', () => {
...
+Cypress.Commands.add('enableToolboxPod', () => {
+  cy.exec(
+    `oc patch ocsinitialization ocsinit -n openshift-storage --type json --patch  '[{ "op": "replace", "path": "/spec/enableCephTools", "value": true }]'`
+  );
+  // wait for the pod CR to be registered
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
+  cy.wait(2500);
+  cy.exec(
+    `oc wait --for=condition=ready pod -l "app=rook-ceph-tools" -n openshift-storage`
+  );
```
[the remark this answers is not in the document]
SanjalKatiyar: also "commandPoll" make 1 call per sec, so it is not too much as well...

cypress/tests/compression.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+} from '../mocks/compression';
+import { navigateToBlockPool } from '../views/block-pool';
+
+describe('Test Pool Compression', () => {
+  before(() => {
+    cy.login();
+    cy.visit('/');
+    cy.install();
+    cy.enableToolboxPod();
+    cy.exec(`echo '${JSON.stringify(compressionPool)}' | oc apply -f -`);
```
SanjalKatiyar, line 16: we have specific UI for this (we are maintain it as well, so we can easily add ids to elements), we should be creating this using a utility function (like we have createPVC, createStorageClass etc), instead of CLI...

cypress/tests/compression.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    cy.exec(
+      `echo '${JSON.stringify(compressionStorageClass)}' | oc apply -f -`
+    );
+    cy.exec(`echo '${fioConfig}' | oc apply -f -`);
+    cy.exec(`echo '${JSON.stringify(fioPVC)}' | oc apply -f -`);
+    cy.exec(`echo '${JSON.stringify(fioJob)}' | oc apply -f -`);
+    // Allow job CR to be registered
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(2500);
+    cy.exec(`oc wait --for=condition=complete job fio`, { timeout: 120000 });
```
SanjalKatiyar, line 26: my suggestion would be to replace all explicit cy.wait conditions with "commandPoll" everywhere we can... there is some drawback of extra network call, but don't have to worry about flakeness due to cy.wait ever...
wdyt ?

cypress/support.ts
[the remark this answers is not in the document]
SanjalKatiyar: or if "commandPoll" won't work,
can we have some utility function similar to "commandPoll", which will continuously poll for a resource to check if it gets created or not instead of explicitly using cy.wait ??

cypress/tests/compression.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    cy.exec(`echo '${JSON.stringify(compressionPool)}' | oc apply -f -`);
+    cy.exec(
+      `echo '${JSON.stringify(compressionStorageClass)}' | oc apply -f -`
+    );
+    cy.exec(`echo '${fioConfig}' | oc apply -f -`);
+    cy.exec(`echo '${JSON.stringify(fioPVC)}' | oc apply -f -`);
+    cy.exec(`echo '${JSON.stringify(fioJob)}' | oc apply -f -`);
+    // Allow job CR to be registered
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(2500);
```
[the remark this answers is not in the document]
bipuladh: As mentioned above we are not waiting for a resource to get created but for the CR to be registered. we could reduce it to 1000ms(1s) that should also work.

cypress/tests/compression.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+describe('Test Pool Compression', () => {
+  before(() => {
+    cy.login();
+    cy.visit('/');
+    cy.install();
+    cy.enableToolboxPod();
+    cy.exec(`echo '${JSON.stringify(compressionPool)}' | oc apply -f -`);
+    cy.exec(
+      `echo '${JSON.stringify(compressionStorageClass)}' | oc apply -f -`
+    );
```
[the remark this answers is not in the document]
SanjalKatiyar: not needed here... I agree...

cypress/tests/compression.spec.ts
SanjalKatiyar: out of scope of this PR, just fyi...
we need to enable pool tests again... I checked just now and realised they are disabled !!

cypress/views/block-pool.ts
```diff
@@ -20,7 +20,7 @@ export const poolMessage: {
 
 export const navigateToBlockPool = () => {
   ODFCommon.visitStorageDashboard();
-  cy.byLegacyTestID('horizontal-link-Storage Systems').click();
+  cy.byTestID('horizontal-link-Storage Systems').click();
```
SanjalKatiyar, line 23: on my local, I can still see attribute as "data-test-id" only. Same is true for other tests as well, they are using "byLegacyTestID", just confirming if it is intentional change or a miss ?

cypress/views/block-pool.ts
```diff
@@ -20,7 +20,7 @@ export const poolMessage: {
...
   cy.byLegacyTestID('item-filter').type('ocs-storagecluster-storagesystem');
   cy.byTestRows('resource-row').get('td a').first().click();
   cy.byLegacyTestID('horizontal-link-BlockPools').click();
```
SanjalKatiyar, line 26: this is "data-test" instead...

alfonsomthd:
@bipuladh @SanjalKatiyar should we create an issue ticket to not forget about this test coverage?

PR #481, opened 2022-10-11
fix: made wizard body height fixed to avoid outer scroll

debjyoti-pandit:
fixes: https://issues.redhat.com/browse/RHSTOR-2161

Before:
image

After:
image

Before: https://drive.google.com/file/d/1u8Sfm-CuKuqVux7L2cEMr4wKOAEpw06x/view?usp=sharing
After: https://drive.google.com/file/d/16aBbk9fwh38xbTPR741EQc1dzKr8U2Qo/view?usp=sharing

External Storage Platform - IBM Flash SC
Before: https://drive.google.com/file/d/18xHg9NxftG-HIFhDfIaqM1iq3S4X_-IZ/view?usp=sharing
After: https://drive.google.com/file/d/10WtYiXiU46iW_gfF6psr6nvluyGX5cMr/view?usp=sharing

### DIFF
--- a/packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
+++ b/packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
@@ -374,7 +374,7 @@ export const CapacityAndNodes: React.FC<CapacityAndNodesProps> = ({
   );
 
   return (
-    <Form className="odf-capacity-and-node-form">
+    <Form>
       {isNoProvisioner ? (
         <SelectedCapacityAndNodes
           storageClassName={storageClass.name || volumeSetName}
--- a/packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
+++ b/packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
@@ -374,7 +374,7 @@ export const CapacityAndNodes: React.FC<CapacityAndNodesProps> = ({
   );
 
   return (
-    <Form>
+    <Form className="odf-capacity-and-node-form">
       {isNoProvisioner ? (
         <SelectedCapacityAndNodes
           storageClassName={storageClass.name || volumeSetName}
--- a/packages/odf/components/create-storage-system/create-storage-system.scss
+++ b/packages/odf/components/create-storage-system/create-storage-system.scss
@@ -21,3 +21,7 @@
     color: var(--pf-global--danger-color--200);
   }
 }
+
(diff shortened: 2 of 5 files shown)


packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -374,7 +374,7 @@ export const CapacityAndNodes: React.FC<CapacityAndNodesProps> = ({
   );
 
   return (
-    <Form>
+    <Form className="odf-capacity-and-node-form">
```
bipuladh, line 377: Why not on the wizard?

debjyoti-pandit: Wizard is common for all the screens, I thought the one which is causing the issue can have custom height.

packages/odf/components/create-storage-system/create-storage-system.scss
```diff
@@ -21,3 +21,7 @@
     color: var(--pf-global--danger-color--200);
   }
 }
+
+.pf-c-wizard__inner-wrap {
+  max-height: 60vh;
```
bipuladh, line 26: Is this the max we can go? we should be able to go more? Also please do not override PF class instead pass a className to Wizard and use that here.

bipuladh (changes requested):

packages/odf/components/create-storage-system/create-storage-system.scss
```diff
@@ -21,3 +21,7 @@
     color: var(--pf-global--danger-color--200);
   }
 }
+
+.odf-create-storage-system-wizard > div > div {
+  max-height: 66vh;
```
GowthamShanmugam, line 26: cc: @bipuladh @SanjalKatiyar Please take a look this CSS change

SanjalKatiyar:
can you try once by zooming the tab to around 125-150% and see if it is still working as expected ??

debjyoti-pandit:
> can you try once by zooming the tab to around 125-150% and see if it is still working as expected ??

Did try with increasing it to 110%. There was 2 scroll bar. Increasing it to 125% will definitely add one extra scroll bar.
If we want to remove the extra scroll bar when in 110% then we have to reduce the height to 60vh which I initially had. I was also considering it.

But the point, it are we going to support 110%. If yes then we should also plan for 125%. In that case, we have to reduce it to 54vh.

Let me know about this. 

In my opinion, we can develop based on 100% as we can control how much % of users we see in their browser.

debjyoti-pandit:
@SanjalKatiyar If we need to support 150% then it will look as below 

image

The footer and the heading is also increasing when we are increasing the zoom level. Which is bringing the scrollbar.

PR #482, opened 2022-10-11
Update Nginx cache configuration

bipuladh:

### DIFF
--- a/default.conf
+++ b/default.conf
@@ -1,10 +1,3 @@
-map $sent_http_content_type $expires {
-    default                    off;
-    text/html                  epoch;
-    text/css                   max;
-    application/javascript     epoch;
-}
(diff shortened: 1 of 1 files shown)


default.conf
```diff
@@ -21,5 +14,9 @@ server {
         root   /usr/share/nginx/html;
     }
     ssi on;
-    expires $expires;
+    add_header Last-Modified $date_gmt;
+    add_header Cache-Control 'no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0';
+    if_modified_since off;
```
SanjalKatiyar, line 19: not blocking the PR, but...
nits: "if_modified_since off", so no need to add "Last-Modified" header to the response...

PR #487, opened 2022-10-12
migrate(obc): css co classes to odf specific

debjyoti-pandit:
**Current**
image

______________________________________________________________________________________

**Current**
image

_______________________________________________________________________________________

**Current**
image

_______________________________________________________________________________________

**Current**
image

### DIFF
--- a/packages/odf/style.scss
+++ b/packages/odf/style.scss
@@ -1,1 +1,1 @@
-@import '../shared/styles/skeleton';
+@import '../shared/style.scss';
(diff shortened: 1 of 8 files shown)


packages/odf/style.scss
```diff
@@ -1 +1,3 @@
+@import '../shared/styles/vars';
+@import '../shared/styles/common';
```
GowthamShanmugam, line 2: Please check style.scss in shared, which will resolve this vars import

GowthamShanmugam: just importing style.scss from shared is enough

debjyoti-pandit: Done

GowthamShanmugam (approved):

debjyoti-pandit:
This latest push doesn't contain any new changes, I have only rebased the branch with master.

bipuladh (approved):
/lgtm

PR #493, opened 2022-10-14
add number increment input for sync interval

SanjalKatiyar:
https://bugzilla.redhat.com/show_bug.cgi?id=2114501
Screenshot from 2022-10-14 16-49-30

### DIFF
--- a/packages/mco/components/disaster-recovery/create-dr-policy/select-replication-type.tsx
+++ b/packages/mco/components/disaster-recovery/create-dr-policy/select-replication-type.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
 import {
   FormGroup,
-  TextInput,
+  NumberInput,
   Dropdown,
   DropdownItem,
   DropdownToggle,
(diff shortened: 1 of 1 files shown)


packages/mco/components/disaster-recovery/create-dr-policy/select-replication-type.tsx
```diff
@@ -38,20 +42,27 @@ const SyncSchedule: React.FC<SyncScheduleProps> = ({ state, dispatch }) => {
...
     const newScheduleFormat = event.target.value;
     setIsOpen(false);
     setSelectedFormat(newScheduleFormat);
     setSyncSchedule(scheduleTime, newScheduleFormat);
   };
 
+  const onChange = (event: React.FormEvent<HTMLInputElement>) => {
+    const target = event.target as HTMLInputElement;
+    const syncTime = isNaN(+target.value) ? minSyncTime : Number(target.value);
+    const normalizedSyncTime = syncTime < minSyncTime ? minSyncTime : syncTime;
```
GowthamShanmugam, line 62: The minimum value won't allow Null, also there is no change of syncTime < minSyncTime. Is these two conditions required?

SanjalKatiyar: I can remove 2. if u think it is not needed... then user can only increase value by 1 at a time...

GowthamShanmugam: yes second one is not required, first one is fine

GowthamShanmugam (approved):

PR #494, opened 2022-10-16
migrate(status-card): css co classes to odf specific

debjyoti-pandit:
image


packages/shared/styles/_common.scss
```diff
@@ -48,3 +48,12 @@
     margin-bottom: 0 !important;
   }
 }
+
+.odf-dashboard-body {
```
GowthamShanmugam, line 52: instead of common lets move this inside dashboard.scss

debjyoti-pandit: Done

packages/shared/styles/_dashboard.scss
```diff
@@ -0,0 +1,8 @@
+.odf-dashboard-body {
```
bipuladh, line 1: Lets create a file named `dashboard.scss` next to `dashboard.tsx` file (same directory) and move this content there. This style might not be useful for every consumer of common.

PR #495, opened 2022-10-16
migrate(nss): css co classes to odf specific

debjyoti-pandit:


packages/odf/style.scss
```diff
@@ -1 +1,6 @@
+@import '../shared/styles/vars';
+@import '../shared/styles/mixin/break-word';
+
+@import '../shared/styles/alerts';
+@import '../shared/styles/common';
```
GowthamShanmugam, line 5: Importing a specific style instead of importing all is good But when someone imports from shared then they have to be very careful about importing whatever is required from shared/styles/

What about moving all shared imports to styles.scss inside the shared directory and import only style.scss here?

debjyoti-pandit: Done

packages/odf/style.scss
```diff
@@ -1 +1 @@
-@import '../shared/styles/skeleton';
+@import '../shared/style.scss';
```
bipuladh, line 1: Is skeleton no more required?

SanjalKatiyar:
if tests are failing try re-basing with latest master...

PR #508, opened 2022-10-20
Configure jest, RTL & add breakdown card unit test

alfonsomthd:
- Added jest configuration + React Testing Library (RTL) to run unit tests.
- Updated GitHub action to include `yarn test-coverage`.
- Added 1st unit test: capacity-breakdown-card.
- NOTE: jest config is kept as a JavaScript file because using a TypeScript file requires upgrading `ts-node` that in turn
  makes `yarn build` too slow (>200 s).


package.json
```diff
@@ -143,6 +148,29 @@
...
+    "transformIgnorePatterns": [
+      "<rootDir>/node_modules/(?!(@openshift-console)/)"
+    ],
+    "testPathIgnorePatterns": [
+      "<rootDir>/cypress/",
+      "<rootDir>/node_modules/",
+      "<rootDir>/plugins/"
+    ],
+    "testRegex": ".*\\.spec\\.[t]sx?$"
+  },
```
SanjalKatiyar, line 173: for later, when we mark this PR up for review...
let's have separate config file for jest as well...

jest.config.js
```diff
@@ -0,0 +1,24 @@
...
+  testEnvironment: 'jsdom',
+  transform: {
+    '\\.[jt]sx?$': '@swc/jest',
+  },
+  transformIgnorePatterns: ['<rootDir>/node_modules/(?!(@openshift-console)/)'],
+  testPathIgnorePatterns: [
+    '<rootDir>/cypress/',
+    '<rootDir>/node_modules/',
+    '<rootDir>/plugins/',
+  ],
```
SanjalKatiyar, line 20: Is it required even if we have added `roots: ['<rootDir>/packages/'],` ??

package.json
```diff
@@ -15,6 +15,9 @@
     "cypress-merge": "mochawesome-merge ./gui-test-screenshots/cypress_report*.json > ./gui-test-screenshots/cypress.json",
     "cypress-generate": "marge -o ./gui-test-screenshots/ -f cypress-report -t 'OpenShift Console Cypress Test Results' -p 'OpenShift Cypress Test Results' --showPassed false --assetsDir ./gui-test-screenshots/cypress/assets ./gui-test-screenshots/cypress.json",
     "cypress-postreport": "yarn cypress-merge && yarn cypress-generate",
+    "test": "jest",
+    "test-silent": "JEST_SILENT_REPORTER_DOTS=true jest --reporters jest-silent-reporter",
```
SanjalKatiyar, line 19: where are these 2 commands are being used ??

packages/shared/table/table.tsx
```diff
@@ -80,7 +80,7 @@ const Table: React.FC<TableProps> = React.memo(
     return (
       <TableComposable aria-label={ariaLabel} translate={null}>
         <Thead translate={null}>
-          <Tr translate={null}>{...headerColumns}</Tr>
+          <Tr translate={null}>{headerColumns}</Tr>
```
SanjalKatiyar, line 83: I wonder why this component was working correctly so far (maybe typescript is doing some magic) !! It's a good catch.

tsconfig.json
```diff
@@ -12,5 +12,6 @@
     "sourceMap": true,
     "noEmitOnError": false
   },
+  "exclude": ["**/*.spec.ts", "**/*.spec.tsx"],
```
SanjalKatiyar, line 15: If test files are of type `.ts` `.tsx` then why are we excluding them from ts config file ?? Maybe I am missing something ??

packages/ocs/dashboards/object-service/capacity-breakdown/capacity-breakdown-card.spec.tsx
```diff
@@ -0,0 +1,42 @@
...
+jest.mock('@openshift-console/dynamic-plugin-sdk/lib/utils/flags', () => ({
+  useFlag: () => true,
+}));
+
+jest.mock(
+  '@odf/shared/hooks/custom-prometheus-poll/custom-prometheus-poll-hook',
+  () => ({
+    useCustomPrometheusPoll: () => [true, false, false],
+  })
+);
```
SanjalKatiyar, line 24: can we add it to setup file as well ??

packages/ocs/dashboards/object-service/capacity-breakdown/capacity-breakdown-card.spec.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+jest.mock('@openshift-console/dynamic-plugin-sdk/lib/extensions', () => {
+  const originalModule = jest.requireActual<
+    typeof import('@openshift-console/dynamic-plugin-sdk/lib/extensions')
+  >('@openshift-console/dynamic-plugin-sdk/lib/extensions');
+  return {
+    ...originalModule,
+    useK8sWatchResource: () => [true, false, false],
+  };
+});
```
SanjalKatiyar, line 13: same this as well to setup file ??
these utilities will be shared by multiple test files...

jest.config.js
alfonsomthd: I think we can drop this as long we keep `roots` (or on reverse). For simplicity, we can keep `roots`.

package.json
alfonsomthd: These scripts are to be run on CLI  for running the unit tests. After a conversation with @bipuladh , we decided for now to not include the tests in pre-commit, so they are run in CI job as `yarn test-coverage`.
Any developer can run `yarn test` to check its code locally and `yarn test-silent` is meant to run your tests without your terminal being cluttered by the output (at Ceph Dashboard we have lots of unit tests and this is very convenient!).

tsconfig.json
alfonsomthd: We're excluding them from the build (you can check which files are to be included in the build by running `yarn tsc --showConfig` and checking "files").
Test files are to be used only by Jest (or any test library).

packages/ocs/dashboards/object-service/capacity-breakdown/capacity-breakdown-card.spec.tsx
alfonsomthd: The thing is for this unit test it's fine to return `[true, false, false]` but I'm not sure if for other tests that payload is OK or even if we need that mock for them in the first place...
I prefer to not promote a mock to the setup file until really needed (or there is a general consensus to do it).
@bipuladh Thoughts?

packages/ocs/dashboards/object-service/capacity-breakdown/capacity-breakdown-card.spec.tsx
alfonsomthd: Same as above.

package.json
SanjalKatiyar: ack !!

tsconfig.json
SanjalKatiyar: make sense... thanks !!

package.json
SanjalKatiyar: created this: https://github.com/red-hat-storage/odf-console/issues/546 for later.

PR #513, opened 2022-10-27
CI: drop Node.js EOL releases

alfonsomthd:
[Node.js EOL releases](https://github.com/nodejs/Release#end-of-life-releases) should not be supported as they do not receive neither security updates nor critical bug fixes.

* The Node.js version selected for CI is the same as in `Dockerfile.prod`.


SanjalKatiyar:
LGTM (just a suggestion: we can add this info to ReadMe on the versions we support and reasoning behind this).
let's wait for @bipuladh as well on this, so that he is in loop before merging this PR !

PR #544, opened 2022-11-22
fix flake with OBC test suite

SanjalKatiyar:


cypress/views/obcPage.ts
```diff
@@ -17,19 +17,21 @@ export class CreateOBCHandler {
...
-    cy.byTestID('item-create').click();
-    cy.byTestID('obc-name').type(this.name);
-    cy.byTestID('loading-indicator').should('not.exist');
-    cy.byTestID('sc-dropdown').should('be.visible').click();
-    cy.contains('openshift-storage.noobaa.io').click();
-    cy.byTestID('obc-create').click();
-    cy.byLegacyTestID('resource-title').contains(this.name, {
-      timeout: MINUTE,
+    cy.get('#page-sidebar').contains('Data Foundation'); // eslint-disable-line cypress/require-data-selectors
+    cy.clickNavLink(['Storage', 'Object Bucket Claims']).then(() => {
```
bipuladh, line 21: is `then` required ?

SanjalKatiyar: not really...
but I added it anyway, seems alright to me... but I can remove that if u suggest ??

bipuladh: lets remove it.

SanjalKatiyar: okay I will remove it...
just to confirm, can there be any issue with using then ??

bipuladh (approved):
/lgtm

PR #552, opened 2022-12-05
Fix relocate or failover flow is blocked from UI

GowthamShanmugam:
- When someone updates preferred cluster on the DRPC manually from CLI, Then UI is blocking the failover or relocate flow.
- UI should not depend on the preferred cluster to decide the target cluster.
- Fetch the target cluster from PlacementRule CR to decide to block or unblock the failover and relocation flow.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2143439


packages/mco/components/data-policies/app-data-policies-status/data-policies-status-popover.tsx
```diff
@@ -23,36 +23,27 @@ import {
...
+const getDRPolicies = (appDRInfoList: ApplicationDRInfo[]) =>
+  appDRInfoList.reduce((obj, appDRInfo) => {
+    const drPolicyName = getDRPolicyName(appDRInfo?.drPolicyControl);
     return {
       ...obj,
       [drPolicyName]: obj.hasOwnProperty(drPolicyName)
-        ? [...obj[drPolicyName], drpc]
-        : [drpc],
+        ? [...obj[drPolicyName], appDRInfo?.drPolicyControl]
+        : [appDRInfo?.drPolicyControl],
```
SanjalKatiyar, line 46: nits: `appDRInfo?.drPolicyControl` used at multiple places, can be stored in a temp variable.

packages/mco/utils/disaster-recovery.ts
```diff
@@ -118,6 +132,38 @@ export const getFilterDRSubscriptions = (
...
+  drPlacementControls?.reduce(
+    (acc, drPlacementControl) =>
+      isObjectRefMatching(
+        drPlacementControl?.spec?.placementRef,
+        Object.keys(subscriptionMap)
+      )
+        ? [
+            ...acc,
+            {
+              drPolicyControl: drPlacementControl,
```
SanjalKatiyar, line 149: do we need to store entire object that we are getting as a response ? can't we just store few fields that are needed ?

SanjalKatiyar: talking about single `drPlacementControl` object...

packages/mco/utils/disaster-recovery.ts
```diff
@@ -118,6 +132,38 @@ export const getFilterDRSubscriptions = (
...
+            },
+          ]
+        : acc,
+    []
+  );
+
+export const getPlacementClusterName = (
+  placementRuleMap: PlacementRuleMap,
+  drpc: DRPlacementControlKind
+) => placementRuleMap[drpc?.spec?.placementRef?.name] || '';
```
SanjalKatiyar, line 165:
```suggestion
) => placementRuleMap?.[drpc?.spec?.placementRef?.name] || '';
```

packages/mco/components/data-policies/app-data-policies-status/data-policies-status-popover.tsx
```diff
@@ -120,42 +111,42 @@ export const DataPoliciesStatusPopover: React.FC<DataPoliciesStatusPopoverProps>
...
         resource,
       ]
     );
 
     const [dataPoliciesStatus, count]: [DataPoliciesStatusType, number] =
       React.useMemo(() => {
-        const activeDRPC: DRPlacementControlKind[] =
+        const appDRInfoList: ApplicationDRInfo[] =
           (drPlacementControlsLoaded &&
             !drPlacementControlsLoadError &&
```
SanjalKatiyar, line 142: don't we need to check for `placementRulesLoaded` & `placementRulesLoadError` as well ??

SanjalKatiyar: if needed, no need to add placementRulesLoaded & placementRulesLoadError to dependency array,  instead just combine:
loaded = placementRulesLoaded  && drPlacementControlsLoaded
error = !!(placementRulesLoadError || drPlacementControlsLoadError)
and pass loaded & error to the dependency array...

packages/mco/components/disaster-recovery/app-failover-relocate/failover-relocate-modal-body.tsx
```diff
@@ -211,6 +187,7 @@ export const FailoverRelocateModalBody: React.FC<FailoverRelocateModalBodyProps>
       placementRulesLoadError,
       subscriptionsLoadError,
       subscriptionMap,
+      placementRuleMap,
```
SanjalKatiyar, line 190: here as well, why not let's combine all errors and loaded into single variable and pass that to dependency:
loaded = drPlacementControlsLoaded && placementRulesLoaded &&  subscriptionsLoaded
error = drPlacementControlsLoadError || placementRulesLoadError || subscriptionsLoadError

packages/mco/components/data-policies/app-data-policies-status/data-policies-status-popover.tsx
GowthamShanmugam: But both are same

GowthamShanmugam: We need to check both are not  loaded  or errors in both case i want dataPoliciesStatus to be empty list, otherwise UI might break.

packages/mco/components/disaster-recovery/app-failover-relocate/failover-relocate-modal-body.tsx
```diff
@@ -147,46 +127,36 @@ export const FailoverRelocateModalBody: React.FC<FailoverRelocateModalBodyProps>
...
-          ? getFilteredDRPlacementRuleNames(placementRules)
-          : [],
-      [placementRules, placementRulesLoaded, placementRulesLoadError]
+    const plsRuleLoaded = placementRulesLoaded && !placementRulesLoadError;
+    const subsLoaded = subscriptionsLoaded && !subscriptionsLoadError;
+    const drpcLoaded =
+      drPlacementControlsLoaded && !drPlacementControlsLoadError;
+
+    const placementRuleMap: PlacementRuleMap = React.useMemo(
+      () => (plsRuleLoaded ? filterDRPlacementRuleNames(placementRules) : {}),
```
SanjalKatiyar, line 136: "placementRules" can be empty as well, in that case "filterDRPlacementRuleNames" will return "undefined"
```suggestion
      () => (plsRuleLoaded && filterDRPlacementRuleNames(placementRules)) || {},
```

SanjalKatiyar:
for both the files "data-policies-status-popover.tsx" and "failover-relocate-modal-body.tsx":
I can see u split single "useMemo" into three diff "useMemo", still for majority of cases all three "useMemo" will gonna run for the same re-render, we are not saving computation instead adding more hooks for the React to take care of.
So, I would say its better to have only a single "useMemo" as all three current "useMemo" are interdependent.

For reducing the size of dependency array we can do the same u have done right now, combining "loaded" and "error" into a single variable and passing that.

GowthamShanmugam:
> for both the files "data-policies-status-popover.tsx" and "failover-relocate-modal-body.tsx": I can see u split single "useMemo" into three diff "useMemo", still for majority of cases all three "useMemo" will gonna run (within same re-render). We are saving little computation, but we are adding more hooks (which are dependent on one another as well) for React to take care of which could affect the performance. So, IMO its better to have only a single "useMemo" like before, as all three current "useMemo" are interdependent (mostly).
> 
> For reducing the size of dependency array we can do the same u are doing right now, combining "loaded" and "error" into a single variable and passing that.

But the logic which I used inside each hook is the problem, I dont want to loop again and again when anyone got changed. especially subscription is keep triggering the render. 

it is not much only 3 hooks. also, i dont want to do any refactoring for backport. it leads to new bug.

PR #556, opened 2022-12-07
Bug 2044157: Fixes storage cluster and scaling device set count issue

bipuladh:
For 4 or more zones it creates 4 or more OSDs


packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -130,6 +137,14 @@ const SelectCapacityAndNodes: React.FC<SelectCapacityAndNodesProps> = ({
     [dispatch]
   );
 
+  const zones = nodes.reduce((acc, curr) => {
+    acc.add(curr.zone);
+    return acc;
+  }, new Set<string>());
+
+  // We create 4 osds or more for more zones
```
GowthamShanmugam, line 145: a comment is not clear, we can specify even more clearly
We create 4 or more OSDs for more than three zones

packages/odf/components/utils/common.ts
```diff
@@ -294,11 +292,32 @@ export const prettifyJSON = (data: string) =>
...
 
+export const getDeviceSetReplica = (
+  isStretchCluster: boolean,
+  isFlexibleScaling: boolean,
+  nodes: WizardNodeState[]
+) => {
+  const uniqueZonesSet: Set<string> = nodes.reduce((acc, curr) => {
+    acc.add(curr.zone);
+    return acc;
+  }, new Set<string>());
```
GowthamShanmugam, line 303: This logic is used in more than one place, it is better create a util function

packages/odf/components/utils/common.ts
```diff
@@ -294,11 +292,32 @@ export const prettifyJSON = (data: string) =>
...
+
+  if (isFlexibleScaling) {
+    return OCS_DEVICE_SET_FLEXIBLE_REPLICA;
+  }
+  if (isStretchCluster) {
+    return uniqueZonesSet.size;
+  }
+
+  return uniqueZonesSet.size;
```
SanjalKatiyar, line 312:
```suggestion
// For both arbiter and normal deployment configurations
  return uniqueZonesSet.size;
```

packages/odf/components/utils/common.ts
```diff
@@ -294,11 +292,32 @@ export const prettifyJSON = (data: string) =>
...
+    acc.add(curr.zone);
+    return acc;
+  }, new Set<string>());
+
+  if (isFlexibleScaling) {
+    return OCS_DEVICE_SET_FLEXIBLE_REPLICA;
+  }
+  if (isStretchCluster) {
+    return uniqueZonesSet.size;
+  }
```
SanjalKatiyar, line 310: @bipuladh in arbiter there are only 3 zones (including arbiter), if we follow this, replica for arbiter will be set to 3 whereas it should be 4

SanjalKatiyar: this seems okay for normal configuration

packages/odf/components/utils/common.ts
```diff
@@ -294,11 +292,32 @@ export const prettifyJSON = (data: string) =>
...
+
+  return uniqueZonesSet.size;
+};
+
```
SanjalKatiyar, line 314:
```suggestion
  return uniqueZonesSet.size;
};

```

SanjalKatiyar: only if comment https://github.com/red-hat-storage/odf-console/pull/556/files#r1042973439 is not correct.

PR #560, opened 2022-12-12
Fix capacity breakdown card (external mode)

SanjalKatiyar:
**Before:**
Screenshot from 2022-12-12 19-52-31

**After:**
Screenshot from 2022-12-12 19-52-06


bipuladh (approved):
/lgtm

PR #584, opened 2023-01-09
Fix DRPolicy list page access issue for application users

GowthamShanmugam:


GowthamShanmugam:
image
Screenshot from 2023-01-10 00-26-13

PR #590, opened 2023-01-10
Refactor: Package file and plugin generation scripts

bipuladh:
- Cleanup package.json file
- Move changing directory details to webpack configuration
- Move to modules for scripts
- Make filepath relative to the script directory instead of the process directory
- Add support for plugin name environment variable
- Add tsm as a dev dependency


scripts/generatePluginPackage.mts
```diff
@@ -0,0 +1,35 @@
...
+import { fileURLToPath } from 'url';
+
+const SUPPORTED_PLUGINS = ['odf', 'mco'];
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const validatePluginName = (pluginName) =>
+  SUPPORTED_PLUGINS.includes(pluginName);
+
+const plguinNameArgument = '--plugin';
```
GowthamShanmugam, line 12:
```suggestion
const pluginNameArgument = '--plugin';
```

bipuladh: updated!

PR #595, opened 2023-01-16
Setup a package for ODF extensions

bipuladh:
Refactors ODF Dashboard

### DIFF
--- a/packages/odf/components/odf-dashboard/dashboard.tsx
+++ b/packages/odf/components/odf-dashboard/dashboard.tsx
@@ -1,33 +1,36 @@
 import * as React from 'react';
+import {
+  DashboardTab,
+  DashboardTabExtensionProps as UnresolvedTabProps,
+  isDashboardTab,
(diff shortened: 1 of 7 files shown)


packages/odf/components/odf-dashboard/dashboard.tsx
```diff
@@ -61,51 +64,52 @@ export const ODFDashboard: React.FC = () => {
...
+          extension.properties.before || extension.properties.after;
+        let pushLocation = updatedPages.findIndex((pg) => pg.id === indexId);
+        if (extension.properties.before && updatedPages.length > 0) {
+          pushLocation -= 1;
+        }
+        if (updatedPages.length > 0) {
+          updatedPages.splice(pushLocation, 0, page);
+        } else {
+          updatedPages.push(page);
+        }
```
SanjalKatiyar, line 106: let's say I want to add before "overview":
pushLocation will be 0, then "pushLocation -= 1;"  will make it -1,
nav will get added to the end instead in this case...

PR #596, opened 2023-01-16
Setup a package for ODF extensions

bipuladh:
Refactors ODF Dashboard

Screenshot:
Screenshot from 2023-01-16 15-13-20

### DIFF
--- a/packages/odf/components/odf-dashboard/dashboard.tsx
+++ b/packages/odf/components/odf-dashboard/dashboard.tsx
@@ -1,33 +1,36 @@
 import * as React from 'react';
+import {
+  DashboardTab,
+  DashboardTabExtensionProps as UnresolvedTabProps,
+  isDashboardTab,
+} from '@odf/odf-plugin-sdk/extensions';
 import PageHeading from '@odf/shared/heading/page-heading';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
 import {
   HorizontalNav,
-  NavPage,
-  useFlag,
+  useResolvedExtensions,
 } from '@openshift-console/dynamic-plugin-sdk';
+import { ResolvedCodeRefProperties } from '@openshift-console/dynamic-plugin-sdk/lib/types';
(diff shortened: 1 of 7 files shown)


packages/odf/components/odf-dashboard/dashboard.tsx
```diff
@@ -61,51 +64,52 @@ export const ODFDashboard: React.FC = () => {
...
+        let pushLocation = updatedPages.findIndex((pg) => pg.id === indexId);
+        if (extension.properties.before && updatedPages.length > 0) {
+          pushLocation -= 1;
+        }
+        if (updatedPages.length > 0) {
+          updatedPages.splice(pushLocation, 0, page);
+        } else {
+          updatedPages.push(page);
+        }
       });
```
SanjalKatiyar, line 107: let's say I want to add before "overview":
pushLocation will be 0, then "pushLocation -= 1;" will make it -1,
nav will get added to the end instead in this case...

SanjalKatiyar: similar issue can occur if we are searching (before/after) for any extension not yet in the list... in this case pushLocation will be  -1 which will get updated to  -2 and may cause unexpected result (depends)...

SanjalKatiyar: it is internal use extension so may not be a big issue... still just pointing out once...

SanjalKatiyar:
@bipuladh just to confirm once, did u test create/details page routes for noobaa resources as well ??
I think "Create" button on list page is still pointing to "/odf/resource/..."

SanjalKatiyar:
> @bipuladh just to confirm once, did u test create/details page routes for noobaa resources as well ?? I think "Create" button on list page is still pointing to "/odf/resource/..."

same for breadcrumbs on details pages...

bipuladh:
@SanjalKatiyar this PR is getting confusing I will send another PR with route changes for now reverting those changes.

PR #602, opened 2023-01-17
use generic names for Data Foundation

umangachapagain:
Updated names from "OpenShift Data Foundation" to "Data Foundation" to keep it generic.


bipuladh (changes requested):
https://github.com/red-hat-storage/odf-console/pull/602/files#diff-bddf1b45c5882191e67bfc176d5c212e3d0ff6f5cc37e3992c82d6726fd185fbL40
Needs changes as well

bipuladh (approved):

GowthamShanmugam (approved):

PR #603, opened 2023-01-17
Adds bucket replication UI for OBC page

vbnrh:
Design: https://www.sketch.com/s/a5c851b3-87c6-4ce6-8eea-28d47aa4fa4c/a/1KbjyWk

Screenshot from 2023-01-23 17-38-24
Screenshot from 2023-01-23 17-37-58


packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -167,13 +198,153 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+      <TextInput
+        data-test="prefix-input"
+        placeholder={t('Enter a prefix')}
+        type="text"
+        id="prefix-input"
+        value={rule.prefix}
+        onChange={onPrefixChange}
+        aria-label={t('Prefix')}
+      />
+    </div>
```
bipuladh, line 345: You could use PF Form elements right?

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -167,13 +198,153 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+  ]);
+  const { t } = useCustomTranslation();
+
+  React.useEffect(() => {
+    updateParentState(rules);
+  }, [rules, updateParentState]);
+
+  const handleNSChange = (ns: NamespaceStoreKind, ruleId: number) => {
+    const newRules = [...rules];
+    const idx = newRules.findIndex((rule) => rule.id === ruleId);
```
bipuladh, line 258:
```suggestion
    const index = newRules.findIndex((rule) => rule.id === ruleId);
```
nit

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -167,13 +198,153 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+    newRules[idx].namespaceStore = ns?.metadata?.name;
```
bipuladh, line 259:
```suggestion
    newRules[idx].namespaceStore = getName(ns);
```
nit

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -167,13 +198,153 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+  namespaceStore: string;
+  prefix: string;
+}
+
+export const ReplicationPolicyForm: React.FC<ReplicationFormProps> = ({
+  namespace,
+  updateParentState,
+}) => {
+  const [rules, setRules] = React.useState<Rule[]>([
+    { id: 1, namespaceStore: '', prefix: '' },
```
bipuladh, line 248:
```suggestion
```
Is this some sort of default rule?

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -167,13 +198,153 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
     </div>
   );
 };
 
+type ReplicationFormProps = {
+  namespace: string;
+  updateParentState: (rules: Rule[]) => void;
+};
+
+interface Rule {
```
bipuladh, line 237:
```suggestion
type Rule {
```
nit

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
vbnrh: its an initial empty rule to show the form

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -13,15 +13,38 @@ import {
   k8sCreate,
 } from '@openshift-console/dynamic-plugin-sdk';
 import * as _ from 'lodash-es';
+import { NamespaceStoreKind } from 'packages/odf/types/mcg';
```
GowthamShanmugam, line 16: path is incorrect

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -51,6 +74,124 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
     dispatch({ type: 'setProvisioner', name: sc?.provisioner });
   };
 
+  const generateOBCNameByNamespaceStore = (name: string): string => {
+    const base64Encoded = btoa(name).substring(0, 10);
+    return `destination-bucket-${name}-${base64Encoded}`.toLowerCase();
+  };
+
+  const generateBucketClassNameByNamespaceStore = (name: string): string => {
+    const base64Encoded = btoa(name).substring(0, 10);
```
GowthamShanmugam, line 83: can't we use generated names? generateName: `generic-bucketclass-${name}-`

vbnrh: I am using this name in the generated obc payload . `generateName` will only create the name at runtime

locales/en/plugin__odf-console.json
```diff
@@ -677,6 +679,14 @@
   "StorageClass": "StorageClass",
   "Defines the object-store service and the bucket provisioner.": "Defines the object-store service and the bucket provisioner.",
   "BucketClass": "BucketClass",
+  "Replication Policy": "Replication Policy",
+  "For higher resiliency, set a replication configuration for objects which are stored in NooBaa namespace buckets": "For higher resiliency, set a replication configuration for objects which are stored in NooBaa namespace buckets",
+  "Replication rules ": "Replication rules ",
+  "Create new NamespaceStore": "Create new NamespaceStore",
+  "Add Rule": "Add Rule",
```
SanjalKatiyar, line 686:
```suggestion
  "Add Rule": "Add rule",
```

locales/en/plugin__odf-console.json
```diff
@@ -677,6 +679,14 @@
...
+  "Namespace store": "Namespace store",
```
SanjalKatiyar, line 687:
```suggestion
  "Namespace store": "NamespaceStore",
```

packages/odf/components/mcg/create-obc.scss
```diff
@@ -1,3 +1,15 @@
...
+  margin-bottom: var(--pf-global--spacer--md);
+}
+
+.odf-mcg__remove-button {
+ margin-top: var(--pf-global--spacer--xl)
+}
+
+.odf-mcg__form {
+  margin-left: var(--pf-global--spacer--xl)
+}
```
SanjalKatiyar, line 15: nits:
```suggestion
.odf-mcg {
  &__namespacestore-dropdown {
    margin-bottom: var(--pf-global--spacer--md);
  }
  
  &__remove-button {
   margin-top: var(--pf-global--spacer--xl)
  }
  
  &__form {
    margin-left: var(--pf-global--spacer--xl)
  }
}
```

packages/odf/components/mcg/state.ts
```diff
@@ -1,5 +1,10 @@
 import { K8sResourceKind } from '@odf/shared/types';
 
+export type ReplicationResources = {
+  bucketClass: K8sResourceKind;
+  objectBucketClaim: K8sResourceKind;
```
SanjalKatiyar, line 5: I think we don't have "ObjectBucketClaimKind", so good to define one now...
```suggestion
  bucketClass: BucketClassKind;
  objectBucketClaim: ObjectBucketClaimKind;
```

packages/odf/components/mcg/state.ts
```diff
@@ -10,8 +15,30 @@ export type State = {
   error: string;
   payload: K8sResourceKind;
   bucketClass: string;
+  replicationRuleFormData: ReplicationRuleFormData[];
+  additionalReplicationResources: ReplicationResources[];
```
SanjalKatiyar, line 19: there is lot of redundant info that we are storing in redux here in "additionalReplicationResources" (example: apiVersion, Kind, spec.namespacePolicy.type,  namespaceStoreName etc...). If it is not too much of a refactoring, I would prefer only storing needed info and make changes in code accordingly...

I agree it is not really too much of an optimization, still not good practice to store unnecessary data...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -51,6 +78,124 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+    return `generic-bucketclass-${name}-${base64Encoded}`.toLowerCase();
+  };
```
SanjalKatiyar, line 89: prefix can be "destination-bucket" or "generic-bucketclass"
```suggestion
  const generateGenericName = (name: string, prefix: string): string => {
    const base64Encoded = btoa(name).substring(0, 10);
    return `${prefix}-${name}-${base64Encoded}`.toLowerCase();
  };
```

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -51,6 +78,124 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+        )}`,
+        filter: {
+          prefix: rule.prefix,
+        },
+      };
+      return repRule;
+    });
+
+    return rules ? JSON.stringify(rules) : '';
+  };
```
SanjalKatiyar, line 109: nits: u can skip this if u want to
```suggestion
  const createReplicationRulesAndStringify = (
    data: ReplicationRuleFormData[],
    namespace: string
  ): string => {
    const rules: Array<OBCReplicationRules> = data.map((rule) => ({
        ruleId: `rule-${rule.ruleNumber}`,
        obName: `obc-${namespace}-${generateOBCNameByNamespaceStore(
          rule.namespaceStore
        )}`,
        filter: {
          prefix: rule.prefix,
        },
      }));

    return rules ? JSON.stringify(rules) : '';
  };
```

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -51,6 +78,124 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+        ruleId: `rule-${rule.ruleNumber}`,
+        obName: `obc-${namespace}-${generateOBCNameByNamespaceStore(
+          rule.namespaceStore
+        )}`,
+        filter: {
+          prefix: rule.prefix,
+        },
+      };
+      return repRule;
+    });
```
SanjalKatiyar, line 106: nits: u can skip this comment, if u want to... not really needed...
```suggestion
    const rules: Array<OBCReplicationRules> = data.map((rule) => ({
        ruleId: `rule-${rule.ruleNumber}`,
        obName: `obc-${namespace}-${generateOBCNameByNamespaceStore(
          rule.namespaceStore
        )}`,
        filter: {
          prefix: rule.prefix,
        },
      }));
```

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -51,6 +78,124 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+        namespacePolicy: {
+          type: NamespacePolicyType.SINGLE,
+          single: {
+            resource: namespaceStoreName,
+          },
+        },
+      },
+    };
+    return bc;
+  };
```
SanjalKatiyar, line 133: nits: u can skip this comment, if u want to... not really needed...
```suggestion
  ): K8sResourceKind => ({
      apiVersion: getAPIVersionForModel(NooBaaBucketClassModel),
      kind: NooBaaBucketClassModel.kind,
      metadata: {
        name,
        namespace,
      },
      spec: {
        namespacePolicy: {
          type: NamespacePolicyType.SINGLE,
          single: {
            resource: namespaceStoreName,
          },
        },
      },
    });
```

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -51,6 +78,124 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+
+  const createNewObjectBucketClaim = (
+    name: string,
+    namespace: string,
+    bucketClassName: string,
+    storageClassName: string
+  ) => {
+    const obc: K8sResourceKind = {
```
SanjalKatiyar, line 141: same here... no need to store in a variable... we can return directly...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -51,6 +78,124 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+      );
+      const res: ReplicationResources = {
+        bucketClass: bc,
+        objectBucketClaim: obc,
+      };
+      return res;
+    });
+    return resources;
+  };
+
```
SanjalKatiyar, line 184: all there functions: `generateOBCNameByNamespaceStore, generateBucketClassNameByNamespaceStore, generateBucketClassNameByNamespaceStore, createNewSingleNamespaceBucketClass, createNewSingleNamespaceBucketClass, generateAdditionalReplicationResources`
move them out of FC and keep them under "odf/utils"...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -51,6 +78,124 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+      },
+    };
+    return obc;
+  };
+
+  const generateAdditionalReplicationResources = (
+    state: State,
+    namespace: string
+  ): Array<ReplicationResources> => {
+    const resources = state.replicationRuleFormData.map((rule) => {
```
SanjalKatiyar, line 164: same here... no need to store in a variable... we can return directly...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -51,6 +78,124 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+  };
+
+  const createReplicationRulesAndStringify = (
+    data: ReplicationRuleFormData[],
+    namespace: string
+  ): string => {
+    const rules: Array<OBCReplicationRules> = data.map((rule) => {
+      const repRule: OBCReplicationRules = {
+        ruleId: `rule-${rule.ruleNumber}`,
+        obName: `obc-${namespace}-${generateOBCNameByNamespaceStore(
```
SanjalKatiyar, line 98: does noobaa automatically adds "obc-${namespace}-" to the OB name even if "generateBucketName" is provided to the OBC CR ??

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -73,7 +218,27 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
       obj.spec.generateBucketName = 'bucket-';
     }
     if (state.bucketClass && isNoobaa) {
-      obj.spec.additionalConfig = { bucketclass: state.bucketClass };
+      if (state.replicationRuleFormData.length >= 1) {
```
SanjalKatiyar, line 221: nits:
```suggestion
      if (!!state.replicationRuleFormData.length) {
```

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -167,13 +333,226 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+                onChange={() => {
+                  // if this checkbox is disabled, then on this point we purge the contents of replication form data
+                  if (replicationEnabled)
+                    dispatch({ type: 'setReplicationRuleFormData', data: [] });
+                  toggleReplication(!replicationEnabled);
+                }}
+              />
+            </div>
+            {replicationEnabled && (
+              <div className="form-group">
```
SanjalKatiyar, line 358: why not add this "div" to "ReplicationPolicyForm" FC itself ?

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -167,13 +333,226 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+    () => launcher(NS_STORE_MODAL_KEY, null),
+    [launcher]
+  );
+
+  const handleNSChange = (ns: NamespaceStoreKind, ruleId: number) => {
+    const newRules = [...rules];
+    const index = newRules.findIndex((rule) => rule.id === ruleId);
+    newRules[index].namespaceStore = getName(ns);
+    setRules(newRules);
+    updateParentState(rules);
```
SanjalKatiyar, line 416: why are we maintaining 2 states everywhere (setRules, updateParentState) ??
we can directly use state stored in redux, no need to local state of this component ??

SanjalKatiyar: finally before adding data to the CR we can filter out empty rules and add correct ones...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -167,13 +333,226 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+    ]);
+  };
+
+  const handleRemoveRule = () => {
+    let newRules = [...rules];
+
+    // will remove the last rule from the array
+    newRules.splice(-1, 1);
+    setRules(newRules);
+    updateParentState(rules);
```
SanjalKatiyar, line 440: same... we can utilize redux state itself... instead of having 2 states for same info...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -167,13 +333,226 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+            id="prefix-input"
+            value={rule.prefix}
+            onChange={onPrefixChange}
+            aria-label={t('Prefix')}
+          />
+        </FormGroup>
+      </GridItem>
+    </>
+  );
+};
```
SanjalKatiyar, line 555: this file is getting too long... let's separate this file in 2 parts...

SanjalKatiyar: one for main body... other for sub components...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -184,6 +563,25 @@ export const CreateOBCPage: React.FC<CreateOBCPageProps> = (props) => {
...
+
+        k8sCreate<K8sResourceKind>({
+          model: NooBaaObjectBucketClaimModel,
+          data: res.objectBucketClaim,
+        }).catch((err) => {
+          dispatch({ type: 'setError', message: err.message });
+          dispatch({ type: 'unsetProgress' });
+        });
+      });
+    }
```
SanjalKatiyar, line 584: 1. Store all resources in a promise object and resolve them at once using "Promise.all" (all these calls will be asynchronous from the UI)...
2. If all above calls gets succeeded and there is no error then move forward with OBC creation synchronously.
Or,
1. Handle all the requests at once, storing in promise object and resolving asynchronously using "Promise.all" and displaying error (if any) accordingly.

SanjalKatiyar: I personally prefer first method... all BC/OBC will be async... once those get created then we move to current OBC creation in sync manner...
but I am fine with second method as well... still use "Promise.all"

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
vbnrh: The idea was to make the component generic and not tied to the OBC creation page only. 
The `updateParentState` updates the parent component's state, this can be any logic that user wants to provide for these rules. 

`setRules` is the component managing its own state. 

I don't prefer passing the state and the dispatch into the component directly. This  will tie the component to its parent component's redux.

 I can reuse this form in future if i want to say, add it in a bucket class page and it will help if the page does not necessarily use redux's state.

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
vbnrh: same as above

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
vbnrh: yep

locales/en/plugin__odf-console.json
vbnrh: This is the title for the dropdown dropdown. I don't think we need to change it. Let me know if this is the pattern we follow for other resource selection form elements too

packages/odf/components/mcg/state.ts
vbnrh: I would not bother with optimizing this space issue since the refactorings may take extra time.
We are only allowing 10 rules on the page, which means a total of 30 CRs in total that will be created at any point of time. 

So if form data and additionalResources are combined. It will only be around  ~60 instances. Hardly any extra space for computing these days.

This is an O(1) space complexity in any case.

I did this separation to segregate the form data from the resources that we need to generate to avoid the messy calculations during the time we create the resources.

vbnrh: Even if i were to calculate it without storing it in the state. After the generating the resources, it would still occupy the same amount of space in the memory.

SanjalKatiyar: It is more about maintainability rather than optimization... I agree (as I commented above as well), in this particular example it won't look bad nut things will pile up if we keep on doing this at multiple places...

also, redux is global memory only gets cleaned when component will unmount, and changes to this state will cause re-render to all the FC... whereas local memory gets GCed whenever from time to time also any changes to local state will not cause entire FC to re-render...

SanjalKatiyar: but I already mentioned `If it is not too much of a refactoring` :)
still today we do here, tomorrow some other place and things will get piled up that way !!

locales/en/plugin__odf-console.json
SanjalKatiyar: yeah, it is fine... we can use `NamespaceStore` for dropdown title... we do the same at other places...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
SanjalKatiyar: okay...

packages/odf/components/mcg/state.ts
SanjalKatiyar: u can always just store necessary info and call util function to create the entire CR before creation... that local object will only be scoped with "onSubmit" function call and gets GCed after that, where as here redux has much larger scope...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
vbnrh: I did not feel like adding unneccesary css elements into the new component form. These may not be used if the component requires reuse in other pages.

SanjalKatiyar: we can pass "className" as a prop to "ReplicationPolicyForm" FC...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
vbnrh: I am going to keep the following changes 
- `generateOBCNameByNamespaceStore` , `generateBucketClassNameByNamespaceStore` , `createNewSingleNamespaceBucketClass`, `createNewSingleNamespaceBucketClass`,  - These are not tied to state of the component and can be reused again and again.
- `generateAdditionalReplicationResources`, `createReplicationRulesAndStringify` will be handled by OBC FC. They are exclusive to that component's state and do not work like a basic utility function.

SanjalKatiyar: make sense... but make sure to move "generateAdditionalReplicationResources and createReplicationRulesAndStringify" out of FC... u can keep in same file but move out of the FC (this is imp)...

SanjalKatiyar: inside FC they will get created again and again... so keep them outside...

PR #607, opened 2023-01-19
Input with field requirements

ivoalmeida:
image

### DIFF
--- a/packages/shared/input-with-requirements/TextInputWithFieldRequirements.tsx
+++ b/packages/shared/input-with-requirements/TextInputWithFieldRequirements.tsx
@@ -3,7 +3,6 @@ import CheckCircleIcon from '@patternfly/react-icons/dist/esm/icons/check-circle
 import ExclamationTriangleIcon from '@patternfly/react-icons/dist/esm/icons/exclamation-triangle-icon';
 import InfoCircleIcon from '@patternfly/react-icons/dist/esm/icons/info-circle-icon';
 import cn from 'classnames';
-// eslint-disable-next-line import/no-extraneous-dependencies
 import { useController, Control, FieldValues } from 'react-hook-form';
 import {
   FormGroup,
@@ -19,12 +18,13 @@ import {
   Icon,
 } from '@patternfly/react-core';
 import useFieldRequirements from './useFieldRequirements';
+import './TextInputWithFieldRequirements.scss';
 
 export type TextInputWithFieldRequirementsProps = {
   fieldRequirements: string[];
   control: Control<FieldValues>;
   formGroupProps: FormGroupProps;
-  textInputProps: TextInputProps;
+  textInputProps: TextInputProps & { ['data-test']: string };
 };
 
 export type ValidationIconProp = {
@@ -32,7 +32,7 @@ export type ValidationIconProp = {
 };
 
 export const ValidationIcon: React.FC<ValidationIconProp> = ({ status }) => {
-  const getStatusIcon = () => {
+  const getStatusIcon = React.useCallback(() => {
     switch (status) {
       case 'error':
         return <ExclamationTriangleIcon />;
@@ -42,9 +42,9 @@ export const ValidationIcon: React.FC<ValidationIconProp> = ({ status }) => {
       default:
         return <InfoCircleIcon />;
     }
-  };
+  }, [status]);
 
-  const getVariant = () => {
+  const getVariant = React.useCallback(() => {
     switch (status) {
       case 'error':
         return 'danger';
@@ -53,7 +53,8 @@ export const ValidationIcon: React.FC<ValidationIconProp> = ({ status }) => {
       default:
         return 'info';
     }
-  };
+  }, [status]);
+
   return <Icon status={getVariant()}>{getStatusIcon()}</Icon>;
 };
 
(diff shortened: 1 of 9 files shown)


packages/shared/input-with-requirements/useFieldRequirements.ts
```diff
@@ -0,0 +1,124 @@
...
+export type State = {
+  fieldRequirements: Record<string, Status>;
+};
+
+export type Dispatch = ReactDispatch<Action>;
+
+export type Action = {
+  type: ActionType;
+  payload?: string;
+  resetType?: 'success' | 'indeterminate';
```
alfonsomthd, line 23: Just a suggestion: maybe 'success', 'indeterminate', 'empty', 'error' could be part of a [string enum](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) (e.g. ValidationStatus).

packages/shared/input-with-requirements/useFieldRequirements.spec.ts
```diff
@@ -0,0 +1,350 @@
...
+
+const withState: (status: any, message?: string) => State = expectedState.bind(
+  null,
+  initialisedState
+);
+
+//#endregion
+
+describe('hook', () => {
+  it('should return state with all given field requirements set to undeterminate when isDirty is false', () => {
```
alfonsomthd, line 40:
```suggestion
  it('should return state with all given field requirements set to indeterminate when isDirty is false', () => {
```

packages/shared/input-with-requirements/useFieldRequirements.spec.ts
```diff
@@ -0,0 +1,350 @@
...
+    const formErrors = [];
+    const dispatchedErrors = [];
+    const dispatch = jest.fn();
+    const expected = [];
+    const actual = dispatchSuccess(formErrors, dispatchedErrors, dispatch);
+    expect(actual).toEqual(expected);
+    expect(dispatch).not.toHaveBeenCalled();
+  });
+
+  it('should not dispatch success while formErrors is reports same dispatched errors', () => {
```
alfonsomthd, line 341:
```suggestion
  it('should not dispatch success while formErrors reports same dispatched errors', () => {
```

packages/shared/input-with-requirements/useFieldRequirements.ts
ivoalmeida: Are you talking about the ```resetType```? If so, there can only ever be two status for a reset. If the user hasn't touched a field it is "indeterminate" if has and no errors are preset then "success".

alfonsomthd: I was talking more generically. Instead of using plain strings for statuses repeated in different files, it could be like sth like this:
`resetType?: ValidationStatus.SUCCESS | ValidationStatus.INDETERMINATE;`

ivoalmeida: Got it. They become types when you're setting them inside a ```type```. You cannot pass any string only either of those defined in the type. A enum would work too obviously but I don't see the benefit in this use case (two short strings where keys are same as values...). It's a bit overkill IMHO :)

alfonsomthd: The suggestion is related to all statuses, not only those in that. I see 2 benefits:

1) Avoiding typos ( e.g. "undeterminate" vs "indeterminate"): instead of repeating the same string here and there, you have an enum.
2) Having all the possible Validation Statuses (not only the 2 ones mentioned there, also, 'error', 'empty' or the appropriate ones) described under a single enum instead of having a lot of repeated, unrelated strings scattered through different files: it makes discovery & refactoring easier.

It's just a suggestion, no blocking this PR in any way :wink:

packages/shared/input-with-requirements/TextInputWithFieldRequirements.tsx
```diff
@@ -0,0 +1,138 @@
...
+  textInputProps: TextInputProps & { ['data-test']: string };
+};
+
+export type ValidationIconProp = {
+  status: FormGroupProps['validated'];
+};
+
+export const ValidationIcon: React.FC<ValidationIconProp> = ({ status }) => {
+  const getStatusIcon = () => {
+    switch (status) {
```
bipuladh, line 36:
```suggestion
  const getStatusIcon = (currenStatus) => {
    switch (currentStatus) {
```
This way the function will be reusable and won't be recreated when the props change or update. If you want to keep it as it is use `useCallback`

packages/shared/input-with-requirements/TextInputWithFieldRequirements.tsx
```diff
@@ -0,0 +1,138 @@
...
+        return <CheckCircleIcon />;
+
+      default:
+        return <InfoCircleIcon />;
+    }
+  };
+
+  const getVariant = () => {
+    switch (status) {
+      case 'error':
```
bipuladh, line 49: Same as above.

packages/shared/input-with-requirements/useFieldRequirements.ts
```diff
@@ -0,0 +1,124 @@
...
+  error?: TError
+) => {
+  const [state, dispatch] = useReducer<Reducer<State, Action>, State>(
+    reducer,
+    initialState,
+    () => init(fieldRequirements)
+  );
+  const [, setErrorStack] = useState<string[]>([]);
+
+  useEffect(() => {
```
bipuladh, line 115:
```suggestion
  React.useEffect(() => {
```
nit: something of a style for code base.

bipuladh (commented):
looks good.

bipuladh (approved):

bipuladh:
/lgtm cancel
waiting for other reviewers

PR #609, opened 2023-01-20
shared/utils/Tabs: adapt behavior for location hashes

alfonsomthd:
This allows the use of Location Hashes in tab href's like this:
image


packages/shared/utils/Tabs.tsx
```diff
@@ -78,6 +78,8 @@ const Tabs: React.FC<TabsProps> = ({
     [hrefToTabMap]
   );
 
+  const isLocationHash = (href: string) => href.startsWith('#');
```
SanjalKatiyar, line 81: let's move this outside FC... else new function will get created everytime component will re-render...
good to do where ever we can...

packages/shared/utils/Tabs.tsx
```diff
@@ -116,10 +119,14 @@ const Tabs: React.FC<TabsProps> = ({
   }, [activeTab, history, hrefToTabMap, location.pathname, tabToHrefMap]);
 
   const onClick = (event, tabIndex) => {
+    const requiredHref = tabToHrefMap[tabIndex];
+    if (isLocationHash(requiredHref)) {
+      setActiveTab(tabIndex);
+      return;
+    }
     // Update the href only and let useEffect handle tabChange
```
SanjalKatiyar, line 127: from the comment "Update the href only and let useEffect handle tabChange" I can that this function is only intended to set href and "useEffect" will take care of setting the active tab...
Is it not possible to the same here as well to maintain initial functionality ??

alfonsomthd: In the case of a location hash we don't want to `history.push`: onle set the active tab, not to update history/location.

SanjalKatiyar:
@bipuladh can add lgtm as well once he has taken a loop...

PR #616, opened 2023-01-31
implemented TextInputWithFieldRequirements component on backing store…

ivoalmeida:
…, bucketclass, namespace store, blockpools, storageclass and object bucket claim forms
image
image
image
image
image
image


packages/ocs/block-pool/body.tsx
```diff
@@ -204,29 +272,28 @@ export const BlockPoolBody = (props: BlockPoolBodyPros) => {
...
+            formGroupProps={{
+              label: t('Pool name'),
+              fieldId: 'pool-name',
+              className: 'ceph-block-pool-body__input',
+              isRequired: true,
+            }}
+            textInputProps={{
+              id: 'pool-name',
+              name: 'newPoolName',
+              ['data-test']: 'new-pool-name-textbox',
```
GowthamShanmugam, line 291: just a doubt, why do we need a square bracket here?

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
```diff
@@ -44,19 +47,73 @@ const GeneralPage: React.FC<GeneralPageProps> = ({ dispatch, state }) => {
 
   const [showHelp, setShowHelp] = React.useState(true);
 
-  const [validated, setValidated] = React.useState<ValidatedOptions>(
-    ValidatedOptions.default
+  const isNamespaceStoreSupported = useFlag(FEATURES.OCS_NAMESPACE_STORE);
+
+  const [existingNames, setExistingNames] = React.useState<string[]>([]);
```
GowthamShanmugam, line 52:
```suggestion
```

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
```diff
@@ -44,19 +47,73 @@ const GeneralPage: React.FC<GeneralPageProps> = ({ dispatch, state }) => {
...
-    } else {
-      setValidated(ValidatedOptions.error);
+  React.useEffect(() => {
+    if (loaded) {
+      const names = data.map((data: BucketClassKind) => getName(data));
+      setExistingNames(names);
     }
-  };
+    if (loadError) setExistingNames([]);
+  }, [data, loaded, loadError]);
```
GowthamShanmugam, line 64:
```
existingNames = React.useMemo(() => (
   data.map((data: BucketClassKind) => getName(data))
  ), [data, loaded, loadError]);
```

we can save one hook here,

packages/ocs/block-pool/body.tsx
```diff
@@ -79,6 +85,68 @@ export const BlockPoolBody = (props: BlockPoolBodyPros) => {
...
+        fieldRequirements[1]
+      )
+      .matches(
+        /^[a-z0-9]+([a-z0-9]|([-.](?![-.])))*[a-z0-9]*$/,
+        fieldRequirements[2]
+      )
+      .test(
+        'unique-name',
+        fieldRequirements[3],
+        (value: string) => !!!existingNames.includes(value)
```
GowthamShanmugam, line 121:
```suggestion
        (value: string) => !existingNames.includes(value)
```
?

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
```diff
@@ -44,19 +47,73 @@ const GeneralPage: React.FC<GeneralPageProps> = ({ dispatch, state }) => {
...
+      .matches(
+        /^[a-z0-9]+([a-z0-9]|([-.](?![-.])))*[a-z0-9]*$/,
+        fieldRequirements[2]
+      )
+      .test(
+        'unique-name',
+        fieldRequirements[3],
+        (value: string) => !!!existingNames.includes(value)
+      ),
+  });
```
GowthamShanmugam, line 91: I think these are fixed for bucketclassname, shall we create separate utils files to maintain this? So we can reuse this if support we need to validate bucketclassname in other places.

i think blockpool has two places(modal and create page), keeping this separate will easy to maintain and reuse in other palces.

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
```diff
@@ -44,19 +47,73 @@ const GeneralPage: React.FC<GeneralPageProps> = ({ dispatch, state }) => {
...
+        fieldRequirements[1]
+      )
+      .matches(
+        /^[a-z0-9]+([a-z0-9]|([-.](?![-.])))*[a-z0-9]*$/,
+        fieldRequirements[2]
+      )
+      .test(
+        'unique-name',
+        fieldRequirements[3],
+        (value: string) => !!!existingNames.includes(value)
```
GowthamShanmugam, line 89:
```suggestion
        (value: string) => !existingNames.includes(value)
```
why three?

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
```diff
@@ -110,40 +167,33 @@ const GeneralPage: React.FC<GeneralPageProps> = ({ dispatch, state }) => {
...
+            fieldId: 'bucketclassname-input',
+            label: t('BucketClass name'),
+            helperText: t(
+              'A unique name for the bucket class within the project.'
+            ),
+            isRequired: true,
+          }}
+          textInputProps={{
+            name: 'bucketclassname-input',
+            ['data-test']: 'bucket-class-name',
```
GowthamShanmugam, line 189: why square bracket, is this will act like optional?

packages/odf/components/create-bs/create-bs.tsx
```diff
@@ -78,9 +69,68 @@ const CreateBackingStoreForm: React.FC<CreateBackingStoreFormProps> = (
     onClose,
   } = props;
 
-  const onSubmit = (event) => {
+  const [existingNames, setNames] = React.useState<string[]>([]);
```
GowthamShanmugam, line 72:
```suggestion
```

packages/odf/components/create-bs/create-bs.tsx
```diff
@@ -78,9 +69,68 @@ const CreateBackingStoreForm: React.FC<CreateBackingStoreFormProps> = (
...
+    NooBaaBackingStoreModel,
+    namespace
+  );
+  React.useEffect(() => {
+    if (loaded) {
+      const names = data.map((data: BackingStoreKind) => getName(data));
+      setNames(names);
+    }
+    if (loadError) setNames([]);
+  }, [data, loaded, loadError]);
```
GowthamShanmugam, line 110:
```suggestion
  const existingNames = React.useMemo(() => (
    data.map((data: BackingStoreKind) => getName(data));
  ), [data, loaded, loadError]);
```

packages/odf/components/namespace-store/namespace-store-form.tsx
```diff
@@ -117,14 +120,65 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
 
   const [inProgress, setProgress] = React.useState(false);
   const [error, setError] = React.useState('');
+  const [existingNames, setExistingNames] = React.useState<string[]>([]);
```
GowthamShanmugam, line 123:
```suggestion
```

packages/odf/components/namespace-store/namespace-store-form.tsx
```diff
@@ -117,14 +120,65 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
...
+    NooBaaNamespaceStoreModel,
+    namespace
+  );
+  React.useEffect(() => {
+    if (loaded) {
+      const names = data?.map((data: NamespaceStoreKind) => getName(data));
+      setExistingNames(names);
+    }
+    if (loadError) setExistingNames([]);
+  }, [data, loaded, loadError]);
```
GowthamShanmugam, line 137:
```suggestion
  const existingNames =  React.useMemo(() => (
     data?.map((data: NamespaceStoreKind) => getName(data));
  ), [data, loaded, loadError]);
```

packages/odf/components/namespace-store/namespace-store-form.tsx
```diff
@@ -117,14 +120,65 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
...
+        fieldRequirements[1]
+      )
+      .matches(
+        /^[a-z0-9]+([a-z0-9]|([-.](?![-.])))*[a-z0-9]*$/,
+        fieldRequirements[2]
+      )
+      .test(
+        'unique-name',
+        fieldRequirements[3],
+        (value: string) => !!!existingNames.includes(value)
```
GowthamShanmugam, line 161:
```suggestion
        (value: string) => !existingNames.includes(value)
```

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -185,31 +190,96 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
     namespace: 'openshift-storage',
   };
 
+  const [existingNames, setExistingNames] = React.useState<string[]>([]);
```
GowthamShanmugam, line 193:
```suggestion
```

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -185,31 +190,96 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+
+  const [data, loaded, loadError] = useK8sList(
+    NooBaaObjectBucketClaimModel,
+    namespace
+  );
+  React.useEffect(() => {
```
GowthamShanmugam, line 199: same here replace this with useMemo

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -185,31 +190,96 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+        fieldRequirements[1]
+      )
+      .matches(
+        /^[a-z0-9]+([a-z0-9]|([-.](?![-.])))*[a-z0-9]*$/,
+        fieldRequirements[2]
+      )
+      .test(
+        'unique-name',
+        fieldRequirements[3],
+        (value: string) => !!!existingNames.includes(value)
```
GowthamShanmugam, line 229:
```suggestion
        (value: string) => !existingNames.includes(value)
```

packages/odf/components/create-storage-system/create-storage-system-steps/create-storage-class-step/create-storage-class-step.tsx
```diff
@@ -38,23 +39,91 @@ export const CreateStorageClass: React.FC<CreateStorageClassProps> = ({
...
+      .matches(
+        /^[a-z0-9]+([a-z0-9]|([-.](?![-.])))*[a-z0-9]*$/,
+        fieldRequirements[2]
+      )
+      .test(
+        'unique-name',
+        fieldRequirements[3],
+        (value: string) => !!!existingNames.includes(value)
+      ),
+  });
```
GowthamShanmugam, line 78: For all places, it is almost the same.
Only the key and max allowed value keep changing.
Is it possible to create a generic function to generate this and take the key and max allowed value as an argument?

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/body.tsx
```diff
@@ -149,6 +157,72 @@ export const LocalVolumeSetBody: React.FC<LocalVolumeSetBodyProps> = ({
     setRadio(value);
   };
 
+  const [existingNames, setExistingNames] = React.useState<string[]>([]);
```
GowthamShanmugam, line 160:
```suggestion
```

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/body.tsx
```diff
@@ -149,6 +157,72 @@ export const LocalVolumeSetBody: React.FC<LocalVolumeSetBodyProps> = ({
...
+
+  const [data, loaded, loadError] =
+    useK8sList<StorageSystemKind>(StorageClassModel);
+  React.useEffect(() => {
```
GowthamShanmugam, line 164: same here, replace this using useMemo

packages/ocs/block-pool/body.tsx
```diff
@@ -79,6 +85,68 @@ export const BlockPoolBody = (props: BlockPoolBodyPros) => {
...
+  const [existingNames, setExistingNames] = React.useState<string[]>([]);
+
+  const [data, loaded, loadError] = useK8sList(CephBlockPoolModel);
+  React.useEffect(() => {
+    if (loaded) {
+      const names = data?.map((data) => getName(data));
+      setExistingNames(names);
+    }
+    if (loadError) setExistingNames([]);
+  }, [data, loaded, loadError]);
```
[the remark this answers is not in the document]
ivoalmeida: done

packages/ocs/block-pool/body.tsx
ivoalmeida: no reason, just using bracket notation but removed it since it appears it's not inline with coding guidelines
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_accessors#bracket_notation

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
ivoalmeida: yeah, good catch. done

packages/ocs/block-pool/body.tsx
ivoalmeida: I made a mistake in one place and copied everywhere. my bad

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
ivoalmeida: I hear you. No truly reusable way of doing this unfortunately because the fieldRequirements is used by another component and it needs the useCustomTranslation hook to set translations. It would  also be error prone to just abstract the schema and send the field requirement as props because order cannot be guaranteed. In the end I abstracted away the regular expressions. 

I will extend the schemas in my next PR to include order fields on each form so they won't look all the same anymore.

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
ivoalmeida: I made a mistake in one place and copied everywhere. my bad

packages/odf/components/create-bs/create-bs.tsx
ivoalmeida: done

packages/odf/components/create-bs/create-bs.tsx
ivoalmeida: done

packages/odf/components/namespace-store/namespace-store-form.tsx
ivoalmeida: done

packages/odf/components/namespace-store/namespace-store-form.tsx
ivoalmeida: done

packages/odf/components/namespace-store/namespace-store-form.tsx
ivoalmeida: done

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
ivoalmeida: done

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
ivoalmeida: done

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
ivoalmeida: done

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/body.tsx
ivoalmeida: done

packages/odf/components/create-storage-system/create-storage-system-steps/create-storage-class-step/create-storage-class-step.tsx
ivoalmeida: I hear you. No truly reusable way of doing this unfortunately because the fieldRequirements is used by another component and it needs the useCustomTranslation hook to set translations. It would also be error prone to just abstract the schema and send the field requirement as props because order cannot be guaranteed. In the end I abstracted away the regular expressions.

I will extend the schemas in my next PR to include order fields on each form so they won't look all the same anymore.

packages/ocs/block-pool/body.tsx
```diff
@@ -79,6 +86,63 @@ export const BlockPoolBody = (props: BlockPoolBodyPros) => {
     []
   );
 
+  const [data, loaded, loadError] = useK8sList(CephBlockPoolModel);
+
+  const existingNames = React.useMemo(() => {
+    return loaded && !loadError ? data?.map((data) => getName(data)) : [];
+  }, [data, loaded, loadError]);
```
GowthamShanmugam, line 93:
```suggestion
  const existingNames = React.useMemo(() => (
    loaded && !loadError ? data?.map((data) => getName(data)) : [];
  ), [data, loaded, loadError]);
```
 [a return is required when we keep more than one line in the function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions)

packages/ocs/block-pool/body.tsx
```diff
@@ -79,6 +86,63 @@ export const BlockPoolBody = (props: BlockPoolBodyPros) => {
...
+    shouldUseNativeValidation: false,
+    delayError: undefined,
+  });
+
+  const poolName = watch('newPoolName');
+
+  React.useEffect(() => {
+    dispatch({
+      type: BlockPoolActionType.SET_POOL_NAME,
+      payload: isValid ? poolName : undefined,
```
GowthamShanmugam, line 142:
```suggestion
      payload: isValid ? poolName : '',
```

the reducer has a default value for poolName is an empty string, and it is better the same.

GowthamShanmugam: also, undefined is not safe, it might break the code.

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
```diff
@@ -44,19 +48,67 @@ const GeneralPage: React.FC<GeneralPageProps> = ({ dispatch, state }) => {
...
-  const onChange = (value: string) => {
-    dispatch({ type: 'setBucketClassName', name: value });
-    if (validateBucketClassName(value)) {
-      setValidated(ValidatedOptions.success);
-    } else {
-      setValidated(ValidatedOptions.error);
-    }
-  };
+  const existingNames = React.useMemo(() => {
+    return loaded && !loadError ? data?.map((data) => getName(data)) : [];
```
GowthamShanmugam, line 58: same here no need return

packages/odf/components/create-bs/create-bs.tsx
```diff
@@ -78,9 +70,63 @@ const CreateBackingStoreForm: React.FC<CreateBackingStoreFormProps> = (
...
   } = props;
 
-  const onSubmit = (event) => {
+  const [data, loaded, loadError] = useK8sList<BackingStoreKind>(
+    NooBaaBackingStoreModel,
+    namespace
+  );
+
+  const existingNames = React.useMemo(() => {
+    return loaded && !loadError ? data?.map((data) => getName(data)) : [];
```
GowthamShanmugam, line 79: no return need to an arrow function

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/body.tsx
```diff
@@ -149,6 +158,67 @@ export const LocalVolumeSetBody: React.FC<LocalVolumeSetBodyProps> = ({
     setRadio(value);
   };
 
+  const [data, loaded, loadError] =
+    useK8sList<StorageSystemKind>(StorageClassModel);
+
+  const existingNames = React.useMemo(() => {
+    return loaded && !loadError ? data?.map((data) => getName(data)) : [];
```
GowthamShanmugam, line 165: same here

packages/odf/components/create-storage-system/create-storage-system-steps/create-storage-class-step/create-storage-class-step.tsx
```diff
@@ -38,23 +40,86 @@ export const CreateStorageClass: React.FC<CreateStorageClassProps> = ({
     [dispatch]
   );
 
+  const [data, loaded, loadError] =
+    useK8sList<StorageSystemKind>(StorageClassModel);
+
+  const existingNames = React.useMemo(() => {
+    return loaded && !loadError ? data?.map((data) => getName(data)) : [];
```
GowthamShanmugam, line 47: same here

packages/odf/components/create-storage-system/create-storage-system-steps/create-storage-class-step/create-storage-class-step.tsx
```diff
@@ -38,23 +40,86 @@ export const CreateStorageClass: React.FC<CreateStorageClassProps> = ({
...
+    delayError: undefined,
+  });
+
+  const storageClassName = watch('storage-class-name');
+
+  React.useEffect(() => {
+    dispatch({
+      type: 'wizard/setStorageClass',
+      payload: {
+        name: isValid ? storageClassName : undefined,
```
GowthamShanmugam, line 98:
```suggestion
        name: isValid ? storageClassName : '',
```
?

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/body.tsx
```diff
@@ -149,6 +158,67 @@ export const LocalVolumeSetBody: React.FC<LocalVolumeSetBodyProps> = ({
...
+    delayError: undefined,
+  });
+
+  const createLvsStorageClassName = watch('create-lvs-storage-class-name');
+
+  React.useEffect(() => {
+    dispatch({
+      type: 'wizard/setStorageClass',
+      payload: {
+        name: isValid ? createLvsStorageClassName : undefined,
```
GowthamShanmugam, line 216:
```suggestion
        name: isValid ? createLvsStorageClassName : '',
```
?

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -185,31 +191,91 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
     namespace: 'openshift-storage',
   };
 
+  const [data, loaded, loadError] = useK8sList(
+    NooBaaObjectBucketClaimModel,
+    namespace
+  );
+
+  const existingNames = React.useMemo(() => {
+    return loaded && !loadError ? data?.map((data) => getName(data)) : [];
```
GowthamShanmugam, line 200: sam here

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -185,31 +191,91 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+    shouldFocusError: true,
+    shouldUnregister: false,
+    shouldUseNativeValidation: false,
+    delayError: undefined,
+  });
+
+  const obcName = watch('obcName');
+
+  React.useEffect(() => {
+    dispatch({ type: 'setName', name: isValid ? obcName : undefined });
```
GowthamShanmugam, line 250:
```suggestion
    dispatch({ type: 'setName', name: isValid ? obcName : '' });
```

packages/odf/components/namespace-store/namespace-store-form.tsx
```diff
@@ -118,13 +122,60 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
...
   const { onCancel, className, redirectHandler, namespace } = props;
 
-  const onSubmit = async (event) => {
+  const [data, loaded, loadError] = useK8sList<NamespaceStoreKind>(
+    NooBaaNamespaceStoreModel,
+    namespace
+  );
+
+  const existingNames = React.useMemo(() => {
+    return loaded && !loadError ? data?.map((data) => getName(data)) : [];
```
GowthamShanmugam, line 133: same here

packages/shared/input-with-requirements/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as TextInputWithFieldRequirements } from './TextInputWithFieldRequirements';
+export { default as useFieldRequirements } from './useFieldRequirements';
```
GowthamShanmugam, line 2:
```suggestion
export TextInputWithFieldRequirements from './TextInputWithFieldRequirements';
export  useFieldRequirements from './useFieldRequirements';
```

packages/shared/yup-validation-resolver/index.ts
```diff
@@ -0,0 +1 @@
+export { default as useYupValidationResolver } from './useYupValidationResolver';
```
GowthamShanmugam, line 1:
```suggestion
export useYupValidationResolver  from './useYupValidationResolver';
```

packages/ocs/block-pool/body.tsx
ivoalmeida: a return IS required within a {} block. using a {} block in an arrow function is a preference (when there's only one return line) and sometimes necessary when you need a breakpoint, which was my case. Regardless, I've removed the {} block and return.

packages/ocs/block-pool/body.tsx
ivoalmeida: done

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
ivoalmeida: done. see previous comment about return within {} block

packages/odf/components/create-bs/create-bs.tsx
ivoalmeida: done. see previous comment about return within {} block

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/body.tsx
ivoalmeida: done. see previous comment about return within {} block

packages/odf/components/create-storage-system/create-storage-system-steps/create-storage-class-step/create-storage-class-step.tsx
ivoalmeida: done. see previous comment about return within {} block

packages/odf/components/create-storage-system/create-storage-system-steps/create-storage-class-step/create-storage-class-step.tsx
ivoalmeida: done

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/body.tsx
ivoalmeida: done

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
ivoalmeida: done. see previous comment about return within {} block

packages/odf/components/namespace-store/namespace-store-form.tsx
ivoalmeida: done. see previous comment about return within {} block

packages/shared/input-with-requirements/index.ts
ivoalmeida: I'm sorry but I think you are confused? Your suggestion is not a valid export and it raises an error straightaway. 

Do you want to elaboarte why you feel the exports needs to be changed?

packages/shared/yup-validation-resolver/index.ts
ivoalmeida: I'm sorry but I think you are confused? Your suggestion is not a valid export and it raises an error straightaway. 

Do you want to elaboarte why you feel the exports needs to be changed?

GowthamShanmugam:
LGTM
/assign @SanjalKatiyar

packages/shared/input-with-requirements/TextInputWithFieldRequirements.tsx
```diff
@@ -35,7 +41,7 @@ export const ValidationIcon: React.FC<ValidationIconProp> = ({ status }) => {
   const getStatusIcon = React.useCallback(() => {
```
SanjalKatiyar, line 41: can we move "getStatusIcon" and "getVariant" functions outside this FC...
I don't really see any need of using "useCallback" here... we can do the same thing without using extra hook, by moving function declaration outside...

SanjalKatiyar, line 41: can we move "getStatusIcon" and "getVariant" functions outside this FC...
I don't really see any need of using "useCallback" here... we can do the same thing without using extra hook, by moving function declaration outside...

SanjalKatiyar: "getVariant" is returning a string, which will be checked by value...
"getStatusIcon" is returning a FC, I doubt that React will treat a returned component as a new one if function is called again and again...

packages/shared/yup-validation-resolver/index.ts
GowthamShanmugam: oh! i will check this

packages/ocs/block-pool/body.tsx
```diff
@@ -79,6 +86,64 @@ export const BlockPoolBody = (props: BlockPoolBodyPros) => {
...
+  const existingNames = React.useMemo(
+    () => (loaded && !loadError ? data?.map((data) => getName(data)) : []),
+    [data, loaded, loadError]
+  );
+
+  const fieldRequirements = [
+    t('No more than 253 characters'),
+    t('Starts and ends with a lowercase letter or number'),
+    t('Only lowercase letters, numbers, non-consecutive periods, or hyphens'),
+    t('Cannot be used before'),
```
SanjalKatiyar, line 100: will not block the PR on this, but this is not very intuitive... we should confirm with UX once...
something like this (only suggestion):
```suggestion
    t('An unique name for this resource'),
```
```suggestion
    t('Should be an unique within same Namespace'),
```

packages/shared/input-with-requirements/TextInputWithFieldRequirements.tsx
ivoalmeida: @SanjalKatiyar useCallback was Bipul's idea from previous PR. Agree with you though. Change done

packages/ocs/block-pool/body.tsx
ivoalmeida: I cannot change the copy as per your suggestion as it came from the the UX writing team. 
https://docs.google.com/document/d/1yWm6NSHKjFslC09NYw82G7_hFvyTTjR9nEeT88QG2E8/

locales/en/plugin__odf-console.json
```diff
@@ -742,6 +746,7 @@
   "Create new NamespaceStore ": "Create new NamespaceStore ",
   "An error has occurred while fetching namespace stores": "An error has occurred while fetching namespace stores",
   "Select a namespace store": "Select a namespace store",
+  "A unique name for the namespace within the project": "A unique name for the namespace within the project",
```
SanjalKatiyar, line 749:
```suggestion
  "An unique name for the namespace within the project": "A unique name for the namespace within the project",
```

packages/ocs/block-pool/body.tsx
SanjalKatiyar: Okay... not an important issue... we can fix later if any QE/customer points this out...

packages/ocs/block-pool/body.tsx
```diff
@@ -79,6 +86,64 @@ export const BlockPoolBody = (props: BlockPoolBodyPros) => {
     []
   );
 
+  const [data, loaded, loadError] = useK8sList(CephBlockPoolModel);
```
SanjalKatiyar, line 89:
```suggestion
  const [data, loaded, loadError] = useK8sList(CephBlockPoolModel, CEPH_NS);
```

packages/ocs/block-pool/body.tsx
SanjalKatiyar: somwhere we are using "Cannot be used before" and somewhere "A unique name for the resource within a project", does not make sense to me :/
 "Cannot be used before" >> should be used for Cluster wide resources
 "A unique name for the resource within a project" >> should be used for Namespace wide resources.

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
```diff
@@ -44,19 +48,68 @@ const GeneralPage: React.FC<GeneralPageProps> = ({ dispatch, state }) => {
 
   const [showHelp, setShowHelp] = React.useState(true);
 
-  const [validated, setValidated] = React.useState<ValidatedOptions>(
-    ValidatedOptions.default
+  const isNamespaceStoreSupported = useFlag(FEATURES.OCS_NAMESPACE_STORE);
+
+  const [data, loaded, loadError] = useK8sList<BucketClassKind>(
+    NooBaaBucketClassModel
   );
```
SanjalKatiyar, line 55:
```suggestion
  const [data, loaded, loadError] = useK8sList<BucketClassKind>(
    NooBaaBucketClassModel,
    namespace
  );
```

SanjalKatiyar: pass namespace as a prop to "GeneralPage", it should be available in the component which is calling "GeneralPage".

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
```diff
@@ -44,19 +48,68 @@ const GeneralPage: React.FC<GeneralPageProps> = ({ dispatch, state }) => {
...
+  const existingNames = React.useMemo(
+    () => (loaded && !loadError ? data?.map((data) => getName(data)) : []),
+    [data, loaded, loadError]
+  );
+
+  const fieldRequirements = [
+    t('3-63 characters'),
+    t('Starts and ends with a lowercase letter or number'),
+    t('Only lowercase letters, numbers, non-consecutive periods, or hyphens'),
+    t('Globally unique name'),
```
SanjalKatiyar, line 66: Can u check the BucketClass CR once ? I am pretty sure it is a Namespace wide resource not a Cluster wide...
in that case we can't show it should be a "Globally unique name"...

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
ivoalmeida: where is the ```namespace``` coming from?

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
ivoalmeida: what do you mean by "check the BucketClass CR"?

packages/shared/yup-validation-resolver/useYupValidationResolver.ts
```diff
@@ -0,0 +1,49 @@
...
+                },
+              };
+            },
+            {}
+          ),
+        })),
+    [validationSchema]
+  );
+
+export default useYupValidationResolver;
```
SanjalKatiyar, line 49: why create a hook ? we can make it a normal JS function instead...
```suggestion
const yupValidationResolver = <T>(validationSchema: AnySchema): Resolver =>
    async (data: T) =>
      validationSchema
        ?.validate(data, {
          abortEarly: false,
        })
        .then((values: T) => ({ values, errors: {} }))
        .catch((errors: any) => ({
          values: {},
          errors: errors?.inner?.reduce(
            (
              allErrors: Record<string, ValidationErrorMessages>,
              currentError: ValidationError
            ) => {
              return {
                ...allErrors,
                [currentError.path]: {
                  type: currentError.type ?? 'validation',
                  message: currentError.message,
                  messages: {
                    ...allErrors[currentError.path]?.messages,
                    [currentError.message]: {
                      type: currentError.type ?? 'validation',
                      field: currentError.path,
                    },
                  },
                },
              };
            },
            {}
          ),
        }));

export default yupValidationResolver;
```

packages/shared/yup-validation-resolver/useYupValidationResolver.ts
```diff
@@ -0,0 +1,49 @@
...
+export type ValidationErrorMessage = { type: string; field?: string };
+
+export type ValidationErrorMessages = {
+  type?: string;
+  message?: string;
+  messages?: Record<string, ValidationErrorMessage>;
+};
+
+const useYupValidationResolver = <T>(validationSchema: AnySchema): Resolver =>
+  useCallback(
```
SanjalKatiyar, line 14: no need for it to be a hook IMO... it can just be a normal JS function... u can use currying to pass on the  "validationSchema" to this function...
```suggestion
const yupValidationResolver = <T>(validationSchema: AnySchema): Resolver =>
```

packages/shared/yup-validation-resolver/useYupValidationResolver.ts
SanjalKatiyar, line 49:
```suggestion
  );

export default yupValidationResolver;
```

packages/ocs/block-pool/body.tsx
```diff
@@ -79,6 +86,64 @@ export const BlockPoolBody = (props: BlockPoolBodyPros) => {
...
+  });
+  const resolver = useYupValidationResolver(schema);
+  const {
+    control,
+    watch,
+    formState: { isValid },
+  } = useForm({
+    mode: 'all',
+    reValidateMode: 'onChange',
+    resolver: resolver,
```
SanjalKatiyar, line 129: use currying with "yupValidationResolver" so to pass on this schema...
```suggestion
    resolver: yupValidationResolver(schema),
```

packages/ocs/block-pool/body.tsx
```diff
@@ -79,6 +86,64 @@ export const BlockPoolBody = (props: BlockPoolBodyPros) => {
...
+        validationRegEx.alphaNumericsPeriodsHyphensNonConsecutive,
+        fieldRequirements[2]
+      )
+      .test(
+        'unique-name',
+        fieldRequirements[3],
+        (value: string) => !existingNames.includes(value)
+      ),
+  });
+  const resolver = useYupValidationResolver(schema);
```
SanjalKatiyar, line 121: no need for a hook, can be a normal JS function...

packages/shared/yup-validation-resolver/useYupValidationResolver.ts
```diff
@@ -0,0 +1,49 @@
+import { useCallback } from 'react';
```
SanjalKatiyar, line 1: do we need a custom resolver, aren't there any libraries for this ??
no issues with custom, just a doubt...

packages/ocs/block-pool/body.tsx
```diff
@@ -79,6 +86,64 @@ export const BlockPoolBody = (props: BlockPoolBodyPros) => {
     []
   );
 
+  const [data, loaded, loadError] = useK8sList(CephBlockPoolModel);
+
+  const existingNames = React.useMemo(
+    () => (loaded && !loadError ? data?.map((data) => getName(data)) : []),
+    [data, loaded, loadError]
+  );
```
SanjalKatiyar, line 94: I will leave this up to you, u can fix it if u want else it is okay currently as well...
but "useMemo" is only for expensive operations or for cases where we want to preserve the reference of the object/list... here both conditions are not met...
I know we are doing this wrong at other places in our repo as well :P

packages/ocs/block-pool/body.tsx
```diff
@@ -79,6 +86,64 @@ export const BlockPoolBody = (props: BlockPoolBodyPros) => {
...
+
+  const poolName = watch('newPoolName');
+
+  React.useEffect(() => {
+    dispatch({
+      type: BlockPoolActionType.SET_POOL_NAME,
+      payload: isValid ? poolName : '',
+    });
+  }, [poolName, dispatch, isValid]);
+
```
SanjalKatiyar, line 146: all of this, seems to me could be shifted a new FucntionalComponent itself...
but again I will leave this up to you... no strict requirement...

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
```diff
@@ -44,19 +48,68 @@ const GeneralPage: React.FC<GeneralPageProps> = ({ dispatch, state }) => {
...
-    if (validateBucketClassName(value)) {
-      setValidated(ValidatedOptions.success);
-    } else {
-      setValidated(ValidatedOptions.error);
-    }
-  };
+  const existingNames = React.useMemo(
+    () => (loaded && !loadError ? data?.map((data) => getName(data)) : []),
+    [data, loaded, loadError]
+  );
```
SanjalKatiyar, line 60: same as "block-pool" comment...

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
```diff
@@ -44,19 +48,68 @@ const GeneralPage: React.FC<GeneralPageProps> = ({ dispatch, state }) => {
...
+  });
+  const resolver = useYupValidationResolver(schema);
+  const {
+    control,
+    formState: { isValid },
+    watch,
+  } = useForm({
+    mode: 'all',
+    reValidateMode: 'onChange',
+    resolver: resolver,
```
SanjalKatiyar, line 96: same as "block-pool" comment...

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
```diff
@@ -44,19 +48,68 @@ const GeneralPage: React.FC<GeneralPageProps> = ({ dispatch, state }) => {
...
+    context: undefined,
+    criteriaMode: 'firstError',
+    shouldFocusError: true,
+    shouldUnregister: false,
+    shouldUseNativeValidation: false,
+    delayError: undefined,
+  });
+
+  const bucketClassName = watch('bucketclassname-input');
```
SanjalKatiyar, line 105: same as "block-pool" comment...
type and whether ref will be same across re renders ?

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
```diff
@@ -110,40 +163,33 @@ const GeneralPage: React.FC<GeneralPageProps> = ({ dispatch, state }) => {
...
+          popoverProps={{
+            headerContent: t('Name requirements'),
+            footerContent: `${t('Example')}: my-multi-cloud-mirror`,
+          }}
+          formGroupProps={{
+            className: 'nb-create-bc-step-page-form__element',
+            fieldId: 'bucketclassname-input',
+            label: t('BucketClass name'),
+            helperText: t(
+              'A unique name for the bucket class within the project.'
```
SanjalKatiyar, line 179: it is a convention: CR names should be mentioned same as they appear in their CR's "kind".
```suggestion
              'A unique name for the BucketClass within the project.'
```

packages/odf/components/create-bs/create-bs.tsx
```diff
@@ -78,9 +70,64 @@ const CreateBackingStoreForm: React.FC<CreateBackingStoreFormProps> = (
...
-  const onSubmit = (event) => {
+  const [data, loaded, loadError] = useK8sList<BackingStoreKind>(
+    NooBaaBackingStoreModel,
+    namespace
+  );
+
+  const existingNames = React.useMemo(
+    () => (loaded && !loadError ? data?.map((data) => getName(data)) : []),
+    [data, loaded, loadError]
+  );
```
SanjalKatiyar, line 81: same as "block-pool" comment...

packages/odf/components/create-bs/create-bs.tsx
```diff
@@ -78,9 +70,64 @@ const CreateBackingStoreForm: React.FC<CreateBackingStoreFormProps> = (
...
+  const resolver = useYupValidationResolver(schema);
+
+  const {
+    control,
+    handleSubmit,
+    formState: { isValid, isValidating },
+  } = useForm({
+    mode: 'all',
+    reValidateMode: 'onChange',
+    resolver: resolver,
```
SanjalKatiyar, line 118: same as "block-pool" comment...

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/body.tsx
```diff
@@ -149,6 +158,68 @@ export const LocalVolumeSetBody: React.FC<LocalVolumeSetBodyProps> = ({
     setRadio(value);
   };
 
+  const [data, loaded, loadError] =
+    useK8sList<StorageSystemKind>(StorageClassModel);
+
+  const existingNames = React.useMemo(
+    () => (loaded && !loadError ? data?.map((data) => getName(data)) : []),
+    [data, loaded, loadError]
+  );
```
SanjalKatiyar, line 167: same as "block-pool" comment... up to you...

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/body.tsx
```diff
@@ -149,6 +158,68 @@ export const LocalVolumeSetBody: React.FC<LocalVolumeSetBodyProps> = ({
...
+  });
+  const resolver = useYupValidationResolver(schema);
+  const {
+    control,
+    watch,
+    formState: { isValid },
+  } = useForm({
+    mode: 'all',
+    reValidateMode: 'onChange',
+    resolver: resolver,
```
SanjalKatiyar, line 202: same as "block-pool" comment...

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/body.tsx
```diff
@@ -149,6 +158,68 @@ export const LocalVolumeSetBody: React.FC<LocalVolumeSetBodyProps> = ({
...
+    context: undefined,
+    criteriaMode: 'firstError',
+    shouldFocusError: true,
+    shouldUnregister: false,
+    shouldUseNativeValidation: false,
+    delayError: undefined,
+  });
+
+  const createLvsStorageClassName = watch('create-lvs-storage-class-name');
```
SanjalKatiyar, line 211: add types...

packages/odf/components/create-storage-system/create-storage-system-steps/create-storage-class-step/create-storage-class-step.tsx
```diff
@@ -38,23 +40,87 @@ export const CreateStorageClass: React.FC<CreateStorageClassProps> = ({
     [dispatch]
   );
 
+  const [data, loaded, loadError] =
+    useK8sList<StorageSystemKind>(StorageClassModel);
+
+  const existingNames = React.useMemo(
+    () => (loaded && !loadError ? data?.map((data) => getName(data)) : []),
+    [data, loaded, loadError]
+  );
```
SanjalKatiyar, line 49: same as "block-pool" comment...

packages/odf/components/create-storage-system/create-storage-system-steps/create-storage-class-step/create-storage-class-step.tsx
```diff
@@ -38,23 +40,87 @@ export const CreateStorageClass: React.FC<CreateStorageClassProps> = ({
...
+
+  const fieldRequirements = [
+    t('No more than 253 characters'),
+    t('Starts and ends with a lowercase letter or number'),
+    t('Only lowercase letters, numbers, non-consecutive periods, or hyphens'),
+    t('Cannot be used before'),
```
SanjalKatiyar, line 55: common text...
even for this: "No more than 253 characters", we can create:

> const charLimitText = (t: Tfunction, limit: number | string) => t(`No more than ${limit} characters`)

packages/odf/components/create-storage-system/create-storage-system-steps/create-storage-class-step/create-storage-class-step.tsx
```diff
@@ -38,23 +40,87 @@ export const CreateStorageClass: React.FC<CreateStorageClassProps> = ({
...
+  });
+  const resolver = useYupValidationResolver(schema);
+  const {
+    control,
+    watch,
+    formState: { isValid },
+  } = useForm({
+    mode: 'all',
+    reValidateMode: 'onChange',
+    resolver: resolver,
```
SanjalKatiyar, line 84: same as "block-pool" comment...

packages/odf/components/create-storage-system/create-storage-system-steps/create-storage-class-step/create-storage-class-step.tsx
```diff
@@ -38,23 +40,87 @@ export const CreateStorageClass: React.FC<CreateStorageClassProps> = ({
...
+    context: undefined,
+    criteriaMode: 'firstError',
+    shouldFocusError: true,
+    shouldUnregister: false,
+    shouldUseNativeValidation: false,
+    delayError: undefined,
+  });
+
+  const storageClassName = watch('storage-class-name');
```
SanjalKatiyar, line 93: type...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -185,31 +191,92 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
 
+  const [data, loaded, loadError] = useK8sList(
+    NooBaaObjectBucketClaimModel,
+    namespace
+  );
+
+  const existingNames = React.useMemo(
+    () => (loaded && !loadError ? data?.map((data) => getName(data)) : []),
+    [data, loaded, loadError]
+  );
```
SanjalKatiyar, line 202: same...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -185,31 +191,92 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+
+  const fieldRequirements = [
+    t('No more than 253 characters'),
+    t('Starts and ends with a lowercase letter or number'),
+    t('Only lowercase letters, numbers, non-consecutive periods, or hyphens'),
+    t('Cannot be used before'),
+  ];
```
SanjalKatiyar, line 209: same...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -185,31 +191,92 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+  const resolver = useYupValidationResolver(schema);
+
+  const {
+    control,
+    watch,
+    formState: { isValid },
+  } = useForm({
+    mode: 'all',
+    reValidateMode: 'onChange',
+    resolver: resolver,
```
SanjalKatiyar, line 239: same...

packages/odf/components/namespace-store/namespace-store-form.tsx
```diff
@@ -118,13 +122,61 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
...
+    () => (loaded && !loadError ? data?.map((data) => getName(data)) : []),
+    [data, loaded, loadError]
+  );
+
+  const fieldRequirements = [
+    t('No more than 43 characters'),
+    t('Starts and ends with a lowercase letter or number'),
+    t('Only lowercase letters, numbers, non-consecutive periods, or hyphens'),
+    t('A unique name for the namespace within the project'),
+  ];
```
SanjalKatiyar, line 142: same...

packages/odf/components/namespace-store/namespace-store-form.tsx
```diff
@@ -118,13 +122,61 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
...
+        'unique-name',
+        fieldRequirements[3],
+        (value: string) => !existingNames.includes(value)
+      ),
+  });
+  const resolver = useYupValidationResolver(schema);
+  const { control, handleSubmit } = useForm({
+    mode: 'all',
+    reValidateMode: 'onChange',
+    resolver: resolver,
```
SanjalKatiyar, line 166: same...

packages/odf/components/namespace-store/namespace-store-form.tsx
```diff
@@ -204,32 +256,32 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
...
+        popoverProps={{
+          headerContent: t('Name requirements'),
+          footerContent: `${t('Example')}: my-namespacestore`,
+        }}
+        formGroupProps={{
+          label: t('Namespace store name'),
+          fieldId: 'namespacestore-name',
+          className: 'nb-endpoints-form-entry',
+          helperText: t(
+            'A unique name for the namespace store within the project'
```
SanjalKatiyar, line 274:
```suggestion
            'A unique name for the NamespaceStore within the project'
```

packages/odf/utils/validation.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Only lowercase letters, numbers, non-consecutive periods, or hyphens
+ */
+export const alphaNumericsPeriodsHyphensNonConsecutive =
+  /^[a-z0-9]+([a-z0-9]|([-.](?![-.])))*[a-z0-9]*$/;
+
+export default {
+  startAndEndsWithAlphanumerics,
+  alphaNumericsPeriodsHyphensNonConsecutive,
+};
```
SanjalKatiyar, line 16: any reason for exporting both as separate and then together as a default as well ??

packages/shared/input-with-requirements/TextInputWithFieldRequirements.tsx
```diff
@@ -23,49 +24,62 @@ import './TextInputWithFieldRequirements.scss';
...
+      return 'success';
+    default:
+      return 'info';
+  }
+};
 
-  return <Icon status={getVariant()}>{getStatusIcon()}</Icon>;
+export const ValidationIcon: React.FC<ValidationIconProp> = ({ status }) => {
+  return <Icon status={getVariant(status)}>{getStatusIcon(status)}</Icon>;
 };
```
SanjalKatiyar, line 65: not really needed to be a separate FC anymore... we can directly call "Icon", instead of calling "ValidationIcon"...

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
SanjalKatiyar: from the match.param, if not present there default is set to "openshift-storage"...
check the component which is calling "GeneralPage"...

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
SanjalKatiyar: CR is Kubernetes terminology... means CustomResource...
in all the "Create" pages we are creating CRs only...
ex: "Create BackingStore" UI flow will create a BackingStore CR for the user based upon all the inputs...

packages/ocs/block-pool/body.tsx
ivoalmeida: I cannot change the copy as per your suggestion as it came from the the UX writing team.
https://docs.google.com/document/d/1yWm6NSHKjFslC09NYw82G7_hFvyTTjR9nEeT88QG2E8/

SanjalKatiyar: coming from UX does not mean it is 100% correct, that is what we need to discuss with UX and get it fixed...
but not blocking the PR on this for now... but we will have to send a follow up...
this is common understanding that:
```
"Cannot be used before" >> should be used for Cluster wide resources
"A unique name for the resource within a project" >> should be used for Namespace wide resources.
```

SanjalKatiyar: OBC/BucketClass/BlockPool/NamespaceStore/BackingStore are namespace wide resource... there names are unique within a particular namespace, but we can create a new CR with same name in different namespace.

SanjalKatiyar: StorageClass is a cluster wide resource... in this case name should always be unique globally...

packages/ocs/block-pool/body.tsx
ivoalmeida: Alternative to this would be to use other 2 hooks, useState to store the values, and useEffect to map the results, calling getName on each one of them set state whenever it changes, so the validation that gets triggered on every key stroke can check the state for already existing names.

SanjalKatiyar: no no.. I am suggesting to even remove this hook as well... adding 2 instead of currently 1 is something I definitely don;t want...

SanjalKatiyar: I meant this to be  just like this:
```
  let existingNames = [];
    if (loaded && !loadError) 
       existingNames = data?.map((data) => getName(data));
```

SanjalKatiyar: that's all... but definitely don't use 2 hook... 1 hook (current) is still better than 2...
what I meant by "I will leave this up to you" was to either keep it as it is or don't even use any hook here...

SanjalKatiyar: because "existingNames" list will not be too (like millions or so, will be in thousand in worst case) big and this itself is not a dependency to any other hook so keep ref same won't affect us...

packages/ocs/block-pool/body.tsx
```diff
@@ -79,6 +86,64 @@ export const BlockPoolBody = (props: BlockPoolBodyPros) => {
...
+    resolver: resolver,
+    context: undefined,
+    criteriaMode: 'firstError',
+    shouldFocusError: true,
+    shouldUnregister: false,
+    shouldUseNativeValidation: false,
+    delayError: undefined,
+  });
+
+  const poolName = watch('newPoolName');
```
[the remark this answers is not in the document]
ivoalmeida: I've added the type to the variables even though is not necessary . ```watch``` would only ever trigger a re-render if the field it subscribes to changes.

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
ivoalmeida: Sorry but I still have no idea what you want me to do here. Does the copy seem incorrect to you? If so I cannot/should not change it myself. UX team need to make a decision on this, then update the document https://docs.google.com/document/d/1yWm6NSHKjFslC09NYw82G7_hFvyTTjR9nEeT88QG2E8/ so QA can test this accordingly

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
ivoalmeida: done

packages/ocs/block-pool/body.tsx
ivoalmeida: done

packages/ocs/block-pool/body.tsx
ivoalmeida: Alternative to this would be to use other 2 hooks, useState to store the values, and useEffect to map the results, calling getName on each one of them set state whenever it changes, so the validation that gets triggered on every key stroke can check the state for already existing names.

packages/ocs/block-pool/body.tsx
ivoalmeida: I've never said that because it came from UX it is correct. I cannot/should not change this myself. UX team need to make a decision on this, then update the document https://docs.google.com/document/d/1yWm6NSHKjFslC09NYw82G7_hFvyTTjR9nEeT88QG2E8/ so QA can test this accordingly and then I can submit a change.

packages/shared/yup-validation-resolver/useYupValidationResolver.ts
ivoalmeida: I actually disagree. This is used in the scope of a React component which depends on a schema object that depends on other hooks inside this scope. so even if it was a function rather than a hook it would still need to be wrapped inside a useCallback hook. having this as a hook means I can have this bit if implementation using useCallback only in one place (the hook) instead of having to repeat this implementation everywhere where I would consume this function.

packages/ocs/block-pool/body.tsx
ivoalmeida: useForm needs to be in the form component. fieldRequirements is used also by TextInputWithFieldRequirements component. The schema will be different for each form (it should be a reflection of the fields of each form that need any kind of validation).

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
ivoalmeida: ```watch``` would only ever trigger a re-render if the field it subscribes to changes.

packages/ocs/block-pool/body.tsx
SanjalKatiyar: I agree, yup... that is why I said "we need to discuss with UX and get it fixed... but not blocking the PR on this for now"

packages/ocs/block-pool/body.tsx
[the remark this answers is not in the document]
SanjalKatiyar: it is helpful when doing the PR reviews... local IDE can tell us the type just by hovering over a variable but that's not the case for github... so good to add at some places...
moveover I was not worried about re render caused by "watch" I was worried about any re render in general...
because we are using "poolName" in a dependency list of a hook so its reference should be same...
that is the reason why I asked for the type... now I know output of "watch('newPoolName')" is only a string so I don;t have to worry about reference anymore :)

packages/odf/utils/validation.ts
ivoalmeida: it's good practice to always have a default export and give option of individual exports

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
ivoalmeida: done

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/body.tsx
ivoalmeida: added type

packages/odf/components/namespace-store/namespace-store-form.tsx
ivoalmeida: done

packages/ocs/block-pool/body.tsx
bipuladh: lets keep it as a hook please.

packages/ocs/block-pool/body.tsx
```diff
@@ -79,6 +87,67 @@ export const BlockPoolBody = (props: BlockPoolBodyPros) => {
...
+
+  const resolver = useYupValidationResolver(schema);
+  const {
+    control,
+    watch,
+    formState: { isValid },
+  } = useForm({
+    mode: 'all',
+    reValidateMode: 'onChange',
+    resolver: resolver,
```
bipuladh, line 133:
```suggestion
    resolver,
```
nit

bipuladh (commented):
Overall the PR looks good. Thank you.

packages/shared/yup-validation-resolver/useYupValidationResolver.ts
```diff
@@ -0,0 +1,49 @@
...
+                    ...allErrors[currentError.path]?.messages,
+                    [currentError.message]: {
+                      type: currentError.type ?? 'validation',
+                      field: currentError.path,
+                    },
+                  },
+                },
+              };
+            },
+            {}
```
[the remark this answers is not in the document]
bipuladh: ignore this. resolving.

packages/shared/yup-validation-resolver/useYupValidationResolver.ts
ivoalmeida: will continue as hook like we agreed on call so  this change is unnecessary

packages/ocs/block-pool/body.tsx
ivoalmeida: will continue as hook like we agreed on call so  this change is unnecessary

packages/ocs/block-pool/body.tsx
ivoalmeida: this was moved inside a useMemo together with schema and fieldRequirements like agreed on call

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
ivoalmeida: this was moved inside a useMemo together with schema and fieldRequirements like agreed on call

packages/odf/components/bucket-class/wizard-pages/general-page.tsx
ivoalmeida: useYupValidationResolver will continue as a hook

packages/odf/components/create-bs/create-bs.tsx
ivoalmeida: this was moved inside a useMemo together with schema and fieldRequirements like agreed on call

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/body.tsx
ivoalmeida: this was moved inside a useMemo together with schema and fieldRequirements like agreed on call

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/body.tsx
```diff
@@ -149,6 +158,68 @@ export const LocalVolumeSetBody: React.FC<LocalVolumeSetBodyProps> = ({
...
+  const existingNames = React.useMemo(
+    () => (loaded && !loadError ? data?.map((data) => getName(data)) : []),
+    [data, loaded, loadError]
+  );
+
+  const fieldRequirements = [
+    t('No more than 253 characters'),
+    t('Starts and ends with a lowercase letter or number'),
+    t('Only lowercase letters, numbers, non-consecutive periods, or hyphens'),
+    t('Cannot be used before'),
```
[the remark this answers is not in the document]
ivoalmeida: done

packages/odf/components/create-storage-system/create-storage-system-steps/create-storage-class-step/create-storage-class-step.tsx
ivoalmeida: this was moved inside a useMemo together with schema and fieldRequirements like agreed on call

packages/odf/components/create-storage-system/create-storage-system-steps/create-storage-class-step/create-storage-class-step.tsx
ivoalmeida: done

packages/odf/components/create-storage-system/create-storage-system-steps/create-storage-class-step/create-storage-class-step.tsx
ivoalmeida: done

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
ivoalmeida: this was moved inside a useMemo together with schema and fieldRequirements like agreed on call

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
ivoalmeida: done

packages/odf/components/namespace-store/namespace-store-form.tsx
ivoalmeida: this was moved inside a useMemo together with schema and fieldRequirements like agreed on call

packages/odf/components/create-bs/create-bs.tsx
ivoalmeida: useYupValidationResolver will continue as a hook

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/body.tsx
ivoalmeida: useYupValidationResolver will continue as a hook

packages/odf/components/create-storage-system/create-storage-system-steps/create-storage-class-step/create-storage-class-step.tsx
ivoalmeida: useYupValidationResolver will continue as a hook

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
ivoalmeida: useYupValidationResolver will continue as a hook

packages/odf/components/namespace-store/namespace-store-form.tsx
ivoalmeida: useYupValidationResolver will continue as a hook

packages/shared/yup-validation-resolver/useYupValidationResolver.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export type ValidationErrorMessages = {
+  type?: string;
+  message?: string;
+  messages?: Record<string, ValidationErrorMessage>;
+};
+
+const useYupValidationResolver = <T>(validationSchema: AnySchema): Resolver =>
+  useCallback(
+    async (data: T) =>
```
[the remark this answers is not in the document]
ivoalmeida: I removed async. This was a leftover from a re-factoring

packages/ocs/block-pool/body.tsx
ivoalmeida: done

packages/ocs/block-pool/body.tsx
SanjalKatiyar: yeah... earlier "schema" ref was changing so it wasn;t making any sense to create a hook as "useCallback" was not doing what it was supposed to do...
had a discussion later and decided to memoise "schema" and other references...

packages/shared/yup-validation-resolver/useYupValidationResolver.ts
SanjalKatiyar: this comment is resolved... had a discussion over call and we are keeping it to be a hook with memoizing its props...

SanjalKatiyar:
need to update the validation text based upon whether a CR is cluster-wide or ns-wide, will be done is subsequent PRs...
/approve
/lgtm

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -185,31 +192,102 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+          (value: string) => !existingNames.includes(value)
+        )
+        .notRequired()
+        .transform((value: string) =>
+          value &&
+          Object.values(value).some(
+            (v) => !(v === null || v === undefined || v === '')
+          )
+            ? value
+            : undefined
```
SanjalKatiyar, line 234: I didn't get this... we are receiving input as string... we are creating an array of characters from that and checking if "some" of the characters are not `null | undefined | ''` ... why ?
can't we just do `!!value ? value : undefined` ??

SanjalKatiyar: can u plz give an example input where above condition will be useful ??

SanjalKatiyar: just curious... that's all...

ivoalmeida: You're right. I was testing this on more generic use cases other than just  for string values hence the Object.values.
I changed the code as per your suggestion.

bipuladh:
Pages are somewhat breaking only on CI. Tested by creating a docker image and stuffs are working fine.

PR #620, opened 2023-02-02
Bug 2165128: removed the subversion check for the mco operator

debjyoti-pandit:
fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2165128

image

### DIFF
--- a/packages/mco/components/disaster-recovery/create-dr-policy/create-dr-policy.tsx
+++ b/packages/mco/components/disaster-recovery/create-dr-policy/create-dr-policy.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { getMajorVersion } from '@odf/mco/utils';
 import { CEPH_STORAGE_NAMESPACE } from '@odf/shared/constants/common';
 import PageHeading from '@odf/shared/heading/page-heading';
 import { useFetchCsv } from '@odf/shared/hooks/use-fetch-csv';
@@ -82,11 +83,7 @@ export const CreateDRPolicy: React.FC<ReRouteResourceProps> = ({
     specName: ODFMCO_OPERATOR,
     cluster: HUB_CLUSTER_NAME,
   });
-  const odfMCOVersion = React.useMemo(() => {
-    const version = csv?.spec?.version;
-    // converting z-stream release version to main release version
-    return version ? version.substring(0, version.lastIndexOf('.')) + '.0' : '';
-  }, [csv]);
+  const odfMCOVersion = getMajorVersion(csv?.spec?.version);
 
   React.useEffect(() => {
     if (state.selectedClusters.length === 2) {
--- a/packages/mco/utils/common.ts
+++ b/packages/mco/utils/common.ts
@@ -33,3 +33,9 @@ export const createRefFromK8Resource = (
   kind: resource?.kind,
   uid: getUID(resource),
 });
+
+export const getMajorVersion = (version: string): string => {
+  return !!version
+    ? version.split('.')[0] + '.' + version.split('.')[1] + '.0'
+    : '';
+};
--- a/packages/mco/components/disaster-recovery/create-dr-policy/select-cluster-list.tsx
+++ b/packages/mco/components/disaster-recovery/create-dr-policy/select-cluster-list.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { isMinimumSupportedODFVersion } from '@odf/mco/utils';
+import { getMajorVersion, isMinimumSupportedODFVersion } from '@odf/mco/utils';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
 import { referenceForModel } from '@odf/shared/utils';
 import { useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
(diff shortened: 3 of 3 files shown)


packages/mco/components/disaster-recovery/create-dr-policy/create-dr-policy.tsx
```diff
@@ -82,11 +83,10 @@ export const CreateDRPolicy: React.FC<ReRouteResourceProps> = ({
...
   });
-  const odfMCOVersion = React.useMemo(() => {
-    const version = csv?.spec?.version;
-    // converting z-stream release version to main release version
-    return version ? version.substring(0, version.lastIndexOf('.')) + '.0' : '';
-  }, [csv]);
+  const odfMCOVersion = React.useMemo(
+    () => getMajorVersion(csv?.spec?.version),
+    [csv]
+  );
```
SanjalKatiyar, line 89:
```suggestion
  const odfMCOVersion = getMajorVersion(csv?.spec?.version);
```
useMemo is for expensive operations. No need here...

packages/mco/utils/common.ts
```diff
@@ -33,3 +35,7 @@ export const createRefFromK8Resource = (
   kind: resource?.kind,
   uid: getUID(resource),
 });
+
+export const getMajorVersion = (version: string): string => {
+  return version ? major(version) + '.' + minor(version) + '.0' : '';
```
SanjalKatiyar, line 40:
```suggestion
  return !!version ? major(version) + '.' + minor(version) + '.0' : '';
```

package.json
```diff
@@ -83,6 +83,7 @@
     "react-router-dom": "5.2.0",
     "react-tagsinput": "^3.19.0",
     "redux": "^4.0.0",
+    "semver": "^7.3.8",
```
SanjalKatiyar, line 86:
```suggestion
```

package.json
```diff
@@ -103,6 +104,7 @@
     "@types/react": "16.8.13",
     "@types/react-helmet": "^6.1.1",
     "@types/react-router-dom": "5.1.2",
+    "@types/semver": "^7.3.13",
```
SanjalKatiyar, line 107:
```suggestion
```

package.json
SanjalKatiyar, line 86: @GowthamShanmugam we need to backport this PR, do u think it is okay to add new dependency ??

SanjalKatiyar: @GowthamShanmugam if so we need to make sure we won't break the new 4.12.z builds...

packages/mco/utils/common.ts
```diff
@@ -10,6 +10,8 @@ import {
   ObjectReference,
   K8sResourceCommon,
 } from '@openshift-console/dynamic-plugin-sdk';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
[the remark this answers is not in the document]
debjyoti-pandit: It was throwing eslint error as it is not part of the dependencies list.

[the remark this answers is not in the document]
SanjalKatiyar: it is part of dependency list (as per what I see in ur PR)... we need to check why we are getting this right...
we can't just ignore and add a comment without figuring out the reasoning behind it...

package.json
GowthamShanmugam: I think it is ok, we are not changing any existing dependency, so we are good here,

packages/mco/utils/common.ts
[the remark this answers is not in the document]
GowthamShanmugam: it might be due to cache

package.json
SanjalKatiyar: actually it will change... many libraries are internally using "semver" as an internal dependency, now since we added "semver" as our own dependency, those internal versions will get updated as well...
I hope it won't cause any issue. We have CI for ODF so we might be good, but for MCO it is good to test that it won;t break older release-4.12 !!

SanjalKatiyar: it is good to merge this for master, but for backport to 4.12, just make sure we don't introduce any regression...

packages/mco/utils/common.ts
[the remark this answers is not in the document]
SanjalKatiyar: cc @GowthamShanmugam

package.json
GowthamShanmugam: agreed, atleast do some basic test.

packages/mco/utils/common.ts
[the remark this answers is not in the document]
GowthamShanmugam: try to clean cache and node_modules and try

package.json
bipuladh: build teams use cachito so messing with dependencies can cause issues on that side.

bipuladh (changes requested):
I really don't see the need to add a new package for this use case.

package.json
debjyoti-pandit: @GowthamShanmugam https://github.com/red-hat-storage/odf-console/pull/620#pullrequestreview-1309043223

GowthamShanmugam (approved):

SanjalKatiyar:
@debjyoti-pandit plz update PR title properly...

GowthamShanmugam (approved):

PR #624, opened 2023-02-06
Bug 2167299: Change version formatting for details card

vbnrh:
This commit changes the way the version is formatted in the details cards (internal, external and object service dashboards)

Instead of odf-operator.v4.12.0, we will just show the semver of the operator.

This is done as part of the bluewash changes

Screenshot from 2023-02-06 15-09-45
Screenshot from 2023-02-06 15-09-33

Edit : 
PR updated to pick the version from the respective operator csvs and updated in all the details card.
The screenshots are the same, just with the `v` removed 

### DIFF
--- a/packages/odf/utils/odf.ts
+++ b/packages/odf/utils/odf.ts
@@ -1,4 +1,3 @@
-import { ODF_OPERATOR } from '@odf/shared/constants';
 import { ClusterServiceVersionKind } from '@odf/shared/types';
 import { K8sResourceKind } from '@odf/shared/types';
 import { StorageClassResourceKind } from '@odf/shared/types';
@@ -37,12 +36,4 @@ export const getStorageClassDescription = (
 };
 
 export const getOperatorVersion = (operator: K8sResourceKind): string =>
-  operator?.status?.installedCSV;
-
-export const getODFVersion = (items: K8sResourceKind[]): string => {
-  const operator: K8sResourceKind = _.find(
-    items,
-    (item) => item?.spec?.name === ODF_OPERATOR
-  );
-  return getOperatorVersion(operator);
-};
+  operator?.spec?.version;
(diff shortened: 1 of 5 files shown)


packages/odf/utils/odf.ts
```diff
@@ -44,5 +44,19 @@ export const getODFVersion = (items: K8sResourceKind[]): string => {
     items,
     (item) => item?.spec?.name === ODF_OPERATOR
   );
-  return getOperatorVersion(operator);
+
+  const version = getOperatorVersion(operator);
+  const temp = version?.split('.');
```
SanjalKatiyar, line 49: r we sure thin assumption will never break ??

vbnrh: Not unless it changes in subscription and the csv. 
I don't have a clear indication that it may change anytime soon in 4.13 or 4.14 unless we will be renaming odf-operator to something else completely.

SanjalKatiyar: why not just read it from `spec.version`

SanjalKatiyar: that way it is guaranteed to be x.y.z format... without csv name added to it...

vbnrh: Updated the code to pick the version from the csv. Has multiple changes, please take a look

SanjalKatiyar:
> Updated the code to pick the version from the csv. Has multiple changes, please take a look

Looks good, but we don't fetch entire list of CSV directly like this... use "useFetchCsv" hook instead... direct polling can be buggy sometimes...
@vbnrh

vbnrh:
> > Updated the code to pick the version from the csv. Has multiple changes, please take a look
> 
> Looks good, but we don't fetch entire list of CSV directly like this... use "useFetchCsv" hook instead... direct polling can be buggy sometimes... @vbnrh

Changes done

SanjalKatiyar:
@vbnrh backport PRs are up, plz add correct BZ number to the title and move BZ status to POST

PR #629, opened 2023-02-07
Bug 2167301: [release-4.12] Use generics names for Data Foundation

umangachapagain:
Backport of #602


locales/en/plugin__odf-console.json
```diff
@@ -352,7 +351,7 @@
   "An error occurred. Pool {{name}} was not created": "An error occurred. Pool {{name}} was not created",
   "Pool {{name}} creation timed out. Please check if odf operator and rook operator are running": "Pool {{name}} creation timed out. Please check if odf operator and rook operator are running",
   "The creation of a StorageCluster is still in progress or has failed. Try again after the StorageCuster is ready to use.": "The creation of a StorageCluster is still in progress or has failed. Try again after the StorageCuster is ready to use.",
-  "Pool management tasks are not supported for default pool and OpenShift Container Storage's external mode.": "Pool management tasks are not supported for default pool and OpenShift Container Storage's external mode.",
+  "Pool management tasks are not supported for default pool and Data Foundation's external mode.": "Pool management tasks are not supported for default pool and Data Foundation's external mode.",
```
SanjalKatiyar, line 354: "Pool management tasks are not supported for default pool and OpenShift Container Storage's external mode." >> this same key is still present in jp/ko/zh locales files... there as well we need to change "OpenShift Container Storage" to "Data Foundation".

umangachapagain:
If this is a flake, can we override the test?

SanjalKatiyar:
> If this is a flake, can we override the test?

yeah seems like flake... it should get merged after few re tries... if not by evening then I will override...

PR #631, opened 2023-02-07
add nfs to the deployment flow

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-4211

Screenshot from 2023-02-07 13-58-57
Screenshot from 2023-02-07 13-59-03
Screenshot from 2023-02-07 13-59-11
Screenshot from 2023-02-07 13-59-20
Screenshot from 2023-02-07 16-17-30


packages/odf/components/create-storage-system/reducer.ts
```diff
@@ -78,6 +79,7 @@ type CreateStorageSystemState = {
   nodes: WizardNodeState[];
   backingStorage: {
     type: BackingStorageType;
+    enableNFS: boolean;
```
GowthamShanmugam, line 82: enableNFS sounds like a function name, either we keep it as isNFSEnabled (or) nfsEnabled

SanjalKatiyar: I agree, but I did it just to keep it consistent with other state variable in this reducer file.
If u will check other fields, they are named as "enableArbiter" "enableTaint" etc... that is why I kept same here...

GowthamShanmugam (approved):

PR #632, opened 2023-02-09
Bug fix appset Apply policy modal

GowthamShanmugam:


GowthamShanmugam:
image

GowthamShanmugam:
Still, one fix is pending, without labels, we should not allow protecting apps.

PR #633, opened 2023-02-09
Bug 2168635: fix redirect link to operator details page

SanjalKatiyar:


vbnrh:
Adding a test coverage for this issue is very important so that we don't get it again.

SanjalKatiyar:
> Adding a test coverage for this issue is very important so that we don't get it again.

created: https://github.com/red-hat-storage/odf-console/issues/634 for this.

packages/ocs/dashboards/object-service/details-card/details-card.tsx
```diff
@@ -77,12 +79,12 @@ export const ObjectServiceDetailsCard: React.FC<{}> = () => {
...
       <CardBody>
         <DetailsBody>
-          <DetailItem
-            key="service_name"
-            title={t('Service name')}
-            isLoading={false}
-          >
-            <Link to={servicePath}>{serviceName}</Link>
+          <DetailItem key="service_name" title={t('Service name')}>
+            {csvLoaded && !csvLoadError ? (
```
vbnrh, line 83: We are either showing 'Data Foundation' or the link for it. 
Is it ok to remove the loading part if we are backporting it ?

vbnrh (changes requested):
Mostly looks good. Just one comment

packages/ocs/dashboards/object-service/details-card/details-card.tsx
SanjalKatiyar: by loading part do u mean "isLoading={false}" ??
it was always false so doesn't seem to be useful
if csv is loaded we will show link now... meanwhile it is loading it will just display name without any redirect link

PR #640, opened 2023-02-14
Dashbaord changes

GowthamShanmugam:


GowthamShanmugam:
image

GowthamShanmugam:
image

GowthamShanmugam:
Screenshot from 2023-02-14 23-59-15

GowthamShanmugam:
image

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
```diff
@@ -278,53 +254,33 @@ export const ReplicationHistorySection: React.FC<ReplicationHistorySectionProps>
...
-          <div className="mco-dashboard__contentRow mco-cluster-app__contentRow--flexEnd">
-            <PVCDropdown
-              remoteNS={remoteNS}
-              pvcsList={pvcsList}
-              pvcSLAData={pvcSLAData}
-              pvc={pvc}
-              setPVC={setPVC}
-            />
-            <UtilizationDurationDropdown />
-          </div>
```
SanjalKatiyar, line 290: only remove "PVCDropdown", we still need "UtilizationDurationDropdown" this is range query graph...

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
```diff
@@ -62,37 +54,78 @@ const getThresholdData = (data: ChartData, syncInterval: string): ChartData => {
...
+
+  React.useEffect(() => {
+    updateProtectedPVC();
+    clearSetIntervalId.current = setInterval(
+      updateProtectedPVC,
+      URL_POLL_DEFAULT_DELAY
+    );
+    return () => clearInterval(clearSetIntervalId.current);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedAppSet, protectedPVCData]);
```
SanjalKatiyar, line 122: all of this is not required... no need for timer or anything...
just keep it same as before, only diff is: earlier we were reading protectedPVC names from "pvcSLAData", now we are reading from "protectedPVCData"...
use "useMemo", if ever "protectedPVCData" will get updated we will update count on dashboard as well... no need to do it every 15s using setInterval...

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/common.tsx
```diff
@@ -50,59 +44,91 @@ const getNSAndNameFromId = (itemId: string): string[] => {
...
+  React.useEffect(() => {
+    updateSummary();
+    !!selectedAppSet && updateSummary();
+    clearSetIntervalId.current = setInterval(
+      updateSummary,
+      URL_POLL_DEFAULT_DELAY
+    );
+    return () => clearInterval(clearSetIntervalId.current);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedAppSet, protectedPVCData]);
```
SanjalKatiyar, line 86: no need for timer...
just keep it simple as before by using single "useMemo" hook, instead of two hooks ("useState" and "useEffect")...
if ever "protectedPVCData" gets updated "useMemo" will run again and data will get updated...
no need to update every 15s, there is no need for that...

packages/mco/components/mco-dashboard/queries.ts
```diff
@@ -16,16 +16,12 @@ export enum StorageDashboard {
...
-export const getPvcSlaPerClusterQuery = (clusterName: string) =>
-  `${PVC_SLA_QUERY}{cluster=~'${clusterName}'}`;
-export const getPvcSlaPerPVCQuery = (
-  clusterName: string,
-  remoteNS: string,
-  pvcName: string
+export const LAST_SYNC_TIME_QUERY = 'ramen_last_sync_timestamp_seconds';
+export const getLastSyncTimeDRPCQuery = (
+  drpcNamespace: string,
+  drpcName: string
```
SanjalKatiyar, line 22:
```suggestion
export const getLastSyncTimeDRPCQuery = (
  drpcNamespace: string,
  drpcName: string,
  clusterName: string,
```

packages/mco/components/mco-dashboard/queries.ts
```diff
@@ -16,16 +16,12 @@ export enum StorageDashboard {
...
 ) =>
-  `${PVC_SLA_QUERY}{cluster=~'${clusterName}',pvc_namespace=~'${remoteNS}',pvc_name=~'${pvcName}'}`;
+  `${LAST_SYNC_TIME_QUERY}{resource_type="drpc",name="'${drpcName}'",namespace="'${drpcNamespace}'"}`;
```
SanjalKatiyar, line 24:
```suggestion
  `${LAST_SYNC_TIME_QUERY}{resource_type='drpc',name='${drpcName}',namespace='${drpcNamespace}', cluster='${clusterName}'}`;
```

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster-app-card.tsx
```diff
@@ -40,77 +50,76 @@ import './cluster-app-card.scss';
...
         <PVCsSection
-          clusterResources={clusterResources}
+          protectedPVCData={protectedPVCData}
           clusterName={clusterName}
         />
       </GridItem>
-      <GridItem rowSpan={4} sm={12}>
-        <VolumeSummarySection pvcSLAData={pvcSLAData} />
+      <GridItem lg={12} rowSpan={8} sm={12}>
+        <VolumeSummarySection protectedPVCData={protectedPVCData} />
```
SanjalKatiyar, line 84: generally we follow 12 x 12 gird system by default... not sure if it is a good practice to break that default...
@bipuladh do u think it is okay to do so if we are using PF "Grid" component ??

SanjalKatiyar: @GowthamShanmugam did u check for diff diff screen sizes, was anything breaking or it is working as expected ??

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster-app-card.tsx
```diff
@@ -40,77 +50,76 @@ import './cluster-app-card.scss';
...
         />
       </GridItem>
-      <GridItem lg={12} rowSpan={5} sm={12}>
-        <ReplicationHistorySection
-          clusterName={clusterName}
-          selectedAppSet={selectedAppSet}
-          pvcSLAData={pvcSLAData}
-        />
+      <GridItem lg={12} rowSpan={3} sm={12}>
+        <ReplicationHistorySection selectedAppSet={selectedAppSet} />
```
SanjalKatiyar, line 122: same doubt/concern here, regarding 12 x 12 grid system...

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster-app-card.tsx
```diff
@@ -122,37 +131,64 @@ export const ClusterAppCard: React.FC = () => {
...
-    query: !!cluster ? getPvcSlaPerClusterQuery(cluster) : null,
-    basePath: ACM_ENDPOINT,
+  const [mcvs, mcvsLoaded, mcvsLoadError] = useK8sWatchResource<
+    ACMManagedClusterViewKind[]
+  >({
+    kind: referenceForModel(ACMManagedClusterViewModel),
+    isList: true,
+    namespace: cluster,
+    namespaced: true,
+    optional: true,
```
SanjalKatiyar, line 141: what does "optional" do ??

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster-app-card.tsx
```diff
@@ -122,37 +131,64 @@ export const ClusterAppCard: React.FC = () => {
...
     cluster: HUB_CLUSTER_NAME,
   });
+  const [lastSyncTimeData, lastSyncTimeError, lastSyncTimeLoading] =
+    useCustomPrometheusPoll({
+      endpoint: 'api/v1/query' as any,
+      query: LAST_SYNC_TIME_QUERY,
```
SanjalKatiyar, line 147: poll it cluster wise, also send "null" or '' (empty string) for cases when cluster is undefined...
```suggestion
     query: !!cluster ? getLastSyncPerClusterQuery(cluster) : null,
```

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster-app-card.tsx
```diff
@@ -161,25 +197,30 @@ export const ClusterAppCard: React.FC = () => {
...
           </CardHeader>
           <CardBody>
-            {!appSet.namespace && appSet.name === ALL_APPS ? (
-              <ClusterWiseCard
-                clusterName={cluster}
-                pvcSLAData={pvcSLAData}
-                csvData={csvData}
-                clusterResources={drClusterAppsMap}
-              />
+            {!!drClusterAppsMap?.[cluster]?.protectedAppSets?.length ? (
```
SanjalKatiyar, line 206: why to block cluster view when no protected apps are present ?
we can still show cluster wise card... user can see the cluster health and that there are "0" apps etc etc...
block app wise card in case no protected apps...
```suggestion
            {!!Object.keys(drClusterAppsMap)?.length ? (
```

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
```diff
@@ -176,70 +201,20 @@ export const SnapshotSection: React.FC<CommonProps> = ({ selectedAppSet }) => {
...
       useCustomPrometheusPoll({
         endpoint: 'api/v1/query_range' as any,
-        query:
-          !!remoteNS && !!pvc
-            ? getPvcSlaPerPVCQuery(clusterName, remoteNS, pvc)
-            : null,
+        query: getLastSyncTimeDRPCQuery(
+          placementInfo?.drpcNamespace,
+          placementInfo?.drpcName
+        ),
```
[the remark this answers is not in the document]
SanjalKatiyar: or this will be hub cluster wise metric with no "cluster" label ??

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster-app-card.tsx
SanjalKatiyar: we can skip the "cluster" part, if this will be hub cluster wise metric with no "cluster" label ??

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
```diff
@@ -176,70 +201,20 @@ export const SnapshotSection: React.FC<CommonProps> = ({ selectedAppSet }) => {
...
-      selectedKey={pvc}
-      selectOptions={pvcDropdownOptions}
-      onChange={setPVC}
-      placeholderText={t('Select PVCs')}
-    />
-  );
-};
-
 export const ReplicationHistorySection: React.FC<ReplicationHistorySectionProps> =
-  ({ clusterName, selectedAppSet, pvcSLAData }) => {
```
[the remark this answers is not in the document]
SanjalKatiyar: skip this if it is hub cluster wise metric, without any "cluster" label...

packages/mco/components/mco-dashboard/queries.ts
SanjalKatiyar: skip "clusterName" part, if it is hub cluster wise metric with any "cluster" label...

packages/mco/components/mco-dashboard/queries.ts
SanjalKatiyar: skip "clusterName" part, if it is hub cluster wise metric with any "cluster" label...

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster.tsx
```diff
@@ -254,20 +257,11 @@ export const PVCsSection: React.FC<PVCsSectionProps> = ({
         (item: PrometheusResult) => item?.metric?.cluster === clusterName
       )?.value[1];
```
SanjalKatiyar, line 258: Instead of filtering here based on cluster name, why don't we just pass cluster name directly to TOTAL_PVC_COUNT_QUERY while polling ??
ex: `TOTAL_PVC_COUNT_QUERY{cluster='${clusterName}'}`

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster.tsx
```diff
@@ -254,20 +257,11 @@ export const PVCsSection: React.FC<PVCsSectionProps> = ({
...
-      clusterResources[clusterName]?.protectedAppSets?.reduce(
-        (acc, protectedAppSetsMap) => {
-          const pvcsList: string[] = getProtectedPVCsFromDRPC(
-            protectedAppSetsMap?.drPlacementControl
-          );
-          return !!pvcsList?.length ? acc + pvcsList.length : acc;
-        },
-        0
-      ) || 0;
+    let protectedPVCsCount = protectedPVCData?.length || 0;
```
SanjalKatiyar, line 260: we don't need "useMemo" anymore (once we only poll total PVCs from a particular cluster, as suggested above)...
const totalPVCsCount = pvcsCount?.data?.result?.[0]?.value[1] || 0;
const protectedPVCsCount = protectedPVCData?.length || 0

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster.tsx
SanjalKatiyar: we don't need "useMemo" anymore (once we only poll total PVCs from a particular cluster, as suggested above)...
```
const totalPVCsCount = pvcsCount?.data?.result?.[0]?.value[1] || 0;
const protectedPVCsCount = protectedPVCData?.length || 0
```

packages/mco/components/mco-dashboard/data-policy/status-card/status-card.tsx
```diff
@@ -57,6 +59,30 @@ const getDRCombinedStatus = (
...
+        )
+          ? { ...acc, [cluster]: item?.value[1] !== '1' ? count : count + 1 }
+          : acc;
+      return acc;
+    }, {});
+  return Object.keys(clusterWiseHealth)?.every(
+    (cluster) => clusterWiseHealth?.[cluster] === clusterOperators.length
+  )
+    ? HealthState.OK
+    : HealthState.ERROR;
```
SanjalKatiyar, line 83: I didn't get it... why worrying about all the clusters ?? Just calculate for the selected cluster ??
We will have only one hub, we will worry about multiple hub when we have multiple, for now no need to do extra calculations...

SanjalKatiyar, line 83: why so much extra calculation ?? is there any issue with this ??
```suggestion
    const isAnyUnhealthy: boolean = !!csvManagedData?.data?.result?.find(
      (item) =>
        drClusterAppsMap.hasOwnProperty(item?.metric.cluster) &&
        !!clusterOperators.find((operator: string) =>
          item?.metric.name.startsWith(operator)
        ) &&
        item?.value[1] !== '1'
    );
    return isAnyUnhealthy ? HealthState.ERROR : HealthState.OK;
```

SanjalKatiyar: it will stop at very first unhealthy operator it will find (in any cluster)...
for your part it will always calculate to the end and then determine the final outcome...

SanjalKatiyar: if u think in future we will need names of all unhealthy clusters, we can make that change in future (it is a small change only)... why do extra calculations now, when we don;t need it...

packages/mco/components/mco-dashboard/mco-common.scss
```diff
@@ -23,6 +23,12 @@
         justify-content: center;
         width: 100%;
     }
+    &__title--size {
+        font-size: 24px;
+    }
+    &__subtitle--size {
+        font-size: 16px;
```
SanjalKatiyar, line 30: use `rem` instead of `px`...

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
GowthamShanmugam: ack

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
GowthamShanmugam: We don't need this earlier because it was the Prometheus call, it will take care of invoking this logic every 15Sec.
Now, it is from VRG CR, To calculate PVC with issues, we need to check the (current time - lastSync time), if lastsyncTime never changes, it wont report PVC with issues.

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/common.tsx
GowthamShanmugam: what if protectedPVCData never updated? lastSyncTime is reported but after it never gets changed then it is a problem. protectedPVCData is based on VRG CR reading. once lastSynctime reported we need to decide wether update the summary or not using (lastSymcTime - currentTime)

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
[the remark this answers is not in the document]
GowthamShanmugam: it is hub cluster-only metrics from the DR hub operator, which we dont need, or we can pass local-cluster as the cluster name.

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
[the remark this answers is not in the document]
GowthamShanmugam: yes, we can pass cluster name as 'local-cluster'

packages/mco/components/mco-dashboard/queries.ts
GowthamShanmugam: yes we dont need this anymore

packages/mco/components/mco-dashboard/queries.ts
GowthamShanmugam: ack

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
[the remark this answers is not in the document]
GowthamShanmugam: ack

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster-app-card.tsx
GowthamShanmugam: i did not checked with different screen, i will check

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster-app-card.tsx
GowthamShanmugam: why i am not maintaining 12, how i missed :)

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster-app-card.tsx
GowthamShanmugam: it is hub cluster metric and it has cluster label, we can still pass hub cluster name

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster-app-card.tsx
GowthamShanmugam: i am little concern about QE, they will start raise a bug like this card empty and that card is empty :), but ya i will think about this.

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster.tsx
GowthamShanmugam: good catch

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster.tsx
GowthamShanmugam: got it

packages/mco/components/mco-dashboard/data-policy/status-card/status-card.tsx
GowthamShanmugam: this logic has a huge problem since csvManagedData is a metric, Anycase if the operator is not reporting the metrics (Some problem in installation) then it will consider only the reported metrics and mark the status as healthy.

In the cluster-wise dashboard we are checking metrics for each operator by a specific name, if no metrics exist for the particular operator then there it will display as unhealthy. 

Basically, over all status card is displaying healthy, clusterwise is displaying unhealthy, it's a mismatch in logic. What i did, per cluster wise, i am looking at metrics for all the cluster operators.

packages/mco/components/mco-dashboard/mco-common.scss
GowthamShanmugam: ack

packages/mco/utils/disaster-recovery.tsx
```diff
@@ -381,3 +405,49 @@ export const getDRStatus = ({
       };
   }
 };
+
+const filterMulticlusterView = (mcvs: ACMManagedClusterViewKind[]) =>
+  mcvs?.filter(
+    (mcv) => mcv?.spec?.scope?.resource === DRVolumeReplicationGroup.kind
+  );
```
SanjalKatiyar, line 412: instead of polling all mcvs and filtering onces with vrg later...
can u try passing "fieldSelector" to "useK8sWatchResource" ??

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster-app-card.tsx
SanjalKatiyar, line 141: can u try this and check if it works ?
```suggestion
    kind: referenceForModel(ACMManagedClusterViewModel),
    isList: true,
    namespace: cluster,
    namespaced: true,
    optional: true,
    fieldSelector: spec.scope.resource = DRVolumeReplicationGroup.kind,
```

SanjalKatiyar, line 141:
```suggestion
    kind: referenceForModel(ACMManagedClusterViewModel),
    isList: true,
    namespace: cluster,
    namespaced: true,
    optional: true,
    fieldSelector: spec.scope.resource = DRVolumeReplicationGroup.kind,
```

SanjalKatiyar: no need to filter later...

packages/mco/models/ramen.ts
```diff
@@ -35,3 +35,15 @@ export const DRPlacementControlModel: K8sModel = {
...
 };
+
+export const DRVolumeReplicationGroup: K8sModel = {
+  label: 'VolumeReplicationGroup',
+  labelPlural: 'VolumeReplicationGroups',
+  apiVersion: 'v1alpha1',
+  apiGroup: 'ramendr.openshift.io',
+  plural: 'volumereplicationgroups',
+  abbr: 'VRG',
+  namespaced: false,
```
SanjalKatiyar, line 46: What I understood was that we have one VRG corresponding to every DRPC...
DRPC is a namespaced resource, yet VRG is cluster wide ??

packages/mco/utils/disaster-recovery.tsx
```diff
@@ -301,20 +310,36 @@ export const matchClusters = (
...
+
+  const unit = splittedArray?.[1] || 's';
+  return (
+    (unit === TIME_UNITS.Days && daysToMilliseconds(interval)) ||
+    (unit === TIME_UNITS.Hours && hourToMilliseconds(interval)) ||
+    (unit === TIME_UNITS.Minutes && minutesToMilliseconds(interval))
+  );
+};
+
 const THRESHOLD = 2;
```
SanjalKatiyar, line 329: we can move this to constants, earlier I added here just for reference... now we are sure that it is "2" so we can move to constants file...

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
```diff
@@ -62,37 +54,78 @@ const getThresholdData = (data: ChartData, syncInterval: string): ChartData => {
...
+  const clearSetIntervalId = React.useRef<NodeJS.Timeout>();
+  const [ProtectedPVC, setProtectedPVC] = React.useState([0, 0]);
+  const [protectedPVCsCount, pvcsWithIssueCount] = ProtectedPVC;
 
-  const [protectedPVCsCount, pvcsWithIssueCount] = React.useMemo(() => {
-    const pvcsList = getProtectedPVCsFromDRPC(
-      selectedAppSet?.drPlacementControl
-    );
-    const remoteNS = getRemoteNSFromAppSet(selectedAppSet?.application);
+  const updateProtectedPVC = () => {
```
SanjalKatiyar, line 95:
```suggestion
  const updateProtectedPVC = React.useCallback(() => {
```

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
```diff
@@ -62,37 +54,78 @@ const getThresholdData = (data: ChartData, syncInterval: string): ChartData => {
...
+          )[0] !== SLA_STATUS.HEALTHY
         )
           return acc + 1;
         else return acc;
       }, 0) || 0;
 
-    return [pvcsList?.length || 0, issueCount];
-  }, [selectedAppSet, pvcSLAData, clusterName]);
+    setProtectedPVC([protectedPVCData?.length || 0, issueCount]);
+  };
```
SanjalKatiyar, line 112:
```suggestion
  }, [selectedAppSet, protectedPVCData];
```

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
SanjalKatiyar, line 122:
```suggestion
  }, [updateProtectedPVC]);
```

SanjalKatiyar: ack... use "useCallback" then and pass the "updateProtectedPVC" to the dependency array of this "useEffect"

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/common.tsx
```diff
@@ -50,59 +44,91 @@ const getNSAndNameFromId = (itemId: string): string[] => {
...
   const { t } = useCustomTranslation();
+  const [summary, setSummary] = React.useState({
+    critical: 0,
+    warning: 0,
+    healthy: 0,
+  });
+  const clearSetIntervalId = React.useRef<NodeJS.Timeout>();
 
-  const { critical, warning, healthy } = React.useMemo(() => {
+  const updateSummary = () => {
```
SanjalKatiyar, line 58:
```suggestion
  const updateSummary = React.useCallback(() => {
```

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/common.tsx
SanjalKatiyar, line 86:
```suggestion
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [selectedAppSet, protectedPVCData]);
```
```suggestion
  }, [updateSummary]);
```

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
```diff
@@ -62,37 +54,78 @@ const getThresholdData = (data: ChartData, syncInterval: string): ChartData => {
...
+
 export const ProtectedPVCsSection: React.FC<ProtectedPVCsSectionProps> = ({
-  clusterName,
-  pvcSLAData,
+  protectedPVCData,
   selectedAppSet,
 }) => {
   const { t } = useCustomTranslation();
+  const clearSetIntervalId = React.useRef<NodeJS.Timeout>();
+  const [ProtectedPVC, setProtectedPVC] = React.useState([0, 0]);
```
SanjalKatiyar, line 92: nits:
```suggestion
  const [protectedPVC, setProtectedPVC] = React.useState([0, 0]);
```

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/common.tsx
SanjalKatiyar: ack... let use "useCallback" here...

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster-app-card.tsx
SanjalKatiyar: if no clusters in cluster dropdown >>  we show "No data available" at cluster card level...
if there are clusters, but no applications >> we still show cluster card, we only show "No data available" at app card level...

cluster card will not have missing info... just there are no protected apps, that is why they see count as "0" everywhere... we need to test this once...

packages/mco/components/mco-dashboard/data-policy/status-card/status-card.tsx
SanjalKatiyar: make sense !!

packages/mco/utils/disaster-recovery.tsx
```diff
@@ -381,3 +405,49 @@ export const getDRStatus = ({
...
+  );
+
+export const getProtectedPVCFromVRG = (
+  mcvs: ACMManagedClusterViewKind[]
+): ProtectedPVCData[] => {
+  const filteredMCVs = filterMulticlusterView(mcvs);
+  return filteredMCVs?.reduce((acc, mcv) => {
+    const drpcName =
+      mcv?.metadata?.annotations?.[
+        'drplacementcontrol.ramendr.openshift.io/drpc-name'
```
SanjalKatiyar, line 421: nits: constants file...

packages/mco/utils/disaster-recovery.tsx
```diff
@@ -381,3 +405,49 @@ export const getDRStatus = ({
...
+      ];
+    const drpcNamespace =
+      mcv?.metadata?.annotations?.[
+        'drplacementcontrol.ramendr.openshift.io/drpc-namespace'
```
SanjalKatiyar, line 425: nits: constants file...

packages/mco/utils/disaster-recovery.tsx
```diff
@@ -301,20 +310,36 @@ export const matchClusters = (
...
 
+const getScheduledSynInterval = (scheduledSyncInterval: string) => {
+  const regex = new RegExp(
+    `([0-9]+)|([${TIME_UNITS.Days}|${TIME_UNITS.Hours}|${TIME_UNITS.Minutes}]+)`,
+    'g'
+  );
+  const splittedArray = scheduledSyncInterval?.match(regex);
+  const interval = Number(splittedArray?.[0] || 0);
+
+  const unit = splittedArray?.[1] || 's';
```
SanjalKatiyar, line 321: nit: we are supporting m/d/h... so default should be one of those...
```suggestion
  const unit = splittedArray?.[1] || 'm';
```

packages/shared/details-page/datetime.ts
```diff
@@ -234,3 +234,18 @@ export const twentyFourHourTime = (
...
+
+export const hourToMilliseconds = (hours: number) => {
+  //            min  sec  ms
+  return hours * 60 * 60 * 1000;
+};
+
+export const minutesToMilliseconds = (seconds: number) => {
+  //              sec  ms
+  return seconds * 60 * 1000;
+};
```
SanjalKatiyar, line 251:
```suggestion
export const minutesToMilliseconds = (minutes: number) => {
  //                           sec  ms
  return minutes * 60 * 1000;
};
```

SanjalKatiyar, line 251:
```suggestion
export const minutesToMilliseconds = (minutes: number) => {
  //              sec  ms
  return minutes * 60 * 1000;
};
```

packages/shared/details-page/datetime.ts
```diff
@@ -234,3 +234,18 @@ export const twentyFourHourTime = (
...
   dates.reduce((a, b) => (new Date(a) > new Date(b) ? a : b));
+
+export const daysToMilliseconds = (days: number) => {
+  //          hour  min  sec  ms
+  return days * 24 * 60 * 60 * 1000;
+};
+
+export const hourToMilliseconds = (hours: number) => {
+  //            min  sec  ms
+  return hours * 60 * 60 * 1000;
```
SanjalKatiyar, line 245:
```suggestion
  return minutesToMilliseconds(hours * 60);
```

packages/shared/details-page/datetime.ts
```diff
@@ -234,3 +234,18 @@ export const twentyFourHourTime = (
 
 export const getLatestDate = (dates: string[]) =>
   dates.reduce((a, b) => (new Date(a) > new Date(b) ? a : b));
+
+export const daysToMilliseconds = (days: number) => {
+  //          hour  min  sec  ms
+  return days * 24 * 60 * 60 * 1000;
```
SanjalKatiyar, line 240:
```suggestion
  return hourToMilliseconds(days * 24);
```

SanjalKatiyar: not really needed, up to you if u want to make this change or not...

packages/shared/details-page/datetime.ts
SanjalKatiyar: not really needed, up to you if u want to make this change or not...

packages/shared/popup/status-popup.tsx
```diff
@@ -20,9 +20,14 @@ export const Status: React.FC<StatusProps> = ({ value, icon, children }) => (
...
-        <div className="text-secondary" data-test="operator-status-icon">
-          {value}
+        <div>
+          <span data-test="operator-status-value">{icon}</span>
+          <span
+            className="text-secondary odf-status-popup__text"
+            data-test="operator-status-icon"
+          >
+            {value}
+          </span>
```
SanjalKatiyar, line 30: this is a shared component, are u sure this change is not breaking the UI at any other place where this component is used ??

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster-app-card.tsx
GowthamShanmugam: I am not breaking the 12*12 rule, it is same, just adjusted values as per UX mockups

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster-app-card.tsx
GowthamShanmugam: The actual usage is, the loaded should be true, even though it raises any  loadError occurs

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster-app-card.tsx
GowthamShanmugam: i checked it is working

packages/mco/utils/disaster-recovery.tsx
GowthamShanmugam: FieldSelector can be used for metadata.name, metadata.namespace, and status.phase: https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/
Also, it dont have labels to filter.

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster-app-card.tsx
GowthamShanmugam: it wont fieldSelector won't work for all fields: https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/

GowthamShanmugam: when i tried i got error, like unsupported field selector-error

packages/mco/models/ramen.ts
GowthamShanmugam: it is namespaced, my mistake

packages/mco/utils/disaster-recovery.tsx
GowthamShanmugam: ack

packages/shared/popup/status-popup.tsx
GowthamShanmugam: those are fine, this is the only place is using value, other places are using icon, which is executed in else part

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
```diff
@@ -8,91 +8,173 @@ import {
...
+        data: {
+          stroke: thresholdColor.value,
+          strokeDasharray: '3,3',
+          fillOpacity: 0,
+        },
+      });
+    }
+  }
+  return { data, chartStyle };
+};
```
SanjalKatiyar, line 105: let's not copy everything... almost entire code is same...
update "getRangeVectorStats" in shared to take optional argument "threashold"...
use "mapLimitsRequests" from shared for "utilizationData", add an optional argument for "description" (if passed we will use passed value else default what it is setting right now)...
only thing new we add to this file is "scheduled SLA" related threshold part, outside of "mapLimitsRequests"...

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
```diff
@@ -176,101 +250,35 @@ export const SnapshotSection: React.FC<CommonProps> = ({ selectedAppSet }) => {
...
+  ({ selectedAppSet }) => {
     const { t } = useCustomTranslation();
     const { duration } = useUtilizationDuration();
-    const [pvc, setPVC] = React.useState<string>();
 
-    const remoteNS = getRemoteNSFromAppSet(selectedAppSet?.application);
-    const pvcsList: string[] = getProtectedPVCsFromDRPC(
-      selectedAppSet?.drPlacementControl
+    const placementInfo = selectedAppSet?.placementInfo?.[0];
+    const [thresholdInSeconds, unit] = convertSyncIntervalToSeconds(
```
SanjalKatiyar, line 259: nits:
```suggestion
    const [thresholdInSeconds, initialUnit] = convertSyncIntervalToSeconds(
```

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
```diff
@@ -176,101 +250,35 @@ export const SnapshotSection: React.FC<CommonProps> = ({ selectedAppSet }) => {
...
-     * ToDo(Sanjal): Find a way to add more info to tooltip of the chart
-     * Also, update utils based upon unit of the data (sec, min etc)
-     */
     const { data, chartStyle } = mapLimitsRequests(
-      pvcSLARangeData,
-      null,
-      null,
-      trimSecondsXMutator,
+      pvcsSLARangeData,
+      thresholdInSeconds,
```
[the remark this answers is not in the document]
GowthamShanmugam: thresholdInSeconds to convert the actual metrics value into date, humanize to display value in chart

packages/shared/charts/utils.ts
```diff
@@ -141,5 +160,31 @@ export const mapLimitsRequests = (
     }
   }
 
+  if (threshold) {
```
GowthamShanmugam, line 163: right now it is used only MCO

PR #650, opened 2023-02-20
Implemented RHF and Yup schema validations on forms

ivoalmeida:
image
image
image
image
image
image


packages/odf/components/attach-obc/AttachOBC.tsx
```diff
@@ -42,6 +44,15 @@ const AttachStorage: React.FC<AttachStorageProps> = (props) => {
...
 
+  const { control } = useForm();
+
+  const fieldRequirements = [
+    fieldRequirementsTranslations.maxChars(t, 253),
+    fieldRequirementsTranslations.startAndEndName(t),
+    fieldRequirementsTranslations.alphaNumericPeriodAdnHyphen(t),
+    fieldRequirementsTranslations.cannotBeUsedBefore(t),
+  ];
+
```
SanjalKatiyar, line 55: can't we have "CreateOBCForm" handle this, rather than initialising here and passing it down as props ??

SanjalKatiyar: also, I did not get it... when we are opening "Create OBC" page from the list page, we are using validation "schema" and everything...
but when we are opening same from "AttachOBC" page... we are not using the "schema" ??

SanjalKatiyar: shouldn't the behaviour be same since it is same form ??
since inputs are same, let "CreateOBCForm" handle "schema", "fieldRequirements", "useForm" etc...

SanjalKatiyar: maybe I am missing anything here...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -225,7 +358,19 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
           (value: string) => !existingNames.includes(value)
         )
         .notRequired()
-        .transform((value: string) => (!!value ? value : undefined)),
+        .transform((value: string) =>
+          value &&
+          Object.values(value).some(
+            (v) => !(v === null || v === undefined || v === '')
+          )
```
SanjalKatiyar, line 365:
```suggestion
          !!value
```

packages/odf/components/create-bs/create-bs.tsx
```diff
@@ -288,6 +303,13 @@ const CreateBackingStoreForm: React.FC<CreateBackingStoreFormProps> = (
       {provider === BC_PROVIDERS.PVC && (
         <PVCType state={providerDataState} dispatch={providerDataDispatch} />
       )}
+      {!isValid && isSubmitted ? (
+        <Alert
+          variant="danger"
+          isInline
+          title={t('Address form errors to proceed')}
+        />
+      ) : null}
```
SanjalKatiyar, line 312: nits:
```suggestion
      {!isValid && isSubmitted && (
        <Alert
          variant="danger"
          isInline
          title={t('Address form errors to proceed')}
        />
      )}
```

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -441,12 +468,21 @@ export const CreateOBCPage: React.FC<CreateOBCPageProps> = (props) => {
...
+          control={control}
+          fieldRequirements={fieldRequirements}
         />
+        {!isValid && isSubmitted ? (
+          <Alert
+            variant="danger"
+            isInline
+            title={t('Address form errors to proceed')}
+          />
+        ) : null}
```
SanjalKatiyar, line 485: nits:
```suggestion
        {!isValid && isSubmitted && (
          <Alert
            variant="danger"
            isInline
            title={t('Address form errors to proceed')}
          />
        )}
```

packages/odf/components/create-storage-system/create-storage-system-steps/create-storage-class-step/create-storage-class-step.tsx
```diff
@@ -72,26 +76,28 @@ export const CreateStorageClass: React.FC<CreateStorageClassProps> = ({
...
+      'endpoint-input': Yup.string()
+        .required()
+        .test(
+          'ip-address',
+          t('The endpoint is not a valid IP address'),
+          (value: string) => isValidIP(value)
+        ),
+      'username-input': Yup.string().required(),
+      'password-input': Yup.string().required(),
+      'poolname-input': Yup.string().required(),
```
SanjalKatiyar, line 88: not blocking the PR for now... but this should come from an extension, just like `Component, displayName`...
this is vendor specific form and each vendor can have different input fields...
cc @bipuladh

packages/shared/form-group-controller/FormGroupController.tsx
```diff
@@ -0,0 +1,48 @@
...
+export type FormGroupControllerRenderProps = {
+  value: any;
+  validated: FormGroupProps['validated'];
+  onChange: (...event: any[]) => void;
+  onBlur: () => void;
+};
+
+export type FormGroupControllerProps = UseControllerProps & {
+  formGroupProps: FormGroupProps;
+  render: (props: FormGroupControllerRenderProps) => React.ReactNode;
```
SanjalKatiyar, line 14: can we add types for "controllerProps" as well ??
"control" and "name" fields are required IMO, so good to add here so that no one misses it...

packages/odf/components/attach-obc/AttachOBC.tsx
ivoalmeida: I needed fieldRequirements in the component that uses displays the FRs and also with the schema. I came up with a better way of sharing this via a hook since I also needed to implement a schema inside AttachOBC as well and it uses the same main field which obc name.

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
ivoalmeida: the merge from previous PR didn't catch this. fixed

packages/odf/components/create-bs/create-bs.tsx
ivoalmeida: done

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
ivoalmeida: done

packages/odf/components/namespace-store/namespace-store-form.tsx
```diff
@@ -315,48 +326,67 @@ const NamespaceStoreForm: React.FC<NamespaceStoreFormProps> = (props) => {
...
+          />
         </>
       )}
+      {!isValid && isSubmitted ? (
+        <Alert
+          variant="danger"
+          isInline
+          title={t('Address form errors to proceed')}
+        />
+      ) : null}
```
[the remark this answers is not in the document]
ivoalmeida: done

packages/shared/yup-validation-resolver/useYupValidationResolver.ts
```diff
@@ -29,7 +38,7 @@ const useYupValidationResolver = <T>(validationSchema: AnySchema): Resolver =>
                 ...allErrors,
                 [currentError.path]: {
                   type: currentError.type ?? 'validation',
-                  message: currentError.message,
+                  message: t(currentError.message),
```
SanjalKatiyar, line 41: this will not work like this... if u need translations make sure "currentError.message" is itself translated at the time when we are initialising "currentError" object with the error messages...
```suggestion
                  message: currentError.message,
```

packages/shared/form-group-controller/FormGroupController.tsx
ivoalmeida: Take another look. The types are there already ```UseControllerProps```. If you don't provide a control prop for example, it will error out.

packages/shared/yup-validation-resolver/useYupValidationResolver.ts
```diff
@@ -10,8 +11,16 @@ export type ValidationErrorMessages = {
   messages?: Record<string, ValidationErrorMessage>;
 };
 
-const useYupValidationResolver = <T>(validationSchema: AnySchema): Resolver =>
-  useCallback(
+setLocale({
+  mixed: {
+    required: 'Required',
+  },
+});
```
SanjalKatiyar, line 18: it is not used anywhere right ?
```suggestion
```

SanjalKatiyar, line 18: I don;t see any use of this as well... we maintain locales in "plugin__odf-console.jspn" under "locales" directory...

packages/shared/yup-validation-resolver/useYupValidationResolver.ts
SanjalKatiyar: which mu assumption is we are already doing `fieldRequirementsTranslations.maxChars(t, 43),` when we are passing `t` functions to the error messages, right ??

packages/shared/form-group-controller/FormGroupController.tsx
SanjalKatiyar: yeah... I missed the `&` part !!

packages/shared/yup-validation-resolver/useYupValidationResolver.ts
ivoalmeida: When a required error is raised the ```currentError.message``` value will be "Required" which is a key already set in ```locales/en/plugin__odf-console.json```.

This is implemented like to avoid repeating ```t('Required')``` multiple times on every single form

packages/shared/yup-validation-resolver/useYupValidationResolver.ts
```diff
@@ -43,7 +54,8 @@ const useYupValidationResolver = <T>(validationSchema: AnySchema): Resolver =>
             {}
           ),
         })),
-    [validationSchema]
+    [t, validationSchema]
```
SanjalKatiyar, line 57:
```suggestion
    [validationSchema]
```

SanjalKatiyar:
ci/prow/odf-console-e2e-aws >> should pass once https://github.com/openshift/release/pull/36731#issuecomment-1445980467 is merged...
ci/prow/images >> I am seeing this error in other PRs as well... seems like CI outage or something...

ivoalmeida:
@SanjalKatiyar 
I fixed the issue with backing class. Basically the schema was not changing after the "Switch to Credentials" button was clicked, which caused the secret input to not be rendered. I moved the state and setState that toggles showSecret to the components that consume S3EndpointType form, that way I can update the schema accordingly.

Also, I made a change to one of the BucketClass name field requirements based on a previous comment you made which was then raised with Chandan:
https://docs.google.com/document/d/1yWm6NSHKjFslC09NYw82G7_hFvyTTjR9nEeT88QG2E8/edit?pli=1&disco=AAAAo6VI-eA

packages/odf/components/create-bs/create-bs.tsx
```diff
@@ -278,6 +293,9 @@ const CreateBackingStoreForm: React.FC<CreateBackingStoreFormProps> = (
         provider === BC_PROVIDERS.IBM ||
         provider === BC_PROVIDERS.AZURE) && (
         <S3EndPointType
+          showSecret={showSecret}
+          setShowSecret={setShowSecret}
+          control={control}
```
SanjalKatiyar, line 298: this `create-bs.tsx` component also uses `GCPEndpointType` which maintains state for secret toggle as well... won;t this issue (schema not re-rendering) happen for that as well ??

SanjalKatiyar: can u check that once ? 
we don;t have e2e test coverage for that, so we won't notice that in CI.

ivoalmeida:
@SanjalKatiyar 
More changes to copy and removal of seemingly redundant helper texts according to Chandan's directions.
Please see the bottom/most up to date comment here: https://docs.google.com/document/d/1yWm6NSHKjFslC09NYw82G7_hFvyTTjR9nEeT88QG2E8/edit?pli=1&disco=AAAAo6VI-d0

SanjalKatiyar:
> @SanjalKatiyar More changes to copy and removal of seemingly redundant helper texts according to Chandan's directions. Please see the bottom/most up to date comment here: https://docs.google.com/document/d/1yWm6NSHKjFslC09NYw82G7_hFvyTTjR9nEeT88QG2E8/edit?pli=1&disco=AAAAo6VI-d0

ack !
@ivoalmeida PTAL: https://github.com/red-hat-storage/odf-console/pull/650#discussion_r1119883920 whether this is working as intended or not...
also plz run `yarn i18n` once...

let me know once done and I will tag the PR... we should be merging the PR by EOD...

ivoalmeida:
> 

I've added the following response to

ivoalmeida:
> > @SanjalKatiyar More changes to copy and removal of seemingly redundant helper texts according to Chandan's directions. Please see the bottom/most up to date comment here: https://docs.google.com/document/d/1yWm6NSHKjFslC09NYw82G7_hFvyTTjR9nEeT88QG2E8/edit?pli=1&disco=AAAAo6VI-d0
> 
> ack ! @ivoalmeida PTAL: [#650 (comment)](https://github.com/red-hat-storage/odf-console/pull/650#discussion_r1119883920) whether this is working as intended or not... also plz run `yarn i18n` once...
> 
> let me know once done and I will tag the PR... we should be merging the PR by EOD...

I've added the following response to that comment but the comment is still pending (not sure why):
_I checked manually. It works fine with either option. Also, the toggle won't matter in this case both inputs are tracked inside the same FormGroupController._

SanjalKatiyar:
> I've added the following response to that comment but the comment is still pending (not sure why): _I checked manually. It works fine with either option. Also, the toggle won't matter in this case both inputs are tracked inside the same FormGroupController._

I am not sure if I get it, as both "GCPEndpointType"  and "S3EndPointType" had its own toggle state...
we lifted toggle state for "S3EndPointType" up, but "GCPEndpointType" is still behaving like before...
but I am tagging the PR for now as u already testing it out and as said that it is working...

ivoalmeida:
> > I've added the following response to that comment but the comment is still pending (not sure why): _I checked manually. It works fine with either option. Also, the toggle won't matter in this case both inputs are tracked inside the same FormGroupController._
> 
> I am not sure if I get it (both these FCs are not wrapped by any FormGroupController), earlier both "GCPEndpointType" and "S3EndPointType" had its own toggle state... u lifted toggle state for "S3EndPointType" up to fix the CI issue, but "GCPEndpointType" is still behaving like before... but I am tagging the PR for now as u already tested it out and as u said that it is working...

On "GCPEndpointType" the showSecret toggle doesn't matter because either input is rendered and controlled by the same FormGroupController that's why I don't need to lift state for this component.

On S3EndPointType it is different because the switch renders two totally different inputs and both of those are not required. I tested creating a new backing stored with "access-key" and "secret-key" empty and it did with no backend validation errors

agarwal-mudit:
Merging manually because of flaky CI, Ivo ran these tests locally and it passed there.

PR #651, opened 2023-02-21
add warning for no StorageClass (wizard)

SanjalKatiyar:
Part of: https://issues.redhat.com/browse/RHSTOR-4367
Screenshot from 2023-02-21 13-08-14


packages/shared/dropdown/ResourceDropdown.tsx
```diff
@@ -212,41 +217,65 @@ const ResourceDropdown: ResourceDropdown = <T extends unknown>({
...
+      {!!inlineError && loadError && (
+        <Alert
+          variant={AlertVariant.danger}
+          isInline
+          isPlain
+          className="odf-resourceDropdown__inline-message"
+          title={inlineError}
+        />
+      )}
+    </>
```
bipuladh, line 278: Why can't this be part of the installation flow? I am not sure this is required anywhere else. Let's keep our components light.

SanjalKatiyar: installation does not know whether there are any resources available or not...
that state is maintained by dropdown itself...

SanjalKatiyar: only thing installation knows is whether any resource is selected from the dropdown or not... if no resources are available in the cluster, installation won;t know about that...

SanjalKatiyar: unless we either lift the state up or pass callback to dropdown to update whether resources are present in cluster or not...
but both these changes again require tweaking this generic component

SanjalKatiyar: let me know which u prefer and I will update PR accordingly...

SanjalKatiyar: installation does not know whether there are any resources (StorageClass here) available or not... that state is maintained by dropdown itself...
only thing installation knows is whether any resource (if present) is selected from the dropdown or not... if no resources are available in the cluster, installation won;t know about that...

I asked UX whether we can skip this inline message at all, but they suggested that it is good to have this inline message...

unless we either lift the state up or pass callback to dropdown to update whether polled resource is empty or not (no resource present in the cluster at all)...
but both these changes again require tweaking this generic component

SanjalKatiyar: there is however one more way, just occurred to me... I can pass these inline messages as a render prop, still I will have to tweak this generic component here as well... but changes will be less...

wdyt ?? should we keep it as it is (current state of the PR) or render props way to go ??

SanjalKatiyar: @bipuladh updated the PR... we still need changes in generic FC, but not those changes are minimal !!

packages/shared/dropdown/ResourceDropdown.tsx
```diff
@@ -212,41 +213,48 @@ const ResourceDropdown: ResourceDropdown = <T extends unknown>({
...
+            onToggle={loaded && !loadError ? onToggle : () => null}
+            toggleIndicator={CaretDownIcon}
+            data-test={dataTest}
+          >
+            {!loaded && <LoadingInline />}
+            {loaded &&
+              !loadError &&
+              (_.isEmpty(resources) ? (
+                <span>{t('No resources available')}</span>
+              ) : (
```
SanjalKatiyar, line 231: new addition to the FC...

packages/shared/dropdown/ResourceDropdown.tsx
```diff
@@ -212,41 +213,48 @@ const ResourceDropdown: ResourceDropdown = <T extends unknown>({
...
+        <TextInput
+          id="search-bar"
+          iconVariant="search"
+          value={searchText}
+          onChange={(val) => setSearchText(val)}
+          autoComplete="off"
+        />
+        {dropdownItems}
+      </Dropdown>
+      {children?.(loaded, loadError, resources)}
```
SanjalKatiyar, line 256: new addition to the FC...

PR #653, opened 2023-02-22
update alert card for mco dashboard

SanjalKatiyar:
Screenshot from 2023-02-22 16-35-20
Screenshot from 2023-02-22 16-35-32
Screenshot from 2023-02-22 16-42-18


packages/shared/alert/AlertsPanel.tsx
```diff
@@ -162,47 +162,45 @@ const AlertsPanel: React.FC<AlertsProps> = ({
       <Title headingLevel="h3">
         {t('Alerts')} ({filteredAlerts.length})
       </Title>
-      {filteredAlerts.length > 0 && (
```
SanjalKatiyar, line 165: @alfonsomthd very minor change...
currently accordions are only visible when there are any alerts else on UI it just display plain text: `Alert (0)`...
remove this minor condition so that accordions are always visible, if no alert count will be zero (check the attached screen shots to this PR)...
I hope it won't affect ur sidebar... it is good to have same functionality for sidebar as well...

packages/mco/utils/disaster-recovery.tsx
```diff
@@ -479,3 +480,6 @@ export const filterPVCDataUsingAppsets = (
         return result;
       })
   );
+
+export const filterDRAlerts = (alert: Alert) =>
+  alert?.annotations?.alert_type === 'DisasterRecovery';
```
GowthamShanmugam, line 485: is ODF using annotation or label?

SanjalKatiyar: annotation...

PR #656, opened 2023-02-23
Bug 2178262: Fix for toggle selection issue in apply policy modal

GowthamShanmugam:
Problem:
  ExpandView is a child component for DualListSelectorListItem, the selection count got increased and decreased even when we click the arrow icon.  Sometimes it confuses the user whether the app got selected or unselected. 
  
Fix:
  The app selection won't work when the user clicks the arrow button, it simply expands the view. 
  For more: https://github.com/patternfly/patternfly-react/blob/v5/packages/react-core/src/components/DualListSelector/DualListSelectorListItem.tsx#L55

### DIFF
--- a/packages/mco/components/modals/drpolicy-apps-apply/apply-policy-dual-list-selector.tsx
+++ b/packages/mco/components/modals/drpolicy-apps-apply/apply-policy-dual-list-selector.tsx
@@ -2,6 +2,8 @@ import * as React from 'react';
 import { CustomDualListSelector } from '@odf/shared/generic/custom-dual-list-selector';
 import { SelectorInput } from '@odf/shared/modals/Selector';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
+import AngleRightIcon from '@patternfly/react-icons/dist/esm/icons/angle-right-icon';
+import classNames from 'classnames';
 import {
   DualListSelectorPane,
   DualListSelectorList,
@@ -11,7 +13,6 @@ import {
   EmptyStateBody,
   Title,
   Label,
-  ExpandableSection,
   FormGroup,
   Form,
 } from '@patternfly/react-core';
@@ -54,6 +55,14 @@ type DualListSelectorChosenListItemProps = {
   drpcPvcLabels: PlacementToDrpcMap;
   setDrpcPvcLabels: (options: PlacementToDrpcMap) => void;
 } & Pick<DualListSelectorAvailablePaneProps, 'onOptionSelect'>;
+
+type DualListSelectorExpandListItemProps = {
+  id: string;
+  appName: string;
+  appNamespace: string;
+  isAlreadyProtected: boolean;
+  children: React.ReactNode;
+};
 
 const defaultLabelsState = {
   drpcName: undefined,
@@ -102,9 +111,11 @@ const DualListSelectorAvailablePane: React.FC<DualListSelectorAvailablePaneProps
                 onOptionSelect={(e) => onOptionSelect(e, optionIndex, false)}
                 translate={false}
               >
-                <div className="mco-apply-policy-dual-selector__listItem">
-                  <div>
-                    {option.appSetName}
+                <div className="mco-apply-policy-dual-selector__listRow">
+                  <div className="mco-apply-policy-dual-selector__listItem">
+                    <span className="mco-apply-policy-dual-selector__listItem--text">
+                      {option.appSetName}
+                    </span>
                     {option.isAlreadyProtected && (
                       <span>
                         <Label isCompact color="green">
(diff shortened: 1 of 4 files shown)


GowthamShanmugam:
ezgif com-video-to-gif

GowthamShanmugam:
After fix:
[Screencast from 2023-02-23 17-51-03.webm](https://user-images.githubusercontent.com/15073590/220904717-acc051b1-87c7-428a-932b-fb2114672236.webm)

package.json
```diff
@@ -56,6 +56,7 @@
     "@patternfly/react-charts": "^6.15.3",
     "@patternfly/react-core": "4.267.5",
     "@patternfly/react-icons": "^4.11.0",
+    "@patternfly/react-styles": "^4.92.6",
```
SanjalKatiyar, line 59: it did not make any changes to "yarn.lock" ??

SanjalKatiyar:
what was the issue with "ExpandableSection" imported from PF ??
I did not get it why we removed that and added a new dependency to make <div> look like "ExpandableSection" FC...

packages/mco/components/modals/drpolicy-apps-apply/apply-policy-dual-list-selector.tsx
```diff
@@ -166,42 +164,74 @@ const DualListSelectorChosenListItem: React.FC<DualListSelectorChosenListItemPro
...
-            onChange={(e) => e.stopPropagation()}
-            onClick={(e) => e.stopPropagation()}
+          <div className={css(styles.expandableSectionToggle)}>
+            <button
+              type="button"
+              className="mco-apply-policy-dual-selector__button"
+              aria-expanded={isExpanded}
+              onClick={(e) => {
+                setIsExpanded(!isExpanded);
+                e.stopPropagation();
```
SanjalKatiyar, line 181: we don't need this right ??

SanjalKatiyar, line 181: we don't need this right ??
```suggestion
```

packages/mco/components/modals/drpolicy-apps-apply/apply-policy-dual-list-selector.tsx
```diff
@@ -166,42 +164,74 @@ const DualListSelectorChosenListItem: React.FC<DualListSelectorChosenListItemPro
...
+              }}
+              onKeyDown={(e: React.KeyboardEvent) => {
+                if (e.key === ' ' || e.key === 'Enter') {
```
SanjalKatiyar, line 184: nits:
```suggestion
                if (e.code === 'Space' || e.code === 'Enter') {
```

packages/mco/components/modals/drpolicy-apps-apply/apply-policy-dual-list-selector.tsx
```diff
@@ -166,42 +164,74 @@ const DualListSelectorChosenListItem: React.FC<DualListSelectorChosenListItemPro
...
+                  (document.activeElement as HTMLElement).click();
```
SanjalKatiyar, line 185: why this ?? are we sure "document.activeElement" will point to same element always, in every case ??
can't we use "setIsExpanded" if we only want to expand the expandable ??

packages/shared/modals/Selector.tsx
```diff
@@ -293,7 +293,11 @@ export class SelectorInput extends React.Component<
             removeKeys={removeKeys}
             inputProps={inputProps}
             renderTag={renderTag}
-            onChange={this.handleChange.bind(this)}
+            onChange={
+              !this.props.disabled
+                ? this.handleChange.bind(this)
+                : () => undefined
+            }
```
SanjalKatiyar, line 300: just a doubt...
setting `disabled={this.props.disabled}` is not enough ??

GowthamShanmugam: Even after setting ```disabled={this.props.disabled}``` the label input box blocks new label input but still allowing to delete the already existing label, To avoid this, i am blocking the onChange.

packages/mco/components/modals/drpolicy-apps-apply/apply-policy-dual-list-selector.tsx
GowthamShanmugam: No issues, we call  setIsExpanded directly also

packages/mco/components/modals/drpolicy-apps-apply/apply-policy-dual-list-selector.tsx
GowthamShanmugam: there is nothing called code in KeyboardEvent type

packages/mco/components/modals/drpolicy-apps-apply/apply-policy-dual-list-selector.tsx
GowthamShanmugam: this is the one that prevents the dual list selection, it will expand and break the onClick call. so when the user clicks expand button, it wont consider as one dual list item selection.

package.json
GowthamShanmugam: do we need to add one more entry in yarn.lock?

SanjalKatiyar: no no `yarn install` should have did it, but I guess there's no internal dependency changes if that does not happen...

packages/mco/components/modals/drpolicy-apps-apply/apply-policy-dual-list-selector.tsx
SanjalKatiyar: I can see it here: https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event#event_properties and saw its usage in console as well !!

GowthamShanmugam:
Screenshot from 2023-02-28 19-24-08
Screenshot from 2023-02-28 19-23-58

packages/mco/components/modals/drpolicy-apps-apply/apply-policy-dual-list-selector.tsx
GowthamShanmugam:
```
 interface KeyboardEvent<T = Element> extends SyntheticEvent<T, NativeKeyboardEvent> {
        altKey: boolean;
        charCode: number;
        ctrlKey: boolean;
        /**
         * See [DOM Level 3 Events spec](https://www.w3.org/TR/uievents-key/#keys-modifier). for a list of valid (case-sensitive) arguments to this method.
         */
        getModifierState(key: string): boolean;
        /**
         * See the [DOM Level 3 Events spec](https://www.w3.org/TR/uievents-key/#named-key-attribute-values). for possible values
         */
        key: string;
        keyCode: number;
        locale: string;
        location: number;
        metaKey: boolean;
        repeat: boolean;
        shiftKey: boolean;
        which: number;
    }
```
i dont see any type called code, it throws error

SanjalKatiyar: ack !!

GowthamShanmugam:
Before Fix:
[Screencast from 2023-03-01 14-19-06.webm](https://user-images.githubusercontent.com/15073590/222089645-1718071e-8c1c-41b4-8d1d-6167bade92c4.webm)

PR #658, opened 2023-02-23
Adds support for syncing deletion and bug fixes

vbnrh:
https://www.sketch.com/s/a5c851b3-87c6-4ce6-8eea-28d47aa4fa4c/a/7ywrDop

Screenshot from 2023-02-27 16-34-38


packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -121,17 +135,31 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
           namespaceStore: rule.namespaceStore,
           prefix: rule.prefix,
+          syncDeletion: rule.syncDeletion,
         });
     });
     dispatch({ type: 'setReplicationRuleFormData', data: rulesFormData });
+    if (
+      !rules.some((rule) => rule.syncDeletion) ||
+      logReplicationInfo?.logLocation ||
+      logReplicationInfo?.logPrefix
```
SanjalKatiyar, line 156:
```suggestion
    if (!_.isEmpty(logReplicationInfo)
```

SanjalKatiyar: also, why do we need `!rules.some((rule) => rule.syncDeletion)` condition ??
if `logReplicationInfo?.logLocation` and `logReplicationInfo?.logPrefix` both are `undefined` why we do want to dispatch if `syncDeletion` is `true` ??

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -121,17 +135,31 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+    ) {
```
SanjalKatiyar, line 157:
```suggestion
    if (
      !_.isEmpty(logReplicationInfo)
    ) {
```

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -364,11 +393,15 @@ export const CreateOBCPage: React.FC<CreateOBCPageProps> = (props) => {
...
-
+  const submitBtnId = 'obc-submit-btn';
   const history = useHistory();
 
-  const save = (e: React.FormEvent<EventTarget>) => {
+  const save = (e: React.SyntheticEvent<HTMLFormElement, SubmitEvent>) => {
     e.preventDefault();
+
+    // I don't really want to do this but the namespace store modal's submit triggers this form also.
+    if (e.nativeEvent.submitter.id !== submitBtnId) return;
```
SanjalKatiyar, line 403: try `stopPropagation()` method...

SanjalKatiyar: ref: https://stackoverflow.com/questions/52163574/how-to-stop-event-capturing-in-javascript

packages/odf/components/mcg/log-replication-info-form.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export const LogReplicationInfoForm: React.FC<LogReplicationInfoFormProps> = ({
+  location,
+  prefix,
+  onPrefixChange,
+  onLogLocationChange,
+}) => {
+  const { t } = useCustomTranslation();
+  const [logsLocation, setLogsLocation] = React.useState(location ?? '');
+  const [logPrefix, setPrefix] = React.useState(prefix ?? '');
```
SanjalKatiyar, line 20: let's not maintain the state twice... this FC is not something which could be re-usable at multiple places... what I understood is that, FC which is calling "LogReplicationInfoForm" is already maintaining the state... u can pass that as a prop, instead of maintain state locally once again...

packages/odf/components/mcg/replication-policy-form.tsx
```diff
@@ -50,6 +57,13 @@ export const ReplicationPolicyForm: React.FC<ReplicationFormProps> = ({
   const [rules, setRules] = React.useState<Rule[]>([
     { id: 1, namespaceStore: '', prefix: '' },
   ]);
+
+  const [logReplicationInfo, setLogReplicationInfo] =
+    React.useState<LogReplicationInfo>({
+      logLocation: '',
+      logPrefix: '',
+    });
```
SanjalKatiyar, line 65: maybe I will not block the PR for now on this particular FC... but this is not right... we are always maintaining state twice without any relevant reason... earlier we did for `rules, setRules` now again we added for `logReplicationInfo, setLogReplicationInfo`...
in future we will do that same again if requirements comes up...

fact that "useReducer" was created on the first place so that we can avoid maintaining so many "useState" variables inside a FC

SanjalKatiyar: and here we are using state from "useReducer" as well as maintaining multiple "useState" variables again inside FC...
which is a repetition as well as defeats the purpose of "useReducer"...

SanjalKatiyar: these are not really generic components... and even if we think that these components can be re-used at other places in future... still at other place as well we can just simply pass down the similar state from "useReducer" like we should have done from here as well...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
vbnrh: The thought around this is that if we uncheck all sync deletion, then i would want the value to be undefined for logreplicationinfo if it was previously set .

packages/odf/components/mcg/replication-policy-form.tsx
vbnrh: Yeah , i was just changing this. I built the underlying component first, then added it to replication form. The state of the log replication form got leftover

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
vbnrh: This method did not work

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -121,17 +139,30 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
         });
     });
     dispatch({ type: 'setReplicationRuleFormData', data: rulesFormData });
+    if (
+      !rules.some((rule) => rule.syncDeletion) ||
+      !_.isEmpty(logReplicationInfo)
+    ) {
+      dispatch({
+        type: 'setLogReplicationInfo',
+        data: { ...logReplicationInfo },
```
SanjalKatiyar, line 163: I hope "logReplicationInfo" will never be "undefined", because then "...logReplicationInfo" will give error...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
SanjalKatiyar: did u try adding it to namespace store modal's submit ??

SanjalKatiyar: it will not work if u add it here...

SanjalKatiyar: I am sure issue is due to event bubbling, but if it still not work add a proper comment, something like:
ex: `NamespaceStore modal's submit is triggering this form as well, added this check to prevent that`

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
SanjalKatiyar: I hope `logReplicationInfo` will never be `undefined`, because then `...logReplicationInfo` will give error...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -159,9 +190,10 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
       if (!!state.replicationRuleFormData.length) {
         const replicationPolicy = createReplicationRulesAndStringify(
           state.replicationRuleFormData,
-          namespace
+          namespace,
+          state.logReplicationInfo
         );
-        obj.spec.additionalConfig = { 'replication-policy': replicationPolicy };
+        obj.spec.additionalConfig = { replication_policy: replicationPolicy };
```
SanjalKatiyar, line 196: `React Hook React.useEffect has a missing dependency: 'state.logReplicationInfo'. Either include it or remove the dependency array`

packages/odf/components/mcg/replication-policy-form.tsx
```diff
@@ -75,13 +89,37 @@ export const ReplicationPolicyForm: React.FC<ReplicationFormProps> = ({
     updateParentState(rules);
   };
 
+  const handleSyncDeletion = (value: boolean, ruleId: number) => {
+    const newRules = [...rules];
+    const index = newRules.findIndex((rule) => rule.id === ruleId);
+    newRules[index].syncDeletion = value;
+    setRules(newRules);
+    let temp = { ...logReplicationInfo };
+    if (!rules.some((rule) => rule.syncDeletion)) {
```
SanjalKatiyar, line 98: `setRules(newRules)` will not update `rules` instantly...
```suggestion
    if (!rules.some((rule) => rule.syncDeletion) && value === false) {
```

SanjalKatiyar: also, we have same check `!rules.some((rule) => rule.syncDeletion) ` at multiple places... do we need at all the places ??

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
vbnrh: Namespacestore modal's submit is connected to the NSForm. It is out of scope for this PR.

vbnrh: `// I don't really want to do this but the namespace store modal's submit triggers this form also` 
This comment is not fine ? : )

SanjalKatiyar: u can add a "ToDo" as well on your name for later...

packages/odf/components/mcg/replication-policy-form.tsx
vbnrh: `setRules(newRules) will not update rules instantly...`  -> i did not see any issues with this ? what do you mean ?
Screenshot from 2023-02-27 20-01-34

vbnrh: `also, we have same check !rules.some((rule) => rule.syncDeletion)  at multiple places... do we need at all the places ??` -> yeah this condition is required in lot of places if even a single sync deletion box is checked, we need the log replication form inputs as well.

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
SanjalKatiyar: this is little hackish way... we have methods to handle such situations...

SanjalKatiyar: "This comment is not fine ? : )" >> not hard a requirement, just `I don't really want to do this` part seems to be little odd :P

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
vbnrh: It won't be if it crosses the second condition right ?

SanjalKatiyar: it is "OR" condition... so if first condition is true, it won't worry about second condition...

packages/odf/components/mcg/replication-policy-form.tsx
SanjalKatiyar: u are setting "setRules(newRules)" on Line#96 then using "rules" on Line#98,
but "rules" on Line#98 will not have the changes u did on Line#96, until component re-renders again...

SanjalKatiyar: either use `newRules.some ...` or `!rules.some((rule) => rule.syncDeletion) && value === false`

vbnrh: ah ok, thanks for clarification.

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
vbnrh: Changed the condition order.

SanjalKatiyar: even changing order won't do anything... say first condition (undefined check) is false, but second is true ("some" one)... it will still execute `...logReplicationInfo` part... that is why I confirmed can it be "undefined" in any case...

vbnrh: It should not be undefined for any reason. Always gets initialized with empty object somewhere.

Even so, I'll fix it if a bug comes up

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -71,19 +73,35 @@ export const isObjectSC = (sc: StorageClassResourceKind) =>
...
       'destination-bucket'
     )}`,
     filter: {
       prefix: rule.prefix,
     },
+    sync_deletions: rule.syncDeletion,
   }));
-  return rules ? JSON.stringify(rules) : '';
+
+  let log_replication_info: OBCLogReplicationInfo | {} = !_.isEmpty(
```
bipuladh, line 91:
```suggestion
  const log_replication_info: OBCLogReplicationInfo | {} = !_.isEmpty(
```

bipuladh, line 91:
```suggestion
  let log_replication_info: OBCLogReplicationInfo = !_.isEmpty(
```

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -71,19 +73,35 @@ export const isObjectSC = (sc: StorageClassResourceKind) =>
...
+    logReplicationInfo
+  )
+    ? {
+        logs_location: {
+          logs_bucket: logReplicationInfo.logLocation,
+          prefix: logReplicationInfo.logPrefix,
+        },
+      }
+    : {};
```
bipuladh, line 100:
```suggestion
    : { logs_location: { logs_bucket: null, prefix: null } };
```

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -364,11 +397,15 @@ export const CreateOBCPage: React.FC<CreateOBCPageProps> = (props) => {
...
-
+  const submitBtnId = 'obc-submit-btn';
   const history = useHistory();
 
-  const save = (e: React.FormEvent<EventTarget>) => {
+  const save = (e: React.SyntheticEvent<HTMLFormElement, SubmitEvent>) => {
     e.preventDefault();
+
+    // NamespaceStore modal's submit is triggering this form as well, added this check to prevent that
+    if (e.nativeEvent.submitter.id !== submitBtnId) return;
```
bipuladh, line 407: in the namespacestore modal stop event propagation

packages/odf/components/mcg/replication-policy-form.tsx
```diff
@@ -201,6 +253,21 @@ export const RuleForm: React.FC<RuleFormProps> = ({
           />
         </FormGroup>
       </GridItem>
+      <GridItem>
+        <FormGroup
+          fieldId="sync-deletions-field"
+          className={'odf-mcg__rule-checkbox'}
```
bipuladh, line 259:
```suggestion
          className='odf-mcg__rule-checkbox'
```

packages/odf/components/mcg/replication-policy-form.tsx
```diff
@@ -201,6 +253,21 @@ export const RuleForm: React.FC<RuleFormProps> = ({
           />
         </FormGroup>
       </GridItem>
+      <GridItem>
+        <FormGroup
+          fieldId="sync-deletions-field"
```
bipuladh, line 258: shouldn't this match the id of checkbox?

packages/odf/components/mcg/replication-policy-form.tsx
```diff
@@ -151,6 +192,15 @@ export const ReplicationPolicyForm: React.FC<ReplicationFormProps> = ({
             </Button>
           </FormGroup>
         )}
+
+        {rules.some((rule) => rule.syncDeletion) && (
```
bipuladh, line 196: I see you using `rules.some` in multiple places in this file. Can you create a variable to track it?

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
SanjalKatiyar: had a discussion regarding this: https://github.com/red-hat-storage/odf-console/pull/658#discussion_r1118781050

packages/odf/components/mcg/replication-policy-form.tsx
vbnrh: there are 3 places we are using this all in different methods. It is out of scope for a single variable. Just keeping it variable for this form

PR #661, opened 2023-02-28
Add tool tips and hide control for unprivileged-users access

TimothyAsirJeyasing:


TimothyAsirJeyasing:
There's no use case as of now for this. we can revisit if needed

PR #663, opened 2023-02-28
Remove peer ready and peer available check for failover action(Do not merge)

GowthamShanmugam:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2173954

Subscription behavior:
  1. Even when peer-ready and peer-available statuses are not true, UI won't block the failover action.
  2. For failover, UI will only auto-select the subscriptions which are all have peer ready and peer available status true.  It is up to the user to continue the failover. UI won't block the user.
  3. For relocation, peer ready and peer available status should be true, otherwise UI will block the action with a proper error message.
  4. UI will display a peer not ready error message only for relocation and not for failover.
 
 ApplicationSet:
  1. Just like subscription, UI will display a peer not ready error message only for relocation and not for failover.
  2. Just like subscription, UI won't block the user from failover when the peer is not available or the peer ready is false.


GowthamShanmugam:
Screenshot from 2023-02-28 22-45-26

PR #665, opened 2023-03-01
Enabling MCO and DR dashboard based on admin privilege

GowthamShanmugam:
1. Only the admin user can access the DR dashboard.
2. Only the admin user can access the storage system dashboard.
3. Restricting the create the DRPolicy page route from non-admin users.

### DIFF
--- a/packages/mco/constants/common.ts
+++ b/packages/mco/constants/common.ts
@@ -11,3 +11,5 @@ export const enum SLA_STATUS {
   WARNING = 'warning',
   HEALTHY = 'healthy',
 }
+export const ACM_OBSERVABILITY_FLAG = 'ACM_OBSERVABILITY';
+export const ADMIN_FLAG = 'ADMIN';
--- a/packages/mco/features.ts
+++ b/packages/mco/features.ts
@@ -1,10 +1,20 @@
+import { SelfSubjectAccessReviewModel } from '@odf/shared/models';
 import {
+  k8sCreate,
   k8sList,
   K8sResourceCommon,
+  SelfSubjectAccessReviewKind,
   SetFeatureFlag,
 } from '@openshift-console/dynamic-plugin-sdk';
-import { SECOND, DEFAULT_SYNC_TIME } from './constants';
-import { AcmMultiClusterObservabilityModel } from './models';
+import * as _ from 'lodash-es';
+import {
+  SECOND,
+  DEFAULT_SYNC_TIME,
+  ACM_OBSERVABILITY_FLAG,
+  ADMIN_FLAG,
+  HUB_CLUSTER_NAME,
+} from './constants';
+import { AcmMultiClusterObservabilityModel, MirrorPeerModel } from './models';
 import { ACMMultiClusterObservability } from './types';
 
 let intervals: NodeJS.Timeout[] = [];
@@ -21,7 +31,37 @@ type FlagController = {
   syncTime?: number;
 };
 
-const acmMcoDetector: FeatureDetector = async (
+// Check the user's access to some resources.
+const ssarChecks = {
+  [ADMIN_FLAG]: {
+    group: MirrorPeerModel.apiGroup,
+    resource: MirrorPeerModel.plural,
+    verb: 'create',
+  },
+};
+
+const handleError = (
+  res: any,
+  flagKey: string,
+  setFlag: SetFeatureFlag,
+  intervalId: any
+) => {
+  if (res?.response instanceof Response) {
+    const status = res?.response?.status;
+    if (_.includes([502], status)) {
+      setFlag(flagKey, undefined);
+      clearInterval(intervals[intervalId]);
+    }
+    if (!_.includes([401, 403, 500], status)) {
+      setFlag(flagKey, undefined);
+    }
+  } else {
+    setFlag(flagKey, undefined);
+    clearInterval(intervals[intervalId]);
+  }
+};
+
+const acmObservabilityDetector: FeatureDetector = async (
   setFlag: SetFeatureFlag,
   flagKey: string,
   id: any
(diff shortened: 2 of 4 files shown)


GowthamShanmugam:
For admin users:
image

For app users:
image

packages/mco/features.ts
```diff
@@ -21,6 +30,15 @@ type FlagController = {
...
+// Check the user's access to some resources.
+const ssarChecks = {
+  [ADMIN_FLAG]: {
+    group: MirrorPeerModel.apiGroup,
+    resource: MirrorPeerModel.plural,
+    verb: 'create',
+  },
+};
+
 const acmMcoDetector: FeatureDetector = async (
```
SanjalKatiyar, line 42: function name should be diff...
it is not MCO detector, but observability detector...

packages/mco/constants/common.ts
```diff
@@ -11,3 +11,5 @@ export const enum SLA_STATUS {
   WARNING = 'warning',
   HEALTHY = 'healthy',
 }
+export const ACM_OBSERVABILITY_FLAG = 'ACM_MCO';
```
SanjalKatiyar, line 14: nits:
```suggestion
export const ACM_OBSERVABILITY_FLAG = 'ACM_OBSERVABILITY';
```

packages/mco/features.ts
GowthamShanmugam: ack, here mco means multicluster observability, but i prefer to change it observability

packages/mco/constants/common.ts
GowthamShanmugam: ack

plugins/mco/console-extensions.json
```diff
@@ -46,6 +46,9 @@
       "path": "/multicloud/data-services/data-policies/ramendr.openshift.io~v1alpha1~DRPolicy/~new",
       "exact": true,
       "component": { "$codeRef": "createDataPolicy.CreateDRPolicy" }
+    },
+    "flags": {
+      "required": ["ADMIN"]
```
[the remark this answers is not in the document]
GowthamShanmugam: yes, it will hide, list page component has built-in check for that  itself

TimothyAsirJeyasing:
LGTM

packages/mco/features.ts
```diff
@@ -46,10 +64,39 @@ const acmMcoDetector: FeatureDetector = async (
...
+    })) as SelfSubjectAccessReviewKind;
+    if (result?.status?.allowed) {
+      setFlag(flagKey, true);
+      clearInterval(intervals[id]);
+    }
+    setFlag(flagKey, result.status?.allowed);
+  } catch (error) {
+    setFlag(flagKey, false);
+  }
+};
```
SanjalKatiyar, line 90: there is an issue with this (same is true for `acmObservabilityDetector` as well):
we are clearing interval when `result?.status?.allowed === true`
but we should clear interval when `result?.status?.allowed === false` as well... else it will keep on checking until it will find `true` (which is wrong as we already got value as `false` meaning user don;t have access)...

SanjalKatiyar: also we have a function `handleError` on odf's `feature.ts` file, it would make sense to use that IMO...

GowthamShanmugam: ack, let me check this

packages/mco/features.ts
```diff
@@ -21,50 +32,127 @@ type FlagController = {
...
+      );
+      if (isEnabled) {
+        setFlag(flagKey, true);
+        clearInterval(intervals[id]);
+      } else {
+        setFlag(flagKey, false);
+        clearInterval(intervals[id]);
+      }
+    }
+  } catch (error) {
```
SanjalKatiyar, line 95: here consider this situation:
Observability CR can be created at any point of time, above it will keep on checking until it will find the CR...
but say CR is created, but it is not in "Ready" phase (taking some time to get into "Ready" state)... in this case u are setting flag as false and doing a clearInterval, even though CR will get "Ready" in later point of time... flag is set to false...

packages/mco/features.ts
```diff
@@ -21,50 +32,127 @@ type FlagController = {
...
+      cluster: HUB_CLUSTER_NAME,
+    })) as SelfSubjectAccessReviewKind;
+    if (!!result) {
+      if (result?.status?.allowed) {
+        setFlag(flagKey, true);
+        clearInterval(intervals[id]);
+      } else {
+        setFlag(flagKey, false);
+        clearInterval(intervals[id]);
+      }
```
SanjalKatiyar, line 123: consider 3 situations:
1. `result?.status?.allowed === true` >> means user is admin and we don;t need to poll anymore so we will clearInterval.
2. `result?.status?.allowed === false` >> means user is not admin and again we don't need to poll anymore so we will clearInterval.
3. `result?.status?.allowed === undefined` >> means for some reason this field is not present in the CR, so we don;t know whether user is admin or not... we need to poll again... we can't clearInterval. But, above it will set flag to false and stop polling further.

packages/mco/features.ts
```diff
@@ -21,50 +32,127 @@ type FlagController = {
...
+  if (res?.response instanceof Response) {
+    const status = res?.response?.status;
+    if (_.includes([403, 502], status)) {
+      setFlag(flagKey, undefined);
+      clearInterval(intervals[intervalId]);
+    }
+    if (!_.includes([401, 403, 500], status)) {
+      if (retryCount > 5) {
+        setFlag(flagKey, false);
+        clearInterval(intervals[intervalId]);
```
SanjalKatiyar, line 60: take example of Observability CR... is status is 404... will it stop polling and set the flag to false after 5 tries ??
Observability CR can be created at any point of time in the cluster... it can give 404 until then multiple times...

packages/mco/features.ts
GowthamShanmugam: We can't avoid the second situation, what if observability is not coming ready state?   Due to some reason if it got stuck then it will keep checking the CR again and again. 

Observability CR can be created at any point in time, above it will keep on checking until it will find the CR -> nope, I am using k8List, so it returns an empty array. so it won't be false. Here, we dont even need this !!acmObservability check, k8List always returns an empty array.

packages/mco/features.ts
GowthamShanmugam: We are not using a hook here, I dont think in any case it can go undefined while using k8sCreate. basically, we dont even need to retry detectSSAR unless until it throws exception

packages/mco/features.ts
SanjalKatiyar: "I am using k8List, so it returns an empty array. so it won't be false. Here, we dont even need this !!acmObservability check, k8List always returns an empty array." >> yeah yeah I know... I was pointing out the current behaviour...

comment here was this:
`We can't avoid the second situation, what if observability is not coming ready state? Due to some reason if it got stuck then it will keep checking the CR again and again.` 
This is rare situation, but situation where observability CR gets created and by the time it will get "Ready" our approach will set the flag to false can occur more frequently IMO.

packages/mco/features.ts
```diff
@@ -21,50 +32,127 @@ type FlagController = {
...
-        queryParams: {},
+        queryParams: { cluster: HUB_CLUSTER_NAME },
       })) as ACMMultiClusterObservability[];
-    const isEnabled = managedClusters?.[0]?.status?.conditions?.some(
-      (condition) => condition.status === 'True' && condition.type === 'Ready'
-    );
-    if (isEnabled) {
-      setFlag(flagKey, true);
-      clearInterval(intervals[id]);
+    if (!!acmObservability) {
```
SanjalKatiyar, line 83: `!![]` is `true`
so it will set flag to false even if Observability CR is not present right ??

packages/mco/features.ts
GowthamShanmugam: We are using k8sList, and it won't check for a specific CR, Exception can occur only when CRD itself is not present (or) the observability namespace itself is not present. We need to hard stop the logic.

in case of any CR case also it is the same. We dont have to retry again and again.

packages/mco/features.ts
GowthamShanmugam: yes

GowthamShanmugam: i will change this logic to set undefined when: 

if(!!acmObservability.length){

} else {
   setFlag(flagKey, undefined);
}

packages/mco/features.ts
SanjalKatiyar: I got ur point... but "handleError" is supposed to be a utility/generic function, right ?
it may work for k8list, but the function which is calling "handleError" may/may not be polling for specific CR...

packages/mco/features.ts
```diff
@@ -21,50 +32,119 @@ type FlagController = {
...
+      data: {
+        apiVersion: 'authorization.k8s.io/v1',
+        kind: 'SelfSubjectAccessReview',
+        spec: { resourceAttributes: ssarChecks[flagKey] },
+      },
+      cluster: HUB_CLUSTER_NAME,
+    })) as SelfSubjectAccessReviewKind;
+    const isAllowed = result?.status?.allowed;
+    setFlag(flagKey, isAllowed);
+    isAllowed && clearInterval(intervals[id]);
```
SanjalKatiyar, line 116: again same point:
if `isAllowed === false` (exactly "false" not "undefined")... then also we should `clearInterval`...

packages/mco/features.ts
```diff
@@ -21,50 +32,119 @@ type FlagController = {
...
+        queryParams: { cluster: HUB_CLUSTER_NAME },
       })) as ACMMultiClusterObservability[];
-    const isEnabled = managedClusters?.[0]?.status?.conditions?.some(
-      (condition) => condition.status === 'True' && condition.type === 'Ready'
-    );
-    if (isEnabled) {
-      setFlag(flagKey, true);
+    if (!!acmObservability.length) {
+      const isEnabled = acmObservability?.[0]?.status?.conditions?.some(
+        (condition) => condition.status === 'True' && condition.type === 'Ready'
```
SanjalKatiyar, line 85: just to confirm: https://github.com/red-hat-storage/odf-console/pull/665#discussion_r1136547482

packages/mco/features.ts
GowthamShanmugam: Users may or May not create the CR, if we are not hard stopping it, in a corner case, it will keep sending one request on a specific interval. 

We can increase the sync interval for this kind of flag check, but it won't help much. 

Okay! i will remove this retry logic

packages/mco/features.ts
GowthamShanmugam: yes yes, i missed

SanjalKatiyar:
/retitle Enabling MCO and DR dashboard based on admin privilege
now we have 4.13 branches, we need to add BZ number to the title for backport PRs now... not for master...

PR #679, opened 2023-03-10
Match the version of consoleplugin to odf operator

TimothyAsirJeyasing:


TimothyAsirJeyasing:
Screenshot from 2023-03-10 22-38-26

TimothyAsirJeyasing:
@SanjalKatiyar Please review

SanjalKatiyar:
@TimothyAsirJeyasing plz update title correctly...

plugins/odf/console-plugin.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "odf-console",
-  "version": "0.0.0",
+  "version": "4.13.0",
```
bipuladh, line 3: Make a script set this up based on some env variable. Keep the build team aware of the ENV_VAR and also update yarn build to run that script.

bipuladh (changes requested):

plugins/odf/console-plugin.json
TimothyAsirJeyasing: Sure, build team agreed to set the environment variable for main and compatibility
https://gitlab.cee.redhat.com/ceph/rhodf/-/commit/010143930adf3b2e1308e47069178fef81d4527c

package.json
```diff
@@ -22,8 +22,9 @@
     "test-cypress-headless": "node --max-old-space-size=4096 node_modules/.bin/cypress run --e2e --config-file ./cypress/cypress.config.ts --env openshift=true --browser ${BRIDGE_E2E_BROWSER_NAME:=chrome} --headless",
     "build:generate": "yarn clean-${PLUGIN} && tsm ./scripts/generatePluginPackage.mts --plugin ${PLUGIN}",
     "build:plugin": "yarn build:generate && yarn ts-node node_modules/.bin/webpack -c ./webpack.config.ts",
-    "build": "NODE_ENV=production PLUGIN=odf I8N_NS=plugin__odf-console yarn build:plugin",
-    "build-mco": "NODE_ENV=production PLUGIN=mco I8N_NS=plugin__odf-multicluster-console yarn build:plugin",
+    "build:version": "sed -i -e 's/\"version\": .*/\"version\": \"'${ODF_CONSOLE_VERSION:=0.0.0}'\",/g'",
```
SanjalKatiyar, line 25: what will be the format of the value of "ODF_CONSOLE_VERSION" ??

SanjalKatiyar: will it change based on z-stream ??

SanjalKatiyar, line 25: is it possible to check if "ODF_CONSOLE_VERSION" is a valid semver or not ??
if not valid we will set it to default value...
I don;t want plugin to fail, in case there is a bug on build script side !! (say they pass incorrect format of the version)....
cc @bipuladh

b-ranto: Yes, it will always match the current z-stream version, we always use the same X.Y.Z version for all the components during the build.

SanjalKatiyar: cool... thanks !

TimothyAsirJeyasing: Ok, How about the validation
[[ $ODF_CONSOLE_VERSION && $ODF_CONSOLE_VERSION =~ ^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-((0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*))?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?$ ]] || ODF_CONSOLE_VERSION="0.0.0"
sed -i -e 's/\"version\": .*/\"version\": \"'$ODF_CONSOLE_VERSION'\",/g' test.json

SanjalKatiyar: cc @alfonsomthd what do u think would be best practice here...
it's a small command so does not make sense to have a separate script file for this...
still long enough to be added directly to package.json...
another thing is we have utilities like `semver.valid()` which we can use, but that would mean adding a new dependency...

alfonsomthd: After a conversation we agree on creating a script that runs the semver validation and it's invoked during `yarn build`

TimothyAsirJeyasing: ack

versionCheck.js
```diff
@@ -0,0 +1,24 @@
+var fs = require('fs');
```
bipuladh, line 1: This should be a typescript file. 
It should use import statements instead of require. 
It should live under `scripts/` directory. 
For reference please check https://github.com/red-hat-storage/odf-console/blob/master/scripts/generatePluginPackage.mts

versionCheck.js
```diff
@@ -0,0 +1,24 @@
...
+const semver = require('semver');
+
+const defaultVer = '0.0.0';
+const file = process.argv[2];
+const version = process.env['ODF_CONSOLE_VERSION'] ?? defaultVer;
```
bipuladh, line 6:
```suggestion
const version = process.env['ODF_CONSOLE_VERSION'] || defaultVer;
```

versionCheck.js
```diff
@@ -0,0 +1,24 @@
...
+
+file &&
```
bipuladh, line 8:
```suggestion
if (file === undefined) {
    console.error("usage: mts versionCheck <pluginName>");
```

versionCheck.js
```diff
@@ -0,0 +1,24 @@
+var fs = require('fs');
+const semver = require('semver');
+
+const defaultVer = '0.0.0';
+const file = process.argv[2];
```
bipuladh, line 5:
```suggestion
const pluginName = process.argv[2];
```

versionCheck.js
```diff
@@ -0,0 +1,24 @@
...
+    if (err) {
+      console.log('File not found!', file);
+      return;
+    }
+    fs.readFile(file, 'utf8', function (err, content) {
+      const validVersion = semver.valid(version) ?? defaultVer;
+      const replace = content.replace(
+        /"version": .*/gi,
+        '"version": "' + validVersion + '",'
+      );
```
bipuladh, line 19: It's a JSON file you can resolve it as a object and then update the field. Why use REGEX?

versionCheck.js
```diff
@@ -0,0 +1,24 @@
...
+      fs.writeFile(file, replace, 'utf-8', function (err) {
+        console.log(err);
+      });
```
bipuladh, line 22: You can later serialize the JSON and write again.

bipuladh (changes requested):

versionCheck.js
TimothyAsirJeyasing: Thank you, will update accordingly

versionCheck.js
TimothyAsirJeyasing: Will update

versionCheck.js
TimothyAsirJeyasing: Done the changes

versionCheck.js
TimothyAsirJeyasing: Done

versionCheck.js
TimothyAsirJeyasing: Sure will use JSON only and agree JSON will be much suitable one here.

versionCheck.js
TimothyAsirJeyasing: Sure will do that

SanjalKatiyar:
once backport PRs are up (after this master PR is merged), add the BZ number to the title of both PRs and make sure both PRs are added to "Links" section of the BZ... after that we can move BZ to "POST" and need to ask for "qa_ack" on the BZ... once done we can merge the backport PRs and move BZ to MODIFIED...

SanjalKatiyar:
fix the build tests...

scripts/versionCheck.mts
```diff
@@ -0,0 +1,38 @@
...
+
+const args = process.argv;
+
+const indexOfPluginArgument = args.indexOf(pluginNameArgument);
+const pluginToBeProcessed = args[indexOfPluginArgument + 1];
+
+if (!validatePluginName(pluginToBeProcessed)) {
+  // eslint-disable-next-line no-console
+  console.error(`Invalid argument ${pluginToBeProcessed}`);
+  process.exit(1);
```
alfonsomthd, line 24: According to [Node.js documentation](https://nodejs.org/api/process.html#process_process_exit_code): `If it is necessary to terminate the Node.js process due to an error condition, throwing an uncaught error and allowing the process to terminate accordingly is safer than calling process.exit().`

```suggestion
  throw new Error(`Invalid argument: '${pluginToBeProcessed}'`);
```

alfonsomthd (changes requested):

scripts/versionCheck.mts
TimothyAsirJeyasing: Sure, will do that

alfonsomthd:
> /approve

@TimothyAsirJeyasing it does not work for me.

alfonsomthd:
@TimothyAsirJeyasing Please rebase.

SanjalKatiyar:
@TimothyAsirJeyasing plz resolve the conflicts...

scripts/versionCheck.mts
```diff
@@ -0,0 +1,36 @@
...
+}
+const pluginDir = `${__dirname}/../plugins/${pluginToBeProcessed}`;
+
+const consolePlugin = JSON.parse(
+  fs.readFileSync(`${pluginDir}/console-plugin.json`, {
+    encoding: 'utf-8',
+  })
+);
+
+const version = process.env['ODF_CONSOLE_VERSION'] || defaultVer;
```
SanjalKatiyar, line 32: we are using same script for mco/odf...
`ODF_CONSOLE_VERSION` is specific to odf only...

b-ranto: No worries there. We use the same ENV variable DS for both odf and mco builds. In fact, we use the exact same `X.Y.Z` version for all the components that we build in ODF. (we could override it for MCO if we wanted to but no reason really)

SanjalKatiyar: @b-ranto can we please use generic name instead like `PLUGIN_VERSION" ??
because mco and odf console are 2 different plugins, using `ODF_CONSOLE_VERSION` for both is a incorrect naming convention (even though logically correct)...

b-ranto: Ack, I agree that `ODF_CONSOLE_VERSION` might like weird for MCO console. Let's use `PLUGIN_VERSION` instead.

SanjalKatiyar:
plz resolve the conflicts...

bipuladh (approved):

TimothyAsirJeyasing:
Manually its applying now to the branches, Could you please cherry pick one more time

PR #683, opened 2023-03-14
Fixes page breaking issue with topology

bipuladh:
Adds optional chaining for times when Node CR is not fully populated.


DanielOsypenko:
After discussion adding collected logs here.
I have a case may be a result of current issue - 
image

stack trace from the page - https://www.google.com/url?sa=j&url=http%3A%2F%2Fpastebin.test.redhat.com%2F1094456&uct=1672476946&usg=37hzztqLj3rWOMSywTFXUr5rmX4.&source=chat

must-gather - logs https://drive.google.com/drive/folders/1aLXswz6mU2EqDK6sAWab936_2hjmhV-k?usp=share_link

PR #695, opened 2023-03-15
CI build: constrain the memory usage of node processes

alfonsomthd:
As the CI memory hard limit has been [increased to 12 GB](https://github.com/openshift/release/pull/37326), here we limit & constrain the memory usage of the node processes to stay below that limit:
- Generic: for any node process during CI `yarn build`: 1GB.
- thread-loader: limit the amount of workers; memory allowed for each worker: 1GB.
- fork-ts-checker-webpack-plugin: the default: 2GB.

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -2,7 +2,7 @@ FROM docker.io/library/node:14.17.3 AS builder
 WORKDIR /app
 COPY . /app
 RUN yarn install
-RUN NODE_OPTIONS="--max-old-space-size=1000" yarn build
+RUN NODE_OPTIONS="--max-old-space-size=1024" yarn build
 
 
 FROM quay.io/ocs-dev/cypress:latest as serverpackage
--- a/Dockerfile.mco.ci
+++ b/Dockerfile.mco.ci
@@ -2,7 +2,7 @@ FROM docker.io/library/node:14.17.3 AS builder
 WORKDIR /app
 COPY . /app
 RUN yarn install
-RUN NODE_OPTIONS="--max-old-space-size=1000" yarn build-mco
+RUN NODE_OPTIONS="--max-old-space-size=1024" yarn build-mco
 
 
 FROM quay.io/ocs-dev/cypress:latest as serverpackage
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -18,6 +18,7 @@ const resolveLocale = (dirName: string, ns: string) =>
 const NODE_ENV = (process.env.NODE_ENV ||
   'development') as webpack.Configuration['mode'];
 const PLUGIN = process.env.PLUGIN;
+const OPENSHIFT_CI = process.env.OPENSHIFT_CI;
 
 if (PLUGIN === undefined) {
   process.exit(1);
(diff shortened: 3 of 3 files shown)


webpack.config.ts
```diff
@@ -57,7 +58,14 @@ const config: webpack.Configuration & DevServerConfiguration = {
           {
             loader: 'thread-loader',
             options: {
-              ...(NODE_ENV === 'development' ? { poolTimeout: Infinity } : {}),
+              ...(NODE_ENV === 'development'
+                ? { poolTimeout: Infinity }
```
GowthamShanmugam, line 62:
```suggestion
                ? { poolTimeout: Infinity, poolRespawn: false }
```

?

webpack.config.ts
```diff
@@ -18,6 +18,7 @@ const resolveLocale = (dirName: string, ns: string) =>
 const NODE_ENV = (process.env.NODE_ENV ||
   'development') as webpack.Configuration['mode'];
 const PLUGIN = process.env.PLUGIN;
+const OPENSHIFT_CI = process.env.OPENSHIFT_CI;
```
bipuladh, line 21: we need to update the `package.json` file as well.

Dockerfile.ci
```diff
@@ -2,7 +2,7 @@ FROM docker.io/library/node:14.17.3 AS builder
 WORKDIR /app
 COPY . /app
 RUN yarn install
-RUN NODE_OPTIONS="--max-old-space-size=1000" yarn build
+RUN NODE_OPTIONS="--max-old-space-size=700" yarn build
```
bipuladh, line 5: Make it 4 gigs.

Dockerfile.mco.ci
```diff
@@ -2,7 +2,7 @@ FROM docker.io/library/node:14.17.3 AS builder
 WORKDIR /app
 COPY . /app
 RUN yarn install
-RUN NODE_OPTIONS="--max-old-space-size=1000" yarn build-mco
+RUN NODE_OPTIONS="--max-old-space-size=700" yarn build-mco
```
bipuladh, line 5: 4 gigs

Dockerfile.ci
alfonsomthd: No need as this constraint only affects to the webpack process that does not consume more than 600mb.
The 4 GB that we have on CI are for the full memory of a pod, so we're are reserving 2048 to fork-ts-checker plugin (it's default, we shouldn't make it less) and for the thread-loader workers.

webpack.config.ts
alfonsomthd: @bipuladh No need of it as this is an env. variable that is [available in OpenShift CI environment](https://docs.ci.openshift.org/docs/architecture/step-registry/#available-environment-variables) (we don't need it in local).

webpack.config.ts
alfonsomthd: Done.

Dockerfile.ci
alfonsomthd: After our conversation, updating it to 1024.

GowthamShanmugam:
it LGTM

SanjalKatiyar:
/lgtm
will let others to add "approve" though, in case anyone want to take a another look

Dockerfile.mco.ci
alfonsomthd: After our conversation, updating it to 1024.

bipuladh (approved):

PR #699, opened 2023-03-17
CI: add .ci-operator.yaml

alfonsomthd:
This PR adds the [.ci-operator.yaml](https://github.com/alfonsomthd/odf-console/pull/new/ci-op-yaml) as prerequisite for using our custom CI runner image so it's noe needed to be built on the fly during every run (which causing errors some times).

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,0 +1,4 @@
+build_root_image:
+  name: odf-console-ci-runner
+  namespace: ci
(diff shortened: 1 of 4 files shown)


SanjalKatiyar:
> @SanjalKatiyar: Overrode contexts on behalf of SanjalKatiyar: ci/prow/odf-console-e2e-aws

not affecting the e2e product's flow...

Issue #671, opened 2023-03-06
[bug] [odf/mco] Fix anti-pattern in SelectorInput


SanjalKatiyar:
SelectorInput (https://github.com/red-hat-storage/odf-console/blob/master/packages/shared/modals/Selector.tsx#L173): used in add labels modal and apply-policy (mco) modal.

Priority of issue is "low" **for now**, still it is an anti-pattern and may cause bug moving forward if this component is used at multiple places.
Check demo: https://codesandbox.io/s/m3w9zn1z8x (same case can happen with SelectorInput as well)
Refer: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#common-bugs-when-using-derived-state

SanjalKatiyar:
cc @bipuladh @GowthamShanmugam (just FYI)

SanjalKatiyar:
commit which introduced this: https://github.com/red-hat-storage/odf-console/pull/579/commits/299a8796b77f9bb833d4346b056fcfad0fcffe84#diff-c6f3325fae0fffcf0cf61edd0c338cd51f59a881106d9000842c4370b6990c5c

TimothyAsirJeyasing:
Shall we replace getDerivedStateFromProps with componentDidUpdate similar to console repo for SelectorInput ?

PR #700, opened 2023-03-17
Fix anti-pattern in SelectorInput

TimothyAsirJeyasing:
https://github.com/red-hat-storage/odf-console/issues/671

### DIFF
--- a/packages/shared/modals/Selector.tsx
+++ b/packages/shared/modals/Selector.tsx
@@ -170,12 +170,11 @@ export class SelectorInput extends React.Component<
     };
   }
 
-  static getDerivedStateFromProps(props, state) {
-    if (!_.isEqual(props.tags, state.tags)) {
-      return { ...state, tags: props.tags };
+  componentDidUpdate = (prevProps) => {
+    if (!_.isEqual(prevProps.tags, this.props.tags)) {
+      this.setState({ tags: this.props.tags });
     }
-    return state;
-  }
+  };
 
   static arrayify(obj) {
     return _.map(obj, (v, k) => (v ? `${k}=${v}` : k));


TimothyAsirJeyasing:
@SanjalKatiyar Please review

SanjalKatiyar:
/hold
@TimothyAsirJeyasing as discussed earlier, please add a small screencast of the working of this component in "apply-policy" modal of MCO. (This FC is used multiple times on the same modal, add PVC labels to all the input boxes on the modal and check/verify functionality is working as expected).

packages/shared/modals/Selector.tsx
```diff
@@ -170,11 +170,11 @@ export class SelectorInput extends React.Component<
...
 
-  static getDerivedStateFromProps(props, state) {
-    if (!_.isEqual(props.tags, state.tags)) {
-      return { ...state, tags: props.tags };
+  componentDidUpdate(prevProps) {
+    if (!_.isEqual(prevProps.tags, this.props.tags)) {
+      return { ...this.state, tags: this.props.tags };
     }
-    return state;
+    return this.state;
```
SanjalKatiyar, line 177: nits:
```suggestion
```

SanjalKatiyar, line 177: `ComponentDidUpdate` should not return anything...

SanjalKatiyar: check the React docs...

packages/shared/modals/Selector.tsx
```diff
@@ -170,11 +170,14 @@ export class SelectorInput extends React.Component<
...
-    if (!_.isEqual(props.tags, state.tags)) {
-      return { ...state, tags: props.tags };
+  componentDidUpdate(prevProps) {
+    if (!_.isEqual(prevProps.tags, this.props.tags)) {
+      this.updateProp(this.props.tags);
     }
-    return state;
+  }
+
+  updateProp(tags: string[]) {
```
alfonsomthd, line 179: Shouldn't this be called `updateTags` ?
```suggestion
  priavte updateTags(tags: string[]) {
```

packages/shared/modals/Selector.tsx
```diff
@@ -170,11 +170,14 @@ export class SelectorInput extends React.Component<
...
-      return { ...state, tags: props.tags };
+  componentDidUpdate(prevProps) {
+    if (!_.isEqual(prevProps.tags, this.props.tags)) {
+      this.updateTags(this.props.tags);
     }
-    return state;
+  }
+
+  private updateTags(tags: string[]) {
+    this.setState({ tags: tags });
```
SanjalKatiyar, line 180: make it inline, no need for separate function... and test for mco apply-policy modal once again, similar to how we did earlier...

TimothyAsirJeyasing:
https://user-images.githubusercontent.com/6670284/229709073-c7766234-7d3e-488c-89e8-50d385b04f55.mp4

packages/shared/modals/Selector.tsx
TimothyAsirJeyasing: i received an lint error when i keep inside saying " Do not use setState in componentDidUpdate". I have seen in console they kept outside function.

SanjalKatiyar: "i received an lint error when i keep inside saying " Do not use setState in componentDidUpdate" >> this does not solve anything, u r still using "setState" in "componentDidUpdate", just now calling it from another location...
"I have seen in console they kept outside function." >> can u send me console link to this ??

alfonsomthd: @SanjalKatiyar I think Tim refers to [this](https://github.com/openshift/console/blob/master/frontend/public/components/utils/selector-input.jsx#L27).

SanjalKatiyar: yes, I checked that...
I meant here that there is no need to create a separate function just to add single line "this.setState({ tags: tags });",
functions are created for either re-usability purpose or to make  "componentDidUpdate" clean... but here "updateTags" is only a single line, single use function which can be made inline in "componentDidUpdate"...

anyway... not a hard requirement, consider this comment as "nits", up to the assignee to fix if what, else no need as currently it is alright as well...

TimothyAsirJeyasing:
https://user-images.githubusercontent.com/6670284/229754353-8d164b9a-db45-407a-9d58-2a77d04e9dd1.mp4

PR #735, opened 2023-03-27
Bug 2181832: added placeholder for resource dropdown

debjyoti-pandit:
fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2181832

**Before selecting**
image

**After selecting**
image

It also changes secret dropdown
image

PVC storage class dropdown
image

GCP secret dropdown
image

PVC for NSS
image

image

### DIFF
--- a/packages/shared/dropdown/ResourceDropdown.tsx
+++ b/packages/shared/dropdown/ResourceDropdown.tsx
@@ -85,14 +85,16 @@ type ResourceDropdownTextProps = {
   showBadge?: boolean;
   resourceModel?: K8sModel;
   text: string;
+  className?: string;
 };
 
 const ResourceDropdownText: React.FC<ResourceDropdownTextProps> = ({
   showBadge = true,
   resourceModel,
   text,
+  className,
 }) => (
-  <span>
+  <span className={className}>
     {text && showBadge && <ResourceBadge model={resourceModel} />}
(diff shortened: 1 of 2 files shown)


packages/shared/dropdown/ResourceDropdown.tsx
```diff
@@ -233,9 +233,13 @@ const ResourceDropdown: ResourceDropdown = <T extends unknown>({
               !loadError &&
               (_.isEmpty(resources) ? (
                 <span>{t('No resources available')}</span>
+              ) : !selectedItem ? (
+                <span className="text-muted">
+                  {t(`Select ${resourceModel.label}`)}
+                </span>
```
SanjalKatiyar, line 239: let `ResourceDropdownText` handle this... instead of adding one more condition here...

SanjalKatiyar:
run `yarn i18n`...

packages/shared/dropdown/ResourceDropdown.tsx
debjyoti-pandit: if we dropdown text, it contains formatting and badges
image

SanjalKatiyar: `ResourceDropdownText` will take care of when to add a badge, when not to...

PR #742, opened 2023-03-28
added missing vault enterprise namespace to csi kms config map

debjyoti-pandit:
fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2181446

as the data are filled, it is reflected in the config map -
image
image
image

content of the secret is the same as that entered
image
image
image

image
image
image
All the content of the secrets are same as those entered
image
image
image
image


SanjalKatiyar:
/retitle added missing vault enterprise namespace to csi kms config map

PR #744, opened 2023-03-28
Topology: fix 'Exit fullscreen' message not shown

alfonsomthd:
When expanding to full screen, 2 flows were started: the `fullscreenchange` event handler and the `requestFullscreen` promise success callback. As these flows were asynchronous and were updating the same value, sometimes the fullscreen status was updated twice (ending up in an inconsistent state).
Now we only rely on the `fullscreenchange` event handler for updating the fullscreen status.


bipuladh:
Great work!
/lgtm
/approve

PR #765, opened 2023-04-04
CI: use nginx as server for console images

alfonsomthd:
- To be merged after https://github.com/red-hat-storage/odf-operator/pull/307
- Replace `http-server` with NGINX (only for `odf` plugin, not `mco`). This fixes this error that happens now and then:
```
Error:
[2/3] STEP 3/4: RUN yarn global add http-server
yarn global v1.22.10
[1/4] Resolving packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/http-server: getaddrinfo EAI_AGAIN registry.yarnpkg.com".
info If you think this is a bug, please open a bug report with the information provided in "/usr/local/share/.config/yarn/global/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/global for documentation about this command.
Error link: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/37436/rehearse-37436-pull-ci-red-hat-storage-odf-console-master-odf-console-e2e-aws/1637717349631004672
```
Additional improvements:
- Upgrade Node.js version (used for build the app) to LTS.
- Add extra check in e2e workflow to check the status of the odf-console container.
- Add missing github action for pushing the CI runner image.
- Add missing documentation for pushing a CI runner beta image.


Dockerfile.ci
```diff
@@ -1,15 +1,10 @@
...
-
-FROM docker.io/library/node:14.17.3 as serverpackage
-RUN yarn global add http-server
-RUN export PATH="$(yarn global bin):$PATH"
-
-FROM serverpackage
+FROM quay.io/ocs-dev/odf-console-ci-server:current
 COPY --from=builder /app/plugins/odf/dist /app
 COPY --from=builder /app/plugins/odf/dist /app/compatibility
 COPY --from=builder /app/http-server.sh .
```
SanjalKatiyar, line 10: had a discussion with @alfonsomthd regarding updating `Dockerfile.ci` to run with `nginx` instead of `http-server` and use the same Dockerfile for upstream `ocs-dev` images as well, so that we can have a setup as close to production as possible (as in downstream images we use `nginx-120` but upstream/CI is still using `http-server`)...
@bipuladh wdyt regarding this ??

alfonsomthd: IMHO the closer to production, the better...

alfonsomthd: @SanjalKatiyar @bipuladh After testing a beta image with NGINX in test cluster, I updated the PR replacing `http-server` with NGINX.

bipuladh (commented):
Changes look good to me.

default.conf
```diff
@@ -1,22 +0,0 @@
...
-    location = /50x.html {
-        root   /usr/share/nginx/html;
-    }
-    ssi on;
-    add_header Last-Modified $date_gmt;
-    add_header Cache-Control 'no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0';
-    if_modified_since off;
-    expires off;
-    etag off;
-}
```
SanjalKatiyar, line 22: mco is still using this... that will break if we remove it right now...

.github/workflows/quay-image.yml
```diff
@@ -16,10 +16,16 @@ jobs:
...
-        run: docker build . --file Dockerfile --tag quay.io/ocs-dev/odf-console:latest
+        run: docker build . --file Dockerfile.ci --tag quay.io/ocs-dev/odf-console:latest
       - name: Push the latest image to ocs-dev
         run: docker push quay.io/ocs-dev/odf-console:latest
+
+        # CI runner.
+      - name: Build the CI runner image
+        run: docker build . --file Dockerfile.ci.runner --tag quay.io/ocs-dev/odf-console-ci:current
+      - name: Push the CI runner image to ocs-dev
+        run: docker push quay.io/ocs-dev/odf-console-ci-runner:current
```
SanjalKatiyar, line 31: do we need to create runner image everytime there is a push to master ??

default.conf
SanjalKatiyar: we need to update mco operator first so that it brings its own config...

SanjalKatiyar: can u check  mco Dockerfile once ?

.github/workflows/quay-image.yml
SanjalKatiyar: no strict issues from me, just confirming...

default.conf
alfonsomthd: > can u check mco Dockerfile once ?

Provided that there is no config for it on the operator side, I'm not sure what you mean...

alfonsomthd: I'll revert this particular change.

.github/workflows/quay-image.yml
alfonsomthd: Right, I'll create a separate job for the image only when its Dockerfile is changed...

default.conf
SanjalKatiyar: eventually mco operator will bring its own config as well...

alfonsomthd: @SanjalKatiyar Ack. Then I'll exclude any MCO-related change from this PR.

SanjalKatiyar:
/approve
we can add "lgtm" once tests passes...

PR #769, opened 2023-04-06
Topology sidebar Alerts panel: toggle alerts accordion on text click

alfonsomthd:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2184984

### DIFF
--- a/packages/shared/alert/AlertsPanel.tsx
+++ b/packages/shared/alert/AlertsPanel.tsx
@@ -57,14 +57,13 @@ const AlertBadge: React.FC<AlertBadgeProps> = ({
   const onClick = () => onToggle(`alert-toggle-${alertSeverity}`);
 
   return (
-    <>
+    <span onClick={onClick}>
       <Badge
         key={key}
         className={`odf-alerts-panel__badge odf-alerts-panel__badge-${alertSeverity}`}
       >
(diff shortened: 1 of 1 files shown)


packages/shared/alert/AlertsPanel.tsx
```diff
@@ -70,7 +70,7 @@ const AlertBadge: React.FC<AlertBadgeProps> = ({
           <StatusIconAndText title={alerts.length.toString()} icon={icon} />
         </Button>
       </Badge>
-      <span className="odf-alerts-panel__badge-text">
+      <span className="odf-alerts-panel__badge-text" onClick={onClick}>
```
bipuladh, line 73: Can we make the whole thing one div and put the `onClick` listener there?

alfonsomthd: I tested with a div and the view broke: with a `<span>` it works. Should we go for `span` ?

alfonsomthd: > I tested with a div and the view broke: with a `<span>` it works. Should we go for `span` ?

@bipuladh replaced fragment with `span`.

bipuladh (approved):
/lgtm

PR #779, opened 2023-04-11
Fixes issues with Alerts and Status

bipuladh:
Screenshot from 2023-04-06 17-39-37


SanjalKatiyar:
@bipuladh LGTM
can u squash the commits ?
also, there is a lint warning (introduced by one of my PRs) for `TopologyNavigation.tsx`, can u plz pass the controller to the dependency list as well ?

SanjalKatiyar:
LGTM
cc @alfonsomthd (for further reviews)

packages/shared/topology/utils/AlertFilters.ts
```diff
@@ -0,0 +1,32 @@
...
+  };
+export const nodeFilter = (name: string) => (alert: Alert) =>
+  alert?.labels?.node === name || commonFilter(name)(alert);
+
+export const filterRelevantAlerts = (
+  resourceName: string,
+  resourceModel: K8sModel,
+  alerts: Alert[]
+): Alert[] => {
+  if (!resourceName || !!alerts) return [];
```
alfonsomthd, line 21: Shouldn't we check is there are no alerts?
```suggestion
  if (!resourceName || _.isEmpty(alerts)) return [];
```

bipuladh: ack

PR #782, opened 2023-04-11
add check for circular dependency

SanjalKatiyar:


SanjalKatiyar:
@bipuladh @alfonsomthd there are other ways as well (like adding a analyser script and create a dependency visualisation etc etc), but webpack seems easiest and relatively fastest...

Issue #672, opened 2023-03-06
[perf-improv] create separate provider for the context and pass its descendants as a "children"


SanjalKatiyar:
Currently used:
https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/topology/Topology.tsx#L502
https://github.com/red-hat-storage/odf-console/blob/master/packages/mco/components/mco-dashboard/data-policy/dr-dashboard.tsx#L163

Refer: https://reactjs.org/docs/context.html#contextprovider
**All consumers** that are descendants of a Provider will re-render whenever the Provider’s value prop changes.
Whenever Provider’s value prop changes **all** the child FCs will get re-rendered **(irrespective of whether a child is a consumer of the context or not)**. This will cause unnecessary re-renders for those FCs which are not consuming any state from the context.
To avoid that, use composition and create a separate FC which will provide the context and pass the consumer FCs as its "children".
example: https://github.com/openshift/console/blob/master/frontend/public/components/app.jsx#L241

SanjalKatiyar:
cc @bipuladh @GowthamShanmugam (just FYI)

black-dragon74:
@SanjalKatiyar can I take this one?

SanjalKatiyar:
> @SanjalKatiyar can I take this one?

/assign @debjyoti-pandit 
Hi,
this is already assigned and WIP.

PR #783, opened 2023-04-12
fix: context merging for dr dashboard

debjyoti-pandit:
fixes: https://github.com/red-hat-storage/odf-console/issues/672

### DIFF
--- a/packages/mco/components/mco-dashboard/data-policy/dr-dashboard-context.tsx
+++ b/packages/mco/components/mco-dashboard/data-policy/dr-dashboard-context.tsx
@@ -10,14 +10,46 @@ export const DRResourcesContext = React.createContext<DRResourcesContextType>(
   {} as DRResourcesContextType
 );
 
-type CSVStatusesContextType = {
+export type CSVStatusesContextType = {
   csvData: PrometheusResponse;
   csvError: any;
   csvLoading: boolean;
 };
 
-type DRResourcesContextType = {
+export type DRResourcesContextType = {
   drClusterAppsMap: DrClusterAppsMap;
   loaded: boolean;
   loadError: any;
 };
+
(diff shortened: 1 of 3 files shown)


SanjalKatiyar:
we simply have to add ex: "CSVStatusesContext.Provider"  to "CSVStatusesContextProvider" FC, which will take values as a prop and pass it down to its "children"... we can reduce the complexity of this PR... most of the thing are not necessarily needed... 

example: https://github.com/openshift/console/blob/master/frontend/public/components/app.jsx#L241
see "ModalProvider" FC, which internally passes value down to "children".

SanjalKatiyar:
> we simply have to add ex: "CSVStatusesContext.Provider" to "CSVStatusesContextProvider" FC, which will take values as a prop and pass it down to its "children"... we can reduce the complexity of this PR... most of the thing are not necessarily needed...
> 
> example: https://github.com/openshift/console/blob/master/frontend/public/components/app.jsx#L241 see "ModalProvider" FC, which internally passes value down to "children".

@debjyoti-pandit

debjyoti-pandit:
> > we simply have to add ex: "CSVStatusesContext.Provider" to "CSVStatusesContextProvider" FC, which will take values as a prop and pass it down to its "children"... we can reduce the complexity of this PR... most of the thing are not necessarily needed...
> > example: https://github.com/openshift/console/blob/master/frontend/public/components/app.jsx#L241 see "ModalProvider" FC, which internally passes value down to "children".
> 
> @debjyoti-pandit

On it, currently working on tests and am stuck. once I and done will fix this PR.

SanjalKatiyar:
can we update this PR @debjyoti-pandit sometime as well ?? pending comments for quite a while...

debjyoti-pandit:
Sure

PR #802, opened 2023-04-18
Fix replication history graph

GowthamShanmugam:
Fix: mod operation is resetting the replication interval.
For example,
  If the second is 5400, and when we call getDuration it returns the result as 30M instead of 90M. The result for minutes is always under 60M.
  
 BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2187736

### DIFF
--- a/packages/shared/utils/humanize.js
+++ b/packages/shared/utils/humanize.js
@@ -1,5 +1,4 @@
 import * as _ from 'lodash-es';
-import { getDuration } from '../details-page/datetime';
 
 export const units = {};
 export const validate = {};
@@ -256,7 +255,7 @@ export const humanizePercentage = (value) => {
 
 export const humanizeMinutes = (seconds, customUnitUnit) => {
   const val = Number(seconds);
-  const { minutes } = getDuration(val * 1000);
+  const minutes = Math.floor(val / 60);
   const unit = customUnitUnit || 'min';
   return {
     string: `${minutes} ${unit}`,
@@ -267,7 +266,7 @@ export const humanizeMinutes = (seconds, customUnitUnit) => {
 
 export const humanizeHours = (seconds, customUnitUnit) => {
   const val = Number(seconds);
-  const { hours } = getDuration(val * 1000);
+  const hours = Math.floor(val / (60 * 60));
   const unit = customUnitUnit || 'hr';
   return {
     string: `${hours} ${unit}`,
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
Before fix:
Screenshot from 2023-04-18 18-45-21

After fix:
Screenshot from 2023-04-18 19-50-23

packages/shared/details-page/datetime.ts
```diff
@@ -105,11 +105,8 @@ export const getDuration = (ms: number) => {
   }
   let seconds = Math.floor(ms / 1000);
   let minutes = Math.floor(seconds / 60);
-  seconds = seconds % 60;
   let hours = Math.floor(minutes / 60);
-  minutes = minutes % 60;
   const days = Math.floor(hours / 24);
-  hours = hours % 24;
```
SanjalKatiyar, line 112: I think there is something we are missing, check: https://github.com/openshift/console/blob/master/frontend/public/components/utils/datetime.ts#L65
console is using the same for like last 2 years, I am curious why no one ever found any bug before, this `getDuration` is used in `fromNow` which itself is used at multiple places...

Though change u have made above make sense, but maybe this `getDuration` is meant for something else and we are using is incorrectly here...
we need to verify it once...

GowthamShanmugam: Right now, getDuration is not used anywhere except in the DR dashboard. I think the old implementation is wrong, 
https://github.com/openshift/console/blob/98523601dd716b8b479e3b0112eaaeda83bf39ee/frontend/packages/pipelines-plugin/src/components/pipelines/pipeline-metrics/pipeline-metrics-utils.ts#L149

from the console what i observed is, from minutes they are using mod to find how many secs, But here mod is used seconds against seconds.

SanjalKatiyar: `from the console what i observed is, from minutes they are using mod to find how many secs, But here mod is used seconds against seconds.` >> yeah I saw 2 implementation in console as well, but they are still using https://github.com/openshift/console/blob/master/frontend/public/components/utils/datetime.ts#L65 at some places...

SanjalKatiyar: which makes me wonder maybe this is for diff use case (which we need to check once)...

GowthamShanmugam: mm agreed, it is better i create new function for MCO

packages/shared/utils/humanize.js
```diff
@@ -256,7 +255,7 @@ export const humanizePercentage = (value) => {
 
 export const humanizeMinutes = (seconds, customUnitUnit) => {
```
SanjalKatiyar, line 256:
```suggestion
export const humanizeMinutes = (seconds, customUnit) => {
```

PR #807, opened 2023-04-20
[WIP] Test redux-store for custom storagecluster name

SanjalKatiyar:


plugins/odf/console-extensions.json
```diff
@@ -671,5 +671,20 @@
...
+      "reducer": { "$codeRef": "storageClusterReducer" }
+    }
+  },
+  {
+    "type": "console.flag/hookProvider",
+    "properties": {
+      "handler": {
+        "$codeRef": "storageClusterReducer.useSetStorageClusterName"
+      }
+    }
```
SanjalKatiyar, line 688: for the first time we can dispatch the StorageClusterName while creating a StorageCluster from the wizard flow, but this extension here will be helpful for the cases when user reloads the browser...
on reload redux will reset and any state dispatched from wizard flow will reset as well, in that case this extension will ensure to dispatch the StorageClusterName value back to redux...

packages/odf/redux-store-reducer/reducer.ts
```diff
@@ -0,0 +1,58 @@
...
+  // console.log("useSetStorageClusterName");
+  // console.log("useSetStorageClusterName");
+  // console.log("useSetStorageClusterName");
+
+  React.useEffect(() => {
+    scNameDetector()
+      .then((res) => {
+        dispatch(res);
+      })
+      .catch((e) => {
```
SanjalKatiyar, line 54: we can use k8sWatch as well...

packages/ocs/dashboards/persistent-internal/status-card/status-card.tsx
```diff
@@ -191,6 +197,7 @@ export const StatusCard: React.FC = () => {
         </Gallery>
       </HealthBody>
       <CephAlerts />
+      <button onClick={() => dispatch('test-name')}>TEST DISPATCH</button>
```
SanjalKatiyar, line 200: ignore this file... it is only for testing...

packages/odf/redux-store-reducer/reducer.ts
SanjalKatiyar: or k8sList...

PR #814, opened 2023-04-21
[WIP] Test context-provider for custom storagecluster name

SanjalKatiyar:


bipuladh:
The use case of showing `StorageCluster` name is pretty much limited. And operations that require storagelcuster name are limited (adding capacity, creating storage system). 
Also as StorageCluster is more of an internal detail. The main interface for users is StorageSystem so maybe we should look at what are the effects of allowing users to set `StorageSystem`. Also we would have to check if either of the operators have hardcoded the storagesystem or storagecluster name. 
/hold

SanjalKatiyar:
`The use case of showing StorageCluster name is pretty much limited. And operations that require storagelcuster name are limited (adding capacity, creating storage system)` >> other than these places, **currently** we use it in StorageClass form and one of dashboard query as well.

`Also as StorageCluster is more of an internal detail. The main interface for users is StorageSystem so maybe we should look at what are the effects of allowing users to set StorageSystem` >> ack !
(this PR was only to quickly test out the extensions which might be needed for this BZ).

`Also we would have to check if either of the operators have hardcoded the storagesystem or storagecluster name.` >> OCS don;t have hardcoded dependency on the StorageCluster name, though it is hardcoded at some places in ODF operator.
Checking this comment: https://bugzilla.redhat.com/show_bug.cgi?id=1840613#c11 seems like we were okay with supporting the ask in the BZ (and in that case this would have been an epic for UI and backend both). But, I will confirm once on the BZ itself.

Thanks for the suggestion.

SanjalKatiyar:
> The use case of showing `StorageCluster` name is pretty much limited. And operations that require storagelcuster name are limited (adding capacity, creating storage system). Also as StorageCluster is more of an internal detail. The main interface for users is StorageSystem so maybe we should look at what are the effects of allowing users to set `StorageSystem`. Also we would have to check if either of the operators have hardcoded the storagesystem or storagecluster name. /hold

got confirmation from @iamniting, even ODF operator don;t have any dependency on hardcoded name !

iamniting:
> got confirmation from @iamniting, even ODF operator don;t have any dependency on hardcoded name !

Yes that is correct we do not have any dependency on hardcoded name
But while providing the functionality to the user we should not allow them to create lengthy names. As we do use the storagecluster names to generate the other objects names and if names get too lengthy k8s does not allow us to create those objects. We have seen that issue even in the external mode. Because the name is a bit longer in the external mode.

PR #815, opened 2023-04-21
CI: runner image: update Node.js version to LTS

alfonsomthd:
To be merged after: https://github.com/openshift/release/pull/38662

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,4 +1,4 @@
 build_root_image:
   name: odf-console-ci-runner
   namespace: ci
-  tag: latest
+  tag: node18
--- a/Dockerfile.ci.runner
+++ b/Dockerfile.ci.runner
@@ -1,8 +1,30 @@
-FROM quay.io/coreos/tectonic-console-builder:v24
-ENV NODE_VERSION="v14.17.3"
-RUN cd /tmp \
-  && wget --quiet -O /tmp/node.tar.gz http://nodejs.org/dist/${NODE_VERSION}/node-${NODE_VERSION}-linux-x64.tar.gz \
-  && tar xf node.tar.gz && rm -f /tmp/node.tar.gz && cd node-* && cp -r lib/node_modules /usr/local/lib/node_modules \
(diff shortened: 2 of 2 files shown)


.ci-operator.yaml
```diff
@@ -1,4 +1,4 @@
 build_root_image:
   name: odf-console-ci-runner
   namespace: ci
-  tag: latest
+  tag: beta
```
SanjalKatiyar, line 4: just commenting so that we remember to do it and don;t miss by mistake: change it to "current" once CI passes...

alfonsomthd: It's already documented in our README that `beta` tag is for testing purposes, as per our conversation the plan is to replace 'beta' with `node18` tag.

SanjalKatiyar: yeah, I commented here so that we don;t tag the PR by mistake unless this is reverted back...
"as per our conversation the plan is to replace 'beta' with node18 tag." >> this is another PR right ? or r u planning to do it in this PR only ?

Dockerfile.ci.runner
```diff
@@ -1,8 +1,22 @@
...
+FROM cypress/base:18.12.1
+
+RUN apt update && apt install -yq curl gnupg2 jq && rm -rf /var/lib/apt/lists/*
+
+RUN curl -LSs https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
+    && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" > /etc/apt/sources.list.d/google-chrome.list' \
+    && apt update && apt install -yq --no-install-recommends google-chrome-stable \
+    && rm -rf /var/lib/apt/lists/*
+
+ARG OC_VERSION="4.13.0-rc.4"
```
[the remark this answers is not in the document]
SanjalKatiyar: looks good otherwise...

.ci-operator.yaml
alfonsomthd: @SanjalKatiyar I updated the tag in this PR (thta is to be backported) and I'll add the github action cron schedule in a follow-up PR to avoid conflicts.

PR #817, opened 2023-04-24
tests: dashboard status card components

debjyoti-pandit:
part of - https://issues.redhat.com/browse/RHSTOR-4488

fixes - https://issues.redhat.com/browse/RHSTOR-4494

### DIFF
--- a/packages/shared/src/dashboards/status-card/HealthItem.spec.tsx
+++ b/packages/shared/src/dashboards/status-card/HealthItem.spec.tsx
@@ -1,0 +1,267 @@
+import * as React from 'react';
+import { HealthState } from '@openshift-console/dynamic-plugin-sdk';
+import { render, screen } from '@testing-library/react';
+import HealthItem from './HealthItem';
+
+const classNames = jest.fn();
+jest.mock('classnames', () => (...props: string[]) => {
+  classNames(props);
+  return props[1] ? props[0] + ' ' + props[1] : props[0];
+});
+
+const mockSecondaryStatusComponent = jest.fn((_unused) => {
+  return <div data-test-id="mocked-secondary-status" />;
+});
+jest.mock(
+  '@odf/shared/status/SecondaryStatus',
+  () => (props) => mockSecondaryStatusComponent(props)
+);
+
+const mockPopOverComponent = jest.fn((_unused) => {
+  return <div data-test-id="mocked-pop-over" />;
+});
+
+const mockButtonComponent = jest.fn((_unused) => {
+  return <div data-test-id="mocked-button" />;
+});
+
+jest.mock('@patternfly/react-core', () => {
+  const originalModule = jest.requireActual('@patternfly/react-core');
+  return {
+    ...originalModule,
+    Popover: (props) => mockPopOverComponent(props),
+    Button: (props) => mockButtonComponent(props),
+  };
+});
+
+jest.mock('./states', () => ({
+  healthStateMapping: {
+    [HealthState.OK]: {
+      icon: 'unit-test-ok-icon',
+    },
+    [HealthState.UNKNOWN]: {
+      icon: 'unit-test-unknown-icon',
+    },
+  },
+  healthStateMessage: () => 'mocked-health-state-message',
+}));
+
+const title = 'unit-tests';
+
+describe('tests for document elements of health item component', () => {
+  it('should render only the title when in loading', () => {
+    render(<HealthItem title={title} state={HealthState.LOADING} />);
+    expect(screen.getByText(`${title}`)).toBeInTheDocument();
+  });
+  it('should not render popover when in loading', () => {
+    const { container } = render(
+      <HealthItem title={title} state={HealthState.LOADING} />
+    );
+    expect(
+      container.querySelector('[data-test-id="mocked-pop-over"]')
+    ).toBeNull();
+    expect(mockPopOverComponent).not.toHaveBeenCalled();
+  });
+  it('should not render the secondary status when in loading', () => {
+    const { container } = render(
+      <HealthItem title={title} state={HealthState.LOADING} />
+    );
+    expect(
+      container.querySelector('[data-test-id="mocked-secondary-status"]')
+    ).toBeNull();
+  });
+  it('should not render the icon status if no icon set to true', () => {
+    const { container } = render(
+      <HealthItem title={title} noIcon={true} state={HealthState.OK} />
+    );
+    expect(
+      container.querySelector(`[data-test="${title}-health-item-icon"]`)
+    ).toBeNull();
+  });
+  it('should render the status icon if no icon is left default as false and no icon is passed', () => {
+    const { container } = render(
+      <HealthItem title={title} state={HealthState.OK} />
+    );
+    expect(
+      container.querySelector(`[data-test="${title}-health-item-icon"]`)
+    ).toHaveTextContent('unit-test-ok-icon');
+  });
+  it('should render the default unknown icon if no icon is left default as false and no icon and state is passed', () => {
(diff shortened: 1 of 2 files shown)


packages/shared/dashboards/status-card/health-item.spec.tsx
```diff
@@ -0,0 +1,141 @@
...
+import HealthItem from './HealthItem';
+
+jest.mock('@odf/shared/status/SecondaryStatus', () => () => (
+  <div data-test-id="secondary-status" />
+));
+
+const popOverComponent = jest.fn();
+jest.mock('@patternfly/react-core', () => ({
+  Popover: (props) => {
+    popOverComponent(props);
```
alfonsomthd, line 13: Not needed.
```suggestion
```

packages/shared/dashboards/status-card/health-item.spec.tsx
```diff
@@ -0,0 +1,141 @@
...
+  it('should render the passed icon if no icon is left default as false and a icon node is passed', () => {
+    const { container } = render(
+      <HealthItem
+        title={title}
+        icon={<div data-test="passed-health-item-icon" />}
+      />
+    );
+    expect(
+      container.querySelector(`[data-test="passed-health-item-icon"]`)
+    ).not.toBeNull();
```
alfonsomthd, line 102:
```suggestion
    ).toBeTruthy();
```

packages/shared/dashboards/status-card/health-item.spec.tsx
```diff
@@ -0,0 +1,141 @@
...
+}));
+
+const title = 'unit-tests';
+
+describe('tests for document elements of health item component', () => {
+  it('should render and have a span saying loading title status when in loading', () => {
+    render(<HealthItem title={title} state={HealthState.LOADING} />);
+    expect(mockedFunction).toHaveBeenCalledWith('Loading {{title}} status', {
+      title,
+    });
```
alfonsomthd, line 48: IMHO this is testing the translation hook so we can place this test in the translation hook test suite. Instead, waht about testing that the component has been rendered?
```suggestion
    expect(
      container.querySelector(`[data-item-id="${title}-health-item"]`)
    ).toBeTruthy();
```

packages/shared/dashboards/status-card/health-item.spec.tsx
```diff
@@ -0,0 +1,141 @@
...
+  },
+  healthStateMessage: () => 'mocked-health-state-message',
+}));
+
+const mockedFunction = jest.fn().mockImplementation(() => undefined);
+jest.mock('@odf/shared/useCustomTranslationHook', () => ({
+  useCustomTranslation: () => {
+    return { t: mockedFunction };
+  },
+}));
```
alfonsomthd, line 39: IMHO this code is for testing the translation hook so it could be placed in `useCustomTranslationHook.spec.ts` new file.

packages/shared/dashboards/status-card/health-item.spec.tsx
```diff
@@ -0,0 +1,141 @@
+import * as React from 'react';
```
alfonsomthd, line 1: Just an observation:
for consistency/alignment the test file should be named like the original (only adding `.spec`): `HealthItem.spec.tsx`.
At least this is the pattern followed in `openshift/console.`

packages/shared/dashboards/status-card/health-item.spec.tsx
debjyoti-pandit: This is needed if we are going to check the param passed on to the mocked component (popover in this case)

packages/shared/dashboards/status-card/health-item.spec.tsx
debjyoti-pandit: here, I am just mocking and so that I can check the params passed on here from this file

packages/shared/dashboards/status-card/health-item.spec.tsx
alfonsomthd: It's meaningless (here you're not testing anything: run the tests without it and they'll pass).

packages/shared/dashboards/status-card/health-item.spec.tsx
alfonsomthd: Here you're testing what is passed to the translation hook which is an internal piece of the component and that is what [we should avoid](https://testing-library.com/docs/#what-you-should-avoid-with-testing-library).
What we have to test is the correct rendering/behavior of the component.

alfonsomthd: Please see this [suggestion](https://github.com/red-hat-storage/odf-console/pull/817#discussion_r1177759756).

packages/shared/dashboards/status-card/health-item.spec.tsx
alfonsomthd: > This is needed if we are going to check the param passed on to the mocked component (popover in this case)

AFAIK `Popover` is a child component and [testing child components is discouraged](https://testing-library.com/docs/#what-you-should-avoid-with-testing-library) (we can write dedicated unit tests for that component).

The unit tests for a component should focus (set the scope) on the behavior of that component.

alfonsomthd: > > This is needed if we are going to check the param passed on to the mocked component (popover in this case)
> 
> AFAIK `Popover` is a child component and [testing child components is discouraged](https://testing-library.com/docs/#what-you-should-avoid-with-testing-library) (we can write dedicated unit tests for that component).
> 
> The unit tests for a component should focus (set the scope) on the behavior of that component.

I misunderstood you: if your only intention is to test what is passed to the child (but not to `test` the child itself), then it's OK.

packages/shared/dashboards/status-card/health-item.spec.tsx
alfonsomthd: > here, I am just mocking and so that I can check the params passed on here from this file

There is no need to mock the translation hook as we can directly check what teh real implementation returns.

packages/shared/dashboards/status-card/health-item.spec.tsx
debjyoti-pandit: If it is ok then can I resolve this comment?

packages/shared/dashboards/status-card/states.spec.tsx
```diff
@@ -0,0 +1,293 @@
...
+  it('should return translated unknown when state is unknown', () => {
+    const result = healthStateMessage(HealthState.UNKNOWN, mockedTFunction);
+    expect(mockedTFunction).toHaveBeenCalledWith('Unknown');
+    expect(result).toBe('Unknown');
+  });
+  it('should return translated pending when state is progressing', () => {
+    const result = healthStateMessage(HealthState.PROGRESS, mockedTFunction);
+    expect(mockedTFunction).toHaveBeenCalledWith('Pending');
+    expect(result).toBe('Pending');
+  });
```
alfonsomthd, line 85: There is no need to mock the translation hook and duplictae the exptaction by cheking also the mock: we shouldn't test the internals.

SanjalKatiyar: let;s use "toHaveBeenCalledTimes(number)" instead of "toHaveBeenCalledWith"...

SanjalKatiyar: ok...

packages/shared/dashboards/status-card/health-item.spec.tsx
alfonsomthd: Sure.

alfonsomthd:
@bipuladh @SanjalKatiyar this needs approval.

PR #827, opened 2023-05-02
tests: create namespace store

TimothyAsirJeyasing:


packages/odf/components/namespace-store/create-namespace-store.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+  useK8sWatchResource: jest.fn(),
+}));
+
+const NamespaceStoreForm = jest.fn();
+jest.mock('./namespace-store-form', () => (props) => {
+  NamespaceStoreForm(props);
+  return <div className="nb-endpoints-page-form__short" />;
+});
+
+it('checkHeaderText', () => {
```
alfonsomthd, line 27: Can you please make the test description mor meaningful, like `shows the correct heading texts`

packages/odf/components/namespace-store/create-namespace-store.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const { container } = render(<CreateNamespaceStore match={{ params }} />);
+  const heading = container.getElementsByClassName(
+    'odf-create-operand__header-text'
+  )[0];
+  expect(heading).toBeInTheDocument();
+  expect(heading).toHaveTextContent(/Create NamespaceStore/);
+});
+
+it('checkTitleText', () => {
```
alfonsomthd, line 36: It seems this text only adda an additonal expectation over the previous so IMHO these text heading expectations cna libe in the same test.

packages/odf/components/namespace-store/create-namespace-store.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+        },
+      },
+    ],
+    true,
+    undefined,
+  ],
+}));
+
+describe('CreateNamespaceStore test', () => {
+  it('shows the correct heading texts', () => {
```
debjyoti-pandit, line 56: Breaking down unit tests into separate tests for different parts is considered a beneficial approach rather than consolidating all the checks in one unit test. By doing so, it becomes easier to locate where the test is failing and subsequently fix the code. An ideal unit test should allow developers or testing teams to identify the problem by just reading the title or description.

debjyoti-pandit: cc @alfonsomthd

packages/odf/components/namespace-store/create-namespace-store.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+    expect(titleHeading).toBeInTheDocument();
+    expect(titleHeading).toHaveTextContent(
+      /Represents an underlying storage to be used as read or write target for the data in the namespace buckets\./
+    );
+    expect(nsStoreFormClass).toBeInTheDocument();
+    expect(screen.getByText('Cancel')).toBeInTheDocument();
+    expect(screen.getByText('Namespace store name')).toBeInTheDocument();
+    expect(screen.getByText('Provider')).toBeInTheDocument();
+    fireEvent.click(screen.getByRole('button', { name: 'Cancel' }));
+    expect(historyMock.location.pathname).toBe('/test');
```
debjyoti-pandit, line 76: My understanding is that the current check is focused on the cancel button's functionality. However, I suggest that we redirect our attention to verifying whether the NamespaceStoreForm is invoked with the appropriate redirect handler. By doing so, we can ensure that the redirect handler's ability to find the last index is also being tested. Additionally, I recommend mocking the referenceForModel to better isolate the behavior under test.

packages/odf/components/namespace-store/create-namespace-store.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+}));
+
+describe('CreateNamespaceStore test', () => {
+  it('shows the correct heading texts', () => {
+    const { container } = render(<CreateNamespaceStore match={{ params }} />);
+    const heading = container.getElementsByClassName(
+      'odf-create-operand__header-text'
+    )[0];
+    const titleHeading = container.getElementsByClassName('help-block')[0];
+    const nsStoreFormClass = container.getElementsByClassName(
```
debjyoti-pandit, line 62: Wouldn't it be beneficial to use a mocked version of the NamespaceStoreForm?

cc @alfonsomthd

packages/odf/components/namespace-store/create-namespace-store.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+  ],
+}));
+
+describe('CreateNamespaceStore test', () => {
+  it('shows the correct heading texts', () => {
+    const { container } = render(<CreateNamespaceStore match={{ params }} />);
+    const heading = container.getElementsByClassName(
+      'odf-create-operand__header-text'
+    )[0];
+    const titleHeading = container.getElementsByClassName('help-block')[0];
```
debjyoti-pandit, line 61: We can also verify the size and heading level of the Title component that is being passed from this component.

packages/odf/components/namespace-store/create-namespace-store.spec.tsx
```diff
@@ -0,0 +1,78 @@
+import * as React from 'react';
+import { fireEvent, render, screen } from '@testing-library/react';
+import CreateNamespaceStore from './create-namespace-store';
+
+const params = {
+  ns: '',
```
debjyoti-pandit, line 6: Can we include a test to confirm that the component's default namespace parameter is set to CEPH_STORAGE_NAMESPACE?

packages/odf/components/namespace-store/create-namespace-store.spec.tsx
alfonsomthd: > Breaking down unit tests into separate tests for different parts is considered a beneficial approach rather than consolidating all the checks in one unit test. By doing so, it becomes easier to locate where the test is failing and subsequently fix the code. An ideal unit test should allow developers or testing teams to identify the problem by just reading the title or description.

I agree (except when the code being tested is very small). In this particular case it seems the test logic keeps growing so it makes sense to split in order to understand the use case (scenario) being tested.

packages/odf/components/namespace-store/create-namespace-store.spec.tsx
alfonsomthd: > Additionally, I recommend mocking the referenceForModel to better isolate the behavior under test.
> 
> cc @alfonsomthd

@debjyoti-pandit we should only mock when necessary: right now (current implementation) the mock is not making any difference: the test is passing without it, so I guess you're suggesting this in order to test the props' values passed to the child component, right?

packages/odf/components/namespace-store/create-namespace-store.spec.tsx
alfonsomthd: @debjyoti-pandit If we want to increase the coverage by testing the props' values passed to it (the child component) then yes.

packages/odf/components/namespace-store/create-namespace-store.spec.tsx
debjyoti-pandit: Yes, true. the props values we are passing to the child component.

packages/odf/components/namespace-store/create-namespace-store.spec.tsx
```diff
@@ -0,0 +1,96 @@
...
+    expect(nsStoreFormClass).toBeInTheDocument();
+  });
+
+  it('Check onCancel and history', () => {
+    render(<CreateNamespaceStore match={{ params }} />);
+    fireEvent.click(screen.getByRole('button', { name: 'Cancel' }));
+    expect(historyMock.location.pathname).toBe('/testpath');
+  });
+
+  it('test get name', () => {
```
alfonsomthd, line 91: @TimothyAsirJeyasing this test (and some other above) does not belong here as we should avoid testing the child components.
Here we should test the `CreateNamespaceStore` behavior and make sure that the child component receives the namespace we expect it (but not testing the child itself). So I suggest:

1. Apply this patch where the component-related tests are kept:
```
diff --git a/packages/odf/components/namespace-store/create-namespace-store.spec.tsx b/packages/odf/components/namespace-store/create-namespace-store.spec.tsx
index 5f4f091..67c0259 100644
--- a/packages/odf/components/namespace-store/create-namespace-store.spec.tsx
+++ b/packages/odf/components/namespace-store/create-namespace-store.spec.tsx
@@ -1,55 +1,18 @@
-import * as selectors from '@odf/shared/selectors';
-import { fireEvent, render, screen } from '@testing-library/react';
+import { CEPH_STORAGE_NAMESPACE } from '@odf/shared/constants';
+import { render } from '@testing-library/react';
 import CreateNamespaceStore from './create-namespace-store';
 
 const params = {
-  ns: 'anyvalue',
+  ns: 'test-ns',
 };
 
-jest.mock('@odf/shared/selectors', () => ({
-  getName: jest.fn().mockReturnValue('test'),
-}));
-
-jest.mock('@openshift-console/dynamic-plugin-sdk', () => {
-  return {
-    ...jest.requireActual('@openshift-console/dynamic-plugin-sdk'),
-    useK8sWatchResource: jest.fn().mockReturnValue([[], false, false]),
-  };
+const mockNamespaceStoreForm = jest.fn();
+jest.mock('./namespace-store-form', () => (props) => {
+  mockNamespaceStoreForm(props);
+  return <mock-NamespaceStoreForm />;
 });
 
-const historyMock = {
-  location: {
-    pathname: '',
-  },
-  goBack: () => {
-    historyMock.location.pathname = '/testpath';
-  },
-  //push: () => jest.fn(),
-};
-
-jest.mock('react-router', () => {
-  return {
-    ...jest.requireActual('react-router'),
-    useHistory: jest.fn().mockImplementation(() => historyMock),
-  };
-});
-
-jest.mock('@odf/shared/hooks/useK8sList', () => ({
-  __esModule: true,
-  useK8sList: () => [
-    [
-      {
-        metadata: {
-          name: 'test',
-        },
-      },
-    ],
-    true,
-    undefined,
-  ],
-}));
-
 describe('CreateNamespaceStore test', () => {
   it('shows the correct heading texts', () => {
     const { container } = render(<CreateNamespaceStore match={{ params }} />);
@@ -74,23 +37,21 @@ describe('CreateNamespaceStore test', () => {
     ).toBeTruthy();
   });
 
-  it('Check namespace store form class found', () => {
-    const { container } = render(<CreateNamespaceStore match={{ params }} />);
-    const nsStoreFormClass = container.getElementsByClassName(
-      'nb-endpoints-page-form__short'
-    )[0];
-    expect(nsStoreFormClass).toBeInTheDocument();
-  });
-
-  it('Check onCancel and history', () => {
+  it('pass the received namespace to the form', () => {
     render(<CreateNamespaceStore match={{ params }} />);
-    fireEvent.click(screen.getByRole('button', { name: 'Cancel' }));
-    expect(historyMock.location.pathname).toBe('/testpath');
+    expect(mockNamespaceStoreForm).toHaveBeenCalledWith(
+      expect.objectContaining({
+        namespace: 'test-ns',
+      })
+    );
   });
 
-  it('test get name', () => {
-    render(<CreateNamespaceStore match={{ params }} />);
-    const getNameSpy = jest.spyOn(selectors, 'getName');
-    expect(getNameSpy).toHaveBeenCalledWith({ metadata: { name: 'test' } });
+  it('pass the default namespace to the form when no namespace is received', () => {
+    render(<CreateNamespaceStore match={{ params: {} }} />);
+    expect(mockNamespaceStoreForm).toHaveBeenCalledWith(
+      expect.objectContaining({
+        namespace: CEPH_STORAGE_NAMESPACE,
+      })
+    );
   });
 });
```
2. Move the `NamespaceStoreForm` related tests that you implemented (the ones deleted on the patch) to the right file:
```namespace-store-form.spec.tsx```

alfonsomthd (changes requested):

alfonsomthd:
Closed in favor of https://github.com/red-hat-storage/odf-console/pull/1086

PR #830, opened 2023-05-02
Fix graph issue - WIP

GowthamShanmugam:


GowthamShanmugam:
Before fix:
image


After fix:
Screenshot from 2023-05-03 00-30-36

packages/shared/charts/utils.ts
```diff
@@ -97,92 +95,85 @@ export type MapLimitsRequest = {
...
-        data: {
-          stroke: requestedColor.value,
-          strokeDasharray: '3,3',
-          fillOpacity: 0,
-        },
-      });
-    }
-  }
-
-  if (threshold) {
```
GowthamShanmugam, line 163: Limit is threshold, no need this extra code

packages/shared/charts/utils.ts
```diff
@@ -97,92 +95,85 @@ export type MapLimitsRequest = {
...
-    description,
-    null,
+    utilizationDescription || t('usage'),
+    utilizationSymbol || null,
     xMutator
   );
   const data = utilizationData ? [...utilizationData] : [];
-  const chartStyle = [null];
+  const chartStyle =
+    utilizationData?.map(() => utilizationChartStyle || null) || [];
```
GowthamShanmugam, line 135: when metrics are getting a new label, utilizationData will have more than one array. Need to equalize the chartStyle, otherwise, it will wrongly take limit chartStyle

packages/shared/charts/utils.ts
```diff
@@ -97,92 +95,85 @@ export type MapLimitsRequest = {
...
       xMutator
     );
-    data.push(...limitData);
     if (limitData.length) {
-      chartStyle.push({
-        data: { strokeDasharray: '3,3', fillOpacity: 0 },
-      });
+      data.push(...limitData);
+      limitData.forEach(() =>
+        chartStyle.push(
```
GowthamShanmugam, line 146: same here, limitData can have more than one array, and need to equalize its chartdata

packages/shared/dashboards/utilization-card/area-chart.tsx
```diff
@@ -72,11 +72,19 @@ export const AreaChart: React.FC<AreaChartProps> = ({
     [humanize, unit, formatDate]
   );
 
-  const legendData = processedData.map((d) => ({
-    childName: d[0].description,
-    name: d[0].description,
-    symbol: d[0].symbol,
-  }));
+  const legendData = React.useMemo(() => {
+    const legendDataWithDuplicate = processedData.map((d) => ({
```
GowthamShanmugam, line 76: when there is more than one list for the same metrics, the same labels will be duplicated here, and it creates confusion in the graph.

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
```diff
@@ -190,11 +194,12 @@ export const ReplicationHistorySection: React.FC<ReplicationHistorySectionProps>
...
 
     const placementInfo = selectedAppSet?.placementInfo?.[0];
-    const [threshold, initialUnit] = convertSyncIntervalToSeconds(
+    const [_, initialUnit] = convertSyncIntervalToSeconds(
       placementInfo?.syncInterval
     );
     const drpcName = placementInfo?.drpcName;
     const drpcNamespace = placementInfo?.drpcNamespace;
+    const drPolicyName = placementInfo?.drPolicyName;
     const [pvcsSLARangeData, pvcsSLARangeError, pvcsSLARangeLoading] =
```
SanjalKatiyar, line 203: nits: we are using "SLA" here in variable naming...

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx
```diff
@@ -206,14 +211,31 @@ export const ReplicationHistorySection: React.FC<ReplicationHistorySectionProps>
...
+      scheduledIntervalRangeLoading,
+    ] = useCustomPrometheusPoll({
+      endpoint: 'api/v1/query_range' as any,
+      query: !!drPolicyName
+        ? getScheduledIntervalDRPolicyQuery(drPolicyName)
+        : null,
+      timespan: duration,
+      basePath: ACM_ENDPOINT,
+      cluster: HUB_CLUSTER_NAME,
+    });
```
SanjalKatiyar, line 226: why not directly from the CR ??

syncInterval will change if older DRPolicy is removed and a new DRPolicy (with new syncInterval) is applied on the same Application...  so this threshold line should not always be linear but can vary... is that why  ??

PR #836, opened 2023-05-04
BZ-2122521: Disable and block double registration of the same storage IP for IBM FS

alonfirestein:


bipuladh (commented):
Could you please squash the commits?

SanjalKatiyar:
all the changes as per this BZ were supposed to be a part of `/packages/ibm/`, that was the original intention for creating a separate package... 
I can see we made changes to `/packages/odf/components/create-storage-system/create-storage-system-steps/create-storage-class-step/create-storage-class-step.tsx` file instead, which is reasonable because we need access to `Yup schema` and `/packages/ibm/` doesn;t have that yet.

Nothing major is wrong with PR, but we have 2 options:
1. We fix our extension point and make refactors for adds functionality to solve above problem, this PR will be on hold till then (we have good amount of time and we should be able to get it merged in 4.14, even if we keep it on hold for a short while).
2. We merge this PR so that BZ can be closed, and do the refactoring later on our end.

cc @bipuladh wdyt we should do ?

bvered:
@SanjalKatiyar - I see some tests has failed. 
Can you please instruct what is failing and how to fix it?

SanjalKatiyar:
> @SanjalKatiyar - I see some tests has failed. Can you please instruct what is failing and how to fix it? @alonfirestein - fyi

@bvered @alonfirestein plz follow up on this PR: https://github.com/red-hat-storage/odf-console/pull/838 for further discussion... sorry for any inconvenience...

SanjalKatiyar:
> @SanjalKatiyar - I see some tests has failed. Can you please instruct what is failing and how to fix it? @alonfirestein - fyi

plz run:
yarn lint-fix
yarn format
yarn build

if we get any error we need to fix and commit the changes...
also, please keep single commit per PR (just convention)...

PR #842, opened 2023-05-08
Fix for the issue DRPolicy doesn't show connected application

GowthamShanmugam:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2196298

From 4.13, I have added an extra check on placementRef namespace. The problem is 4.12 UI is not adding this namespace info on DRPC placementRef, so UI is ignoring this from the connected application list.


bipuladh (commented):
/lgtm

PR #852, opened 2023-05-18
fix status card for nfs

SanjalKatiyar:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2208029
Screenshot from 2023-05-18 21-59-58
discussion details can be found here: https://docs.google.com/document/d/1ZQN_1XlBm6EN12ViA-ktSR6Y8Ttb5aW52mk-SMiyxDE/edit?disco=AAAAxObzgrM

### DIFF
--- a/packages/ocs/dashboards/network-file-system/status-card/status-card.tsx
+++ b/packages/ocs/dashboards/network-file-system/status-card/status-card.tsx
@@ -1,48 +1,43 @@
 import * as React from 'react';
+import { CEPH_STORAGE_NAMESPACE } from '@odf/shared/constants';
 import HealthItem from '@odf/shared/dashboards/status-card/HealthItem';
-import {
-  useCustomPrometheusPoll,
-  usePrometheusBasePath,
-} from '@odf/shared/hooks/custom-prometheus-poll';
+import { PodModel } from '@odf/shared/models';
+import { PodKind } from '@odf/shared/types';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
 import {
-  HealthState,
-  SubsystemHealth,
+  referenceForModel,
+  getPodStatus,
+  getPodHealthState,
+} from '@odf/shared/utils';
+import {
+  useK8sWatchResource,
+  WatchK8sResource,
 } from '@openshift-console/dynamic-plugin-sdk';
 import { Card, CardBody, CardHeader, CardTitle } from '@patternfly/react-core';
-import { NFS_STATUS_QUERY, NFSDashboardQuery } from '../../../queries';
 
-export const getServerHealthState = (
-  state: string,
-  loading,
-  loadError
-): SubsystemHealth => {
-  if (loading) {
-    return { state: HealthState.LOADING };
-  }
-  if (loadError) {
-    return { state: HealthState.NOT_AVAILABLE };
-  }
-  if (!!state && state !== '0') {
-    return { state: HealthState.OK };
-  }
-  return { state: HealthState.ERROR };
+const nfsPodResource: WatchK8sResource = {
+  isList: true,
+  kind: referenceForModel(PodModel),
+  namespace: CEPH_STORAGE_NAMESPACE,
+  selector: {
+    matchLabels: {
+      app: 'rook-ceph-nfs',
+      ceph_nfs: 'ocs-storagecluster-cephnfs',
+    },
+  },
 };
 
 export const StatusCard: React.FC = () => {
   const { t } = useCustomTranslation();
-  const [workerThreads, workerThreadsError, workerThreadsLoading] =
-    useCustomPrometheusPoll({
-      endpoint: 'api/v1/query' as any,
-      query: NFS_STATUS_QUERY[NFSDashboardQuery.WORKER_THREADS_TOTAL],
-      basePath: usePrometheusBasePath(),
-    });
+  const [nfsPods, nfsPodsLoaded, nfsPodsLoadError] =
+    useK8sWatchResource<PodKind[]>(nfsPodResource);
 
-  const serverHealthState = getServerHealthState(
-    workerThreads?.data?.result[0]?.value[1],
-    workerThreadsLoading,
-    workerThreadsError
+  const serverHealthState = getPodHealthState(
+    getPodStatus(nfsPods?.[0]),
+    !nfsPodsLoaded,
+    nfsPodsLoadError
   );
+
   return (
     <Card data-test="nfs-status-card">
       <CardHeader>
--- a/packages/shared/types/k8s.ts
+++ b/packages/shared/types/k8s.ts
@@ -127,7 +127,13 @@ enum ImagePullPolicy {
 }
 
 // https://github.com/kubernetes/api/blob/release-1.16/core/v1/types.go#L2411-L2432
-type PodPhase = 'Pending' | 'Running' | 'Succeeded' | 'Failed' | 'Unknown';
+export enum PodPhase {
+  Pending = 'Pending',
+  Failed = 'Failed',
+  Running = 'Running',
+  Succeeded = 'Succeeded',
+  Unknown = 'Unknown',
+}
 
 type VolumeMount = {
   mountPath: string;
@@ -281,7 +287,7 @@ type PodCondition = {
 } & K8sResourceCondition;
 
 type PodStatus = {
-  phase: PodPhase;
+  phase: keyof typeof PodPhase;
   conditions?: PodCondition[];
   message?: string;
   reason?: string;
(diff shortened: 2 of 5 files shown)


packages/ocs/dashboards/network-file-system/status-card/status-card.tsx
```diff
@@ -1,48 +1,40 @@
...
-    return { state: HealthState.NOT_AVAILABLE };
-  }
-  if (!!state && state !== '0') {
-    return { state: HealthState.OK };
-  }
-  return { state: HealthState.ERROR };
+const nfsPodResource: WatchK8sResource = {
+  isList: true,
+  kind: referenceForModel(PodModel),
+  namespace: 'openshift-storage',
```
GowthamShanmugam, line 18: shall replace 'openshift-storage' with some constant?

packages/ocs/dashboards/network-file-system/status-card/status-card.tsx
```diff
@@ -1,48 +1,40 @@
...
+  selector: {
+    matchLabels: {
+      app: 'rook-ceph-nfs',
+      ceph_nfs: 'ocs-storagecluster-cephnfs',
+    },
```
GowthamShanmugam, line 23: same here, lets use constants, if it is specific to this file, then let keep the constant on top of this file

packages/ocs/dashboards/network-file-system/status-card/status-card.tsx
SanjalKatiyar: ack

packages/ocs/dashboards/network-file-system/status-card/status-card.tsx
SanjalKatiyar: ack

packages/shared/utils/pod-health.tsx
```diff
@@ -0,0 +1,34 @@
+import {
+  HealthState,
+  SubsystemHealth,
+} from '@openshift-console/dynamic-plugin-sdk';
+
+// containers have not been started or have terminated
+const POD_PHASE_DEGRADED = ['Pending', 'Failed'];
```
bipuladh, line 7: Please use enums for Phase values.

packages/shared/types/k8s.ts
```diff
@@ -126,7 +126,7 @@ enum ImagePullPolicy {
   IfNotPresent = 'IfNotPresent',
 }
 
-// https://github.com/kubernetes/api/blob/release-1.16/core/v1/types.go#L2411-L2432
+// https://github.com/kubernetes/api/blob/release-1.27/core/v1/types.go#LL2733C1-L2751C2
```
bipuladh, line 129: Unless the values have updates there lets keep it as it is.

packages/ocs/dashboards/network-file-system/status-card/status-card.tsx
```diff
@@ -1,48 +1,40 @@
...
-    });
+  const [nfsPod, nfsPodLoaded, nfsPodLoadError] =
+    useK8sWatchResource<PodKind>(nfsPodResource);
 
-  const serverHealthState = getServerHealthState(
-    workerThreads?.data?.result[0]?.value[1],
-    workerThreadsLoading,
-    workerThreadsError
+  const serverHealthState = getRunningPodHealthState(
+    getOperandStatus(nfsPod?.[0])?.value,
```
bipuladh, line 33: Although it might work for this case here. It's a bad practice to use functions that are meant for a different type. Please create another function to pull the status of PodKind.

bipuladh (commented):
Please update the commit message with proper details of the fix. Also add the BZ number to the commit message.

packages/ocs/dashboards/network-file-system/status-card/status-card.tsx
SanjalKatiyar: I think it is fine, we won;t be using this key-value any where else...

packages/shared/utils/pod-health.tsx
```diff
@@ -0,0 +1,41 @@
+import { PodKind } from '@odf/shared/types';
```
bipuladh, line 1: I don't see any `JSX` here we can keep the file as `ts`. Please add this file to `index.ts` file under utils.

bipuladh (commented):
A few comments

packages/ocs/dashboards/network-file-system/status-card/status-card.tsx
```diff
@@ -1,48 +1,43 @@
...
 export const StatusCard: React.FC = () => {
   const { t } = useCustomTranslation();
-  const [workerThreads, workerThreadsError, workerThreadsLoading] =
-    useCustomPrometheusPoll({
-      endpoint: 'api/v1/query' as any,
-      query: NFS_STATUS_QUERY[NFSDashboardQuery.WORKER_THREADS_TOTAL],
-      basePath: usePrometheusBasePath(),
-    });
+  const [nfsPod, nfsPodLoaded, nfsPodLoadError] =
+    useK8sWatchResource<PodKind>(nfsPodResource);
```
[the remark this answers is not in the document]
bipuladh: nit: Also `nfsPods`

packages/shared/types/k8s.ts
```diff
@@ -281,7 +287,7 @@ type PodCondition = {
 } & K8sResourceCondition;
 
 type PodStatus = {
-  phase: PodPhase;
+  phase: keyof typeof PodPhase;
```
bipuladh, line 290:
```suggestion
  phase: PodPhase;
```
Wouldn't this work?

packages/shared/utils/pod-health.ts
```diff
@@ -0,0 +1,31 @@
+import { PodPhase, PodKind } from '@odf/shared/types';
+import {
+  HealthState,
+  SubsystemHealth,
+} from '@openshift-console/dynamic-plugin-sdk';
+
+export const getPodStatus = (obj: PodKind): PodPhase =>
+  (obj?.status?.phase as PodPhase) || PodPhase.Unknown;
+
+export const getRunningPodHealthState = (
```
bipuladh, line 10:
```suggestion
export const getPodHealthState = (
```
Would this be okay?

bipuladh (commented):
Few more comments.

packages/shared/types/k8s.ts
SanjalKatiyar: it will work, but that will make type of pod's status.phase as an "enum" (PodPhase)... where as right now it is still string, which is analogous to what Pod CR's reports...

packages/shared/types/k8s.ts
```diff
@@ -127,7 +127,13 @@ enum ImagePullPolicy {
 }
 
 // https://github.com/kubernetes/api/blob/release-1.16/core/v1/types.go#L2411-L2432
-type PodPhase = 'Pending' | 'Running' | 'Succeeded' | 'Failed' | 'Unknown';
+export enum PodPhase {
+  Pending = 'Pending', // containers have not been started
+  Failed = 'Failed', // containers have terminated
+  Running = 'Running', // containers have been started
+  Succeeded = 'Succeeded', // containers have voluntarily terminated
+  Unknown = 'Unknown', // state of the pod could not be obtained
```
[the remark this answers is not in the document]
SanjalKatiyar: I can remove it if u want, but I don;t see any issue as comments gets removed anyway in production build... so up to you, if u want I can remove ??

packages/shared/utils/pod-health.ts
SanjalKatiyar: that was intentional on my end... because if pod's status is "Succeeded", I am still reporting it as "error" of the dashboard...

SanjalKatiyar: normally "Succeeded" is healthy as well, but not in this case... that is why I named it "RunningPodHealth"...

packages/shared/types/k8s.ts
SanjalKatiyar: Pod CR reports us string not an enum...

packages/shared/types/k8s.ts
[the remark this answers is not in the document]
bipuladh: lets remove them. I think these terms are well known and don't require much explanation.

packages/shared/utils/pod-health.ts
SanjalKatiyar: it is not NFS specific IMO... it depends on usecase, for example: let's say we want to show ODF health status based on the odf-operator pod (not the CSV or any metric), there as well we will reuse this "getRunningPodHealthState" because these pods are never supposed to be "Succeeded"... in fact most of the pods don;t...

only few cases like osd creation pods, they are regarded as healthy even if they have "Succeeded" and we have running osds...

SanjalKatiyar: as per discussion, removing "Succeeded" from "error" and adding under healthy...

PR #859, opened 2023-05-22
Fixes storageCluster status when Alerts are firing

bipuladh:
Screenshot from 2023-05-22 14-04-16

### DIFF
--- a/packages/shared/models/common.ts
+++ b/packages/shared/models/common.ts
@@ -248,3 +248,16 @@ export const DaemonSetModel: K8sKind = {
   labelPlural: 'DaemonSets',
   labelPluralKey: 'DaemonSets',
 };
+
+export const StorageClusterModel: K8sKind = {
+  label: 'Storage Cluster',
+  labelPlural: 'Storage Clusters',
+  apiVersion: 'v1',
+  apiGroup: 'ocs.openshift.io',
+  plural: 'storageclusters',
+  abbr: 'OCS',
+  namespaced: true,
+  kind: 'StorageCluster',
+  id: 'ocscluster',
+  crd: true,
+};
--- a/packages/shared/topology/utils/AlertFilters.ts
+++ b/packages/shared/topology/utils/AlertFilters.ts
@@ -1,6 +1,7 @@
 import { Alert, K8sModel } from '@openshift-console/dynamic-plugin-sdk';
 import * as _ from 'lodash-es';
-import { NodeModel } from '../../models';
+import { NodeModel, StorageClusterModel } from '../../models';
+import { cephFilter, noobaaFilter, rgwFilter } from '../../utils';
 
 export const commonFilter =
   (name: string) =>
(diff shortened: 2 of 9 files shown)


packages/shared/models/common.ts
```diff
@@ -248,3 +248,16 @@ export const DaemonSetModel: K8sKind = {
...
+  labelPlural: 'Storage Clusters',
+  apiVersion: 'v1',
+  apiGroup: 'ocs.openshift.io',
+  plural: 'storageclusters',
+  abbr: 'OCS',
+  namespaced: true,
+  kind: 'StorageCluster',
+  id: 'ocscluster',
+  crd: true,
+};
```
SanjalKatiyar, line 263: we have "OCSStorageClusterModel" in "shared" and "StorageClusterModel" in "ocs"...

packages/shared/topology/utils/AlertFilters.ts
```diff
@@ -10,6 +10,29 @@ export const commonFilter =
...
+  _.get(alert, 'annotations.storage_type') === 'NooBaa';
+
+export const filterNooBaaAlerts = (alerts: Alert[]): Alert[] =>
+  alerts?.filter(noobaaFilter);
+
+const rgwFilter = (alert: Alert) => alert?.annotations?.storage_type === 'RGW';
+
+export const filterRGWAlerts = (alerts: Alert[]): Alert[] =>
+  alerts?.filter(rgwFilter);
+
```
SanjalKatiyar, line 35: we should remove these from "ocs/utils/metrics.ts" and let "ocs" import these from "shared" as well..

SanjalKatiyar: if it is shared between "topology" and "ocs", we can keep them in general purpose utils...

packages/shared/models/common.ts
bipuladh: Yeah, looks like all these models belong in `shared`. The reason being our codebase is completely ODF centric and all our packages at some point will need these models. I could make the changes in this PR but that would make a huge backport PR so I will create an issue to track this.

packages/shared/topology/utils/AlertFilters.ts
```diff
@@ -23,6 +46,10 @@ export const filterRelevantAlerts = (
   if (_.isEqual(resourceModel, NodeModel)) {
     filter = nodeFilter(resourceName);
   }
+  if (_.isEqual(StorageClusterModel, resourceModel)) {
+    filter = (alert: Alert) =>
+      rgwFilter(alert) || noobaaFilter(alert) || cephFilter(alert);
```
SanjalKatiyar, line 51: no need for this here ??
```suggestion
    filter = (alert: Alert) =>
      rgwFilter(alert) || noobaaFilter(alert) || cephFilter(alert) || commonFilter(alert);
```

SanjalKatiyar, line 51: no need for " || commonFilter" here ??

SanjalKatiyar:
@bipuladh PR needs rebase...

SanjalKatiyar:
@bipuladh we need manual backport...

PR #860, opened 2023-05-23
tests: for inventory and icons in shared status folder

debjyoti-pandit:


packages/shared/status/Inventory.spec.tsx
```diff
@@ -0,0 +1,132 @@
...
+jest.mock('./Node', () => ({
+  nodeStatus: (resource) => resource.status,
+}));
+
+describe('getStatusGroups', () => {
+  const resources = [
+    { id: 1, status: 'Not Ready' },
+    { id: 2, status: 'Ready' },
+    { id: 3, status: 'UNKNOWN' },
+    { id: 4, status: 'Not Ready' },
```
bipuladh, line 18: Don't we have `enums` for these Statuses that we can use here?

packages/shared/status/Inventory.spec.tsx
```diff
@@ -0,0 +1,132 @@
...
+  getPVStatusGroups,
+  getStatusGroups,
+  InventoryStatusGroup,
+} from './Inventory';
+
+jest.mock('./Node', () => ({
+  nodeStatus: (resource) => resource.status,
+}));
+
+describe('getStatusGroups', () => {
```
bipuladh, line 13: Please write better descriptions

packages/shared/status/Inventory.spec.tsx
debjyoti-pandit: We have enums, but these are the test values that will get mapped to the proper enums which we have.

Also, ready and not ready like the below pics written in text getting mapped to the enums which are present.

Pic 1 - 
image

Pic 2 -
image

PR #861, opened 2023-05-23
Fix cant initiate failover of applicationSet based apps after hub recovery

GowthamShanmugam:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2209288

Fix explanation:
  1. Earlier, Failover/Relocate component is using placement ownerRef to match placement and its placement decision.  It is failing in some cases. So Switching to label-based matching.
  
Other errors observed from QE setup once after fixing the above one:
  2. Failover/Relocation readiness text in UI should not rely only on DRPC peer-ready conditions and available conditions. It is common for all errors, So if any error occurs the readiness should display  "Not Ready".
  3. While Failing/Relocating ramen is removing the cluster name from the placement rule, and adding it back once the failover/relocation operation is completed.  In between if the cluster went down, and if a user wants to trigger Failover, Then UI is unable to find out the target cluster and deployment cluster.
  The fix is, Find the deployment cluster name from DRPC, which is only possible at the time of failover and relocation operation is ongoing.
  - If the status is in the FailingOver state then the target cluster can be found using failoverCluster spec from DRPC.
  - If the status is in the Relocating state then the target cluster can be found using preferedCluster spec from DRPC.
  - If a target cluster is found then finding a deployment cluster is possible using the list of DR Clusters.

### DIFF
--- a/packages/mco/components/modals/app-failover-relocate/failover-relocate-modal.tsx
+++ b/packages/mco/components/modals/app-failover-relocate/failover-relocate-modal.tsx
@@ -102,6 +102,7 @@ export const FailoverRelocateModal: React.FC<FailoverRelocateModalProps> = (
       <ModalBody>
         <FailoverRelocateModalBody
           {...props}
+          canInitiate={canInitiate}
           setCanInitiate={setCanInitiate}
           setPlacement={setPlacement}
         />
--- a/packages/mco/utils/disaster-recovery.tsx
+++ b/packages/mco/utils/disaster-recovery.tsx
@@ -54,6 +54,7 @@ import {
   DRVolumeReplicationGroupKind,
   ProtectedPVCData,
   ProtectedAppSetsMap,
+  ACMPlacementDecisionKind,
 } from '../types';
 
 export type PlacementRuleMap = {
@@ -244,13 +245,13 @@ export const checkDRActionReadiness = (
 
 export const findCluster = (
   clusters: K8sResourceCommon[],
-  deploymentClusterName: string,
-  isDeploymentCluster = false
+  matchClusterName: string,
+  matchCluster = false
 ) =>
   clusters?.find((cluster) =>
-    isDeploymentCluster
-      ? getName(cluster) === deploymentClusterName
-      : getName(cluster) !== deploymentClusterName
+    matchCluster
+      ? getName(cluster) === matchClusterName
+      : getName(cluster) !== matchClusterName
   );
 
 export const findDRType = (drClusters: DRClusterKind[]) =>
@@ -488,3 +489,28 @@ export const filterPVCDataUsingAppsets = (
 
 export const filterDRAlerts = (alert: Alert) =>
   alert?.annotations?.alert_type === 'DisasterRecovery';
+
+export const findDeploymentClusterName = (
+  plsDecision: ACMPlacementDecisionKind,
+  drpc: DRPlacementControlKind,
+  drClusters: DRClusterKind[]
+): string => {
+  let deploymentClusterName =
+    plsDecision?.status?.decisions?.[0]?.clusterName || '';
+  // During failover/relocating ramen removing cluser name from placement decision
+  if (!deploymentClusterName) {
+    const currStatus: DRPC_STATUS = drpc?.status?.phase as DRPC_STATUS;
+    if (currStatus === DRPC_STATUS.Relocating) {
+      // While relocating, preferredCluster spec is the target cluster
+      // Find deployment cluster from drclusters using target cluster
+      const cluster = findCluster(drClusters, drpc?.spec?.preferredCluster);
+      deploymentClusterName = getName(cluster);
+    } else if (currStatus === DRPC_STATUS.FailingOver) {
+      // While failover, failoverCluster spec is the target cluster
+      // Find deployment cluster from drclusters using target cluster
+      const cluster = findCluster(drClusters, drpc?.spec?.failoverCluster);
+      deploymentClusterName = getName(cluster);
+    }
+  }
+  return deploymentClusterName;
+};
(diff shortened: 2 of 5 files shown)


packages/mco/utils/disaster-recovery.tsx
```diff
@@ -488,3 +489,36 @@ export const filterPVCDataUsingAppsets = (
...
+    if (currStatus === DRPC_STATUS.Relocating) {
+      const cluster = findCluster(
+        drClusters,
+        drpc?.spec?.preferredCluster,
+        false
+      );
+      deploymentClusterName = getName(cluster);
+      // While failing over failoverCluster cluster is the target cluster
+      // Find deployment cluster using dr cluster list
+    } else if (currStatus === DRPC_STATUS.FailingOver) {
```
SanjalKatiyar, line 514: nits
```suggestion
    } else if (currStatus === DRPC_STATUS.FailingOver) {
          // While failing over failoverCluster cluster is the target cluster
      // Find deployment cluster using dr cluster list
```

packages/mco/utils/disaster-recovery.tsx
```diff
@@ -488,3 +489,36 @@ export const filterPVCDataUsingAppsets = (
...
+    plsDecision?.status?.decisions?.[0]?.clusterName || '';
+  if (!deploymentClusterName) {
+    const currStatus: DRPC_STATUS = drpc?.status?.phase as DRPC_STATUS;
+    // While relocating preferredCluster is the target cluster
+    // Find deployment cluster using dr cluster list
+    if (currStatus === DRPC_STATUS.Relocating) {
+      const cluster = findCluster(
+        drClusters,
+        drpc?.spec?.preferredCluster,
+        false
```
SanjalKatiyar, line 509: this is confusing:
1. "matchCluster" argument for the "findCluster" function is false by default, so no need to pass "false" again.
2. second argument for the "findCluster" function is named as "deploymentClusterName", but in reality u want to figure out the deployment cluster, parameter that u are passing is target cluster naming convention is wrong/confusing).

packages/mco/utils/disaster-recovery.tsx
```diff
@@ -488,3 +489,36 @@ export const filterPVCDataUsingAppsets = (
 
 export const filterDRAlerts = (alert: Alert) =>
   alert?.annotations?.alert_type === 'DisasterRecovery';
+
+export const findDeploymentClusterName = (
+  plsDecision: ACMPlacementDecisionKind,
```
SanjalKatiyar, line 494: no issue with current approach, but if "drpc" is the best way to determine the deployment cluster, do we even need to have check for "plsDecision" ?? Can't we always check based upon "drpc" only ??

packages/mco/components/modals/app-failover-relocate/failover-relocate-modal.tsx
```diff
@@ -53,7 +53,7 @@ export const FailoverRelocateModal: React.FC<FailoverRelocateModalProps> = (
     ModalFooterStatus.INITIAL
   );
   const [placement, setPlacement] = React.useState<PlacementProps>({});
-  const [canInitiate, setCanInitiate] = React.useState(false);
+  const [canInitiate, setCanInitiate] = React.useState<Boolean>(undefined);
```
SanjalKatiyar, line 56: nits: if we are defining the type to be Boolean, then shouldn't; we initialise it with false ??

GowthamShanmugam: The required UX flow is, to keep Failover/Relocate readiness text as unknown, And display Ready/NotReady only when all pre-checks are evaluated. 

Instead of canInitiate, it is better to use the errorMessage field to decide this.  Let's not make the boolean field undefined unnecessarily.

packages/mco/utils/disaster-recovery.tsx
GowthamShanmugam: - PlacementDecision is our source of truth to find out the deployment cluster.
- Finding the deployment cluster name from DRPC won't work always, User may accidently change the "Prefered cluster / Failover cluster" on the DRPC spec without changing the spec "Action". 
- Also, if we find from DRPC, these are the following challenges:
    1) If the current DRPC status is Deployed, then the deployment cluster is a preferred cluster spec.
    2) If the current DRPC status is Initiating, then:
         If the Action is Failover, then the deploy deployment cluster is another cluster from the failover cluster.
         If the Action is Relocate, then the deploy deployment cluster is another cluster from the preferred cluster.
    3) If the current DRPC status is FailingOver, then the deployment cluster is another cluster from the failover cluster.
    4) If the current DRPC status is Relocating, then the deployment cluster is another cluster from the preferred cluster.
    5)  If the current DRPC status is FailedOver, then the deployment cluster is the preferred cluster.
    6) If the current DRPC status is Relocated, then the deployment cluster is the preferred cluster. 
- If any new state comes then we need to handle that also, If the user changes the preferred cluster/failover cluster spec alone then it's a problem. 
- Instead of checking these many conditions, I decided to check placementRule. Which always gives me the correct deployment cluster name. 
- But one corner case is happening here, ramen is removing the cluster names from placementRule, I decided to check DRPC only for this corner case, (i.e. Relocating/Failing Over). I am assuming that Relocating/FailingOver users won't change the DRPC spec. 
- Other alternate solution is to ask user to input cluster name from UI, But if we introduce bulk failover and relocate in future this will be a problem.

GowthamShanmugam: This fix will at least save us from raising blocker BZ-like failover/relocation is blocked (Or) like UI is not displaying target cluster name on the UI.

packages/mco/utils/disaster-recovery.tsx
GowthamShanmugam: correct, the name is wrong.

PR #893, opened 2023-06-07
Bump core sdk + update "odf.dashboard/tab" extension

SanjalKatiyar:
Changes added:
1. Bumps core sdk to latest 0.0.19, to use types for exposed "NamespaceBar" FC.
2.  Updates "odf.dashboard/tab" to a generic "odf.horizontalNav/tab" extension point, which will support adding a new tab to any odf specific page based upon filtering on the "contextId" field.

### DIFF
--- a/packages/odf/components/odf-dashboard/dashboard.tsx
+++ b/packages/odf/components/odf-dashboard/dashboard.tsx
@@ -1,8 +1,8 @@
 import * as React from 'react';
 import {
-  DashboardTab,
-  DashboardTabExtensionProps as UnresolvedTabProps,
-  isDashboardTab,
+  HorizontalNavTab,
+  HorizontalNavTabExtensionProps as UnresolvedTabProps,
+  isHorizontalNavTab,
 } from '@odf/odf-plugin-sdk/extensions';
 import PageHeading from '@odf/shared/heading/page-heading';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
@@ -25,6 +25,8 @@ import { StatusCard } from './status-card/status-card';
 import SystemCapacityCard from './system-capacity-card/capacity-card';
 import { convertDashboardTabToNav, sortPages } from './utils';
 import './dashboard.scss';
+
+const CONTEXT_ID = 'odf-dashboard';
 
 type ODFDashboardPageProps = {
   history: RouteComponentProps['history'];
@@ -71,17 +73,21 @@ const ODFDashboardPage: React.FC<ODFDashboardPageProps> = (props) => {
       href: '',
       name: t('Overview'),
       component: ODFDashboard,
+      contextId: CONTEXT_ID,
     },
     {
       id: 'systems',
       href: 'systems',
       name: t('Storage Systems'),
       component: StorageSystemListPage,
+      contextId: CONTEXT_ID,
     },
   ]);
 
+  const isTab = React.useMemo(() => isHorizontalNavTab(CONTEXT_ID), []);
+
   const [extensions, isLoaded, error] =
-    useResolvedExtensions<DashboardTab>(isDashboardTab);
+    useResolvedExtensions<HorizontalNavTab>(isTab);
 
   React.useEffect(() => {
     const updatedPages = [...pages];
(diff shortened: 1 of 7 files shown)


bipuladh (commented):
A question around the changes to the extensions. 
Why do we need to change the current existing Dashboard tab? 
Also what is going to use this updated `HorizontalNavTab`?

SanjalKatiyar:
> A question around the changes to the extensions. Why do we need to change the current existing Dashboard tab? Also what is going to use this updated `HorizontalNavTab`?

1. Currently "odf.dashboard/tab" (as name suggests) adds new tabs to odf dashboard. But now changed it to "odf.horizontalNav/tab" so that it can add tabs to any page where we will use horizontalNav. It will do so using a "contextId" that's all, minor change. Dashboard has its own unique contentId that's the only change to existing dashboard.

2. Now I will use same extension to add pages to new "Object Storage" tab as well. I will assign a new unique contextId for horizontalNav on Object Service page and filter tab specific to it using that contextID.

bipuladh:
> > A question around the changes to the extensions. Why do we need to change the current existing Dashboard tab? Also what is going to use this updated `HorizontalNavTab`?
> 
> 1. Currently "odf.dashboard/tab" (as name suggests) adds new tabs to odf dashboard. But now changed it to "odf.horizontalNav/tab" so that it can add tabs to any page where we will use horizontalNav. It will do so using a "contextId" that's all, minor change. Dashboard has its own unique contentId that's the only change to existing dashboard.
> 2. Now I will use same extension to add tabs to new "Object Storage" page as well. I will assign a new unique contextId for Object Service page and filter tab specific to it using that contextId.
> 
Why do we need to conditionally add tabs to Object Storage? The whole of it should be present based on the presence of Noobaa? 
> Each extension will have e.properties.contexId field, based upon on which page that extension needs to get added.

SanjalKatiyar:
> Why do we need to conditionally add tabs to Object Storage? The whole of it should be present based on the presence of Noobaa?

Not really:
BS/NS/BC pages are based on MCG presence.
OB/OBC pages are based on MCG or RGW presence.
Also, BS/NS/BC are to be displayed only for admin whereas OB/OBC are to be displayed for non-admins as well.

packages/odf/components/odf-dashboard/dashboard.tsx
```diff
@@ -26,6 +26,8 @@ import SystemCapacityCard from './system-capacity-card/capacity-card';
 import { convertDashboardTabToNav, sortPages } from './utils';
 import './dashboard.scss';
 
+const CONTEXT_ID = 'odf-dashboard';
```
bipuladh, line 29: ODF_DASHBOARD_CONTEXT

packages/odf/components/odf-dashboard/dashboard.tsx
```diff
@@ -71,17 +73,21 @@ const ODFDashboardPage: React.FC<ODFDashboardPageProps> = (props) => {
...
       component: StorageSystemListPage,
+      contextId: CONTEXT_ID,
     },
   ]);
 
+  const isTab = React.useMemo(() => isHorizontalNavTab(CONTEXT_ID), []);
+
   const [extensions, isLoaded, error] =
-    useResolvedExtensions<DashboardTab>(isDashboardTab);
+    useResolvedExtensions<HorizontalNavTab>(isTab);
```
bipuladh, line 90: compose(isTab, (ext) => ext.contextID === DASHBOARD_CONTEXT_ID);

bipuladh (commented):
Please send a follow up for this

packages/odf/components/odf-dashboard/dashboard.tsx
SanjalKatiyar: I don't think this will work:
lets call `isTab as "a"` and `(ext) => ext.contextID === DASHBOARD_CONTEXT_ID as "b"`

compose will return `a(b(...args))` >> meaning output of inner function as an input to outer function, and output of inner function will be boolean.

SanjalKatiyar: unless u r not talking about compose from "redux", rather some other function that will chain the result (passing same arguments to all the functions ??

packages/odf/components/odf-dashboard/dashboard.tsx
SanjalKatiyar: ack...

packages/odf/components/odf-dashboard/dashboard.tsx
SanjalKatiyar: checking...

PR #896, opened 2023-06-09
Add pvc+namespace based used-capacity queries

aruniiird:


bipuladh (approved):

packages/ocs/queries/ceph-storage.ts
```diff
@@ -187,6 +189,14 @@ export const utilizationPopoverQueryMap = [
...
 ];
 
+export const getPVCNamespaceQuery = (namespace: string) => {
+  const queries = {
+    [StorageDashboardQuery.PVC_NAMESPACES_BY_USED]: `sum by (namespace, persistentvolumeclaim) (kubelet_volume_stats_used_bytes{namespace='${namespace}'} * on (namespace, persistentvolumeclaim) group_left(storageclass, provisioner) (kube_persistentvolumeclaim_info * on (storageclass) group_left(provisioner) kube_storageclass_info {provisioner=~"(.*rbd.csi.ceph.com)|(.*cephfs.csi.ceph.com)|(ceph.rook.io/block)"}))`,
+    [StorageDashboardQuery.PVC_NAMESPACES_TOTAL_USED]: `sum(sum by (namespace, persistentvolumeclaim) (kubelet_volume_stats_used_bytes{namespace='${namespace}'} * on (namespace, persistentvolumeclaim) group_left(storageclass, provisioner) (kube_persistentvolumeclaim_info * on (storageclass) group_left(provisioner) kube_storageclass_info {provisioner=~"(.*rbd.csi.ceph.com)|(.*cephfs.csi.ceph.com)|(ceph.rook.io/block)"})))`,
+  };
+  return queries;
+};
+
```
SanjalKatiyar, line 199: @bipuladh @aruniiird wasn;t the ask of this epic to improve the metrics (from the epic description) ??
"kubelet_volume_stats_used_bytes" does not reports `snapshots, clones, or in case of using 
a reclaim policy of Retain` and users are confused by that.

We are adding new option to filter based on PVC, but underlying issue mentioned above is still not solved right ?? (or am I missing something)

SanjalKatiyar: discussed offline: that is what the current ask is as per the UX.

packages/ocs/queries/ceph-storage.ts
```diff
@@ -187,6 +189,14 @@ export const utilizationPopoverQueryMap = [
   },
 ];
 
+export const getPVCNamespaceQuery = (namespace: string) => {
```
debjyoti-pandit, line 192: @SanjalKatiyar @bipuladh 
Can we refactor this function and also the other const to something like this - https://github.com/debjyoti-pandit/odf-console/blob/d484d2bf6982804bccb0fffd3691585798fb4376/packages/ocs/queries/ceph-storage.ts#L128

Please have a look at the [change log](https://github.com/red-hat-storage/odf-console/pull/905/files#diff-ed343f84dc48e8ceb65c9dfa205c01c2c2ee5a9b8ac5b2cab9f49afdf0dc66a4) of my PR, I have added them temporarily as this PR is not yet merged.

debjyoti-pandit: Or should I incorporate these changes as it is in my PR and merge this PR with minimum changes?

debjyoti-pandit: I will rebase my PR on top of this.

packages/ocs/queries/ceph-storage.ts
```diff
@@ -187,6 +189,14 @@ export const utilizationPopoverQueryMap = [
...
+  const queries = {
+    [StorageDashboardQuery.PVC_NAMESPACES_BY_USED]: `sum by (namespace, persistentvolumeclaim) (kubelet_volume_stats_used_bytes{namespace='${namespace}'} * on (namespace, persistentvolumeclaim) group_left(storageclass, provisioner) (kube_persistentvolumeclaim_info * on (storageclass) group_left(provisioner) kube_storageclass_info {provisioner=~"(.*rbd.csi.ceph.com)|(.*cephfs.csi.ceph.com)|(ceph.rook.io/block)"}))`,
+    [StorageDashboardQuery.PVC_NAMESPACES_TOTAL_USED]: `sum(sum by (namespace, persistentvolumeclaim) (kubelet_volume_stats_used_bytes{namespace='${namespace}'} * on (namespace, persistentvolumeclaim) group_left(storageclass, provisioner) (kube_persistentvolumeclaim_info * on (storageclass) group_left(provisioner) kube_storageclass_info {provisioner=~"(.*rbd.csi.ceph.com)|(.*cephfs.csi.ceph.com)|(ceph.rook.io/block)"})))`,
```
aruniiird, line 195: @bipuladh , @SanjalKatiyar , in the above TWO queries (both are similar), we show only the PVCs which are backed by storageclasses provisioned by `RBD` or CephFS` or `CephBlock`. As the intention here is to show the details of all the mounted PVCs, shall we remove that condition and show all ?

SanjalKatiyar: I think it is okay to have it like this... it is ODF dashboard so make sense to show ODF specific PVCs...
other cards work same way as well !

PR #897, opened 2023-06-12
feat: added annotations for encrypted rbd storage class

debjyoti-pandit:
for: https://issues.redhat.com/browse/RHSTOR-4529


packages/odf/utils/mutators.ts
```diff
@@ -0,0 +1,14 @@
+import { StorageClass } from '@odf/shared/types';
+
+export const addKubevirtAnnotations = (sc: StorageClass): StorageClass => {
```
GowthamShanmugam, line 3: just a doubt, why this needs to be util? can we keep this under sc-form.tsx?

debjyoti-pandit: I kept it under utils, as it doesn't belong to any screen and it will be used as a utility function.

GowthamShanmugam: A common utils file is not a correct place until it is reused by other places. I suggested moving this file under the storage-class folder with filename mutators.ts

SanjalKatiyar: move under `/ocs/storage-class/`

bipuladh (commented):
/hold

debjyoti-pandit:
remove /hold

SanjalKatiyar:
/lgtm
cc @bipuladh for approval

PR #899, opened 2023-06-14
fix: cypress tests

debjyoti-pandit:


debjyoti-pandit:
fixed by: https://github.com/red-hat-storage/odf-console/pull/904

PR #900, opened 2023-06-15
Adds support for single replica pool

bipuladh:
Screenshot from 2023-06-15 14-29-27


debjyoti-pandit:
@bipuladh won't we need to show that we enabled single replica pool in the review and create step also?

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -111,18 +111,46 @@ type EnableTaintNodesProps = {
   enableTaint: WizardState['capacityAndNodes']['enableTaint'];
 };
 
+const EnableSingleReplicaPoolCheckbox: React.FC<EnableSingleReplicaPoolCheckboxProps> =
```
debjyoti-pandit, line 114: Since it says `EnableSingleReplicaPoolCheckbox`, do we need to add the checkbox suffix?

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -111,18 +111,46 @@ type EnableTaintNodesProps = {
...
+        description={t(
+          'Enabling this feature creates a single replica pool without data replication, increasing the risk of data loss, data corruption, and potential system instability.'
+        )}
+        id="single-replica-pool"
+        data-checked-state={enableSingleReplicaPool}
+        isChecked={enableSingleReplicaPool}
+        onChange={() =>
+          dispatch({
+            type: 'capacityAndNodes/enableSingleReplicaPool',
+            payload: !enableSingleReplicaPool,
```
GowthamShanmugam, line 129: name says enableSingleReplicaPool but we are adding !enableSingleReplicaPool? 
its confusing

bipuladh:
> @bipuladh all it does is to create an extra pool with 1-way replication (default pool with 3-way still gets created as part of normal deployment) ?? And creating StorageClass is upto users ??

Everything is handled by the ocs-operator. Users cannot create single replica pools by themselves.

SanjalKatiyar:
> > @bipuladh all it does is to create an extra pool with 1-way replication (default pool with 3-way still gets created as part of normal deployment) ?? And creating StorageClass is upto users ??
> 
> Everything is handled by the ocs-operator. Users cannot create single replica pools by themselves.

I meant from user point of view, once this checkbox is selected then: `all it does is to create an extra pool with 1-way replication (default pool with 3-way still gets created as part of normal deployment)`

bipuladh:
> > > @bipuladh all it does is to create an extra pool with 1-way replication (default pool with 3-way still gets created as part of normal deployment) ?? And creating StorageClass is upto users ??
> > 
> > 
> > Everything is handled by the ocs-operator. Users cannot create single replica pools by themselves.
> 
> I meant from user point of view, once this checkbox is selected then: `all it does is to create an extra pool with 1-way replication (default pool with 3-way still gets created as part of normal deployment)`

yes

PR #901, opened 2023-06-15
Adds dashboard for ocs client operator

bipuladh:
Screenshot from 2023-06-21 19-29-29


plugins/client/console-extensions.json
```diff
@@ -0,0 +1,20 @@
+[
+  {
+    "type": "console.navigation/href",
+    "properties": {
+      "id": "odfdashboard",
+      "insertBefore": "persistentvolumes",
+      "section": "storage",
+      "name": "Data Foundation",
```
SanjalKatiyar, line 8: we still want to call it "Data Foundation" (even though there is no ODF) or ODF is present ??

bipuladh: It is data foundation client operator.

package.json
```diff
@@ -12,6 +12,7 @@
     "clean": "yarn clean-mco && yarn clean-odf",
     "clean-mco": "cd plugins/mco && rm -rf ./dist",
     "clean-odf": "cd plugins/odf && rm -rf ./dist",
+    "clean-client": "cd plugins/client && rm -rf ./dist",
```
SanjalKatiyar, line 15: will it make sense to add all these "scripts" as a part of package's package.json file instead of root ??
(same for mco, we can refactor mco's later)

package.json
```diff
@@ -31,6 +33,7 @@
     "analyze-mco": "PLUGIN=mco yarn analyze-common && PLUGIN=mco yarn ts-node ./analyzeTest.ts",
     "dev": "PLUGIN=odf I8N_NS=plugin__odf-console yarn server:plugin",
     "dev-mco": "PLUGIN=mco I8N_NS=plugin__odf-multicluster-console yarn server:plugin",
+    "dev-client": "PLUGIN=client I8N_NS=plugin__odf-client yarn server:plugin",
```
SanjalKatiyar, line 36: We are always shipping a single combined locales files containing texts which are not even related to that particular plugin.
Currently locales has more than 1000 keys, but all are not needed by "client" plugin. Same is true for mco/odf.

We now have multiple packages, it might make sense to have locales per package.
cc @GowthamShanmugam @bipuladh wdyt ??

packages/client/src/components/dashboards/cards/details-card.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const { t } = useCustomTranslation();
+
+  const [csv, csvLoaded, csvError] = useFetchCsv({
+    specName: 'ocs-client-operator',
+    namespace: CEPH_STORAGE_NAMESPACE,
+  });
+
+  const subscriptionVersion = getOperatorVersion(csv);
+
+  const serviceName = t('Data Foundation');
```
SanjalKatiyar, line 23: no need to add "Client" to it ??

packages/client/src/components/dashboards/cards/status-card.tsx
```diff
@@ -0,0 +1,84 @@
...
+    case 'Initializing':
+    case 'Progressing':
+    case 'Onboarding':
+    case 'Offboarding':
+      return { state: HealthState.PROGRESS, message: phase };
+    case 'Connected':
+      return { state: HealthState.OK, message: phase };
+    case 'Failed':
+      return { state: HealthState.ERROR, message: phase };
+  }
```
SanjalKatiyar, line 54: no `default` case...

packages/client/src/components/dashboards/cards/status-card.tsx
```diff
@@ -0,0 +1,84 @@
...
+};
+
+export const StatusCard: React.FC = () => {
+  const { t } = useCustomTranslation();
+  const [data, loaded, loadError] =
+    useK8sWatchResource<StorageClient[]>(clientResource);
```
SanjalKatiyar, line 60: what does `StorageClient` CR represents ?? does it corresponds to a  single provider cluster ??

packages/client/src/components/dashboards/cards/status-card.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  const cephHealth = getStorageClientHealth(data[0], loaded, loadError);
```
SanjalKatiyar, line 62: will be always have single `StorageClient` CR, even in case of multiple providers ??
```suggestion
  const cephHealth = getStorageClientHealth(data?.[0], loaded, loadError);
```

packages/client/src/components/dashboards/client-dashboard.tsx
```diff
@@ -0,0 +1,49 @@
...
+import ActivityCard from '@odf/ocs/dashboards/persistent-internal/activity-card/activity-card';
+import InventoryCard from '@odf/ocs/dashboards/persistent-internal/inventory-card';
+import PageHeading from '@odf/shared/heading/page-heading';
+import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
+import Tabs from '@odf/shared/utils/Tabs';
+import ProviderClusterBreakdownCard from './cards/breakdown-card';
+import DetailsCard from './cards/details-card';
+import StatusCard from './cards/status-card';
+
+const PersistentDashboard: React.FC = () => {
```
SanjalKatiyar, line 13:
```suggestion
const PersistentExternalDashboard: React.FC = () => {
```
Or,
```suggestion
const ClientDashboard: React.FC = () => {
```

packages/client/src/models.ts
```diff
@@ -0,0 +1,13 @@
...
+export const StorageClientModel: K8sModel = {
+  apiGroup: 'ocs.openshift.io',
+  apiVersion: 'v1alpha1',
+  kind: 'StorageClient',
+  crd: true,
+  namespaced: false,
+  abbr: 'SC',
+  plural: 'storageclients',
+  label: 'Storage Client',
+  labelPlural: 'Storage CLients',
```
SanjalKatiyar, line 12:
```suggestion
  labelPlural: 'Storage Clients',
```

packages/client/src/components/dashboards/client-dashboard.tsx
```diff
@@ -0,0 +1,49 @@
+import * as React from 'react';
+import { CommonDashboardRenderer } from '@odf/ocs/dashboards/ocs-system-dashboard';
+import UtilizationCard from '@odf/ocs/dashboards/persistent-external/utilization-card';
+import ActivityCard from '@odf/ocs/dashboards/persistent-internal/activity-card/activity-card';
```
SanjalKatiyar, line 4: aren;t we polling for StorageCluster and OCS operator subscription in this Card ??

SanjalKatiyar:
@bipuladh I hope client operator is adding label to the openshift-storage NS as well: https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/features.ts#L237 then only UI will poll from odf-managed prom, else it will poll from OCP prom.

packages/client/src/components/dashboards/client-dashboard.tsx
```diff
@@ -0,0 +1,49 @@
...
+import InventoryCard from '@odf/ocs/dashboards/persistent-internal/inventory-card';
```
SanjalKatiyar, line 5: this shows "ceph" nodes, will nodes have same label in case of client operator as well ??

SanjalKatiyar: same for PV/PVCs...

package.json
bipuladh: I think it works like a hash-map. So it really doesn't matter IMO.

packages/client/src/components/dashboards/cards/details-card.tsx
bipuladh: We are doing minimal changes so no client required.

packages/client/src/components/dashboards/cards/status-card.tsx
bipuladh: Yes it represents a single provider cluster.

bipuladh:
> @bipuladh I hope client operator is adding label to the openshift-storage NS as well: https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/features.ts#L237 then only UI will poll from odf-managed prom, else it will poll from OCP prom.

We need to poll from OCP prom

packages/client/src/components/dashboards/client-dashboard.tsx
bipuladh: PV/PVCs are based on provisioners and we use the same provisioners. Nodes are not labelled so we can remove that field.

package.json
SanjalKatiyar: searching is not an issue... I meant from point of view of the file size !!

SanjalKatiyar: larger the json object, larger the file size in the bundle...

bipuladh: `asset locales/en/plugin__odf-client.json 92.5 KiB [emitted]` I think for now we are good. We can worry about it in future.

bipuladh:
```
assets by path locales/ 380 KiB
  asset locales/ja/plugin__odf-client.json 106 KiB [emitted] [from: locales/ja/plugin__odf-console.json] [copied]
  asset locales/ko/plugin__odf-client.json 97.2 KiB [emitted] [from: locales/ko/plugin__odf-console.json] [copied]
  asset locales/en/plugin__odf-client.json 92.5 KiB [emitted] [from: locales/en/plugin__odf-console.json] [copied]
  asset locales/zh/plugin__odf-client.json 84.3 KiB [emitted] [from: locales/zh/plugin__odf-console.json] [copied]
```

bipuladh: Also since it's all text content gzip should compress it with high efficiency.

SanjalKatiyar: ack !

packages/client/src/components/dashboards/cards/activity-card.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const ocsEventsFilter = React.useCallback(
+    () => isPersistentStorageEvent(memoizedPVCNames),
+    [memoizedPVCNames]
+  );
+
+  const eventObject = {
+    data: events,
+    loaded: eventsLoaded && pvcLoaded,
+    kind: 'Event',
+    loadError: null,
```
SanjalKatiyar, line 72: don;t want to send "loadError" ??

packages/client/src/components/dashboards/cards/details-card.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <CardHeader>
+        <CardTitle>{t('Details')}</CardTitle>
+      </CardHeader>
+      <CardBody>
+        <DetailsBody>
+          <DetailItem key="service_name" title={t('Service name')}>
+            {serviceName}
+          </DetailItem>
+          <DetailItem key="mode" title={t('Mode')}>
+            Client
```
SanjalKatiyar, line 35:
```suggestion
            t('Client')
```

SanjalKatiyar, line 35: nits:  to initialise it in variable and use that or wrap around `t` function (not using `t` function here seems intentional, so good on that)...

packages/client/src/components/dashboards/client-dashboard.tsx
```diff
@@ -0,0 +1,46 @@
...
+      href: '/',
+      title: t('Overview'),
+      component: PersistentClientDashboard,
+    },
+  ];
+
+  return (
+    <>
+      <PageHeading title={t('Data Foundation')} />
+      <Tabs id="odf-dashboard-tab" tabs={pages} basePath="overview" />
```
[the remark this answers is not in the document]
bipuladh: I could set the page path as overview or use basePath. I guess anyone could work. If we ever introduce another tab. We would then make `Overview` page as `/persistent` and so on.

PR #902, opened 2023-06-15
Bug 2216707: Fixes for topology view for external mode

bipuladh:
Screenshot from 2023-06-15 19-08-19

### DIFF
--- a/locales/en/plugin__odf-console.json
+++ b/locales/en/plugin__odf-console.json
@@ -857,6 +857,7 @@
   "Expand to fullscreen": "Expand to fullscreen",
   "Exit fullscreen": "Exit fullscreen",
   "Back to main view": "Back to main view",
+  "Topology view is not supported for External Mode": "Topology view is not supported for External Mode",
   "No StorageCluster found": "No StorageCluster found",
   "Set up a storage cluster to view the topology": "Set up a storage cluster to view the topology",
   "A minimal cluster deployment will be performed.": "A minimal cluster deployment will be performed.",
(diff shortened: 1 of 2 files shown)


locales/en/plugin__odf-console.json
```diff
@@ -843,7 +843,9 @@
   "Expand to fullscreen": "Expand to fullscreen",
   "Exit fullscreen": "Exit fullscreen",
   "Back to main view": "Back to main view",
+  "Topology view is not supported for External Mode": "Topology view is not supported for External Mode",
   "No StorageCluster found": "No StorageCluster found",
+  "Topology view is not supported for external mode": "Topology view is not supported for external mode",
```
SanjalKatiyar, line 848: same text twice ?? with "E" and "e" in external

packages/odf/components/topology/Topology.tsx
```diff
@@ -575,7 +579,13 @@ const Topology: React.FC = () => {
   );
 };
 
-const TopologyViewEmptyState: React.FC = () => {
+type TopologyViewEmptyStateProps = {
+  isExternalMode?: boolean;
+};
+
+const TopologyViewErrorMessage: React.FC<TopologyViewEmptyStateProps> = ({
```
debjyoti-pandit, line 586: shouldn't here the component name and the props type name be the same?

packages/odf/components/topology/Topology.tsx
```diff
@@ -597,23 +607,34 @@ const TopologyViewEmptyState: React.FC = () => {
...
+        {isExternalMode
+          ? t('Topology view is not supported for External Mode')
+          : t('No StorageCluster found')}
       </Title>
       <EmptyStateBody>
-        {t('Set up a storage cluster to view the topology')}
+        {isExternalMode
+          ? t('Topology view is not supported for external mode')
+          : t('Set up a storage cluster to view the topology')}
       </EmptyStateBody>
```
SanjalKatiyar, line 618: same title and description for external mode ??
also, "E" and "e" in external in both texts...

PR #903, opened 2023-06-19
Policy configuration view for application

GowthamShanmugam:
On top of https://github.com/red-hat-storage/odf-console/pull/894

PR #906, opened 2023-06-22
Assign policy wizard

GowthamShanmugam:
On top of: https://github.com/red-hat-storage/odf-console/pull/903


packages/shared/dropdown/multiselectdropdown.tsx
```diff
@@ -7,14 +7,12 @@ export const MultiSelectDropdown: React.FC<MultiSelectDropdownProps> = ({
   placeholder,
   id,
   options,
-  defaultSelected = [],
+  selected = [],
+  variant,
 }) => {
   const [isOpen, setOpen] = React.useState(false);
-  const [selected, setSelected] = React.useState<string[]>(
```
GowthamShanmugam, line 13: In the Assign policy modal, the use case is to clear all selected labels for the clear button.  So i want to control setSelected from outside of this component.

packages/shared/utils/NameValueEditor.tsx
```diff
@@ -190,6 +193,8 @@ export const NameValueEditor: React.FC<NameValueEditorProps> =
       toolTip,
       nameString,
       valueString,
+      extraProps,
+      PairElementComponent = PairElement,
```
GowthamShanmugam, line 197: I am passing my own custom component to change the NameValueEditor body, But functionality is re-used.

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/body.tsx
```diff
@@ -337,10 +338,6 @@ export const LocalVolumeSetBody: React.FC<LocalVolumeSetBodyProps> = ({
             onChange={(selectedValues: string[]) =>
               formHandler('deviceType', selectedValues)
             }
-            defaultSelected={[
```
GowthamShanmugam, line 340: Since I removed setSelected from multiselecteddropdown.tsx, it no longer has defaultSelected props. I am expecting the state variable "selected" should be pre-initialized with the default variable.

state.deviceType has default redux initialized value of this defaultSelected, no need to pass explicitly.

packages/mco/components/modals/app-manage-policies/utils/common.ts
```diff
@@ -0,0 +1,137 @@
+import { DRPC_STATUS } from '@odf/mco/constants';
+import { DisasterRecoveryFormatted } from '@odf/mco/hooks';
```
GowthamShanmugam, line 2: I can reuse this file for a subscription app also, so created separate util for that

GowthamShanmugam:
[Screencast from 2023-06-28 15-57-10.webm](https://github.com/red-hat-storage/odf-console/assets/15073590/6748a646-5b49-43fc-9a5c-5dd499fe1c0a)

packages/mco/hooks/argo-application-set.ts
```diff
@@ -202,12 +202,12 @@ export type ArgoApplicationSetFormattedKind = {
   placements: {
     placement: ACMPlacementKind;
     placementDecision: ACMPlacementDecisionKind;
-    drPlacementControl: DRPlacementControlKind;
-    drPolicy: DRPolicyKind;
-    drClusters: DRClusterKind[];
+    drPlacementControl?: DRPlacementControlKind;
+    drPolicy?: DRPolicyKind;
+    drClusters?: DRClusterKind[];
```
GowthamShanmugam, line 207: drpc resources are optional for unprotected apps

packages/shared/src/dropdown/multiselectdropdown.tsx
```diff
@@ -7,14 +7,13 @@ export const MultiSelectDropdown: React.FC<MultiSelectDropdownProps> = ({
   placeholder,
   id,
   options,
-  defaultSelected = [],
+  selected = [],
+  variant,
 }) => {
   const [isOpen, setOpen] = React.useState(false);
-  const [selected, setSelected] = React.useState<string[]>(
```
GowthamShanmugam, line 13: I have a use case like removing the selected Placement from outside of this component, 
image

locales/en/plugin__odf-console.json
```diff
@@ -214,7 +225,17 @@
   "Assigned on": "Assigned on",
   "View configurations": "View configurations",
   "No activity": "No activity",
-  "Back": "Back",
+  "Delete": "Delete",
+  "Select  a placement": "Select  a placement",
```
SanjalKatiyar, line 229:
```suggestion
  "Select a placement": "Select a placement",
```

packages/mco/components/modals/app-manage-policies/app-manage-policies-modal.tsx
```diff
@@ -14,11 +16,34 @@ import {
...
+  if (modalViewContext === ModalViewContext.ASSIGN_POLICY_VIEW) {
+    return {
+      title: t('Assign policy'),
+      description: t(
+        'Secure your application by assigning a policy from the available policy templates.'
+      ),
+    };
+  } else {
+    return {
+      title: t('Manage data policy'),
```
SanjalKatiyar, line 35: are we still calling it `Manage data policy` even though we only allow assign as of now ?

GowthamShanmugam: Ya, But Title and description needs to be decided, doc team is working on this

locales/en/plugin__odf-console.json
GowthamShanmugam: ack

packages/mco/components/modals/app-manage-policies/assign-policy-view.tsx
```diff
@@ -0,0 +1,206 @@
...
+  ) =>
+    dispatch({
+      type: ManagePolicyStateType.SET_PLACEMENT_CONTROLS,
+      context: ModalViewContext.ASSIGN_POLICY_VIEW,
+      payload: drPlacementControls,
+    });
+
+  const assignPolicy = () => {
+    setAssignDisabled(true);
+    const updateContext = async (
```
SanjalKatiyar, line 121:
```suggestion
    const updateContext = (
```

packages/mco/components/modals/app-manage-policies/helper/pvc-details-wizard-content.tsx
```diff
@@ -0,0 +1,358 @@
...
+const LABEL = 'label';
+const SPLIT_CHAR = '; ';
+
+const findPlacement = (placements: PlacementType[], name: string) =>
+  placements.find((placement) => getName(placement) === name);
+
+const getPlacementTags = (drpcs: DRPlacementControlType[]) =>
+  !!drpcs.length
+    ? drpcs.map(
+        (drpc) => [getName(drpc.placementInfo), drpc?.pvcSelector] || []
```
SanjalKatiyar, line 42: u will never reach this condition...
```suggestion
        (drpc) => [getName(drpc.placementInfo), drpc?.pvcSelector]
```

packages/mco/components/modals/app-manage-policies/helper/pvc-details-wizard-content.tsx
```diff
@@ -0,0 +1,358 @@
...
+      )
+    : [[]];
+
+const getLabelsFromSearchResult = (searchResult: SearchResult): string[] => {
+  const pvcLabels =
+    searchResult?.data.searchResult?.[0]?.items.reduce(
+      (acc, item) => [...acc, ...item?.[LABEL]?.split(SPLIT_CHAR)],
```
SanjalKatiyar, line 49: spread will give error when `item?.[LABEL]?.split(SPLIT_CHAR)` is `undefined`...

SanjalKatiyar, line 49: this should work ??
```suggestion
      (acc, item) => [...acc, ...item[LABEL]?.split(SPLIT_CHAR)],
```

packages/mco/components/modals/app-manage-policies/helper/pvc-details-wizard-content.tsx
```diff
@@ -0,0 +1,358 @@
...
+
+const getLabelsFromTags = (tags: TagsType, currIndex: number): string[] =>
+  tags.reduce((acc, tag, index) => {
+    const labels = tag?.[1] || [];
+    return currIndex !== index ? [...acc, ...labels] : acc;
+  }, []) as string[];
+
+const getPlacementsFromTags = (tags: TagsType, currIndex: number): string[] =>
+  tags.reduce((acc, tag, index) => {
+    const placement = tag?.[0];
```
SanjalKatiyar, line 63: this is not an array like labels ??

packages/mco/components/modals/app-manage-policies/helper/pvc-details-wizard-content.tsx
```diff
@@ -0,0 +1,358 @@
...
+});
+
+const NameValueEditorComponent = (props) => (
+  <AsyncLoader
+    loader={() =>
+      import('@odf/shared/utils/NameValueEditor').then((c) => c.NameValueEditor)
+    }
+    {...props}
+  />
+);
```
SanjalKatiyar, line 122: nits: better to export this as `LazyNameValueEditor` from `shared` and use that everywhere...

SanjalKatiyar: also do we even need lazy import ??

packages/mco/components/modals/app-manage-policies/utils/acm-utils.ts
```diff
@@ -0,0 +1,31 @@
...
+          {
+            property: 'namespace',
+            values: [workloadNamespace],
+          },
+          {
+            property: 'cluster',
+            values: clusterNames,
+          },
+        ],
+        limit: 20,
```
SanjalKatiyar, line 25: 1. do we need limit ??
2. in the `workloadNamespace`, can there be PVCs which are not related to this Application, how we decide whether a PVC is independent or belongs to an ACM Application ??

packages/mco/components/modals/app-manage-policies/utils/parser-utils.ts
```diff
@@ -99,3 +101,32 @@ export const generateApplicationInfo = (
   placements: plsInfo,
   dataPolicies: drPolicyInfo,
 });
+
+export const getClusterNamesFromPlacements = (placements: PlacementType[]) =>
+  placements?.reduce(
+    (acc, placemnt) => [...acc, ...placemnt?.deploymentClusters],
+    []
+  );
```
SanjalKatiyar, line 109: just to confirm, r u sure we already do not have this function defined somewhere ?? seems like something we have done at other places as well.

packages/mco/constants/acm.ts
```diff
@@ -0,0 +1,29 @@
...
+// "~1" is used to represent a "/", else any "patch" call will treat prefix as a path
+export const PROTECTED_APP_ANNOTATION_WO_SLASH =
+  'cluster.open-cluster-management.io~1experimental-scheduling-disable';
+
+// Placement label
+export const PLACEMENT_REF_LABEL =
+  'cluster.open-cluster-management.io/placement';
+
+export enum APPLICATION_TYPE {
+  APPSET = 'ApplicationSet',
```
SanjalKatiyar, line 20: we already have these constants right defined ?? why are we doing again in this file ??

packages/mco/components/modals/app-manage-policies/helper/pvc-details-wizard-content.tsx
```diff
@@ -0,0 +1,358 @@
...
+      [placementControInfo]
+    );
+
+    // Selected placement and labels
+    const [tags, setTags] = React.useState<TagsType>(
+      getPlacementTags(selectedPlacementControls)
+    );
+
+    // ACM search proxy api call
+    const [searchResult] = useACMSafeFetch(
```
SanjalKatiyar, line 246: what about handling `loading` and `error` ??

packages/mco/components/modals/app-manage-policies/helper/pvc-details-wizard-content.tsx
```diff
@@ -0,0 +1,358 @@
...
+  pvcSelector: [],
+});
+
+const updateLabels = (
+  drPlacementControl: DRPlacementControlType,
+  labels: string[]
+): DRPlacementControlType => ({
+  ...drPlacementControl,
+  pvcSelector: labels,
+});
```
SanjalKatiyar, line 113: can't we re-use `getDRPCKindObj` from `utils` ??

packages/mco/components/modals/app-manage-policies/helper/pvc-details-wizard-content.tsx
GowthamShanmugam: tag is an array type and it has only two indexes, one is the placement of string type, other one is the label of string[] type.

type TagsType = (string | string[])[][];

basically tag?.[0] is string, tag?.[1]; is array

packages/mco/hooks/acm-safe-fetch.ts
```diff
@@ -0,0 +1,37 @@
...
+        body: JSON.stringify(searchQuery),
+      },
+    })
+      .then((data) => {
+        setResponse(data);
+      })
+      .catch((err) => {
+        setError(err);
+      });
+  }, [setResponse, searchQuery, safeFetch]);
```
SanjalKatiyar, line 33:
```suggestion
  }, [setResponse, searchQuery]);
```

packages/mco/components/modals/app-manage-policies/helper/pvc-details-wizard-content.tsx
GowthamShanmugam: We are using lazy loading against  NameValueEditor component in multiple places for code splitting: https://legacy.reactjs.org/docs/code-splitting.html

To avoid increasing the size of the bundle using some complex library. 

Thanks, @bipuladh for the ref link

SanjalKatiyar: yup, I got that... but my doubt was whether we are saving lot of bundle size by doing splitting here ??
Default `NameValueEditor` had code for `PairElement`, but in MCO we are passing that as a prop as well `PairElementComponent` (which means it is getting bundled in same code as the one which is calling the FC)... so is there any major advantage of splitting here...

SanjalKatiyar: only code we are saving is the one from `NameValueEditor` FC itself, which is not too much ??

SanjalKatiyar: anyway, no issues with using lazy loading, but don;t see any specific reason as well here...

packages/mco/components/modals/app-manage-policies/utils/acm-utils.ts
GowthamShanmugam: Yes, UI API read needs a limit. UI can't read large amounts of PVCs. It will affect the performance. As per PM and QE, the number of PVCs per app is around 20 only, the count may go a little bit higher in the future but never read all PVCs blindly. 

UI can't decide which PVCs are related to the app and which PVCs are not realted. UI is just assuming all PVCs under the namespace are realted to the app only. It is up to the user to decide which label to select.

GowthamShanmugam: UI needs a limit, If they wants to use more PVC, then they have to use CLI way

packages/mco/components/modals/app-manage-policies/utils/parser-utils.ts
GowthamShanmugam: this is for the custom object type, the old one for the PlacementDecisionKind object which takes from clusters from status?.decisions

packages/mco/components/modals/app-manage-policies/helper/pvc-details-wizard-content.tsx
GowthamShanmugam: This is a custom-type object for wizard flow,
All the components under the manage policy modal will understand this custom type, 
So I am keeping this custom type to reuse components, for example, policy configuration view (used for review and create page, as well as to display policy configurations)
At the end of the assign policy wizard i am converting this custom type to using getDRPCKindObj to create DRPC.

packages/mco/components/modals/app-manage-policies/utils/acm-utils.ts
SanjalKatiyar: my question here would be `limit` just tells us how my data points the `response` from the API call will be right ??
If limit=20 or limit=200, in both cases it will still be a single API request right ?? so will it make diff to hard limit it to be `20` PVCs ?? @GowthamShanmugam

SanjalKatiyar: in list pages, say `Secrets` a cluster can have 300 secrets as well... all we display on UI and we send a single API call which gives us all 300 secret in a single response...

SanjalKatiyar: UI just need to maintain a list of all PVCs, any other performance impact which I might be missing ??

GowthamShanmugam: We are sending an API request to the local cluster, But here ACM search API is sending an API request to the managed clusters.

ACM is sending an API request to its managed cluster agent, and an agent will receive the API request and read all PVC from the namespace and convert the result to a custom response object and send it back to the hub cluster. 

if we give a limit of 20 then it will read only first 20 PVCs from the namespace. it means the API request will be waiting for the 20 PVC results to be parsed. If we increase the limit then it will take increase the API response time. 

Per namespace, I dont think the user will create more than 20 PVCs, If QE thinks the number is not enough i will increase more.

GowthamShanmugam: The idea here is, not all PVCs should have a unique namespace, the user has to reuse the labels as much as possible between PVCs

packages/mco/hooks/acm-safe-fetch.ts
GowthamShanmugam: safeFetch is also one dependency

SanjalKatiyar: not needed as it won;t change (check `useURLPoll`)...
if needed anyway then wrap `safeFetch` in a `useCallback` before returning it from the hook...

SanjalKatiyar: currently if u add it as it is then it will cause issues...

SanjalKatiyar: cc @GowthamShanmugam

GowthamShanmugam: not sure how it works for useURLPoll, i am getting lint error if i am not adding it in dependency

SanjalKatiyar: u can disable the lint error... or if u do not want to do that let;s use `useCallback`

SanjalKatiyar: for disabling use `// eslint-disable-next-line <error_name>`

SanjalKatiyar: if want to use `useCallback` u need to use it in the hook itself that returns `safeFetch`.

bipuladh (commented):
lgtm

packages/mco/components/modals/app-manage-policies/helper/pvc-details-wizard-content.tsx
GowthamShanmugam: Maybe i am wrong here but I thought bundle creation is nothing to do with runtime calls, MCO calls the PairElementComponent, or  PairElement is decided at runtime,  so it will still affect the bundle size when i use NameValueEditor directly.

PR #910, opened 2023-06-28
Transform shared folder into shared library

alfonsomthd:
Fixes: https://issues.redhat.com/browse/RHSTOR-4666


packages/shared/package.json
```diff
@@ -1,5 +1,116 @@
...
+    "react-hook-form": "^7.42.1",
+    "react-i18next": "^11.11.4",
+    "react-linkify": "^0.2.2",
+    "react-router": "5.2.0",
+    "react-router-dom": "5.2.0",
+    "react-tagsinput": "^3.19.0",
+    "resolve-url-loader": "^5.0.0",
+    "victory-core": "^35.4.4",
+    "yup": "^0.32.11"
+  },
```
SanjalKatiyar, line 50: are we adding same dep and devDep twice ?? once in root and once in shared ??

packages/shared/package.json
```diff
@@ -1,5 +1,116 @@
...
+  "license": "MIT",
+  "main": "src/index.ts",
+  "files": [
+    "src"
+  ],
+  "scripts": {
+    "build": "yarn build:prod",
+    "build:prod": "yarn run tsnode node_modules/.bin/webpack --mode=production -c ./webpack.config.ts",
+    "build:dev": "yarn run tsnode node_modules/.bin/webpack --mode=development -c ./webpack.config.ts",
+    "tsnode": "ts-node -O '{\"module\":\"commonjs\"}'"
```
SanjalKatiyar, line 14: doubt here: why do we need these "scripts" in `shared` ?? Isn't this should be part of root ??

webpack.config.ts
```diff
@@ -49,6 +49,9 @@ const config: webpack.Configuration & DevServerConfiguration = {
   },
   resolve: {
     extensions: ['.ts', '.tsx', '.js', '.jsx'],
+    alias: {
+      '@odf/shared': path.resolve(__dirname, './packages/shared/src/'),
+    },
```
SanjalKatiyar, line 54: just a doubt: was there any build error without this ??

packages/shared/webpack.config.ts
```diff
@@ -0,0 +1,98 @@
+/* eslint-env node */
+
+import * as path from 'path';
+import { ForkTsCheckerWebpackPlugin } from 'fork-ts-checker-webpack-plugin/lib/plugin';
+import { Configuration } from 'webpack';
+
+const config: Configuration = {
```
SanjalKatiyar, line 7: Screenshot 2023-06-28 at 1 34 55 PM

wanna try adding publicPath here ?? or will test it as part of diff story/task ??
(talking about error jest is giving in fusion-native-console)

packages/shared/package.json
```diff
@@ -1,5 +1,116 @@
...
+  },
+  "peerDependencies": {
+    "@openshift-console/dynamic-plugin-sdk": "^0.0.19",
+    "@openshift-console/dynamic-plugin-sdk-internal": "^0.0.7"
```
SanjalKatiyar, line 18: shouldn't react/react-dom etc be peerDep as well ??

packages/shared/webpack.config.ts
```diff
@@ -0,0 +1,98 @@
...
+  entry: './src/index.ts',
+  output: {
+    path: path.resolve(__dirname, 'build'),
+    filename: 'index.js',
+    libraryTarget: 'umd',
```
SanjalKatiyar, line 12: use `output.library.type` instead: https://webpack.js.org/configuration/output/#outputlibrarytarget (warning part)

packages/shared/webpack.config.ts
```diff
@@ -0,0 +1,98 @@
...
+    library: '@odf/shared',
```
SanjalKatiyar, line 13:
```suggestion
    library: {
      name: '@odf/shared',
      type: 'umd',
    },
```

packages/shared/webpack.config.ts
```diff
@@ -0,0 +1,98 @@
...
+      {
+        test: /\.tsx?$/,
+        use: {
+          loader: 'ts-loader',
+          options: {
+            configFile: path.resolve(__dirname, 'tsconfig.json'),
+            transpileOnly: false,
+            happyPackMode: false,
+          },
+        },
```
SanjalKatiyar, line 34: doubt: no need to exclude `node_modules` ??

packages/shared/package.json
SanjalKatiyar: I am assuming this is more of what's required by shared, rather all the deps present in root (some may not be used by shared)... in that case, feel free to ignore this comment.

packages/shared/webpack.config.ts
alfonsomthd: @SanjalKatiyar That error does not belong to the shared library as the unit tests are passing.

SanjalKatiyar: I am talking about tests in fusion repo, for the files those import any component from `odfshared`

SanjalKatiyar: u mean u wrote some unit tests in fusion repo(for any component which import from odfshared) and test was passing ??

packages/shared/package.json
alfonsomthd: @SanjalKatiyar In fact, since you are using [Yarn Workspaces](https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/) and it manages the dependencies of all projects (workspaces), we should add dependencies of each project to its own package.json, not the workspace root.

SanjalKatiyar: yes, but deps like yup/react/lodash etc will be used by all the packages... so might make sense to add them on root

SanjalKatiyar: and only add those deps and devDeps to a package, which are unique to that specific package...

packages/shared/webpack.config.ts
alfonsomthd: I mean the unit tests of all the packages (including the library) are passing.

packages/shared/package.json
alfonsomthd: I think @ivoalmeida 's intention here was to make it clear that when installing the library, those packages should also be required. Anyway, I'll move them to devDeps as we use them in some modules.

packages/shared/package.json
alfonsomthd: As we want to publish that package, AFAIK if all the required deps are not on in its package.json, you'll get errors about missing packages.

packages/shared/package.json
SanjalKatiyar: any consumer of this odf-shared library will be having these deps install, so make sense to keep them (sdk) in peerDeps... my question was around react and other react related deps, shoudn;t we add them in peer as well ??

packages/shared/webpack.config.ts
SanjalKatiyar: make sense to target this as separate story/task...

packages/shared/package.json
```diff
@@ -1,5 +1,116 @@
...
-  "version": "0.0.0"
+  "version": "0.0.0",
+  "license": "MIT",
+  "main": "src/index.ts",
+  "files": [
+    "src"
+  ],
+  "scripts": {
+    "build": "yarn build:prod",
+    "build:prod": "yarn run tsnode node_modules/.bin/webpack --mode=production -c ./webpack.config.ts",
```
GowthamShanmugam, line 12: The translation namesapce for MCO and ODF console  is passed as a variable (I8N_NS=plugin__odf-multicluster-console, I8N_NS=plugin__odf-console) at build time.

The webpack.config.ts will recive this variables and create ENV variable:
new webpack.DefinePlugin({
      'process.env.I8N_NS': JSON.stringify(process.env.I8N_NS),
}),

This ENV variable will be recived by useCustomTranslator.tsx hook in shared.

My doubt is, is this flow will still work when we seprate shared as a library? is shared can still able to recive these ENV variables from odf-console and mco-console during build time?

please makesure this flow is not breaking.

packages/shared/package.json
SanjalKatiyar: cc @bipuladh wdyt ?

SanjalKatiyar:
/lgtm
Few points which needs to be addressed in upcoming PRs:
1. Use `output.library.type` in webpack config: https://github.com/red-hat-storage/odf-console/pull/910#discussion_r1247500237
2. Figure out issue with unit tests (in fusion repo) which are giving error for components with odfshared imports: https://github.com/red-hat-storage/odf-console/pull/910#discussion_r1247481393
3. Update peerDeps (if needed): https://github.com/red-hat-storage/odf-console/pull/910#discussion_r1247612900
4. Making translations work for the consumer of the library.

SanjalKatiyar:
> /lgtm Few points which needs to be addressed in upcoming PRs:
> 
> 1. Use `output.library.type` in webpack config: [Transform shared folder into shared library #910 (comment)](https://github.com/red-hat-storage/odf-console/pull/910#discussion_r1247500237)
> 2. Figure out issue with unit tests (in fusion repo) which are giving error for components with odfshared imports: [Transform shared folder into shared library #910 (comment)](https://github.com/red-hat-storage/odf-console/pull/910#discussion_r1247481393)
> 3. Update peerDeps (if needed): [Transform shared folder into shared library #910 (comment)](https://github.com/red-hat-storage/odf-console/pull/910#discussion_r1247612900)
> 4. Making translations work for the consumer of the library.

cc @bipuladh for approval

packages/shared/package.json
alfonsomthd: @GowthamShanmugam we agreed to address the translations issue in a follow-up PR.

bipuladh (approved):

PR #913, opened 2023-07-04
Fix KMIP file upload error message

puiterwijk:
Turns out that the drop is also rejected if the file extension is not '.pem' (even though PEM encoded files can be in many names, e.g. cert-manager uses .crt in its secret keys). Let's make sure the error message represents this, to avoid wasting more time from other people.

### DIFF
--- a/locales/en/plugin__odf-console.json
+++ b/locales/en/plugin__odf-console.json
@@ -748,7 +748,7 @@
   "Client certificate": "Client certificate",
   "CA certificate": "CA certificate",
   "Client private key": "Client private key",
-  "Maximum file size exceeded. File limit is 4MB.": "Maximum file size exceeded. File limit is 4MB.",
+  "Invalid file: File size exceeds 4MB limit or incorrect file extension (only .pem files allowed).": "Invalid file: File size exceeds 4MB limit or incorrect file extension (only .pem files allowed).",
   "File read failed. Please try again.": "File read failed. Please try again.",
   "Upload a .PEM file here...": "Upload a .PEM file here...",
   "Unique identifier": "Unique identifier",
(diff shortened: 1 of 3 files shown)


SanjalKatiyar:
Thanks a lot for pointing this out and raising a PR :)

1. Though we need to make change at one more place: https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/modals/advanced-kms-modal/advanced-vault-modal.tsx#L68
2. Also, no matter how small the change is, any textual changes are verified by UX designers and Doc team, I will ping them for the same and will let you know.

puiterwijk:
> Thanks a lot for pointing this out and raising a PR :)
> 
> 1. Though we need to make change at one more place: https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/modals/advanced-kms-modal/advanced-vault-modal.tsx#L68
> 2. We need to run `yarn i18n` command, after any textual changes for the parser to make respective change in the json locales file.

Thanks for the reply!
I will fix items 1 and 2 and look forward to the response of 3!

SanjalKatiyar:
> Thanks for the reply! I will fix items 1 and 2 and look forward to the response of 3!

Hi, I checked with them and `Invalid file: File size exceeds 4MB limit or incorrect file format (only .pem files allowed)` is their suggestion.

puiterwijk:
@SanjalKatiyar thanks. Will update now.

puiterwijk:
Though technically the file *format* doesn't have to be wrong, the error is in the file *extension*.
I can absolutely have a PEM-encoded file in a file with a .crt extension (again, look at the default cert secrets used by openshift - those call it `ca.crt`, `cert.crt` and `ca.key`).

SanjalKatiyar:
> Though technically the file _format_ doesn't have to be wrong, the error is in the file _extension_. I can absolutely have a PEM-encoded file in a file with a .crt extension (look at the default cert secrets used by openshift - those call it `ca.crt`, `tls.crt` and `tls.key`).
> 
> So would this also be fine: `Invalid file: File size exceeds 4MB limit or incorrect file extension (only .pem files allowed).` ?
> 
> (replacing `format` with `extension`.)

looks good to me !

PR #918, opened 2023-07-06
Set @odf-console/shared as source library

alfonsomthd:
Publish the source code of the library so it can be recompiled in another projects (thus making the shared translations work).
- Erased the namespace of some translations so any project i18n parser can add those translations to its own namespace (so we don't generate >1 translation files).


packages/shared/package.json
```diff
@@ -1,33 +1,44 @@
...
   "scripts": {
-    "build": "yarn build:prod",
-    "build:prod": "yarn run tsnode node_modules/.bin/webpack --mode=production -c ./webpack.config.ts",
-    "build:dev": "yarn run tsnode node_modules/.bin/webpack --mode=development -c ./webpack.config.ts",
-    "tsnode": "ts-node -O '{\"module\":\"commonjs\"}'"
-  },
-  "peerDependencies": {
-    "@openshift-console/dynamic-plugin-sdk": "^0.0.19",
-    "@openshift-console/dynamic-plugin-sdk-internal": "^0.0.7"
+    "build": "rm -rf build && mkdir build && cp -r src package.json LICENSE build"
```
SanjalKatiyar, line 19: doubt: if we want to publish source code so that consumer can compile on their end, so why we need to add `build` here as well ??

alfonsomthd: Because build contains only what we want to publish: `yarn publish --access public build`

packages/client/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "0.0.0",
   "dependencies": {
     "@odf/ocs": "0.0.0",
-    "@odf/shared": "0.0.0"
+    "@odf/shared": "file:../shared"
```
ivoalmeida, line 7: why not set this as "*"? would that not work?

packages/shared/src/models/common.ts
```diff
@@ -5,21 +5,21 @@ import {
 
 export const PersistentVolumeModel: K8sKind = {
   label: 'PersistentVolume',
-  // t('plugin__odf-console~PersistentVolume')
+  // t('PersistentVolume')
```
ivoalmeida, line 8: why are you keeping this t fn calls commented out? why not remove them?

SanjalKatiyar: these are like markups for the parser: https://github.com/openshift/console/blob/master/INTERNATIONALIZATION.md#internationalization-guidelines

alfonsomthd: @ivoalmeida as @SanjalKatiyar mentioned, without that coomments, the parser don't get the models' translations.

packages/client/package.json
alfonsomthd: @ivoalmeida It does not: in order to publish under our `@odf-console` org., the package has been renamed to `@odf-console/shared`.
Currently we have lots of `from '@odf/shared'` imports so in order to keep them working, marking that as a local dependency does the trick.
If we want,  we can set up a follow-up refactoring/cleanup task to align the imports/naming cc @bipuladh

packages/shared/package.json
```diff
@@ -45,72 +56,9 @@
...
-    "tsm": "^2.3.0",
     "typescript": "^4.8",
-    "webpack": "5.74.0",
-    "webpack-bundle-analyzer": "^4.6.1",
-    "webpack-cli": "4.5.x",
-    "webpack-dev-server": "^4.9.2"
+    "victory-core": "^35.4.4",
+    "yup": "^0.32.11"
   }
 }
```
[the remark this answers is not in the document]
alfonsomthd: Done.

SanjalKatiyar:
/lgtm
cc @bipuladh @ivoalmeida for approval...

PR #921, opened 2023-07-12
update selectable table to support column tooltip/popover

SanjalKatiyar:


packages/shared/src/table/selectable-table.tsx
```diff
@@ -102,6 +103,7 @@ export const SelectableTable: SelectableTableProps = <
           />
           {columns?.map((col, index) => (
             <Th
+              {...(!!col.info ? { info: col.info } : {})}
```
SanjalKatiyar, line 106: for supporting popover/tooltip on column name (header): https://www.patternfly.org/v4/components/table#composable-custom-row-wrapper-header-tooltips--popovers

bipuladh (approved):

PR #924, opened 2023-07-18
added selectProps to MultiSelectDropdown

ivoalmeida:
Adding selectProps: SelectProps as I need to set more props in this component than it is currently allowed.

### DIFF
--- a/packages/shared/src/dropdown/multiselectdropdown.tsx
+++ b/packages/shared/src/dropdown/multiselectdropdown.tsx
@@ -1,14 +1,20 @@
 import * as React from 'react';
-import { Select, SelectOption, SelectVariant } from '@patternfly/react-core';
+import {
+  Select,
+  SelectOption,
+  SelectVariant,
+  SelectProps,
+} from '@patternfly/react-core';
 import { useCustomTranslation } from '../useCustomTranslationHook';
 
 export const MultiSelectDropdown: React.FC<MultiSelectDropdownProps> = ({
   onChange,
-  placeholder,
+  placeholderText,
   id,
   options,
-  selected = [],
+  selections = [],
   variant,
+  ...rest
 }) => {
   const [isOpen, setOpen] = React.useState(false);
 
@@ -16,11 +22,12 @@ export const MultiSelectDropdown: React.FC<MultiSelectDropdownProps> = ({
     _event: React.MouseEvent | React.ChangeEvent,
     selection: string
   ) => {
-    let cSelected: string[] = selected;
-    if (selected.includes(selection)) {
-      cSelected = selected.filter((item) => item !== selection);
+    const prevSelection = selections as string[];
+    let cSelected = prevSelection;
+    if (prevSelection.includes(selection)) {
+      cSelected = prevSelection.filter((item) => item !== selection);
     } else {
-      cSelected = [...selected, selection];
+      cSelected = [...prevSelection, selection];
     }
     onChange(cSelected);
   };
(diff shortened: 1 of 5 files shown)


packages/shared/src/dropdown/multiselectdropdown.tsx
```diff
@@ -56,4 +63,5 @@ export type MultiSelectDropdownProps = {
   placeholder?: string;
   variant?: SelectVariant;
   onChange: (selected: string[]) => void;
+  selectProps?: SelectProps;
```
bipuladh, line 66: Don't we have overlap between the already existing props and `SelectProps`. Could we perhaps remove the common ones and do something like 
```
SelectProps & {
 whateverPropsAreNotPartOfSelectProps: any;
}

```

ivoalmeida: Done

packages/shared/src/dropdown/multiselectdropdown.tsx
```diff
@@ -43,17 +49,17 @@ export const MultiSelectDropdown: React.FC<MultiSelectDropdownProps> = ({
...
       isCheckboxSelectionBadgeHidden
+      {...rest}
     >
       {items}
     </Select>
   );
 };
 
-export type MultiSelectDropdownProps = {
+export type MultiSelectDropdownProps = SelectProps & {
```
GowthamShanmugam, line 59: selectProps selections and placeholderText have no effect due to the selected, options.  shall we remove these custom props?

just asking for your opinion

bipuladh (approved):

packages/shared/src/dropdown/multiselectdropdown.tsx
GowthamShanmugam, line 59: selectProps selections, placeholder text, variant, has no effect due to selected, options, and variant.  shall we remove these custom props?

just asking for your opinion 
```suggestion
 type Props = Omit<SelectProps, 'placeholderText' | 'selections'>
export type MultiSelectDropdownProps = Props & {
```

packages/shared/src/dropdown/multiselectdropdown.tsx
```diff
@@ -1,26 +1,32 @@
...
   const [isOpen, setOpen] = React.useState(false);
 
   const onSelect = (
     _event: React.MouseEvent | React.ChangeEvent,
     selection: string
   ) => {
-    let cSelected: string[] = selected;
-    if (selected.includes(selection)) {
-      cSelected = selected.filter((item) => item !== selection);
+    let cSelected: string[] = selections as string[];
```
GowthamShanmugam, line 25:
```suggestion
    const prevSelection = selections as string[];
    let cSelected= prevSelection;
```

packages/shared/src/dropdown/multiselectdropdown.tsx
```diff
@@ -1,26 +1,32 @@
...
+    if ((selections as string[]).includes(selection)) {
+      cSelected = (selections as string[]).filter((item) => item !== selection);
```
GowthamShanmugam, line 27:
```suggestion
    if ((prevSelection).includes(selection)) {
      cSelected = (prevSelection).filter((item) => item !== selection);
```

packages/shared/src/dropdown/multiselectdropdown.tsx
```diff
@@ -1,26 +1,32 @@
...
     } else {
-      cSelected = [...selected, selection];
+      cSelected = [...(selections as string[]), selection];
```
GowthamShanmugam, line 29:
```suggestion
      cSelected = [...prevSelection, selection];
```

GowthamShanmugam (approved):

PR #926, opened 2023-07-19
Selectable table bug fix

GowthamShanmugam:


packages/shared/src/table/selectable-table.tsx
```diff
@@ -50,6 +63,8 @@ export const SelectableTable: SelectableTableProps = <
     RowComponent,
     extraProps,
     isSelectableHidden,
+    loaded,
+    loadError,
   } = props;
   const { t } = useCustomTranslation();
   const {
```
SanjalKatiyar, line 70: need to maintain same reference for `new Set(selectableRows?.map(getUID)),` (https://github.com/red-hat-storage/odf-console/pull/926/files#diff-a776ebfbd967acff8db4670cb1c227fa5d8dac4d7739f9dbb1cab12da92eb5d1R81) across re-render, which is an argument for `useSelectList` below...

SanjalKatiyar: same is true for `selectableRows` as well, need to maintain the reference...

SanjalKatiyar: cc @GowthamShanmugam

GowthamShanmugam: i forgot to push this change :)

GowthamShanmugam: changes done

PR #927, opened 2023-07-20
Bug 2213550: Correct the display names of rbd image statuses on BlockPools page

GowthamShanmugam:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2213550


SanjalKatiyar:
/retitle Bug 2213550: Correct the display names of rbd image statuses on BlockPools page

PR #931, opened 2023-07-20
for ceph-storage

debjyoti-pandit:


alfonsomthd:
@debjyoti-pandit It seems the PR title is incomplete...

packages/shared/src/dashboards/breakdown-card/breakdown-body.spec.tsx
```diff
@@ -0,0 +1,223 @@
+import React from 'react';
+import { PersistentVolumeClaimModel } from '@odf/shared/models';
+import { render } from '@testing-library/react';
+import { BreakdownCardBody, BreakdownBodyProps } from './breakdown-body';
+
+jest.mock('../../useCustomTranslationHook', () => ({
+  useCustomTranslation: () => ({
+    t: (key: string) => key,
+  }),
+}));
```
alfonsomthd, line 10: @debjyoti-pandit IMHO we can move this mock to `setupJest.ts` where where it already lives a i18n mock also...

packages/shared/src/dashboards/breakdown-card/breakdown-body.spec.tsx
```diff
@@ -0,0 +1,223 @@
...
+      ...defaultProps,
+      isLoading: false,
+      hasLoadError: false,
+      isPersistentInternal: false,
+    };
+    const { container } = render(<BreakdownCardBody {...props} />);
+
+    expect(
+      container.querySelector('[data-test-id="mocked-total-capacity-body"]')
+    ).toBeInTheDocument();
```
alfonsomthd, line 212: @debjyoti-pandit We also should ensure that we get the expected text content.

debjyoti-pandit: I think getting the expected text content is the responsibility of TotalCapacityBody component.
Since in my tests, it is mocked. We are not getting the text content in the document. We can check whether we are passing proper text to that component or not.

alfonsomthd: > [...] We can check whether we are passing proper text to that component or not.
That's what I meant.

alfonsomthd: It's OK as I see we're checking the data received by CapacityBody.

bipuladh:
/lgtm cancel 
please fix the typos in commit messages

SanjalKatiyar:
> /lgtm cancel please fix the typos in commit messages

cc @debjyoti-pandit

setupJest.ts
```diff
@@ -16,3 +16,9 @@ jest.mock('react-i18next', () => ({
     };
   },
 }));
+
+jest.mock('@odf/shared/useCustomTranslationHook', () => ({
+  useCustomTranslation: () => ({
+    t: (key: string) => key,
+  }),
+}));
```
SanjalKatiyar, line 24: https://github.com/red-hat-storage/odf-console/pull/508#discussion_r1032375606
cc @alfonsomthd @bipuladh do we want to move this here now ??

packages/shared/src/dashboards/breakdown-card/breakdown-body.tsx
```diff
@@ -60,9 +60,9 @@ export const BreakdownCardBody: React.FC<BreakdownBodyProps> = ({
     <Grid>
       <GridItem span={4}>
         <TotalCapacityBody
-          prefix={isPersistentInternal && t('Total requests: ')}
+          prefix={isPersistentInternal ? t('Total requests: ') : undefined}
```
SanjalKatiyar, line 63: `{...(isPersistentInternal ? { prefix: t('Total requests: ') } : {})}`

packages/shared/src/dashboards/breakdown-card/breakdown-body.tsx
```diff
@@ -60,9 +60,9 @@ export const BreakdownCardBody: React.FC<BreakdownBodyProps> = ({
...
           capacity={humanize(metricTotal).string}
-          suffix={!isPersistentInternal && t('used')}
+          suffix={!isPersistentInternal ? t('used') : undefined}
```
SanjalKatiyar, line 65: `{...(!isPersistentInternal ? { suffix: t('used ') } : {})}`

packages/shared/src/dashboards/breakdown-card/breakdown-body.tsx
```diff
@@ -71,7 +71,7 @@ export const BreakdownCardBody: React.FC<BreakdownBodyProps> = ({
         {capacityAvailable && (
           <TotalCapacityBody
             capacity={humanize(capacityAvailable).string}
-            suffix={!isPersistentInternal && t('available')}
+            suffix={!isPersistentInternal ? t('available') : undefined}
```
SanjalKatiyar, line 74: `{...(!isPersistentInternal ? { suffix: t('available ') } : {})}`

setupJest.ts
alfonsomthd: > [#508 (comment)](https://github.com/red-hat-storage/odf-console/pull/508#discussion_r1032375606) cc @alfonsomthd @bipuladh do we want to move this here now ??

In this case the return value is always the same and the purpose of moving it here is because will be usually needed in a lot of tests.

SanjalKatiyar:
either squash the commits or if planning to have 2 commits (one for "feat" other for "test") then there shouldn't be overlapping between changes, review comment should be fixed in the commit where that change belongs :)

PR #932, opened 2023-07-24
After hub recovery, cant initiate failover

GowthamShanmugam:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2209288

### DIFF
--- a/packages/mco/utils/acm.ts
+++ b/packages/mco/utils/acm.ts
@@ -1,11 +1,20 @@
-import { getLabel, getName, getNamespace } from '@odf/shared/selectors';
-import { PLACEMENT_REF_LABEL } from '../constants';
+import {
+  getAnnotations,
+  getLabel,
+  getName,
+  getNamespace,
+} from '@odf/shared/selectors';
+import {
+  LAST_APP_DEPLOYMENT_CLUSTER_ANNOTATION,
+  PLACEMENT_REF_LABEL,
+} from '../constants';
 import {
   ArgoApplicationSetKind,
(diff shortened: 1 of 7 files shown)


packages/mco/utils/acm.ts
```diff
@@ -65,3 +74,23 @@ export const getClustersFromDecisions = (
...
+) =>
+  getAnnotations(drPlacementControl)?.[
+    LAST_APP_DEPLOYMENT_CLUSTER_ANNOTATION
+  ] || '';
+
+export const findDeploymentClusters = (
+  placementDecision: ACMPlacementDecisionKind,
+  drPlacementControl: DRPlacementControlKind
+): string[] => {
+  if (placementDecision?.status.decisions.length > 0) {
```
SanjalKatiyar, line 89: do we need to do this or its guaranteed that if "status" is present then "decisions" will be present as well ??
```suggestion
  if ((placementDecision ?? {}).status?.decisions?.length > 0) {
```

SanjalKatiyar:
so, this annotation will never get removed ?? not even during hub recovery ??

GowthamShanmugam:
> so, this annotation will never get removed ?? not even during hub recovery ??

When ramen starts to reconcile, it will add the annotation immediately. In hub recovery they will get this info from s3 object storage and display it as a last remembered application cluster.

packages/mco/utils/acm.ts
GowthamShanmugam: ack

PR #936, opened 2023-07-25
Fix selectable table sorting issue

GowthamShanmugam:


packages/shared/src/table/selectable-table.tsx
```diff
@@ -24,9 +25,9 @@ export const sortRows = (
   sortField: string
 ) => {
   const negation = c !== SortByDirection.asc;
-  const aValue = a?.[sortField] || '';
-  const bValue = b?.[sortField] || '';
-  const sortVal = (aValue as string).localeCompare(bValue as string);
+  const aValue = _.get(a, sortField, '').toString();
```
GowthamShanmugam, line 28: a?.[sortField] won't work when sortField has nested object keys (eg: metadata.name)

packages/shared/src/table/selectable-table.tsx
```diff
@@ -72,7 +73,7 @@ export const SelectableTable: SelectableTableProps = <
     sortIndex: activeSortIndex,
     sortDirection: activeSortDirection,
     sortedData: sortedRows,
-  } = useSortList<T>(rows, columns, true);
+  } = useSortList<T>(rows, columns, false);
```
SanjalKatiyar, line 76: do u think we need to pass true/false as a prop to `SelectableTable` ?? I mean anyone who is using `SelectableTable`, will they have to configure it in any case or it should always be `false` ??

SanjalKatiyar: cc @GowthamShanmugam

GowthamShanmugam: No need to pass this true or false, it is based on how the component is implemented, For the selectable table it should be always false.

PR #940, opened 2023-07-26
Bug 2213552: Modified to a standard time format in DR pages

GowthamShanmugam:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2213552

### DIFF
--- a/packages/mco/components/modals/app-data-policies-status/subscriptions/dr-status-table.tsx
+++ b/packages/mco/components/modals/app-data-policies-status/subscriptions/dr-status-table.tsx
@@ -2,8 +2,8 @@ import * as React from 'react';
 import { pluralize } from '@odf/core/components/utils';
 import {
   getLatestDate,
-  utcDateTimeFormatterWithTimeZone,
   fromNow,
+  utcDateTimeFormatter,
 } from '@odf/shared/details-page/datetime';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
 import { StatusIconAndText } from '@openshift-console/dynamic-plugin-sdk';
@@ -68,7 +68,7 @@ const getSyncStatus = (syncTime: string) =>
   !!syncTime
     ? {
         text: fromNow(syncTime),
-        toolTip: utcDateTimeFormatterWithTimeZone.format(new Date(syncTime)),
+        toolTip: utcDateTimeFormatter.format(new Date(syncTime)),
       }
     : {};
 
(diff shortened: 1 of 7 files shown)


SanjalKatiyar:
/retitle Bug 2213552: Modified to a standard time format in DR pages

SanjalKatiyar:
@GowthamShanmugam I have a doubt here (no need to make any changes in this PR)... but everywhere in ODF and now MCO while formatting time we are considering it as UTC timezone, is that alright to always convert to UTC... cluster can be in other timezones as well...

GowthamShanmugam:
> @GowthamShanmugam I have a doubt here (no need to make any changes in this PR)... but everywhere in ODF and now MCO while formatting time we are considering it as UTC timezone, is it all alright to always convert to UTC... cluster can be in other timezones as well... need to check how OCP does it...

Yes, clusters can be deployed in different zones, To resolve this time format issue, OCP is following UTC as a common time zone. Everywhere on the list page, they are using local time to display but popover  UTC is used as a common timezone.

SanjalKatiyar:
> > @GowthamShanmugam I have a doubt here (no need to make any changes in this PR)... but everywhere in ODF and now MCO while formatting time we are considering it as UTC timezone, is it all alright to always convert to UTC... cluster can be in other timezones as well... need to check how OCP does it...
> 
> Yes, clusters can be deployed in different zones, To resolve this time format issue, OCP is following UTC as a common time zone. Everywhere on the list page, they are using local time to display but popover UTC is used as a common timezone.

so if OCP is using local time for page, it means we are doing it wrong in ODF ?? I think in ODF we always use UTC (for list as well as popover) right ??

GowthamShanmugam:
> > > @GowthamShanmugam I have a doubt here (no need to make any changes in this PR)... but everywhere in ODF and now MCO while formatting time we are considering it as UTC timezone, is it all alright to always convert to UTC... cluster can be in other timezones as well... need to check how OCP does it...
> > 
> > 
> > Yes, clusters can be deployed in different zones, To resolve this time format issue, OCP is following UTC as a common time zone. Everywhere on the list page, they are using local time to display but popover UTC is used as a common timezone.
> 
> so if OCP is using local time for the list page, it means we are doing it wrong in ODF ?? I think in ODF we always use UTC (for list as well as popover) right ??

OCP is using UTC + local time. LocalTime to display text and UTC as a tooltip popover text. I think the place we can't provide popover its is correct to use UTC.

SanjalKatiyar:
> OCP is using UTC + local time. LocalTime to display text and UTC as a tooltip popover text. I think the place we can't provide popover its is correct to use UTC.

I meant in ODF/MCO list pages as well we are using UTC instead of local time... anyway that's not related to this PR...

PR #941, opened 2023-07-28
Bug 2209258: Fixes sidebar content length dynamically based on screen size

bipuladh:
Screenshot from 2023-07-28 13-04-44
Screenshot from 2023-07-28 13-04-50

### DIFF
--- a/packages/shared/src/hooks/useAutoExpand.ts
+++ b/packages/shared/src/hooks/useAutoExpand.ts
@@ -1,0 +1,33 @@
+import * as React from 'react';
+
+const useAutoExpand = () => {
+  const ref = React.useRef(null);
+  const [height, setHeight] = React.useState(0);
+
+  const updateHeight = () => {
+    if (ref.current) {
+      const windowHeight = window.innerHeight;
+      const divTopOffset = ref.current.getBoundingClientRect().top;
+      const availableHeight = windowHeight - divTopOffset;
+      setHeight(availableHeight);
+    }
+  };
+
+  React.useEffect(() => {
+    updateHeight(); // Set initial height on mount
+
+    const handleResize = () => {
+      updateHeight(); // Update height on window resize
+    };
+
+    window.addEventListener('resize', handleResize);
+
+    return () => {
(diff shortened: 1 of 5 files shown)


packages/shared/src/hooks/useAutoExpand.ts
```diff
@@ -0,0 +1,33 @@
...
+const useAutoExpand = () => {
+  const ref = React.useRef(null);
+  const [height, setHeight] = React.useState(0);
+
+  const updateHeight = () => {
+    if (ref.current) {
+      const windowHeight = window.innerHeight;
+      const divTopOffset = ref.current.getBoundingClientRect().top;
+      const availableHeight = windowHeight - divTopOffset;
+      setHeight(availableHeight);
```
SanjalKatiyar, line 12: just a nit (can be skipped as well):
```suggestion
      availableHeight !== height && setHeight(availableHeight);
```

packages/shared/src/hooks/useAutoExpand.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+  };
+
+  React.useEffect(() => {
+    updateHeight(); // Set initial height on mount
```
SanjalKatiyar, line 17: won't `ref.current` be `null` during initial mount and then later ref will get attached to the sideBar component ?? (`updateHeight` only runs when ref is non-empty)

bipuladh: Hence we check for ref.current in line 8.

packages/shared/src/hooks/useAutoExpand.ts
bipuladh: This should only be triggered when the window size is changed so the guard might not be necessary.

packages/shared/src/hooks/useAutoExpand.ts
SanjalKatiyar: I think it will get attached first... due to lifecycle of useEffect...

SanjalKatiyar: nvm ignore this comment...

PR #946, opened 2023-08-01
fix: added Rack info for the topology details screen

debjyoti-pandit:
fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2214023

when only zone is present
image

when only rack is present
image

just in case we have both of them
image

### DIFF
--- a/packages/shared/src/constants/common.ts
+++ b/packages/shared/src/constants/common.ts
@@ -7,3 +7,4 @@ export const ONE_MINUTE = 60 * ONE_SECOND;
 export const ONE_HOUR = 60 * ONE_MINUTE;
 export const ALL_NAMESPACES = 'all-namespaces';
 export const DEFAULT_NS = 'default';
+export const RACK_LABEL = 'topology.rook.io/rack';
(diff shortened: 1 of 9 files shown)


SanjalKatiyar:
LGTM (but we are now maintaining  `RACK_LABEL` and `getRack` at 2 diff packages, should be always under `shared` and be removed from other place)

debjyoti-pandit:
> LGTM (but we are now maintaining `RACK_LABEL` and `getRack` at 2 diff packages, should be always under `shared` and be removed from other place)

True, should I refactor it in this PR itself only?

SanjalKatiyar:
> > LGTM (but we are now maintaining `RACK_LABEL` and `getRack` at 2 diff packages, should be always under `shared` and be removed from other place)
> 
> True, should I refactor it in this PR itself only?

yup...

SanjalKatiyar:
plz fix the build tests...

SanjalKatiyar:
update the PR title correctly as per the guidelines...

packages/shared/src/constants/common.ts
```diff
@@ -7,3 +7,4 @@ export const ONE_MINUTE = 60 * ONE_SECOND;
 export const ONE_HOUR = 60 * ONE_MINUTE;
 export const ALL_NAMESPACES = 'all-namespaces';
 export const DEFAULT_NS = 'default';
+export const RACK_LABEL = 'topology.rook.io/rack';
```
SanjalKatiyar, line 10: remove this from odf/src/constants...

SanjalKatiyar:
squash the commits...

PR #947, opened 2023-08-01
Remove link and remove cluster name

TimothyAsirJeyasing:
https://bugzilla.redhat.com/show_bug.cgi?id=2223702

### DIFF
--- a/packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster.tsx
+++ b/packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster.tsx
@@ -275,6 +275,7 @@ export const SnapshotUtilizationCard: React.FC<SnapshotUtilizationCardProps> =
           customDateTimeFormatter={dateTimeFormatterWithWeekDay.format}
           hideCurrentHumanized
           hideHorizontalBorder
+          disableGraphLink
           showLegend
           CustomUtilizationSummary={CustomUtilizationSummary}
         />
--- a/packages/shared/src/dashboards/utilization-card/utilization-item.tsx
+++ b/packages/shared/src/dashboards/utilization-card/utilization-item.tsx
@@ -56,6 +56,7 @@ export const UtilizationItem: React.FC<UtilizationItemProps> = React.memo(
     description,
     hideCurrentHumanized,
     hideHorizontalBorder,
+    disableGraphLink,
     showLegend,
     CustomUtilizationSummary,
   }) => {
@@ -100,7 +101,7 @@ export const UtilizationItem: React.FC<UtilizationItemProps> = React.memo(
         ariaChartTitle={title}
         data={data}
         loading={!error && isLoading}
-        query={query}
+        query={!disableGraphLink && query}
         // Todo(bipuladh): Make humanize type Humanize once unit.js is converted
         humanize={humanizeValue as Humanize}
         byteDataType={byteDataType}
@@ -272,6 +273,7 @@ type UtilizationItemProps = {
   customDateTimeFormatter?: (date: Date) => string;
   description?: string | ((result: PrometheusResult, index: number) => string);
   hideCurrentHumanized?: boolean;
+  disableGraphLink?: boolean;
   showLegend?: boolean;
   hideHorizontalBorder?: boolean;
   CustomUtilizationSummary?: React.FC<CustomUtilizationSummaryProps>;
--- a/packages/shared/src/dashboards/utilization-card/prometheus-utilization-item.tsx
+++ b/packages/shared/src/dashboards/utilization-card/prometheus-utilization-item.tsx
@@ -40,6 +40,7 @@ export const PrometheusUtilizationItem: React.FC<PrometheusUtilizationItemProps>
     description,
     hideCurrentHumanized,
     hideHorizontalBorder,
+    disableGraphLink,
     showLegend,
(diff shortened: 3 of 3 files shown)


TimothyAsirJeyasing:
/retitle Bug 2223702: Remove link and remove cluster name

TimothyAsirJeyasing:
Screenshot from 2023-08-01 15-03-51

TimothyAsirJeyasing:
@SanjalKatiyar, @GowthamShanmugam Please review

SanjalKatiyar:
shorten commit title and fix build test...

packages/shared/src/dashboards/utilization-card/area-chart.tsx
```diff
@@ -128,6 +128,15 @@ export const AreaChart: React.FC<AreaChartProps> = ({
     CustomChartGroup = ChartGroup;
   }
 
+  const removeParenthesesFromLegendData = (legData) => {
```
bipuladh, line 131:
```suggestion
  const removeParenthesesFromLegendData = (legendData) => {
```
Add types

packages/shared/src/dashboards/utilization-card/area-chart.tsx
```diff
@@ -144,7 +153,13 @@ export const AreaChart: React.FC<AreaChartProps> = ({
             themeColor={ChartThemeColor.multiUnordered}
             scale={{ x: 'time', y: 'linear' }}
             padding={{ top: 20, left: 70, bottom: 60, right: 0 }}
-            legendData={chartType ? legendData : null}
+            legendData={
+              chartType
+                ? showLegend
```
bipuladh, line 158: Where is showLegend coming from and in which cases is it false?

bipuladh (commented):
Could you perhaps look into the usage of AreaChart and instead make the changes related to removing parenthesis there? I hardly think making changes to the parenthesis is something that is required in the common packages.

packages/shared/src/dashboards/utilization-card/area-chart.tsx
TimothyAsirJeyasing: As per UX, we need to show the legend with time, value and the line name (name of the particular line in the graph) in parenthesis. So when we pass the line name with parenthesis as a parameter  to patternfly chart module it also shows the line names in parenthesis at the bottom (bottom line description). So i am passing the name in parenthesis so as to display in the legend and removing them for bottom description.  Also here the variable name legendData is used already so to avoid the compile issue i used a different name. I feel this as a kind of hack only. I do not find any other possibility to solve this issue. Please suggest me if another better possibility to deal this.

packages/shared/src/dashboards/utilization-card/area-chart.tsx
TimothyAsirJeyasing: Currently the Chart module will display if there are 2 set of data (lines) and will not show any graph if there is only one data series. We have added this showLegend variable to allow showing graph for single line of graph also. Its used only for ceph RBD graph, so i though of using the variable here. One other suggestion i am having is like passing the line name in parenthesis and also to pass the removeParanthesis function to use here. Could you please provide your suggestion ?

packages/shared/src/dashboards/utilization-card/area-chart.tsx
GowthamShanmugam: Removing removeParenthesesFromLegendData is a specific use case, this whole function is messing up to common logic. For each string operation, you can't keep adding a new function.

packages/shared/src/dashboards/utilization-card/area-chart.tsx
GowthamShanmugam: i really dont like this logic, just to remove parentheses we are messing up the common logic here

packages/shared/src/dashboards/utilization-card/utilization-item.tsx
```diff
@@ -272,6 +273,7 @@ type UtilizationItemProps = {
   customDateTimeFormatter?: (date: Date) => string;
   description?: string | ((result: PrometheusResult, index: number) => string);
   hideCurrentHumanized?: boolean;
+  hideLink?: boolean;
```
GowthamShanmugam, line 276: disableGraphLink

SanjalKatiyar:
update the PR title correctly as per the guidelines...

TimothyAsirJeyasing:
/retitle Remove link and remove cluster name

TimothyAsirJeyasing:
image

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/cluster.tsx
```diff
@@ -253,7 +253,7 @@ export const PVCsSection: React.FC<PVCsSectionProps> = ({
 
 const getDescription = (result: PrometheusResult, _index: number) =>
   // Returning cluster name as a description
-  result.metric?.['cluster'] || '';
+  `${result.metric?.['cluster'] || ''}`;
```
GowthamShanmugam, line 256: Why this change, it looks no change from old code.

TimothyAsirJeyasing: Yes, i will use the earlier code.

GowthamShanmugam (approved):

PR #962, opened 2023-08-09
OB/OBC: fix list filter

alfonsomthd:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2230334
Before:
Screenshot from 2023-08-09 09-28-51
After:
Screenshot from 2023-08-09 12-21-24


packages/odf/components/mcg/ObjectBucketClaim.tsx
```diff
@@ -268,12 +268,15 @@ export const OBCListPage: React.FC<ObjectBucketClaimsPageProps> = (props) => {
     ...(namespace !== ALL_NAMESPACES_KEY ? { namespace } : {}),
   });
 
-  const [data, filteredData, onFilterChange] = useListPageFilter(obc);
+  const rowFilters = [obcStatusFilter(t)];
```
SanjalKatiyar, line 271: it is added to the dependency list in useListPageFilter... so better to keep ref same across re-renders...

PR #971, opened 2023-08-16
Refactors Modal usage

bipuladh:
Notable changes: 
Uses useModal hook exposed as part of SDK in favour of temporary workaround
Kebab Menu internally handles launching modal and figuring out the correct modal.
Components will handle the translation of kebab menu items externally instead of passing a function that takes `t` as an argument.


packages/ocs/block-pool/BlockPoolDetailsPage.tsx
```diff
@@ -62,57 +59,57 @@ export const BlockPoolDetailsPage: React.FC<BlockPoolDetailsPageProps> = ({
...
-              name: 'Events',
-              component: EventStreamWrapped,
-            },
-          ]}
-        />
-      )}
-    </>
+  return !loaded ? (
+    <LoadingBox />
+  ) : (
```
SanjalKatiyar, line 89: `DetailsPage` takes `loaded` and `loadError` props to internally display the `LoadingBox`...
```suggestion
  return 
```

bipuladh: This code is from before. Let's refactor this in a different PR.

packages/mco/components/modals/app-data-policies-status/subscriptions/data-policies-status-popover.tsx
```diff
@@ -175,9 +178,8 @@ export const DataPoliciesStatusPopover: React.FC<DataPoliciesStatusPopoverProps>
         return [{ drPolicies: {} }, 0];
       }, [drPlacementControls, drpcLoaded, subscriptionMap, placementMap]);
 
-    const [Modal, modalProps, launcher] = useModalLauncher(modalMap);
     const launchModal = React.useCallback(
-      () => launcher(STATUS_MODAL, dataPoliciesStatus),
+      () => launcher(DataPoliciesStatusModal, dataPoliciesStatus),
```
SanjalKatiyar, line 182: don't we need to pass `isOpen = true` here as well ?? I don't think `dataPoliciesStatus` has that key...

packages/odf/components/bucket-class/create-bc.tsx
```diff
@@ -35,23 +35,21 @@ enum CreateStepsBC {
...
   const { ns = CEPH_STORAGE_NAMESPACE } = match.params;
 
-  const [Modal, modalProps, launcher] = useModalLauncher(modalMap);
+  const launcher = useModal();
 
   const history = useHistory();
 
   const launchModal = React.useCallback(
-    () => launcher(NS_STORE_MODAL_KEY, null),
+    () => launcher(NamespaceStoreCreateModal, null),
```
SanjalKatiyar, line 52: same doubt here... `NamespaceStoreCreateModal` is expecting `isOpen` prop, don't we need to pass it ?

SanjalKatiyar: earlier `useModalLauncher` was passing it... I think now we might have to do it explicitly...

packages/odf/components/mcg/ObjectBucketClaim.tsx
```diff
@@ -397,27 +386,30 @@ export const OBCDetailsPage: React.FC<ObjectBucketClaimDetailsPageProps> = ({
...
+                import(
+                  '../../modals/attach-deployment/attach-deployment-obc-modal'
+                )
+            ),
           },
         ]}
       />
     );
-  }, [launchModal, resource, namespace]);
+  }, [resource, namespace]);
```
SanjalKatiyar, line 412:
```suggestion
  }, [t, resource, namespace]);
```

packages/shared/src/details-page/DetailsPage.tsx
```diff
@@ -338,9 +340,7 @@ export const ResourceSummary: React.FC<ResourceSummaryProps> = ({
         obj={resource}
         path="metadata.labels"
         valueClassName="details-item__value--labels"
-        onEdit={() =>
-          launchModal(ModalKeys.EDIT_LABELS, { resource, resourceModel })
-        }
+        onEdit={() => launchModal(EditLabelModal, { resource, resourceModel })}
```
SanjalKatiyar, line 343: `isOpen =  true` ??

packages/shared/src/details-page/DetailsPage.tsx
```diff
@@ -358,7 +358,7 @@ export const ResourceSummary: React.FC<ResourceSummaryProps> = ({
               type="button"
               isInline
               onClick={() =>
-                launchModal(ModalKeys.EDIT_ANN, {
+                launchModal(AnnotationsModal, {
```
SanjalKatiyar, line 361: `isOpen =  true` ??
Ideally Modal itself should use isOpen as true by default, but okay if we don't want to make changed to modal in this PR... but we might need to pass the prop explicitly ??

packages/ocs/block-pool/BlockPoolDetailsPage.tsx
SanjalKatiyar: okay...

PR #972, opened 2023-08-16
Running OCP console as a container

GowthamShanmugam:
A basic env variable to run:

For odf-console:
---------------------------------------------------
export I8N_NS=plugin__odf-console
export PLUGIN=odf
export BRIDGE_PLUGINS="odf-console=http://localhost:9001"  (Optional, by default script will use odf-console plugin)
yarn dev:c

(or)

I8N_NS=plugin__odf-console PLUGIN=odf BRIDGE_PLUGINS="odf-console=http://localhost:9001"  yarn dev:c

For mco-console:
--------------------------------------------
export I8N_NS=plugin__odf-multicluster-console
export PLUGIN=mco
export BRIDGE_PLUGINS="mce=http://localhost:3001,acm=http://localhost:3002,odf-multicluster-console=http://localhost:9001" 
yarn dev:c

(or)

I8N_NS=plugin__odf-multicluster-console PLUGIN=mco BRIDGE_PLUGINS="mce=http://localhost:3001,acm=http://localhost:3002,odf-multicluster-console=http://localhost:9001"  yarn dev:c

"yarn dev:c" is an generic command for all ODF realted plugins

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,8 @@ ODF Console is the UI plugin for Openshift Data Foundation Operator. It works as
 ## Running in Development Mode
 
 ODF console works as a remote bundle for OCP console. To run ODF Console there should be a instance of OCP console up and running.
-Follow these steps to run OCP Console in development mode:
+
+### Steps to run OCP Console as a server in development mode:
 
 1. Follow everything as mentioned in the console [README.md](https://github.com/openshift/console) to build the application.
 2. Run the console bridge as follows `./bin/bridge -plugins odf-console=http://localhost:9001/`
@@ -18,6 +19,15 @@ After the OCP console is set as required by ODF Console. Performs the following 
 3. Clone this repo.
 4. Pull all required dependencies by running `yarn install`.
 5. Run the development mode of odf-console using `yarn run dev`. This runs a webserver in port 9001.
+
+### Steps to run OCP Console as a container in development mode:
+
+1. Install ODF Operator.
+2. Create a Storage System.
+3. Clone this repo.
+4. Pull all required dependencies by running `yarn install`.
+5. Run the development mode of odf-console using `I8N_NS=plugin__odf-console PLUGIN=odf yarn run dev:c`. This runs a container running both the console bridge and a webserver in port 9001.
+6. For more OCP container related environment variables [Refer](https://github.com/red-hat-storage/odf-console/scripts/start-ocp-console.sh).
 
 ### Unit Tests
 
--- a/package.json
+++ b/package.json
@@ -50,7 +50,9 @@
     "po-to-i18n": "node ./scripts/i18n/po-to-i18n.js",
     "export-pos": "./scripts/i18n/export-pos.sh",
     "memsource-upload": "./scripts/i18n/memsource-upload.sh",
-    "memsource-download": "./scripts/i18n/memsource-download.sh"
+    "memsource-download": "./scripts/i18n/memsource-download.sh",
+    "ocp-console": "./scripts/start-ocp-console.sh",
+    "dev:c": "yarn ocp-console & PLUGIN=${PLUGIN} I8N_NS=${I8N_NS} yarn server:plugin & wait"
   },
   "dependencies": {
     "@openshift-console/dynamic-plugin-sdk": "0.0.19",
--- a/scripts/start-ocp-console.sh
+++ b/scripts/start-ocp-console.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+CONSOLE_VERSION=${CONSOLE_VERSION:=4.14}
+CONSOLE_PORT=${CONSOLE_PORT:=9000}
+CONSOLE_IMAGE="quay.io/openshift/origin-console:${CONSOLE_VERSION}"
+
+echo "Starting local OpenShift console..."
+
+BRIDGE_BASE_ADDRESS=${BRIDGE_BASE_ADDRESS:="http://localhost:9000"}
+BRIDGE_USER_AUTH=${BRIDGE_USER_AUTH:="disabled"}
+BRIDGE_K8S_MODE=${BRIDGE_K8S_MODE:="off-cluster"}
+BRIDGE_K8S_AUTH=${BRIDGE_K8S_AUTH:="bearer-token"}
+BRIDGE_K8S_MODE_OFF_CLUSTER_SKIP_VERIFY_TLS=${BRIDGE_K8S_MODE_OFF_CLUSTER_SKIP_VERIFY_TLS:=true}
+BRIDGE_K8S_MODE_OFF_CLUSTER_ENDPOINT=$(oc whoami --show-server)
+BRIDGE_PLUGINS=${BRIDGE_PLUGINS:=odf-console="http://localhost:9001"}
+BRIDGE_PLUGIN_PROXY=${BRIDGE_PLUGIN_PROXY:=""}
+
+# The monitoring operator is not always installed (e.g. for local OpenShift). Tolerate missing config maps.
+set +e
+BRIDGE_K8S_MODE_OFF_CLUSTER_THANOS=$(oc -n openshift-config-managed get configmap monitoring-shared-config -o jsonpath='{.data.thanosPublicURL}' 2>/dev/null)
+BRIDGE_K8S_MODE_OFF_CLUSTER_ALERTMANAGER=$(oc -n openshift-config-managed get configmap monitoring-shared-config -o jsonpath='{.data.alertmanagerPublicURL}' 2>/dev/null)
+set -e
+BRIDGE_K8S_AUTH_BEARER_TOKEN=$(oc whoami --show-token 2>/dev/null)
+BRIDGE_USER_SETTINGS_LOCATION="localstorage"
+
+# Don't fail if the cluster doesn't have gitops.
+set +e
+GITOPS_HOSTNAME=$(oc -n openshift-gitops get route cluster -o jsonpath='{.spec.host}' 2>/dev/null)
+set -e
+if [ -n "$GITOPS_HOSTNAME" ]; then
+    BRIDGE_K8S_MODE_OFF_CLUSTER_GITOPS="https://$GITOPS_HOSTNAME"
+fi
+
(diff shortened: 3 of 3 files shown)


start-ocp-console.sh
```diff
@@ -0,0 +1,52 @@
...
+# Prefer podman if installed. Otherwise, fall back to docker.
+if [ -x "$(command -v podman)" ]; then
+    podman run \
+      --pull always --rm -p "$CONSOLE_PORT":9000 \
+      --rm --network=host \
+      --env-file <(set | grep BRIDGE) \
+      $CONSOLE_IMAGE
+else
+    docker run \
+      --pull always --rm -p "$CONSOLE_PORT":9000 \
```
alfonsomthd, line 48: Same as above.

alfonsomthd (changes requested):

SanjalKatiyar:
1. we have `scripts` directory, might be a good place for this file as well
2. also, can we have separate commands for running OCP as container or locally ??

start-ocp-console.sh
```diff
@@ -0,0 +1,52 @@
...
+BRIDGE_K8S_AUTH_BEARER_TOKEN=$(oc whoami --show-token 2>/dev/null)
+BRIDGE_USER_SETTINGS_LOCATION="localstorage"
+
+# Don't fail if the cluster doesn't have gitops.
+set +e
+GITOPS_HOSTNAME=$(oc -n openshift-gitops get route cluster -o jsonpath='{.spec.host}' 2>/dev/null)
+set -e
+if [ -n "$GITOPS_HOSTNAME" ]; then
+    BRIDGE_K8S_MODE_OFF_CLUSTER_GITOPS="https://$GITOPS_HOSTNAME"
+fi
```
SanjalKatiyar, line 33: isn't it more of a related to mco, and not concerning to odf ?

start-ocp-console.sh
```diff
@@ -0,0 +1,52 @@
...
+
+echo "Starting local OpenShift console..."
+
+BRIDGE_BASE_ADDRESS="http://localhost:9000"
+BRIDGE_USER_AUTH="disabled"
+BRIDGE_K8S_MODE="off-cluster"
+BRIDGE_K8S_AUTH="bearer-token"
+BRIDGE_K8S_MODE_OFF_CLUSTER_SKIP_VERIFY_TLS=true
+BRIDGE_K8S_MODE_OFF_CLUSTER_ENDPOINT=$(oc whoami --show-server)
+BRIDGE_PLUGINS="odf-console=http://localhost:9001"
```
alfonsomthd, line 17: Let's read the active plugins from environment:
```suggestion
BRIDGE_PLUGINS=${BRIDGE_PLUGINS:=odf-console=http://localhost:9001}
```

alfonsomthd (changes requested):

start-ocp-console.sh
GowthamShanmugam: I dont know why OCP keeping this on a common file: https://github.com/openshift/console/commits/master/contrib/oc-environment.sh

I need to figure the purpose of this

package.json
```diff
@@ -31,6 +31,8 @@
     "analyze-odf": "PLUGIN=odf yarn analyze-common && PLUGIN=odf yarn ts-node ./analyzeTest.ts",
     "analyze-mco": "PLUGIN=mco yarn analyze-common && PLUGIN=mco yarn ts-node ./analyzeTest.ts",
     "dev": "PLUGIN=odf I8N_NS=plugin__odf-console yarn server:plugin",
+    "dev:c": "yarn ocp-console & PLUGIN=odf I8N_NS=plugin__odf-console yarn server:plugin",
+    "dev-mco:c": "yarn ocp-console & PLUGIN=mco I8N_NS=plugin__odf-multicluster-console yarn server:plugin",
```
SanjalKatiyar, line 35: if I just run `yarn dev-mco:c` it will bridge to `odf-console` not `mco-console`, why not pass the value of `BRIDGE_PLUGINS` from here...
and for cases where we need to run multiple plugins together, maybe a new command for that or just document it in the ReadMe...

alfonsomthd: BRIDGE_PLUGINS config is read from the environment, [it shouldn't be hardcoded](https://12factor.net/config).

SanjalKatiyar: that is fine as well, point here is `if I just run yarn dev-mco:c it will bridge to odf-console not mco-console`

SanjalKatiyar: and to document the steps...

alfonsomthd: @SanjalKatiyar If I understand you correctly, your suggestion would be to add the proper `default` to the command (but still having preference the environment), sth like adding:
`BRIDGE_PLUGINS=${BRIDGE_PLUGINS:=odf-multicluster-console="http://localhost:9002"}`.
It makes sense.

SanjalKatiyar: "BRIDGE_PLUGINS" is nothing but constant values like "PLUGIN" or "I8N_NS"... which are not gonna change...

SanjalKatiyar: "BRIDGE_PLUGINS" is nothing but constant values like "PLUGIN" or "I8N_NS"... which are not gonna change much often...

SanjalKatiyar:
@GowthamShanmugam for running locally can't we serve mco as diff port ?? right now both odf/mco runs on 9001

package.json
alfonsomthd: BRIDGE_PLUGINS is part of the config (not constant), so I always should be able to run this:
`BRIDGE_PLUGINS="odf-console=http://localhost:9001,fusion-console=http://localhost:9010" yarn dev:c` so it makes sense to set sensible defaults but never hardcoding: it's against the best practices.

SanjalKatiyar: `so it makes sense to set sensible defaults but never hardcoding: it's against the best practices.` >> I think there is some communication gap here, passing it as a process.env is what I was suggesting (which is not a case right now if u run `yarn dev-mco:c`) and not to hardcode on `.sh` file...

also, followup on the last comment, it is little bit confusing, this is "odf-console" repo, but we want to pass "fusion-console" to a script which is defined in entirely diff repo (I can understand it will work)... but @GowthamShanmugam do we need to add script to this repo, doesn't OCP themselves provides a way/script to run it as a container ?? Why do we want to run `yarn ocp-console` from our repo, if console provides the script we can just run from there ??

SanjalKatiyar: btw these are just some questions from my side on quickly going over this PR... I don't have any major issues from my side...

SanjalKatiyar: `yarn dev-mco:c` >> passes value for mco-console
`yarn dev:c` >> passes value for odf-console
and other cases we might document the steps... which will to run `BRIDGE_PLUGINS="odf-console=http://localhost:9001,fusion-console=http://localhost:9010"  yarn ocp-console`

start-ocp-console.sh
GowthamShanmugam: https://github.com/red-hat-storage/odf-console/pull/972#issue-1852974243

GowthamShanmugam:
> @GowthamShanmugam for running locally can't we serve mco at diff port ?? right now both odf/mco runs on 9001

For now, we won't have any use case like running ODF and MCO together, it becomes more complicated when we need to run ODF on multiple clusters. For ODF console i user public URL and for MCO i use local setup.

package.json
GowthamShanmugam: may you can add this script on fusion-console repo also

GowthamShanmugam: I really dont like keeping separate yarn commands for different consoles, lets make everything generic and let all values come from the Env:
"dev:c": "yarn ocp-console & PLUGIN=${PLUGIN} I8N_NS=${I8N_NS} yarn server:plugin"

alfonsomthd: > also, followup on the last comment, it is little bit confusing, this is "odf-console" repo, but we want to pass "fusion-console" to a script which is defined in entirely diff repo (I can understand it will work)... but 

It's just an example (it could another related plugin), don't be so literal :wink: The point here is that config should not be hardcoded as can change based on different deploys/envs/needs.

README.md
```diff
@@ -19,6 +20,15 @@ After the OCP console is set as required by ODF Console. Performs the following
 4. Pull all required dependencies by running `yarn install`.
 5. Run the development mode of odf-console using `yarn run dev`. This runs a webserver in port 9001.
 
+### Steps to run OCP Console as a container in development mode:
+
+1. Install ODF Operator.
+2. Create a Storage System.
+3. Clone this repo.
+4. Pull all required dependencies by running `yarn install`.
+5. Run the development mode of odf-console using `I8N_NS=plugin__odf-console PLUGIN=odf yarn run dev:c`. This runs a webserver in port 9001.
```
alfonsomthd, line 29:
```suggestion
5. Run the development mode of odf-console using `I8N_NS=plugin__odf-console PLUGIN=odf yarn run dev:c`. This runs a container running both the console bridge and a webserver in port 9001.
```

GowthamShanmugam: ack

GowthamShanmugam:
/assign @SanjalKatiyar 
need your approval

package.json
```diff
@@ -50,7 +50,9 @@
     "po-to-i18n": "node ./scripts/i18n/po-to-i18n.js",
     "export-pos": "./scripts/i18n/export-pos.sh",
     "memsource-upload": "./scripts/i18n/memsource-upload.sh",
-    "memsource-download": "./scripts/i18n/memsource-download.sh"
+    "memsource-download": "./scripts/i18n/memsource-download.sh",
+    "ocp-console": "./scripts/start-ocp-console.sh",
+    "dev:c": "yarn ocp-console & PLUGIN=${PLUGIN} I8N_NS=${I8N_NS} yarn server:plugin & wait"
```
SanjalKatiyar, line 55: this will start a local server for plugin as well, why do we need `wait` here ?? do we need to wait for some process completion ??

GowthamShanmugam: Here, We are using (&) to run the two concurrent servers (OCP and odf-console). But (&) on prevents the command from responding to C-c/C-z and also prevents its return code from propagating in case of a failure.

So it will keep running the one server even when another one is failed. We need to kill the job to stop the other server. 

Here, Wait command will wait for the children to be killed and it will makesure clean exit.

PR #983, opened 2023-08-22
Fetching alerts from OCP thanos instead of ACM thanos

GowthamShanmugam:
Blocked by: https://github.com/RamenDR/ramen/pull/1006
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2221660

For 4.14, We dont have any alert from ACM thanos, Replication delay alert is moved to OCP thanos by ramen operation to raise the alert even when observability is not enabled.

In future, UI need to fetch alert from both OCP thanos as well as ACM thanos.

### DIFF
--- a/packages/mco/utils/disaster-recovery.tsx
+++ b/packages/mco/utils/disaster-recovery.tsx
@@ -18,6 +18,7 @@ import {
   K8sResourceCommon,
   GreenCheckCircleIcon,
   Alert,
+  AlertStates,
 } from '@openshift-console/dynamic-plugin-sdk';
 import {
   Operator,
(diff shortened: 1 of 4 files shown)


GowthamShanmugam:
Screenshot from 2023-08-23 16-35-57

packages/mco/utils/disaster-recovery.tsx
```diff
@@ -524,7 +525,8 @@ export const filterPVCDataUsingAppsets = (
   );
 
 export const filterDRAlerts = (alert: Alert) =>
-  alert?.annotations?.alert_type === 'DisasterRecovery';
+  alert?.annotations?.alert_type === 'DisasterRecovery' &&
+  alert.state === AlertStates.Firing;
```
bipuladh, line 529: Would it be better to create another common filter `filterFiringAlerts` and apply these two filter functions.

GowthamShanmugam: You want a separate function to check alert is firing or not, (or) filter all firing alerts. Filtering all firing alerts in separate functions will lead to one more looping.

PR #986, opened 2023-08-23
Add initial default sorting for SelectableTable.

alfonsomthd:
- Add `skeleton` prop to StatusBox inside SelectableTable component.

### DIFF
--- a/packages/shared/src/hooks/sort-list.ts
+++ b/packages/shared/src/hooks/sort-list.ts
@@ -1,12 +1,14 @@
 import * as React from 'react';
 import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk-internal/lib/extensions/console-types';
+import * as _ from 'lodash-es';
 import { SortByDirection } from '@patternfly/react-table';
 
 export const useSortList = <R extends K8sResourceCommon>(
   data: R[],
   columns: any[],
   // True indicate the column index is starting from 1
-  onSelect: boolean
+  onSelect: boolean,
+  initialSortIndex?: number
 ): {
   onSort: (
     event: React.MouseEvent,
@@ -17,7 +19,9 @@ export const useSortList = <R extends K8sResourceCommon>(
   sortDirection: SortByDirection;
   sortedData: R[];
 } => {
-  const [sortIndex, setSortIndex] = React.useState(-1);
+  const [sortIndex, setSortIndex] = React.useState(
+    _.isInteger(initialSortIndex) ? initialSortIndex : -1
+  );
   const [sortDirection, setSortDirection] = React.useState<SortByDirection>(
     SortByDirection.asc
   );
--- a/packages/shared/src/table/selectable-table.tsx
+++ b/packages/shared/src/table/selectable-table.tsx
@@ -16,6 +16,12 @@ import { useSelectList } from '../hooks/select-list';
 import { useSortList } from '../hooks/sort-list';
 import { getUID } from '../selectors';
 
+export type TableColumnProps = ThProps & {
+  thProps?: TableThProps;
+  columnName: string;
+  sortFunction?: <T>(a: T, b: T, c: SortByDirection) => number;
+};
+
 export const sortRows = (
   a: any,
   b: any,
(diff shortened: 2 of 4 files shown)


packages/shared/src/table/selectable-table.tsx
```diff
@@ -51,6 +57,15 @@ const isRowSelected = <T extends K8sResourceCommon>(
   selectedRows: T[]
 ) => selectedRows.some((row) => getUID(row) === rowId);
 
+const canApplyInitialSort = (
+  rows: any[],
+  columns: TableColumnProps[],
+  initialSortIndex?: number
+): boolean =>
+  !_.isEmpty(rows) &&
+  _.isInteger(initialSortIndex) &&
```
SanjalKatiyar, line 66: given useSort already has `_.isInteger(initialSortIndex)` check and if its false default is `-1`, do we need to do it again ??

SanjalKatiyar: no issues though, we can do it... just might be irrelevant...

alfonsomthd: @SanjalKatiyar as in the next truth check we're using `initialSortIndex` as array index I prefer to avoid potential errors.

packages/shared/src/table/selectable-table.tsx
```diff
@@ -18,6 +18,12 @@ import { useSortList } from '../hooks/sort-list';
 import { getUID } from '../selectors';
 import { useCustomTranslation } from '../useCustomTranslationHook';
 
+export type TableColumnProps = ThProps & {
+  thProps?: TableThProps;
+  columnName: string;
+  sortFunction?: (a: any, b: any, c: SortByDirection) => any;
```
bipuladh, line 24:
```suggestion
  sortFunction?: (a: any, b: any, c: SortByDirection) => any;
```
Can we add types for these?

packages/shared/src/hooks/sort-list.ts
```diff
@@ -17,7 +19,9 @@ export const useSortList = <R extends K8sResourceCommon>(
   sortDirection: SortByDirection;
   sortedData: R[];
 } => {
-  const [sortIndex, setSortIndex] = React.useState(-1);
+  const [sortIndex, setSortIndex] = React.useState(
+    _.isInteger(initialSortIndex) ? initialSortIndex : -1
```
bipuladh, line 23:
```suggestion
    initialSortIndex ?? -1
```

packages/shared/src/table/selectable-table.tsx
```diff
@@ -51,6 +57,12 @@ const isRowSelected = <T extends K8sResourceCommon>(
   selectedRows: T[]
 ) => selectedRows.some((row) => getUID(row) === rowId);
 
+const canApplyInitialSort = (
+  columns: TableColumnProps[],
+  initialSortIndex: number
+): boolean =>
+  _.isInteger(initialSortIndex) && !!columns[initialSortIndex]?.sortFunction;
```
bipuladh, line 64:
```suggestion
  !!columns[initialSortIndex]?.sortFunction;
```
Shouldn't this be enough?

packages/shared/src/hooks/sort-list.ts
alfonsomthd: @bipuladh the initialSortIndex has `number` type but I want to make sure we're receiving an integer, not any other type of number.

packages/shared/src/table/selectable-table.tsx
alfonsomthd: Done.

SanjalKatiyar:
/lgtm
@bipuladh for approval as I can see there was a comment from his side on the PR...

packages/shared/src/hooks/sort-list.ts
alfonsomthd: Example: someone could accidentally write `initialSortColumnIndex={0.2}` then we set -1.
As `useSortList` can be used outside `SelectableTable` IMHO this check makes sense.

alfonsomthd: @bipuladh the hook is a complete different piece from SelectableTable and `sortIndex` is meant to be an integer so wherever the hook is called we have to make sure we don't set anything other than integer.
IMHO this safe check makes sense and it's harmless.

PR #1001, opened 2023-08-28
Bug 2203362: [release-4.11] Fix the capacity queries

aruniiird:
Capacity queries are using 'kubelet_volume_stats_used_bytes', which gives the result in a per node basis (evenwhen the nodes are sharing a single PVC).
Corrected the query to give only one result per <namespace>+<PVC> combination.

Signed-off-by: Arun Kumar Mohan <amohan@redhat.com>
(cherry picked from commit d2d2a28a17fbb86a02841f9677e5f67625ff9f41)


SanjalKatiyar:
cc @bipuladh

SanjalKatiyar:
added a cherry-pick comment here: https://github.com/red-hat-storage/odf-console/pull/1001#issuecomment-1695056389
we need to make sure PR for 4.11-compatibility branch is merged as well before moving BZ to MODIFIED...

PR #1004, opened 2023-08-28
Changes the text from SDN to OVN

bipuladh:
Changes the labels for multus selections
Screenshot from 2023-08-28 16-20-12


SanjalKatiyar:
/retitle Changes the text from SDN to OVN

PR #1005, opened 2023-08-28
Bug 2233505: Fix provider details for NS details page

vbnrh:
Screenshot 2023-08-08 at 1 31 22 PM


SanjalKatiyar:
Bug 2233505: Fix provider details for NS details page

PR #1019, opened 2023-09-01
Improves warning message handling for selector input

vbnrh:
This blocks the save button if there are any warning messages

Screenshot 2023-09-01 at 4 32 55 PM


SanjalKatiyar:
> /hold

need to check consecutive dots scenario:
either we can make experience similar to OCP edit labels form or if k8s allows consecutive dots we can talk to UX and update the warning message...

PR #1046, opened 2023-09-14
fix Create StorageClass form

SanjalKatiyar:
https://bugzilla.redhat.com/show_bug.cgi?id=2239033


SanjalKatiyar:
@GowthamShanmugam PTAL.

PR #1060, opened 2023-10-02
Update lint rule: no-extraneous-deps

alfonsomthd:
- Include all missing packages.
- Do not allow devDeps except in tests files.


SanjalKatiyar:
LGTM,
re ran `Build Test / resolve-dependencies (18) (pull_request)` as it failed initially...

PR #1064, opened 2023-10-05
libuv update in 18.18.0 breaks webpack's thread-loader

GowthamShanmugam:
Ref: https://github.com/nodejs/node/issues/49911

Reverting back the node version to 18.17.1 for a temporary fix.


GowthamShanmugam:
New PR: https://github.com/red-hat-storage/odf-console/pull/1068

PR #1075, opened 2023-10-10
Fix unit test for 4.14

GowthamShanmugam:


SanjalKatiyar:
force merging this PR (as it is needed to un-block other 4.14 blocker PRs), this PR just fixes the unit tests and do not impact ODF in any way...

PR #1081, opened 2023-10-17
Fix CVE-2023-43646 get-func-name module security issue

TimothyAsirJeyasing:
broccoli-output-wrapper uses get-func-name module version 2.0.0 which has some security issues. 
Its fixed in version 2.0.1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,6 +129,7 @@
     "@types/node": "^14.14.34",
     "@typescript-eslint/eslint-plugin": "^5.42",
     "@typescript-eslint/parser": "^5.42",
+    "chai": "^4.3.10",
     "comment-json": "4.x",
     "cypress": "^10.10.0",
     "cypress-multi-reporters": "^1.5.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2801,6 +2801,11 @@ assert-plus@1.0.0, assert-plus@^1.0.0:
   resolved "https://registry.yarnpkg.com/assert-plus/-/assert-plus-1.0.0.tgz#f12e0f3c5d77b0b1cdd9146942e4e96c1e4dd525"
   integrity sha1-8S4PPF13sLHN2RRpQuTpbB5N1SU=
 
+assertion-error@^1.1.0:
+  version "1.1.0"
+  resolved "https://registry.yarnpkg.com/assertion-error/-/assertion-error-1.1.0.tgz#e60b6b0e8f301bd97e5375215bda406c85118c0b"
+  integrity sha512-jgsaNduz+ndvGyFt3uSuWqvy4lCnIJiovtouQN5JZHOKCS2QuhEdbcQHFhVksz2N2U9hXJo8odG7ETyWlEeuDw==
+
 ast-types-flow@^0.0.7:
   version "0.0.7"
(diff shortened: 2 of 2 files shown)


TimothyAsirJeyasing:
@SanjalKatiyar , @bipuladh Please review

SanjalKatiyar:
this is not a correct fix, assuming description of PR is correct and "broccoli-output-wrapper" is affected then you need to fix the existing package which is the cause of this instead of adding a new package directly to your dependencies.

SanjalKatiyar:
also u mentioned `broccoli-output-wrapper uses get-func-name`, I don't find `get-func-name` to be an internal dep of `broccoli-output-wrapper`, how does it use that can you point me to the file ?

TimothyAsirJeyasing:
Sorry, its a sub dependency for broccoli-output-wrapper used by chai module
node_modules/broccoli-output-wrapper/yarn.lock:get-func-name@^2.0.0

SanjalKatiyar:
> Sorry, its a sub dependency for broccoli-output-wrapper used by chai module node_modules/broccoli-output-wrapper/yarn.lock:get-func-name@^2.0.0

can u send me the yarn.lock file's exact line (in our odf-console) where `get-func-name` is defined as an internal dep of `broccoli-output-wrapper` ??

SanjalKatiyar:
> Sorry, its a sub dependency for broccoli-output-wrapper used by chai module node_modules/broccoli-output-wrapper/yarn.lock:get-func-name@^2.0.0

or maybe, the place where we are defining `chai` as a dep in odf-console ??

package.json
```diff
@@ -69,6 +69,7 @@
     "@types/react-dnd-html5-backend": "^3.0.2",
     "@types/lodash-es": "^4.17.4",
     "buffer": "^6.0.3",
+    "get-func-name": "^2.0.1",
```
bipuladh, line 72: I don't understand the need to introduce a new package.

bipuladh (changes requested):

package.json
TimothyAsirJeyasing: └─┬ i18next-parser@6.6.0
  └─┬ broccoli-plugin@4.0.7
    └── broccoli-output-wrapper@3.2.5

The recent available version of broccoli-output-wrapper package holds a devDependency to chai ^4.2.0, which included a version of get-func-name vulnerable to CVE-2023-43646. This CVE is fixed in get-func-name version 2.0.1 and above and its fixed/the higher version used only in Chai version 4.3.10  (https://github.com/chaijs/chai/pull/1540). I could see the entry of get-func-name only in node_modules/broccoli-output-wrapper/yarn.lock after i do yarn install. And i agree, we do not need this package and this CVE will not affect our component as we do not directly consume this anywhere.

bipuladh (commented):
Why are we introducing chai?

TimothyAsirJeyasing:
odf-console is not affected by "get-func-name" for production environment, it just used as a internal dependency for chai. 

└─┬ i18next-parser@6.6.0
└─┬ broccoli-plugin@4.0.7
└── broccoli-output-wrapper@3.2.5
        └── (devDependency) chai@4.2.0
                                                 └── get-func-name@2.0.0

We are not at all using the packages chai or get-func-name and (we do not have any entry either in yarn.lock). Chai is used only as a dev-dependency for broccoli-output-wrapper. Initially i was a bit confused about the sub dependencies. There is no fix required for this. I will close this pr and move the bz to on-qa with this note.

PR #1082, opened 2023-10-19
Fix mpath filtering issue

SanjalKatiyar:
https://bugzilla.redhat.com/show_bug.cgi?id=2182351

Screenshot 2023-10-19 at 1 14 49 PM

### DIFF
--- a/packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/selected-capacity.tsx
+++ b/packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/selected-capacity.tsx
@@ -32,7 +32,9 @@ const getTotalCapacity = (disks: DiscoveredDisk[]): number =>
 
 const isAvailableDisk = (disk: DiscoveredDisk): boolean =>
   disk?.status?.state === AVAILABLE &&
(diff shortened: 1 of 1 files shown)


packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/selected-capacity.tsx
```diff
@@ -32,7 +32,9 @@ const getTotalCapacity = (disks: DiscoveredDisk[]): number =>
 
 const isAvailableDisk = (disk: DiscoveredDisk): boolean =>
   disk?.status?.state === AVAILABLE &&
-  (disk.type === DiskType.RawDisk || disk.type === DiskType.Partition);
+  (disk.type === DiskType.RawDisk ||
+    disk.type === DiskType.Partition ||
+    disk.type === DiskType.Multipath);
```
SanjalKatiyar, line 37: nits: we can use `includes` here as well... but since I will now have to test again... let's move forward with 3 checks for now...

SanjalKatiyar: we can refactor if we support more in future...

PR #1083, opened 2023-10-19
Updating range value to 15m to match default ACM observability scraping value

GowthamShanmugam:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2244791


GowthamShanmugam:
image

PR #1086, opened 2023-10-19
Unit tests for CreateNamespaceStore & NamespaceStoreForm

alfonsomthd:
- Refactoring of https://github.com/red-hat-storage/odf-console/pull/827
- Prefer `@testing-library/user-event` over usage of `fireEvent` (as recommended).

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,6 +125,7 @@
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "^12",
     "@testing-library/react-hooks": "^8.0.1",
+    "@testing-library/user-event": "^14.5.1",
     "@types/jest": "29.2.2",
     "@types/node": "^14.14.34",
     "@typescript-eslint/eslint-plugin": "^5.42",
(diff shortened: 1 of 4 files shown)


alfonsomthd:
> > * Refactoring of [tests: create namespace store #827](https://github.com/red-hat-storage/odf-console/pull/827)
> > * Prefer `@testing-library/user-event` over usage of `fireEvent` (as recommended).
> 
> @alfonsomthd do we want to open a Github issue to remove usage of `fireEvent` from all the `spec.tsx` files and use `userEvent` instead ?

We can, but I don't see it as an urgency and in some cases we may need to keep the current implementation. From the [docs](https://testing-library.com/docs/user-event/intro#differences-from-fireevent):

> There are, however, some user interactions or aspects of these [that aren't yet implemented and thus can't yet be described with user-event](https://github.com/testing-library/user-event/issues?q=is%3Aopen+label%3Aaccuracy%2Cenhancement). In these cases you can use fireEvent to dispatch the concrete events that your software relies on.

PR #1089, opened 2023-10-20
Fix Volume replication health shows healthy post failover

GowthamShanmugam:
To align the dashboard with ramen alerts, the Dashboard will reflect volume replication health is critical for the last group sync time empty.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2244842

### DIFF
--- a/packages/mco/components/mco-dashboard/data-policy/cluster-app-card/common.tsx
+++ b/packages/mco/components/mco-dashboard/data-policy/cluster-app-card/common.tsx
@@ -6,6 +6,7 @@ import * as React from 'react';
 import {
   ALL_APPS,
   ALL_APPS_ITEM_ID,
+  LEAST_SECONDS_IN_PROMETHEUS,
   VOLUME_REPLICATION_HEALTH,
 } from '@odf/mco/constants';
 import {
@@ -74,8 +75,11 @@ export const VolumeSummarySection: React.FC<VolumeSummarySectionProps> = ({
     const volumeHealth = { critical: 0, warning: 0, healthy: 0 };
     const placementInfo: PlacementInfo = selectedAppSet?.placementInfo?.[0];
     protectedPVCData?.forEach((pvcData) => {
+      const pvcLastSyncTime = pvcData?.lastSyncTime;
       const health = getVolumeReplicationHealth(
-        getTimeDifferenceInSeconds(pvcData?.lastSyncTime),
+        !!pvcLastSyncTime
+          ? getTimeDifferenceInSeconds(pvcLastSyncTime)
+          : LEAST_SECONDS_IN_PROMETHEUS,
         pvcData?.schedulingInterval
       )[0];
       if (!!selectedAppSet) {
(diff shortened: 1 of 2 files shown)


GowthamShanmugam:
image

GowthamShanmugam:
image

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/common.tsx
```diff
@@ -74,8 +75,11 @@ export const VolumeSummarySection: React.FC<VolumeSummarySectionProps> = ({
     const volumeHealth = { critical: 0, warning: 0, healthy: 0 };
     const placementInfo: PlacementInfo = selectedAppSet?.placementInfo?.[0];
     protectedPVCData?.forEach((pvcData) => {
+      const pvcLastSyncTime = pvcData?.lastSyncTime;
       const health = getVolumeReplicationHealth(
-        getTimeDifferenceInSeconds(pvcData?.lastSyncTime),
+        !!pvcLastSyncTime
+          ? getTimeDifferenceInSeconds(pvcLastSyncTime)
```
bipuladh, line 81: I see this function being used https://github.com/GowthamShanmugam/odf-console/blob/master/packages/mco/components/mco-dashboard/data-policy/cluster-app-card/argo-application-set.tsx#L74 as well. 
Are the changes required there as well?

packages/mco/constants/dashboard.ts
```diff
@@ -14,3 +14,6 @@ export const enum VOLUME_REPLICATION_HEALTH {
   WARNING = 'warning',
   HEALTHY = 'healthy',
 }
+
+// Prometheus time() - 0 in seconds
+export const LEAST_SECONDS_IN_PROMETHEUS = 1697788182;
```
bipuladh, line 19: Where are we getting this value? Maybe we can make the comment more clearer.

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/common.tsx
GowthamShanmugam: you are referring my local branch the old master code, this component is removed from this file.

packages/mco/constants/dashboard.ts
GowthamShanmugam: This is happening on Ramen alerting rule: https://github.com/RamenDR/ramen/blob/5b80317c82cb484f6a639e24967967adb38d708d/config/prometheus/alerts.yaml#L14

- To convert the replication delay into seconds ramen is using  (time() - ramen_last_sync_timestamp_seconds) on alerting rule. 
- Here, UI is reading the replication delay for each PVC from CR. Just consider the values are already in seconds.
- In one corner case, When ramen_last_sync_timestamp_seconds metric is empty or zero, The expression (time() - 0) will consider the least Prometheus second, i.e 1970 which is equal to 1697788182.
- So the alert will be raised for empty value also, To match that in UI, For empty or zero seconds, I am using 1697788182  as a default value in UI.
- So the UI also will reflect the replication health issue in the dashboard.

GowthamShanmugam: I can simply consider empty value as  warning or critical for health status. But when someone read code, they dont understand why we are using empty as critical/warning.

PR #1092, opened 2023-10-28
Syncing latest changes from master for odf-console

df-build-team:
PR containing the latest commits from master branch


Nikhil-Ladha:
@red-hat-storage/odf-console-admins can we please get this PR merged, if everything looks fine to you all?

bipuladh:
@Nikhil-Ladha I am not sure what is happening here? Why do we need to manually merge the syncing PR? Also we have multiple release branches so I am not sure this is going to be enough. We have a compatibility branch as well.

Nikhil-Ladha:
> @Nikhil-Ladha I am not sure what is happening here? Why do we need to manually merge the syncing PR? Also we have multiple release branches so I am not sure this is going to be enough. We have a compatibility branch as well.

By merging, I meant adding the `lgtm` and `approve` labels.
This syncing is happening is only for the upcoming release version, that too only till dev freeze.
And, regarding the compatibility branch, I saw that we are picking up the commits for both `odf-console` and `odf-console-compatibility` from the main itself, that's the reason I didn't create the `compatibility` branch.

bipuladh:
@Nikhil-Ladha I still don't understand it if we need to sync release branches then we need to sync the compatibility branch too for every release. Also if PR creation really required? I think Openshift teams sync release branches without creating these PRs. Is that something we can do as well?

PR #1099, opened 2023-11-13
refactoring (support ODF in any Namespace)

SanjalKatiyar:
Adds a new package "immer".
Adds a new extension "console.redux-reducer", for redux state.
Adds a new hook "useOprInstallNamespaceSelector", for detecting ODF install namespace.
Removes dependency on "openshift-storage" namespace.


packages/odf/redux/reducers/opr-install-namespace.ts
```diff
@@ -0,0 +1,22 @@
...
+};
+
+export const oprInstallNamespaceReducer = (
+  state = initialState,
+  action: nsActionTypes
+): nsPayload => {
+  switch (action.type) {
+    case nsActions.SetOprInstallNs:
+      return {
+        ...state,
```
bipuladh, line 16: This performs a shallow copy. The internal properties will be equal by reference, this can cause bugs. Use `immutable` to help you work with state.
see https://immutable-js.com/

bipuladh (changes requested):

packages/odf/redux/reducers/opr-install-namespace.ts
SanjalKatiyar: done, added `immer` as per the suggestion...

packages/odf/hooks/useSafeK8sGet.ts
```diff
@@ -0,0 +1,30 @@
...
+import { getValidK8sOptions } from '@odf/shared/utils';
+import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
+import { K8sKind } from '@openshift-console/dynamic-plugin-sdk/lib/api/common-types';
+import { useOprInstallNamespaceSelector } from '../redux';
+
+/**
+ * Wrapper around "useK8sGet" hook.
+ * Ensures no API request is made unless its "safe" to do so (ODF installed namespace is fetched successfully).
+ */
+export const useSafeK8sGet = <R extends K8sResourceCommon = K8sResourceCommon>(
```
bipuladh, line 11: Is this required for the usecase when the user is not an admin?

packages/odf/components/utils/safety-box.tsx
```diff
@@ -0,0 +1,62 @@
...
+};
+
+const SafetyBox: React.FC<SafetyBoxProps> = ({
+  children,
+  isAllSafe,
+  isAllLoaded,
+  anyError,
+}) =>
+  !isAllSafe ? (
+    <StatusBox loaded={isAllLoaded} loadError={anyError} />
```
bipuladh, line 24: Doesn't `StatusBox` take a component as child?

packages/odf/hooks/useSafeK8sGet.ts
SanjalKatiyar: this can use used for any case (given we have required RBACs), it saves some unnecessary API calls... in case of non-admin, if we want we can pass `allowFallback = true` argument which would default to `openshift-storage`...

packages/odf/components/utils/safety-box.tsx
SanjalKatiyar: it does, but current way makes it possible to display `children` even in case of any error (upto consumer of the FC to decide what to do)...
one of the example:
`<StatusBox loaded={isAllLoaded} loadError={anyError} />{children}</StatusBox>` >> then in case of non-admin this will always show error as `anyError` will not be empty (RBAC issue), but if I pass `allowFallback` argument and use ternary condition (current way) I can still display `children`, if I want to.

packages/ocs/storage-class/sc-form.tsx
```diff
@@ -119,9 +129,9 @@ export const CephFsNameComponent: React.FC<ProvisionerProps> = ({
           type="text"
           value={parameterValue}
           disabled={!isExternal}
-          onChange={(e) =>
-            onParamChange(parameterKey, e.currentTarget.value, false)
-          }
+          onChange={(e) => {
+            onParamChange(parameterKey, e.currentTarget.value, false);
+          }}
```
bipuladh, line 134: Why this change?

packages/ocs/storage-class/sc-form.tsx
```diff
@@ -108,6 +110,14 @@ export const CephFsNameComponent: React.FC<ProvisionerProps> = ({
     }
   }, [sc, scLoaded, scLoadError, parameterKey]);
 
+  // ToDo (epic 4422): Need to pass the namespace where ceph cluster is deployed (remove from here, add dropdown)
+  React.useEffect(() => {
+    sessionStorage.setItem(
+      CEPH_CLUSTER_NAMESPACE_SESSION_STORAGE,
```
bipuladh, line 116:
```suggestion
      CEPH_CLUSTER_NAMESPACE,
```

packages/odf/hooks/useSafeK8sGet.ts
```diff
@@ -0,0 +1,30 @@
...
+  name?: string,
+  namespace?: string,
+  cluster?: string,
+  allowFallback = false
+): [R, boolean, unknown] => {
+  const { oprInstallNs, isNsSafe, isFallbackSafe } =
+    useOprInstallNamespaceSelector();
+
+  const canUseFallback = allowFallback && isFallbackSafe;
+  const useK8sGetArgs = getValidK8sOptions(
```
bipuladh, line 22:
```suggestion
  const K8sGetArgs = getValidK8sOptions(
```
 `use` should be prefix for hooks.

packages/odf/hooks/useSafeK8sGet.ts
```diff
@@ -0,0 +1,30 @@
+import { useK8sGet } from '@odf/shared';
+import { getValidK8sOptions } from '@odf/shared/utils';
+import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
+import { K8sKind } from '@openshift-console/dynamic-plugin-sdk/lib/api/common-types';
+import { useOprInstallNamespaceSelector } from '../redux';
+
+/**
+ * Wrapper around "useK8sGet" hook.
+ * Ensures no API request is made unless its "safe" to do so (ODF installed namespace is fetched successfully).
```
bipuladh, line 9: IIUC this assumption is incorrect. A fetch request will go but it will just be a useless call. Nevertheless we are adding this call to our stack. If you really want to avoid this I suggest you use the fetch functions such as `k8sGetResource` instead of `useK8sGet` hook. This applies for other safe functions as well.

packages/odf/redux/dispatchers/opr-install-namespace.ts
```diff
@@ -0,0 +1,17 @@
+import { useCallback } from 'react';
+import { useDispatch } from 'react-redux';
+import { nsPayload, setOprInstallNs } from '../actions';
+
+type UseOprInstallNamespaceDispatch = () => (payload: nsPayload) => void;
+
+export const useOprInstallNamespaceDispatch: UseOprInstallNamespaceDispatch =
```
bipuladh, line 7: I don't see much value of having a separate component. It's used in only one place. We can just simply dispatch the action with some args. Is this used in more than one place?

packages/odf/redux/provider-hooks/useOprInstallNamespace.ts
```diff
@@ -0,0 +1,69 @@
+import * as React from 'react';
+import { SubscriptionModel } from '@odf/shared/models';
+import { getNamespace } from '@odf/shared/selectors';
+import { SubscriptionKind } from '@odf/shared/types';
+import { k8sList } from '@openshift-console/dynamic-plugin-sdk';
+import { useOprInstallNamespaceDispatch } from '../dispatchers';
```
bipuladh, line 6: Please see that you are not adding circular dependencies with imports.

packages/odf/redux/selectors/opr-install-namespace.ts
```diff
@@ -0,0 +1,30 @@
...
+  isNsSafe: boolean;
+  isFallbackSafe: boolean;
+} => {
+  const { oprInstallNs, isOprNsLoaded, oprNsLoadError } = useSelector(
+    getOprInstallNamespace
+  );
+
+  return {
+    // installed operator namespace
+    oprInstallNs,
```
bipuladh, line 20: Could we do a better name `odfNamespace`? Also how are we handling usecases for MCO? Do we need to detect MCO as well?

packages/odf/redux/selectors/opr-install-namespace.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const useOprInstallNamespaceSelector = (): nsPayload & {
+  isNsSafe: boolean;
+  isFallbackSafe: boolean;
+} => {
+  const { oprInstallNs, isOprNsLoaded, oprNsLoadError } = useSelector(
+    getOprInstallNamespace
+  );
+
+  return {
```
bipuladh, line 18: If you return it as an array end users will not have to use the same name for variables. It might be better as we might have to support multiple namespaces for multiple operators.

packages/odf/hooks/useSafeK8sGet.ts
SanjalKatiyar: it won't send any api call at all because of the resource object that I am passing to `useK8sGet` (I defined `getValidK8sOptions` for the very same reason)...
and I already updated `useK8sGet` as well, there is a condition `if (kind) fetch();` that would ensure no api call...

SanjalKatiyar: same is true for other "safe" hook wrapper that I defined...

packages/ocs/storage-class/sc-form.tsx
SanjalKatiyar: I was testing something, then forgot to remove this... will update...

packages/ocs/storage-class/sc-form.tsx
SanjalKatiyar: I agree it is a long name, but it is not really a `CEPH_CLUSTER_NAMESPACE`, it is infact the "key" that we use to store the namespace in the session... will update anyway...

packages/odf/hooks/useSafeK8sGet.ts
SanjalKatiyar: ack...

packages/odf/redux/dispatchers/opr-install-namespace.ts
SanjalKatiyar: not really, but it is more of a directory structure to keep things logically grouped together...
in future if we need another dispatcher (diff use case) then we can create a separate file for that (here under `/redux/dispatchers`) and export via `index`...
maybe `useOprInstallNamespaceDispatch` is not reused, but others (in future) could be... so it's more of a way to group things under new `redux` directory...

packages/odf/hooks/useSafeK8sGet.ts
bipuladh: I don't see `if (kind) fetch();` anywhere. Can you point me to it?

packages/odf/redux/provider-hooks/useOprInstallNamespace.ts
SanjalKatiyar: I don;t think we have any circular dependency, maybe I missed somewhere ??
also we use `CircularDependencyPlugin` in webpack config... so I am assuming we are safe...

packages/odf/redux/selectors/opr-install-namespace.ts
SanjalKatiyar: but array would mean to always read "return"ed values in same order... and if I don't want to read all values leaving gaps such as: `const [oprInstallNs, _, _, isSafe] = useOprInstallNamespaceSelector();`...

if we want to rename we can use `const { oprInstallNs: test } = useOprInstallNamespaceSelector();` right ??

packages/odf/redux/selectors/opr-install-namespace.ts
SanjalKatiyar: ack !
we don't need it in MCO, there are just couple of places where we need this namespace and for that as well we need not to rely on this hook/redux (anyway MCO will be on hub and ODF on managed cluster, so this redux won't work for MCO)...

packages/shared/src/hooks/k8s-get-hook.ts
```diff
@@ -34,7 +34,8 @@ export const useK8sGet = <R extends K8sResourceCommon = K8sResourceCommon>(
         setLoaded(true);
       }
     };
-    fetch();
+    if (kind) fetch();
+    else setLoaded(true);
```
SanjalKatiyar, line 38: cc @bipuladh

SanjalKatiyar: same is true for `useK8sList.ts`

packages/odf/hooks/useSafeK8sGet.ts
SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/1099/files#r1401607319

packages/odf/redux/selectors/opr-install-namespace.ts
SanjalKatiyar: gonna be lot of changes to rename everything though !!

SanjalKatiyar: done...

bipuladh (approved):

PR #1104, opened 2023-11-20
Changes to support multiple ODF cluster from the same OCP cluster

GowthamShanmugam:


SanjalKatiyar:
for `setSubSystemMap` and `setHealthData` functions in `status-card.tsx`, should be define key of the map as `item?.metric.cluster + namespace` as well (for safety) ?? or do u think that's not needed ??

GowthamShanmugam:
> for `setSubSystemMap` and `setHealthData` functions in `status-card.tsx`, should be define key of the map as `item?.metric.cluster + namespace` as well (for safety) ?? or do u think that's not needed ??

`item?.metric.cluster + namespace`  key is only useful when we display system health status namespace-wise.  Our current UI design is like either cluster wise, or storage system wise. 

For storage system wise i used a key systemName - namesapce - cluster, for cluster wise you are already using cluster name as a key.

SanjalKatiyar:
> m health status namespace-wis

not really, what I meant is for `status-card.tsx` where we show StorageSystem status and incorporate sub-system status as well into it... in case of multiple StorageSystem there could be multiple sub-system as well and in case that happens, just using `item?.metric.cluster` will not work as same cluster can have multiple sub-systems (in multiple namespaces)...

GowthamShanmugam:
> > m health status namespace-wis
> 
> what I meant is for `status-card.tsx` where we show StorageSystem status and incorporate sub-system status as well into it... in case of multiple StorageSystem there could be multiple sub-system as well and in case that happens, just using `item?.metric.cluster` will not work as same OCP cluster can have multiple sub-systems (in multiple namespaces)...
> 
> so in `setSubSystemMap` and `setHealthData` where we can maintaining map of the sub-systems health, we might want to update the "key" of the map to be combination of `item?.metric.cluster + namespace` instead of currently `item?.metric.cluster`...

I updated the PR.

packages/mco/components/mco-dashboard/queries.ts
```diff
@@ -28,8 +29,9 @@ export const getLastSyncPerClusterQuery = () =>
 
 // ToDo (epic 4422): Need to update as per updates in the metrics
 export const CAPACITY_QUERIES = {
-  [StorageDashboard.TOTAL_CAPACITY_FILE_BLOCK]: `(label_replace(odf_system_map{target_namespace="openshift-storage"} , "managedBy", "$1", "target_name", "(.*)"))  * on (namespace, managedBy, cluster) group_right(storage_system, target_kind) ${TOTAL_CAPACITY_FILE_BLOCK_METRIC}`,
-  [StorageDashboard.USED_CAPACITY_FILE_BLOCK]: `(label_replace(odf_system_map{target_namespace="openshift-storage"} , "managedBy", "$1", "target_name", "(.*)"))  * on (namespace, managedBy, cluster) group_right(storage_system, target_kind) ${USED_CAPACITY_FILE_BLOCK_METRIC}`,
+  // ToDo: For 4.15, Assuming managedBy is unique for each StorageSystem. Need to add targer_namesapce as an another key.
```
SanjalKatiyar, line 32: I know it is a very odd comment (sorry for that), but I am prefixing all these comment with `// ToDo (epic 4422):` so that I can easily search for them later...
```suggestion
  // ToDo (epic 4422): For 4.15, Assuming "managedBy" is unique for each StorageSystem. Need to add "target_namesapce" as an another key.
```

packages/mco/components/mco-dashboard/storage-system/system-capacity-card/capacity-card.tsx
```diff
@@ -274,13 +281,15 @@ const SystemCapacityCard: React.FC = () => {
         ? usedCapacity?.data?.result?.reduce(
             (acc: CapacityMetricDatumMap, usedMetric: PrometheusResult) => {
               const systemName = usedMetric?.metric?.storage_system;
+              const namespace = usedMetric?.metric?.namespace;
```
SanjalKatiyar, line 284: what I meant was to update status and capacity queries so that, ex: `usedMetric` response will have "target_namespace" to it (`odf_system_map` has this label)...

we can map `odf_system_map` with health/capacity metrics on `managedBy` and `namespace` (as we do now) with assumption that `managedBy` is unique...

SanjalKatiyar, line 284:
```suggestion
              const namespace = usedMetric?.metric?.target_namespace;
```

SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/1104#discussion_r1403262280

SanjalKatiyar: later once health/capacity metrics are updated with new label, we can then update our queries to have mapping of `odf_system_map` with health/capacity metrics on `managedBy` and `target_namespace`...

PR #1110, opened 2023-11-28
Blocking DRPolicy creation for multiple ODF cluster

GowthamShanmugam:
1. Replaced most of the custom component implementation with Shared components.
2. Replaced clusterClaim logic by watching MulticCusterView of ODF config map type.
3. Blocking DRPolicy creation when more than one configMap is found under OCP. 
4. Unittesting.

JIRA: https://issues.redhat.com/browse/RHSTOR-4943


GowthamShanmugam:
Screenshot from 2023-11-28 19-42-33

PR #1115, opened 2023-12-05
[RHSTOR-5133] Adds Provider Mode support

bipuladh:
- Adds list page for Clients
 - Adds Modal for generating onboarding token for clients
 - Updates ODF dashboard to add Client Status
 - Adds flags for Provider Mode
 
Screenshots: 
1) List Page
 
list-page

2) Onboarding token modal
modal

3) Client mismatch popover
client-version-popover

4) Status Card
status-card

 
 

bipuladh:
Mock links: https://www.figma.com/proto/Afl7ntvufZTshGyuX701u7/Provider-mode?page-id=0%3A1&type=design&node-id=1-330&viewport=60%2C485%2C0.38&t=ngq3FBzQrHNQyEh6-1&scaling=min-zoom&starting-point-node-id=1%3A330

packages/odf/components/odf-dashboard/status-card/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+      break;
+    case StorageConsumerState.Disabled:
+    case StorageConsumerState.Failed:
+      health = HealthState.ERROR;
+      break;
+    default:
+      health = HealthState.UNKNOWN;
+  }
+  return health;
+};
```
SanjalKatiyar, line 27: are we using this anywhere ??

packages/odf/components/odf-dashboard/status-card/status-card.tsx
```diff
@@ -129,6 +138,23 @@ export const StatusCard: React.FC = () => {
...
+  const isProviderMode = useFlag(PROVIDER_MODE);
+
+  const [clients, clientsLoaded, clientsLoadError] = useK8sWatchResource<
+    StorageConsumerKind[]
+  >({
+    kind: referenceForModel(StorageConsumerModel),
+    isList: true,
+  });
+
+  const clientAggregateHealth = getAggregateClientHealthState(clients);
```
SanjalKatiyar, line 150: should we for safety to prevent type error ?? or `useK8sWatchResource` will guaranteed return empty list meanwhile when resources are loading or if in error ??
```suggestion
  const clientAggregateHealth = getAggregateClientHealthState(clients || []);
```

packages/odf/components/odf-dashboard/status-card/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+  if (totalClientsCount === healthyClientsCount) {
+    return HealthState.OK;
+  }
+  if (totalClientsCount > healthyClientsCount && healthyClientsCount > 0) {
+    return HealthState.ERROR;
+  }
+  if (totalClientsCount === 0) {
+    return HealthState.NOT_AVAILABLE;
+  }
+  return HealthState.UNKNOWN;
```
SanjalKatiyar, line 50: if `totalClientsCount = 2` and `healthyClientsCount = 0`, we are reporting it as `UNKNOWN` instead of `ERROR` ??

SanjalKatiyar: why do we need `&& healthyClientsCount > 0` part ??

packages/odf/components/odf-dashboard/status-card/status-card.tsx
SanjalKatiyar: or maybe we can directly do `state={getAggregateClientHealthState(clients)}`, because then this function will only be called when request is loaded with no error...

packages/odf/components/odf-dashboard/status-card/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+export const getClientText = (clients: StorageConsumerKind[], t: TFunction) => {
+  const connectedClients = clients.filter(
+    (client) => client.status.state === StorageConsumerState.Ready
+  );
+
+  const healthyClients = connectedClients.filter(
+    (client) => getTimeDifferenceInSeconds(client.status.lastHeartbeat) < 120
+  );
+  const healthyClientsCount = healthyClients.length;
+  const totalClientsCount = clients.length;
```
SanjalKatiyar, line 62: this entire calculation can be a separate function I guess, which takes `clients` and return healthy and total... u r also using it in `getAggregateClientHealthState`...

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -0,0 +1,304 @@
...
+      },
+      {
+        title: t('Cluster ID'),
+        props: {
+          className: tableColumns[1].className,
+        },
+        id: tableColumns[1].id,
+      },
+      {
+        title: t('Openshift Version'),
```
SanjalKatiyar, line 93:
```suggestion
        title: t('Openshift version'),
```

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -0,0 +1,304 @@
...
+      Row={StorageClientRow}
+    />
+  );
+};
+
+const getOpenshiftVersion = (obj: StorageConsumerKind) =>
+  obj.status.client.platformVersion;
+
+const getDataFoundationVersion = (obj: StorageConsumerKind) =>
+  obj.status.client.operatorVersion;
```
SanjalKatiyar, line 143: no optional changing ?? for the cases if `status` or maybe versions are missing from the CR...

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -0,0 +1,304 @@
...
+  const difference = getTimeDifferenceInSeconds(heartbeat);
+  const Component = React.useMemo(() => {
+    if (difference > 120 && difference <= 180) {
+      return YellowExclamationTriangleIcon;
+    }
+    if (difference > 180) {
+      return RedExclamationCircleIcon;
+    }
+    return null;
+  }, [difference]);
```
SanjalKatiyar, line 158: does it matter to use `useMemo` here ?? It's not like `Component`'s ref is getting changed or any thing...

SanjalKatiyar: it can either be `YellowExclamationTriangleIcon` or `RedExclamationCircleIcon`, which are not getting created on each re-render...

also this `useMemo` is gonna get executed multiple times anyway, as `difference` can highly likely to be diff across re-renders...

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -0,0 +1,304 @@
...
+  const clientVersion = getDataFoundationVersion(obj);
+  const isVersionMismatch =
+    getMajorMinorVersion(clientVersion) !==
+    getMajorMinorVersion(currentVersion);
+  return isVersionMismatch ? (
+    <Popover
+      position={PopoverPosition.top}
+      headerContent={t('Client version is out of date')}
+      bodyContent={t(
+        'Due to the misatch in the client and provider version. This provider cluster cannot be upgraded.'
```
SanjalKatiyar, line 187:
```suggestion
        'Due to the mismatch in the client and provider version this provider cluster cannot be upgraded.'
```

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -0,0 +1,304 @@
...
+        {getName(obj).split('storageconsumer-')[1]}
+      </TableData>
+      <TableData {...tableColumns[1]} activeColumnIDs={activeColumnIDs}>
+        {getOpenshiftVersion(obj)}
+      </TableData>
+      <TableData {...tableColumns[2]} activeColumnIDs={activeColumnIDs}>
+        <DataFoudationVersion obj={obj} currentVersion={currentVersion} />
+      </TableData>
+      <TableData {...tableColumns[3]} activeColumnIDs={activeColumnIDs}>
+        <LastHeartBeat heartbeat={obj.status.lastHeartbeat} />
```
SanjalKatiyar, line 227: no optional chaining ?? for case if `lastHeartbeat` is not present at all in the CR (some bug maybe)...

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -0,0 +1,304 @@
...
+  );
+};
+
+type ClientListPageProps = {
+  showTitle?: boolean;
+  namespace?: string;
+  selector?: any;
+  hideLabelFilter?: boolean;
+  hideNameLabelFilters?: boolean;
+  hideColumnManagement?: boolean;
```
SanjalKatiyar, line 239: all props are optional ??

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -0,0 +1,304 @@
...
+
+  const [data, filteredData, onFilterChange] = useListPageFilter(
+    storageClients,
+    rowFilters
+  );
+
+  const onClick = React.useCallback(() => {
+    const modalComponentProps = { isOpen: true };
+    launchModal(ClientOnBoardingModal, modalComponentProps);
+  }, [launchModal]);
```
SanjalKatiyar, line 277: do we need `useCallback` here ??
https://kentcdodds.com/blog/usememo-and-usecallback

packages/odf/components/storage-consumers/onboarding-modal.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export const ClientOnBoardingModal: ClientOnBoardingModalProps = ({
+  isOpen,
+  closeModal,
+}) => {
+  const { t } = useCustomTranslation();
+  const MODAL_TITLE = t('Client onboarding token');
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [ticket, _setTicket] = React.useState('dummy--text');
```
SanjalKatiyar, line 27: I am assuming that u r setting initial state to `dummy--text` and have `eslint-disable-next-line` comment as this modal is still WIP... and u did not forget to remove this...

packages/odf/features.ts
```diff
@@ -81,6 +85,7 @@ export const setOCSFlags = async (setFlag: SetFeatureFlag) => {
             'standalone'
         );
         setFlag(OCS_NFS_ENABLED, storageCluster?.spec?.nfs?.enable === true);
+        setFlag(PROVIDER_MODE, isProviderMode(storageCluster));
```
SanjalKatiyar, line 88: just FYI, we will need to move this to redux later... `useODFSystemFlags` is already taking care of this, u just need to add a new field for `isProviderMode` in the object (exactly same as other flags)...

plugins/odf/console-extensions.json
```diff
@@ -651,5 +657,25 @@
         "$codeRef": "obc.OBCDetailsPage"
       }
     }
+  },
+  {
+    "type": "console.navigation/href",
+    "properties": {
+      "id": "storageclients",
+      "name": "%plugin__odf-console~Storage Clients%",
```
SanjalKatiyar, line 665: u need to add a comment in `odf/i18n-comments.ts` file to make it really work... I mean that's the correct way...

SanjalKatiyar: here it will work without adding a comment to `i18n-comments.ts` file as well because u already parsed `Storage Clients` in `status-card.tsx`, but that is a coincidence.

SanjalKatiyar (changes requested):

packages/odf/components/storage-consumers/client-list.tsx
bipuladh: it should be always present. if it's not we will raise a bug on backend.

packages/odf/components/storage-consumers/client-list.tsx
bipuladh: we will raise a bug on the backend

packages/odf/components/storage-consumers/client-list.tsx
bipuladh: yes

packages/odf/components/storage-consumers/onboarding-modal.tsx
bipuladh: `_setTicket` iis not being used so need this

packages/odf/features.ts
bipuladh: We need it is a flag to disable extensions for Provider mode.

plugins/odf/console-extensions.json
bipuladh: ack

packages/odf/components/odf-dashboard/status-card/utils.ts
bipuladh: removing

packages/odf/components/odf-dashboard/status-card/status-card.tsx
bipuladh: I will add a default value in the function parameter.

packages/odf/components/storage-consumers/client-list.tsx
bipuladh: using `useMemo` as we don't need to update the value if the difference is same.

SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/1115#discussion_r1420020048

SanjalKatiyar: even then, it doesn't matter right ? I mean `Component` can either be `YellowExclamationTriangleIcon` or `RedExclamationCircleIcon`... all we are memoizing is two `if` conditions which are not expensive at all...

packages/odf/features.ts
SanjalKatiyar: aahhh !!
so how will u set this in case of multiple storage cluster CR then ?? right now we are just taking first cluster and checking for the flag, this will not work for multiple clusters...

bipuladh: are we supporting multiple clusters as part of 4.15?

packages/odf/components/odf-dashboard/status-card/utils.ts
SanjalKatiyar: u missed this one ??

packages/odf/features.ts
SanjalKatiyar: yeah 1 internal + 1 external... so multiple StorageCluster CRs can be present at a time...

SanjalKatiyar: we can filter internal one though... that will always be single

packages/odf/components/odf-dashboard/status-card/utils.ts
bipuladh: yeah. updated the first case to make sure that we have more than 0 clients.

packages/odf/components/storage-consumers/client-list.tsx
SanjalKatiyar: ack

packages/odf/components/storage-consumers/client-list.tsx
bipuladh: removed memoization

packages/odf/components/storage-consumers/client-list.tsx
bipuladh: removed hook

SanjalKatiyar:
LGTM cc @GowthamShanmugam

nehaberry:
> Mock links: https://www.figma.com/proto/Afl7ntvufZTshGyuX701u7/Provider-mode?page-id=0%3A1&type=design&node-id=1-330&viewport=60%2C485%2C0.38&t=ngq3FBzQrHNQyEh6-1&scaling=min-zoom&starting-point-node-id=1%3A330

Do I need special login /password to access this?

Also, are these UI changes available in the provider base cluster dashboard?

PR #1132, opened 2023-12-15
Hide PVC label selector that should not work for DR

GowthamShanmugam:
4.14 BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2253497
4.15 BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2256580

### DIFF
--- a/packages/mco/constants/acm.ts
+++ b/packages/mco/constants/acm.ts
@@ -35,3 +35,9 @@ export enum ClusterClaimTypes {
 export const MANAGED_CLUSTER_CONDITION_AVAILABLE =
   'ManagedClusterConditionAvailable';
 export const MANAGED_CLUSTER_JOINED = 'ManagedClusterJoined';
+
+// Search result labels
+export const LABEL = 'label';
+export const LABELS_SPLIT_CHAR = '; ';
+export const LABEL_SPLIT_CHAR = '=';
+export const DR_BLOCK_LISTED_LABELS = ['app.kubernetes.io/instance'];
--- a/packages/mco/components/modals/app-manage-policies/helper/pvc-details-wizard-content.tsx
+++ b/packages/mco/components/modals/app-manage-policies/helper/pvc-details-wizard-content.tsx
@@ -1,4 +1,10 @@
 import * as React from 'react';
+import {
+  LABEL,
+  LABELS_SPLIT_CHAR,
+  DR_BLOCK_LISTED_LABELS,
+  LABEL_SPLIT_CHAR,
+} from '@odf/mco/constants';
 import { useACMSafeFetch } from '@odf/mco/hooks/acm-safe-fetch';
 import { SearchResult } from '@odf/mco/types';
 import { MultiSelectDropdown } from '@odf/shared/dropdown/multiselectdropdown';
@@ -33,9 +39,6 @@ import { PlacementType } from '../utils/types';
 import '../../../../style.scss';
 import '../style.scss';
 
-const LABEL = 'label';
-const SPLIT_CHAR = '; ';
-
 const getPlacementTags = (pvcSelectors: PVCSelectorType[]) =>
   !!pvcSelectors.length
     ? pvcSelectors.map((pvcSelector) => [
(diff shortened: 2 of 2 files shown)


SanjalKatiyar:
I can see CSS issue in the screenshot attached to the bug, can we fix that in this PR as well ??

packages/mco/constants/acm.ts
```diff
@@ -35,3 +35,9 @@ export enum ClusterClaimTypes {
 export const MANAGED_CLUSTER_CONDITION_AVAILABLE =
   'ManagedClusterConditionAvailable';
 export const MANAGED_CLUSTER_JOINED = 'ManagedClusterJoined';
+
+// Search result labels
+export const LABEL = 'label';
+export const LABELS_SPLIT_CHAR = '; ';
```
SanjalKatiyar, line 41: why this ?? I don't think `;` is allowed character for labels...

packages/mco/constants/acm.ts
```diff
@@ -35,3 +35,9 @@ export enum ClusterClaimTypes {
...
+export const LABEL_SPLIT_CHAR = '=';
+export const DR_BLOCK_LISTED_LABELS = ['app.kubernetes.io/instance'];
```
SanjalKatiyar, line 43: This is one specific example, we can have too many such "well known" K8s labels (copying link from the bug comments: https://kubernetes.io/docs/reference/labels-annotations-taints/#labels-annotations-and-taints-used-on-api-objects). You will see same bug again if now label on PVCs are say `app.kubernetes.io/component`.
Also, not sure if it is a good idea to maintain a long list of such labels on the UI.

SanjalKatiyar: or is it guaranteed that in DR only "well known" label will be `app.kubernetes.io/component` and we don't have to block others from the above long list.

SanjalKatiyar:
need a bug for 4.15 as well...

packages/mco/constants/acm.ts
GowthamShanmugam: In ACM search API  result; char is used to separate the labels like: key1=value1;key2=value2

GowthamShanmugam:
> I can see CSS issue in the screenshot attached to the bug (labels are getting hidden if they are too long), can we fix that in this PR as well ??

ack

packages/mco/constants/acm.ts
GowthamShanmugam: yes, I realized that app.kubernetes.io/instance is one of the labels that was never suited for DR, similarly every label that changes at runtime dynamically won't suit DR.

So either we need to filter out and maintain a long list, or hide the label that has kubernetes.io as a substring.

SanjalKatiyar: `kubernetes.io` substring check can still fail for edge cases, but seems better than adding long list of names to me, @GowthamShanmugam @bipuladh wdyt ??

packages/mco/constants/acm.ts
SanjalKatiyar: ack...

packages/mco/constants/acm.ts
ShyamsundarR: The label list to maintain would always be a set of well known k8s (or Openshift) labels. Among which, only those that are unsuitable for grouping should be removed (e.g instance).

The other example `component` being discussed is potentially  very useful label to use, to separate components in a workload and hence only DR protect some components etc.

Currently there does not seem to be any other label [here](https://kubernetes.io/docs/reference/labels-annotations-taints/#labels-annotations-and-taints-used-on-api-objects) that needs special attention. So I would not filter out all `k8s.io` labels as that is not correct.

GowthamShanmugam: @SanjalKatiyar for now, I can see only the instance is causing issue, in case if we find anything we will update this list. When the list grows using a pattern I will try to optimize this solution.

SanjalKatiyar:
> > I can see a CSS issue in the screenshot attached to the bug (labels are getting hidden if they are too long), can we fix that in this PR as well ??
> 
> As I checked, I dont think we can't do much here, it is happening due to the wizard component. It displays all the contents inside of it. Adding x overflow looks very odd.

so you mean that we can not apply properties like `word-break`, `word-wrap` etc to the dropdown items ?? or maybe make modal a bit wider ("large" variant) ??

SanjalKatiyar:
> > > I can see a CSS issue in the screenshot attached to the bug (labels are getting hidden if they are too long), can we fix that in this PR as well ??
> > 
> > 
> > As I checked, I dont think we can't do much here, it is happening due to the wizard component. It displays all the contents inside of it. Adding x overflow looks very odd.
> 
> so you mean that we can not apply properties like `word-break`, `word-wrap` etc to the dropdown items ?? or maybe make modal a bit wider ("large" variant) ??

keeping this as part of different PR and un-blocking this one...

PR #1137, opened 2023-12-20
Bug 2255340: Updates CephMonCountLow alert modal errors

bipuladh:


packages/odf/modals/configure-mons/configure-mons.tsx
```diff
@@ -42,27 +47,31 @@ const LowMonAlertModal: ModalComponent = ({ closeModal }) => {
   const [inProgress, setProgress] = React.useState(false);
 
   // ToDo (epic 4422) (bipuladh): Update it to use information from the alert.
-  const [storageCluster, storageClusterLoaded, storageCluserLoadError] =
-    useSafeK8sGet(StorageClusterModel, null, odfNamespace);
+  const [storageClusters, storageClusterLoaded, storageCluserLoadError] =
+    useSafeK8sList<StorageClusterKind>(StorageClusterModel, odfNamespace);
```
SanjalKatiyar, line 51: there won't be multiple StorageClusters in same namespace (if there are all will be in ignore state, except first one)...
u might want to do this: https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/redux/provider-hooks/useODFSystemFlags.ts#L34

packages/odf/modals/configure-mons/configure-mons.tsx
```diff
@@ -42,27 +47,31 @@ const LowMonAlertModal: ModalComponent = ({ closeModal }) => {
...
   const [nodes, nodesLoaded, nodesLoadError] = useK8sList<NodeKind>(NodeModel);
   const failureDomains =
     nodesLoaded && !nodesLoadError ? groupNodesByZones(nodes)?.length : [];
   const Header = <ModalHeader>{t('Configure monitor count')}</ModalHeader>;
 
+  const internalCluster = storageClusters?.find((sc) => !isExternalCluster(sc));
```
SanjalKatiyar, line 57:
```suggestion
  const internalCluster = storageClusters?.find((sc) => !!isClusterIgnored && !isExternalCluster(sc));
```

packages/odf/modals/configure-mons/configure-mons.tsx
SanjalKatiyar: or just https://github.com/red-hat-storage/odf-console/pull/1137/files#r1432417326 would work as well...

PR #1143, opened 2023-12-28
Updates Onboarding modal to connect with provider proxy

bipuladh:
Screenshot: 
Screenshot from 2023-12-27 18-55-12


packages/odf/components/storage-consumers/onboarding-modal.tsx
```diff
@@ -23,22 +25,45 @@ export const ClientOnBoardingModal: ClientOnBoardingModalProps = ({
...
-  const [ticket, _setTicket] = React.useState('dummy--text');
-  // Todo(bipuladh): Add a HTTP request to proxy to get ticket
+  const [ticket, setTicket] = React.useState('');
 
   const onCopyToClipboard = () => {
     navigator.clipboard.writeText(ticket);
   };
 
+  React.useEffect(() => {
+    consoleFetch(
```
SanjalKatiyar, line 35: can we use `useSafeFetch` instead (we have hook already under `shared`) ?? which implements abort-controller as well...

packages/odf/components/storage-consumers/onboarding-modal.tsx
```diff
@@ -23,22 +25,45 @@ export const ClientOnBoardingModal: ClientOnBoardingModalProps = ({
...
+          throw new Error('Network response is not ok!');
+        }
+        return response.text();
+      })
+      .then((text) => {
+        setTicket(text);
+      })
+      .catch((error) => {
+        // eslint-disable-next-line no-console
+        console.error('Provider proxy is not working as expected', error);
```
SanjalKatiyar, line 52: we don't wanna show error to the user (just loading state or response) ??

packages/odf/components/storage-consumers/onboarding-modal.tsx
```diff
@@ -23,22 +25,45 @@ export const ClientOnBoardingModal: ClientOnBoardingModalProps = ({
...
+      });
+  }, []);
+
```
GowthamShanmugam, line 55: In MCO, I am also using similar hook like this ```useACMSafeFetch```. Shall we move my hook to shared and use in both the places?

It will reduce code size.

packages/odf/components/storage-consumers/onboarding-modal.tsx
```diff
@@ -23,22 +25,45 @@ export const ClientOnBoardingModal: ClientOnBoardingModalProps = ({
...
 
   const onCopyToClipboard = () => {
     navigator.clipboard.writeText(ticket);
   };
 
+  React.useEffect(() => {
+    consoleFetch(
+      '/api/proxy/plugin/odf-console/provider-proxy/onboarding-tokens',
+      {
+        method: 'post',
```
GowthamShanmugam, line 38: this is POST method but does not have any body ?

packages/odf/components/storage-consumers/onboarding-modal.tsx
```diff
@@ -23,22 +25,45 @@ export const ClientOnBoardingModal: ClientOnBoardingModalProps = ({
...
-  // Todo(bipuladh): Add a HTTP request to proxy to get ticket
+  const [ticket, setTicket] = React.useState('');
 
   const onCopyToClipboard = () => {
     navigator.clipboard.writeText(ticket);
   };
 
+  React.useEffect(() => {
+    consoleFetch(
+      '/api/proxy/plugin/odf-console/provider-proxy/onboarding-tokens',
```
GowthamShanmugam, line 36: Upto ```/api/proxy/plugin``` is common for all proxy redirection lets declare this as a common constant.

`${CONSOLE_PROXY_ENDPOINT}/${ODF_CONSOLE_PLUGIN}/provider-proxy/onboarding-tokens` 

you can declare provider proxy endpoint also as constants but optional.

packages/odf/components/storage-consumers/onboarding-modal.tsx
bipuladh: we can't. we use consoleFetchJSON underneath safeFetch. the server replies with plain/text. the whole thing crashes.

packages/odf/components/storage-consumers/onboarding-modal.tsx
bipuladh: The error should never occur.

packages/odf/components/storage-consumers/onboarding-modal.tsx
bipuladh: the problem with that is provider mode uses plain/text whereas everywhere else JSON is the standard. Since because of that we need to use fetch as raw as possible.

packages/odf/components/storage-consumers/onboarding-modal.tsx
bipuladh: Yes. This is how they have designed the endpoints for provider server.

packages/odf/components/storage-consumers/onboarding-modal.tsx
bipuladh: this requires a minor refactor to all three packages. We should raise either a Jira/.GH issue. It would be better to create a function that returns this URL. This PR needs to be backported up to 4.14. So the less the changes the safer it is for us.

packages/odf/components/storage-consumers/onboarding-modal.tsx
SanjalKatiyar: there could be network errors, right ??

SanjalKatiyar:
LGTM (cc @GowthamShanmugam for tagging)

GowthamShanmugam (approved):

Issue #879, opened 2023-05-29
Update "@odf/odf-plugin-sdk" to support Yup object schema


SanjalKatiyar:
https://github.com/red-hat-storage/odf-console/pull/862

Details: https://github.com/red-hat-storage/odf-console/pull/836#issuecomment-1534494046

PR #1149, opened 2024-01-03
Fix default RBD StorageClass option for external mode

SanjalKatiyar:
https://bugzilla.redhat.com/show_bug.cgi?id=2256566

`spec.managedResources.cephBlockPools.defaultStorageClass` field was not getting set for external mode StorageCluster CR.

### DIFF
--- a/packages/odf-plugin-sdk/extensions/external-storage/external-storage.ts
+++ b/packages/odf-plugin-sdk/extensions/external-storage/external-storage.ts
@@ -61,6 +61,7 @@ export type CreatePayload<S = ExternalState> = (payloadOptions: {
   namespace: string;
   storageClassName?: string;
   inTransitStatus?: boolean;
+  shouldSetCephRBDAsDefault?: boolean;
 }) => Payload[];
 
 export type Payload = { model: Model; payload: K8sResourceKind };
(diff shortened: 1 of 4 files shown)


GowthamShanmugam (commented):
LGTM

packages/odf-plugin-sdk/extensions/external-storage/external-storage.ts
```diff
@@ -61,6 +61,7 @@ export type CreatePayload<S = ExternalState> = (payloadOptions: {
   namespace: string;
   storageClassName?: string;
   inTransitStatus?: boolean;
+  shouldSetCephRBDAsDefault?: boolean;
```
bipuladh, line 64: Lets make this a mandatory field so that it's either set as false or true.

SanjalKatiyar: it optional because `CreatePayload` is used by IBM package as well (via custom extension), but they won't need `inTransitStatus` or `shouldSetCephRBDAsDefault`, similarly `storageClassName` is needed by IBM package but not RHCS...

SanjalKatiyar: Maybe we can have better types for the custom extension, I have an issue for that: https://github.com/red-hat-storage/odf-console/issues/879. We can look into it then, I guess ?

bipuladh: ack

GowthamShanmugam (approved):

Issue #1119, opened 2023-12-07
Update ReadMe with details about Redux


SanjalKatiyar:
`useODFNamespace` and `useODFSystemFlags` are 2 hooks at store info to the redux, add a documentation regarding the same and their general usage.

SanjalKatiyar:
not needed, closing it.

PR #1150, opened 2024-01-03
add redux details to the ReadMe

SanjalKatiyar:
https://github.com/red-hat-storage/odf-console/issues/1119

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,9 +27,9 @@ After the OCP console is set as required by ODF Console. Performs the following 
 3. Clone this repo.
 4. Pull all required dependencies by running `yarn install`.
 5. Run the development mode of odf-console using `I8N_NS=plugin__odf-console PLUGIN=odf yarn run dev:c`. This runs a container running both the console bridge and a webserver in port 9001.
-6. For more OCP container related environment variables [Refer](https://github.com/red-hat-storage/odf-console/scripts/start-ocp-console.sh).
+6. For more OCP container related environment variables refer the [script](https://github.com/red-hat-storage/odf-console/scripts/start-ocp-console.sh).
 
-### Unit Tests
+## Unit Tests
 
 Run all unit tests:
 
@@ -39,7 +39,7 @@ yarn test
 yarn test-coverage
 ```
 
-### E2E Tests
+## E2E Tests
 
 E2E tests are written in [Cypress](https://www.cypress.io/).
 
@@ -59,7 +59,7 @@ yarn test-cypress-headless
 
 By default, it will look for Chrome in the system and use it, but if you want to use Firefox instead, set BRIDGE_E2E_BROWSER_NAME environment variable in your shell with the value firefox.
 
-#### Build the CI runner image
+## Build the CI runner image
 
 Build a beta for testing:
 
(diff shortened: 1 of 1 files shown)


SanjalKatiyar:
cc @GowthamShanmugam @bipuladh

GowthamShanmugam (approved):
LGTM

README.md
```diff
@@ -75,3 +75,15 @@ docker push quay.io/ocs-dev/odf-console-ci-runner:beta
...
 Only for testing purposes (don't merge this), add a separate commit updating `.ci-operator.yaml`:
 `tag: beta`
+
+## Redux
+
+- Console's redux store is accessed via an extension point [console.redux-reducer](https://github.com/openshift/console/blob/master/frontend/packages/console-dynamic-plugin-sdk/docs/console-extensions.md#consoleredux-reducer), refer the [file](https://github.com/red-hat-storage/odf-console/blob/master/plugins/odf/console-extensions.json).
+- Currently we inject 2 [reducers](https://github.com/red-hat-storage/odf-console/tree/master/packages/odf/redux/reducers):
+  1.  `odfNamespaceReducer`: For storing the Namespace where ODF operator is installed.
+  2.  `odfSystemFlagsReducer`: For setting the feature flags, used for enabling/disabling certain console components.
+
```
bipuladh, line 85: Remove this. We don't need to update our Readme everytime we add reducer extensions.

README.md
```diff
@@ -75,3 +75,15 @@ docker push quay.io/ocs-dev/odf-console-ci-runner:beta
...
+**Note:**
+
+1. Only rely on the redux store if data needs to be accessed by multiple un-related components. Whereever possible use [Context](https://legacy.reactjs.org/docs/context.html) API instead.
```
bipuladh, line 88: Lets point to latest docs https://react.dev/learn/passing-data-deeply-with-context

README.md
```diff
@@ -75,3 +75,15 @@ docker push quay.io/ocs-dev/odf-console-ci-runner:beta
...
+2. [useODFSystemFlags](https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/redux/provider-hooks/useODFSystemFlags.ts#L32) hook opens up certain websocket connections as soon as plugin is enabled. Utilize these at different places (example: dashboard, wizard etc) and poll for entire list of `StorageCluster`, `CephCluster`, `CephObjectStore` and `NooBaa` CRs and filter later.
```
bipuladh, line 89: This is better as a comment next to the hook. Adding it as a doctext will allow users of this hook to glance these details directly from the editor.

bipuladh (commented):
/hold

README.md
SanjalKatiyar: but this Note is not only for the user of `useODFSystemFlags` hook... it is general FYI, where any component which need to access `StorageCluster`, `CephCluster`, `CephObjectStore` and `NooBaa` CRs can utilise the existing connections...

README.md
SanjalKatiyar: ack...

README.md
SanjalKatiyar: ack...

README.md
bipuladh: but this is something that is not required in the README file. We can provide information about the redux extensions a bit but we don't need to explain everything here.

SanjalKatiyar: okay, will remove it... but we need to take care of this during PR reviews and make sure we are passing same `k8sResource` object as done in `useODFSystemFlags`, so that we can re-use existing connections...

SanjalKatiyar: so it is not really and explanation, but more of a practice that we should follow...

SanjalKatiyar:
not needed for now...

PR #1154, opened 2024-01-03
Bug 2253497:Hide PVC label selector that should not work for DR

GowthamShanmugam:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2253497


GowthamShanmugam:
4.14 backport is canceled, BZ also closed

PR #1156, opened 2024-01-03
[MDR] Application view saying PVCs unhealthy but  cluster view shows healthy

GowthamShanmugam:
4.14 BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2256633
4.15 BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2257236

### DIFF
--- a/packages/mco/components/mco-dashboard/data-policy/cluster-app-card/common.tsx
+++ b/packages/mco/components/mco-dashboard/data-policy/cluster-app-card/common.tsx
@@ -7,6 +7,7 @@ import {
   ALL_APPS,
   ALL_APPS_ITEM_ID,
   LEAST_SECONDS_IN_PROMETHEUS,
+  REPLICATION_TYPE,
   VOLUME_REPLICATION_HEALTH,
 } from '@odf/mco/constants';
 import {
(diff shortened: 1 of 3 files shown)


GowthamShanmugam:
image
image

packages/mco/components/mco-dashboard/data-policy/cluster-app-card/common.tsx
```diff
@@ -76,12 +77,15 @@ export const VolumeSummarySection: React.FC<VolumeSummarySectionProps> = ({
...
-      )[0];
+      const health =
+        pvcData.replicationType === REPLICATION_TYPE.ASYNC
+          ? getVolumeReplicationHealth(
+              !!pvcLastSyncTime
+                ? getTimeDifferenceInSeconds(pvcLastSyncTime)
+                : LEAST_SECONDS_IN_PROMETHEUS,
+              pvcData?.schedulingInterval
+            )[0]
+          : VOLUME_REPLICATION_HEALTH.HEALTHY;
```
SanjalKatiyar, line 88: couple of questions here:
1. `VolumeSummarySection` and `ProtectedPVCsSection`, same cards are shared by both app-wise and cluster-wise view. So how one was reporting PVC as healthy and other unhealthy ??
2. If its MDR (as u r checking on condition `REPLICATION_TYPE.ASYNC`), are PVCs always healthy in that ??

GowthamShanmugam: VolumeSummarySection and ProtectedPVCsSection are reporting a wrong status for an application as well as the cluster view. 

Half part of the dashboard is reporting applications are healthy: Total applications, And half part is reporting unhealthy: VolumeSummarySection and ProtectedPVCsSection,


REPLICATION_TYPE.ASYNC is ensure the replication is RDR, only for RDR we need to check volume replication health, MDR does not have volume replication(it is same ceph strech cluster), So for MDR we are displaying always healthy.

PR #1159, opened 2024-01-08
[RDR] Optimise cluster for RDR dialog fails with error msg

GowthamShanmugam:
BZ 4.15: https://bugzilla.redhat.com/show_bug.cgi?id=2257222

### DIFF
--- a/packages/ocs/dashboards/persistent-internal/status-card/osd-migration/osd-migration-progress.tsx
+++ b/packages/ocs/dashboards/persistent-internal/status-card/osd-migration/osd-migration-progress.tsx
@@ -10,10 +10,7 @@ import { CephClusterKind } from '@odf/shared/types';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
 import { ViewDocumentation } from '@odf/shared/utils';
 import { HealthState } from '@openshift-console/dynamic-plugin-sdk';
-import {
-  HealthBody,
-  HealthItem,
-} from '@openshift-console/dynamic-plugin-sdk-internal';
+import { HealthItem } from '@openshift-console/dynamic-plugin-sdk-internal';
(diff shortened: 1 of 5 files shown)


packages/odf/constants/data-protection.ts
```diff
@@ -1,6 +1,10 @@
 export const DISASTER_RECOVERY_TARGET_ANNOTATION =
   'ocs.openshift.io/clusterIsDisasterRecoveryTarget';
 
+// "~1" is used to represent a "/", else any "patch" call will treat prefix as a path
+export const DISASTER_RECOVERY_TARGET_ANNOTATION_WO_SLASH =
+  'ocs.openshift.io~1clusterIsDisasterRecoveryTarget';
+
```
SanjalKatiyar, line 7: when we create StorageSystem, then also we might want to fix this ??

SanjalKatiyar: https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/utils/common.ts#L486

GowthamShanmugam: Initial deployment is working fine from 4.14, only path based update is a problem

packages/ocs/dashboards/persistent-internal/status-card/osd-migration/osd-migration-progress.tsx
```diff
@@ -60,78 +57,76 @@ export const OSDMigrationProgress: React.FC<OSDMigrationProgressProps> = ({
   return (
     <>
       {migrationStatus !== OSDMigrationStatus.PENDING && <Divider />}
-      <HealthBody>
```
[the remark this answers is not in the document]
GowthamShanmugam: I moved OSDMigrationProgress component inside HealthBody on status-card.ts

PR #1171, opened 2024-01-17
Bug 2257236: Application showing wrong status of protected PVC but main dashboard showing healthy

GowthamShanmugam:


GowthamShanmugam:
4.14 BZ was closed and no need to backport

PR #1204, opened 2024-02-02
Rename data policies navigation to disaster recovery

GowthamShanmugam:
Issue: https://issues.redhat.com/browse/RHSTOR-5388


GowthamShanmugam:
image

plugins/mco/console-extensions.json
```diff
@@ -30,11 +30,11 @@
   {
     "type": "console.navigation/href",
     "properties": {
-      "id": "mco-data-policies",
+      "id": "mco-disaster-recovery",
       "perspective": "acm",
       "section": "mco-data-services",
-      "name": "%plugin__odf-multicluster-console~Data policies%",
-      "href": "/multicloud/data-services/data-policies"
+      "name": "%plugin__odf-multicluster-console~Disaster recovery%",
```
SanjalKatiyar, line 36: there is a file `i18n-comments.ts` in odf, we need to create another one under "mco" and mark texts like `Disaster recovery`, `Data Services` etc there... else these will not get translated...

SanjalKatiyar: it is a small thing, not blocking this PR... but something to keep notice of and fix later...

GowthamShanmugam: ack

PR #1210, opened 2024-02-06
Discovered application DR enroll Wizard - Step1

GowthamShanmugam:
Issue: https://issues.redhat.com/browse/RHSTOR-4954


GowthamShanmugam:
image

packages/shared/src/table/selectable-table.tsx
```diff
@@ -115,71 +117,76 @@ export const SelectableTable: SelectableTableProps = <
...
-      loaded={loaded}
-      EmptyMsg={emptyRowMessage}
-      data={sortedRows}
-      skeleton={<div className="loading-skeleton--table" />}
-    >
-      <TableComposable
-        translate={null}
-        aria-label="Selectable table"
-        variant="compact"
+    <div className={className}>
```
GowthamShanmugam, line 120: moved inside div

SanjalKatiyar: can't we pass this `className` to `TableComposable` ?? do we have to add a new `div` ??

packages/mco/components/discovered-application-wizard/enroll-discovered-application.scss
```diff
@@ -0,0 +1,7 @@
+@import '../../style.scss';
+
+.mco-discovered-application {
+    &__wizard--height {
+        max-height: 74vh;
+    }
+}
```
SanjalKatiyar, line 7: will replacing this directly with any of https://www.patternfly.org/utility-classes/sizing#usage-1 works ?? If it does then we don't have to create a separate scss file/class just for height.

packages/mco/components/discovered-application-wizard/footer.tsx
```diff
@@ -0,0 +1,145 @@
...
+              '1 or more mandatory fields are empty. To proceed, fill in the required information.'
+            )}
+            variant={AlertVariant.danger}
+            isInline
+          />
+        )}
+        <WizardFooter>
+          <Button
+            isLoading={requestInProgress}
+            isDisabled={requestInProgress}
```
SanjalKatiyar, line 105:
```suggestion
            isDisabled={requestInProgress || validationError}
```

packages/mco/components/discovered-application-wizard/footer.tsx
```diff
@@ -0,0 +1,145 @@
...
+            variant="primary"
```
SanjalKatiyar, line 106: PF provides `ButtonVariant` for this...
```suggestion
            variant="primary"
```

packages/mco/components/discovered-application-wizard/footer.tsx
```diff
@@ -0,0 +1,145 @@
...
+            {stepName ===
+            enrollDiscoveredApplicationStepNames[
+              EnrollDiscoveredApplicationSteps.Review
+            ]
+              ? t('Assign')
+              : t('Next')}
+          </Button>
+          {/* Disabling the back button for the first step in wizard */}
+          <Button
+            variant="secondary"
```
SanjalKatiyar, line 119: PF provides `ButtonVariant` for this...

packages/mco/components/discovered-application-wizard/footer.tsx
```diff
@@ -0,0 +1,145 @@
...
+            onClick={onBack}
+            isDisabled={stepId === 1 || requestInProgress}
+          >
+            {t('Back')}
+          </Button>
+          <Button
+            variant="link"
```
SanjalKatiyar, line 126: same here `ButtonVariant`...

packages/shared/src/table/selectable-table.tsx
GowthamShanmugam: As per the design, i want the same max height, border, margin for for Table as well as the StatusBox(while loading skeleton and for displaying an empty message). Unfortunately, StatusBox does not take any class name, And i checked the StatusBox code it has lot of return. So i have only two options left, add div inside SelectableTable or add div while calling Selectable table.

All examples:
image

image

image

packages/mco/components/discovered-application-wizard/footer.tsx
GowthamShanmugam: No need to disable the button, The code won't allow to the next page,  it is just like  ACM Appset wizard flow.

packages/mco/components/discovered-application-wizard/utils/reducer.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export type EnrollDiscoveredApplicationState = {
+  namespace: {
+    // Single or multi namespaces of the discovered application
+    namespaces: K8sResourceCommon[];
+    // Cluster name of the discovered application
+    clusterName: string;
+    // All eligible polcies for the cluster
+    drPolicies: DRPolicyKind[];
```
SanjalKatiyar, line 18: do we need to store entire list here ?? can't we just calculate it in step-3 when we need it for displaying a dropdown ??

packages/mco/components/discovered-application-wizard/footer.tsx
SanjalKatiyar: ack !

packages/mco/components/discovered-application-wizard/utils/reducer.ts
SanjalKatiyar: nits: or maybe pass as a prop instead... it is not really a state for this create form...

packages/mco/components/discovered-application-wizard/wizard-steps/namespace-step/namesapce-table.tsx
```diff
@@ -0,0 +1,186 @@
...
+    </Td>
+  );
+};
+
+export const NamespaceSelectionTable: React.FC<NamespaceSelectionTableProps> =
+  ({ namespaces, clusterName, policies, isValidationEnabled, dispatch }) => {
+    const { t } = useCustomTranslation();
+
+    const [drPlacements, drpcLoaded, drpcLoadError] =
+      useK8sList<DRPlacementControlKind>(DRPlacementControlModel);
```
SanjalKatiyar, line 80: nits: pass namespace as well, assuming all imperative DRPC will be created in a single namespace (say `openshift-operators`)... this will reduce the number of drpcs on which we need to loop over later...

packages/mco/components/discovered-application-wizard/wizard-steps/namespace-step/namesapce-table.tsx
```diff
@@ -0,0 +1,186 @@
...
+  drPlacements: DRPlacementControlKind[],
+  policies: DRPolicyKind[]
+): string[] => {
+  if (!!drPlacements?.length) {
+    // Find all the protected namespaces from DRPCs.
+    // By comparing the eligible DRPolicies, filtering the DRPCs and protected namespaces for the peer clusters.
+    const policyNames: string[] = policies.map(getName);
+    return drPlacements.reduce((acc, drpc) => {
+      if (policyNames.includes(drpc.spec.drPolicyRef.name)) {
+        acc.push(
```
SanjalKatiyar, line 45: shouldn't we use "set" here instead of "list" ? or each drpc will guaranteed provide unique set of namespaces ?

SanjalKatiyar: I think it is latter, right ?

packages/mco/components/discovered-application-wizard/wizard-steps/namespace-step/namespace-step.scss
```diff
@@ -0,0 +1,23 @@
...
+        font-size: var(--pf-global--FontSize--lg);
+    }
+    &__dropdown--width {
+        width: 27rem;
+    }
+    &__bullseye--padding {
+        padding: 5rem 0;
+    }
+    &__alert--margin {
+        margin-top: var(--pf-global--spacer--sm);
```
SanjalKatiyar, line 21: we can reduce scss on our side by utilising: https://www.patternfly.org/utility-classes/spacing, things like margin/padding/width etc can easily be set using `pf-` based classes...

packages/mco/components/discovered-application-wizard/wizard-steps/namespace-step/namesapce-table.tsx
SanjalKatiyar, line 80: also, u don't wanna use watch resources for some reason ?

packages/mco/components/discovered-application-wizard/wizard-steps/namespace-step/namespace-step.tsx
```diff
@@ -0,0 +1,154 @@
...
+  ));
+
+export const NamespaceSelection: React.FC<NamespaceSelectionProps> = ({
+  state,
+  isValidationEnabled,
+  dispatch,
+}) => {
+  const { t } = useCustomTranslation();
+
+  const [drPolicies, loaded, loadError] =
```
SanjalKatiyar, line 58: same here, is there any reason we don't wanna use watch resources ??

packages/mco/utils/acm-search-quries.ts
```diff
@@ -0,0 +1,94 @@
...
+          },
+        ],
+        limit: 20000, // search said not to use unlimited results
+      },
+    ],
+  },
+  query: searchFilterQuery,
+});
+
+export const convertSearchResultToK8sObject = (
```
SanjalKatiyar, line 65: can we rename it to specifically the Object we are converting to ? like namespace... as diff object will have diff struct...

packages/shared/src/table/selectable-table.tsx
SanjalKatiyar: for status box, u can do this `skeleton={<div className="loading-skeleton--table pf-u-mt-lg" />}` directly... `pf-u-mt-lg` will resolve the issue with skeleton... same for empty message, that class can go in the `emptyRowMessage` FC.

and for other cases u can directly pass the class to `TableComposable`, do we still need extra div ??

SanjalKatiyar: we can make it more generic where we allow passing all the props to `TableComposable`...  like we do here: https://github.com/red-hat-storage/odf-console/blob/master/packages/shared/src/table/selectable-table.tsx#L20

SanjalKatiyar: for status box, u can do this skeleton={<div className="loading-skeleton--table pf-u-mt-lg" />} directly... pf-u-mt-lg will resolve the issue with skeleton...

SanjalKatiyar: and for other case, IMO we can add `div` while calling Selectable table, because I don't think it is needed by all components that will be using this table... so whereever needed we can add it, use per basis...

SanjalKatiyar: just "nits", upto you...

locales/en/plugin__odf-console.json
```diff
@@ -50,6 +50,32 @@
...
+  "Back": "Back",
+  "Name": "Name",
+  "{{count}} results found for {{clusterName}}_one": "{{count}} results found for {{clusterName}}",
+  "{{count}} results found for {{clusterName}}_other": "{{count}} results found for {{clusterName}}",
+  "0 results found": "0 results found",
+  "Select a namespace": "Select a namespace",
+  "There are no namespaces to display. Please select a cluster first to view namespaces.": "There are no namespaces to display. Please select a cluster first to view namespaces.",
+  "There are no namespaces to display.": "There are no namespaces to display.",
+  "This list does not include namespaces where applications are enrolled separately under disaster recovery protection.": "This list does not include namespaces where applications are enrolled separately under disaster recovery protection.",
+  "Choose a DR cluster for selecting your namespace": "Choose a DR cluster for selecting your namespace",
```
SanjalKatiyar, line 70: this text is a but diff in the latest mock-ups...

packages/shared/src/table/selectable-table.tsx
GowthamShanmugam: ack, i will try this

packages/mco/components/discovered-application-wizard/enroll-discovered-application.scss
GowthamShanmugam: i am getting multiple scroll for pf-v5-u-h-75vh and pf-v5-u-h-66vh is small, i want the custom size.

packages/mco/components/discovered-application-wizard/footer.tsx
GowthamShanmugam: ack

packages/mco/components/discovered-application-wizard/wizard-steps/namespace-step/namespace-step.tsx
GowthamShanmugam: policies won't change very often, per cluster users may use one or two policies. And From policies, i am not using the spec/status which changes very frequently.

packages/mco/components/discovered-application-wizard/wizard-steps/namespace-step/namesapce-table.tsx
GowthamShanmugam: in my case it is different, I have to read all DRPC to makesure the app is not an application-specific protected as well as namespace protected

for example, Appset application1 can be app-specific protected from the ACM page, same can be namespace protected from the namespace wizard. I have to look for other namespaces DRPC to makesure the namespace is not already protected.

packages/mco/utils/acm-search-quries.ts
GowthamShanmugam: only common info of K8sResourceCommon

PR #1211, opened 2024-02-07
Remove support for Single Replica pool from wizard flow

bipuladh:
Removes single replica pool feature from cluster creation flow


SanjalKatiyar:
/retitle Remove support for Single Replica pool from wizard flow

GowthamShanmugam:
LGTM

PR #1212, opened 2024-02-07
Adds fallback layer for multi cluster code in External mode

bipuladh:


packages/ocs/dashboards/persistent-external/utilization-card.tsx
```diff
@@ -29,7 +30,9 @@ export const UtilizationContent: React.FC = () => {
   const { systemFlags } = useODFSystemFlagsSelector();
 
   // name of created StorageClasses are prefix by StorageCluster name
-  const storageClassNamePrefix = systemFlags[clusterNs]?.ocsClusterName;
+  // name of created StorageClasses are prefix by StorageCluster name
```
SanjalKatiyar, line 33: nits
```suggestion
```

packages/ocs/dashboards/persistent-external/fallback-hook.ts
```diff
@@ -0,0 +1,33 @@
...
+export const storageClusterResource: WatchK8sResource = {
+  isList: true,
+  kind: referenceForModel(OCSStorageClusterModel),
+};
+
+const fsProvisionerName = '.cephfs.csi.ceph.com';
+
+const useFallbackHook = (storageClusterName) => {
+  const [storageClasses, storageClassesLoaded, storageClassLoadError] =
+    useK8sWatchResource<StorageClassResourceKind[]>(storageClusterResource);
```
SanjalKatiyar, line 18: don't u wanna pass StorageClass resource instead...
```suggestion
    useK8sWatchResource<StorageClassResourceKind[]>(storageClassResource);
```

packages/ocs/dashboards/persistent-external/fallback-hook.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  const provisioners = storageClasses.map((sc) => sc.provisioner);
+  const fileSystemProvisioner = provisioners.find((item) =>
+    item.includes(fsProvisionerName)
+  );
+  const clusterName = fileSystemProvisioner?.split(fsProvisionerName)[0];
+  return storageClassesLoaded && !storageClassLoadError
+    ? clusterName
+    : 'ocs-storagecluster';
```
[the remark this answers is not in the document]
SanjalKatiyar: also, during multiple StorageClusyter epic we removed all explicit reference to term `ocs-external-storagecluster` & `ocs-storagecluster` (as name can be configurable in future)... can we read this value from some CR instead ??

packages/ocs/dashboards/persistent-external/fallback-hook.ts
```diff
@@ -0,0 +1,33 @@
...
+  WatchK8sResource,
+  useK8sWatchResource,
+} from '@openshift-console/dynamic-plugin-sdk';
+
+export const storageClusterResource: WatchK8sResource = {
+  isList: true,
+  kind: referenceForModel(OCSStorageClusterModel),
+};
+
+const fsProvisionerName = '.cephfs.csi.ceph.com';
```
SanjalKatiyar, line 14: nits
```suggestion
const fsProvisionerPostFix = '.cephfs.csi.ceph.com';
```

packages/ocs/dashboards/persistent-external/fallback-hook.ts
GowthamShanmugam: My confusion is for a storageClass watcher why storageClusterModel is used? correct me if I am missing anything about multi-cluster.

packages/ocs/dashboards/persistent-external/utilization-card.tsx
```diff
@@ -29,7 +30,9 @@ export const UtilizationContent: React.FC = () => {
...
+  const storageClassName = systemFlags?.[clusterNs]?.ocsClusterName;
+  const storageClassNamePrefix = useFallbackHook(storageClassName);
```
SanjalKatiyar, line 35: nits
```suggestion
  const ocsClusterName = systemFlags?.[clusterNs]?.ocsClusterName;
  const storageClassNamePrefix = useFallbackHook(ocsClusterName);
```

packages/ocs/dashboards/persistent-external/breakdown-card.tsx
```diff
@@ -51,7 +52,8 @@ export const BreakdownCard: React.FC = () => {
   const { systemFlags } = useODFSystemFlagsSelector();
 
   // name of created StorageClasses are prefix by StorageCluster name
-  const storageClassNamePrefix = systemFlags[clusterNs]?.ocsClusterName;
+  const storageClassName = systemFlags?.[clusterNs]?.ocsClusterName;
+  const storageClassNamePrefix = useFallbackHook(storageClassName);
```
SanjalKatiyar, line 56: nits
```suggestion
  const ocsClusterName = systemFlags?.[clusterNs]?.ocsClusterName;
  const storageClassNamePrefix = useFallbackHook(ocsClusterName);
```

packages/ocs/dashboards/persistent-external/fallback-hook.ts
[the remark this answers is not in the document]
bipuladh: no they use `ocs-storagecluster` as default for it client mode

packages/ocs/dashboards/persistent-external/fallback-hook.ts
```diff
@@ -0,0 +1,33 @@
...
+import { referenceForModel } from '@odf/shared/utils';
+import {
+  WatchK8sResource,
+  useK8sWatchResource,
+} from '@openshift-console/dynamic-plugin-sdk';
+
+export const storageClusterResource: WatchK8sResource = {
+  isList: true,
+  kind: referenceForModel(OCSStorageClusterModel),
+};
```
GowthamShanmugam: this watch resource variable is not used in more than one place, i would prefer to use the parameters directly on the watch

packages/ocs/dashboards/persistent-external/fallback-hook.ts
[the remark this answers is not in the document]
SanjalKatiyar: is client/consumer a external mode cluster, they are still using `ocs-storagecluster` as a prefix ??

packages/ocs/dashboards/persistent-external/fallback-hook.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const fsProvisionerName = '.cephfs.csi.ceph.com';
+
+const useFallbackHook = (storageClusterName) => {
```
SanjalKatiyar, line 16: can we rename the hook to maybe `useClientFallback` ?? to specify that it is only used for Client use case and not needed by ODF...

packages/ocs/dashboards/persistent-external/fallback-hook.ts
bipuladh: yes. mistake

packages/ocs/dashboards/persistent-external/fallback-hook.ts
bipuladh: I will add a comment explaining that instead.

packages/ocs/dashboards/persistent-external/fallback-hook.ts
bipuladh: Keeps the hook code shorter.

packages/ocs/dashboards/persistent-external/fallback-hook.ts
bipuladh: it's a mistake.

packages/ocs/dashboards/persistent-external/fallback-hook.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const fsProvisionerPostFix = '.cephfs.csi.ceph.com';
+
+/**
+ * This hooks is for ODF Client Mode use case only.
+ * ODF client mode doesn't deploy ODF StorageCluster hence falling back to StorageClases.
+ */
+const useFallbackHook = (storageClusterName) => {
+  const [storageClasses, storageClassesLoaded, storageClassLoadError] =
+    useK8sWatchResource<StorageClassResourceKind[]>(scResource);
```
SanjalKatiyar, line 22: this way it won't even poll when we already have `storageClusterName`...
```suggestion
   useK8sWatchResource<StorageClassResourceKind[]>(getValidWatchK8sResourceObj(scResource, !storageClusterName));
```

packages/ocs/dashboards/persistent-external/fallback-hook.ts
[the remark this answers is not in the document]
bipuladh: We already poll for storageClass in the same page. so we can live with this poll. Essentially the connection is going to be shared.

packages/ocs/dashboards/persistent-external/fallback-hook.ts
SanjalKatiyar: okay, thing is we are returning `ocs-storagecluster` (as edge case)... but in ODF the name will be `ocs-external-storagecluster`, so this is a bit client specific hook...

packages/ocs/dashboards/persistent-external/fallback-hook.ts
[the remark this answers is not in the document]
bipuladh: client isn't external mode cluster.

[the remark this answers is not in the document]
bipuladh: Screenshot from 2024-02-07 17-17-51

packages/ocs/dashboards/persistent-external/fallback-hook.ts
bipuladh: changed the name

packages/ocs/dashboards/persistent-external/fallback-hook.ts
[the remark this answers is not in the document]
SanjalKatiyar: ack !

packages/ocs/dashboards/persistent-external/fallback-hook.ts
bipuladh: this resource is already called in the same page for inventory card. so we can live with this.

packages/ocs/dashboards/persistent-external/fallback-hook.ts
```diff
@@ -0,0 +1,37 @@
...
+  WatchK8sResource,
+  useK8sWatchResource,
+} from '@openshift-console/dynamic-plugin-sdk';
+
+const scResource: WatchK8sResource = {
+  isList: true,
+  kind: referenceForModel(StorageClassModel),
+};
+
+const fsProvisionerPostFix = '.cephfs.csi.ceph.com';
```
GowthamShanmugam, line 14: under common.ts let's create an enum for each ceph provisioner type and reuse the same here, in this way we can replace the cephStorageProvisioners list under common also can be replaced with enums.

packages/ocs/dashboards/persistent-external/fallback-hook.ts
SanjalKatiyar: make sense then !! (I did not check the parent component)

SanjalKatiyar:
LGTM

packages/ocs/dashboards/persistent-external/fallback-hook.ts
bipuladh: Tracking at https://github.com/red-hat-storage/odf-console/issues/1215

packages/ocs/dashboards/persistent-external/fallback-hook.ts
```diff
@@ -0,0 +1,37 @@
...
+const useClientFallback = (storageClusterName) => {
+  const [storageClasses, storageClassesLoaded, storageClassLoadError] =
+    useK8sWatchResource<StorageClassResourceKind[]>(scResource);
+  if (storageClusterName) {
+    return storageClusterName;
+  }
+
+  const provisioners = storageClasses.map((sc) => sc.provisioner);
+  const fileSystemProvisioner = provisioners.find((item) =>
+    item.includes(fsProvisionerPostFix)
```
GowthamShanmugam, line 29: shall we replace this with a util function like isCephFSProvisioner?

packages/ocs/dashboards/persistent-external/fallback-hook.ts
GowthamShanmugam: ack

GowthamShanmugam (approved):

packages/ocs/dashboards/persistent-external/fallback-hook.ts
bipuladh: we can do that but I don't think we would have a very specific use case. We could probably go with isOCSProvisioner with more args. Adding it to #1215

SanjalKatiyar:
/retitle Adds fallback layer for multi cluster code in External mode

PR #1240, opened 2024-02-28
Custom name input field for the protected imperative application

GowthamShanmugam:
on top of: https://github.com/red-hat-storage/odf-console/pull/1210


GowthamShanmugam:
image

GowthamShanmugam:
@SanjalKatiyar @bipuladh Please review

packages/shared/src/input-with-requirements/TextInputWithFieldRequirements.tsx
```diff
@@ -28,7 +28,9 @@ export type TextInputWithFieldRequirementsProps = {
   fieldRequirements: string[];
   control: Control<FieldValues>;
   defaultValue?: any;
-  formGroupProps: FormGroupProps & { helperText?: string };
+  // FromGroup don't have helpertext, using lowecase to appear in the DOM as a custom attribute.
+  // This change is to avoid unit test warning
+  formGroupProps: FormGroupProps & { helpertext?: string };
```
SanjalKatiyar, line 33: this is not correct, it can cause bug at multiple places... multiple files are currently passing this `helperText` (camel-case) prop, but here we are expecting `helpertext` (lower-case)...

instead we should:
1. Either make `helperText` a separate prop (moving it outside of `formGroupProps`).
2. Or, if that is lot of unrelated change for this PR then, for now, remove `helperText` from `formGroupProps` object then pass it here: https://github.com/red-hat-storage/odf-console/blob/master/packages/shared/src/input-with-requirements/TextInputWithFieldRequirements.tsx#L122.

packages/mco/components/discovered-application-wizard/wizard-steps/namespace-step/namespace-step.tsx
```diff
@@ -44,17 +58,121 @@ const clusterOptions = (clusters: string[]): JSX.Element[] =>
...
+  const newName: string = watch('name-input');
+
+  React.useEffect(
+    () =>
+      dispatch({
+        type: EnrollDiscoveredApplicationStateType.SET_NAME,
+        payload: isValid ? newName : '',
+      }),
+    [newName, isValid, dispatch]
+  );
```
SanjalKatiyar, line 127:
```suggestion
  React.useEffect(
    () =>
      dispatch({
        type: EnrollDiscoveredApplicationStateType.SET_NAME,
        payload: newName,
      }),
    [newName, isValid, dispatch]
  );
```
couple of things here:
1. I don't think it is good UX to automatically reset the state when input is invalid. We can disable the "Next" button and show warning/error messages... but should not reset input to `'' (empty string)` automatically.
(let me know if I misunderstood and that was not the intention here).

2. We can eliminate use of `useEffect` entirely IMO (it is good practice to eliminate useEffect where ever possible). `textInputProps` takes `onChange` prop and we can directly pass `dispatch` there. I know other places we are using `useEffect` for this, but to me it seems like a miss and overkill.
(let me know if it does not works, or there is any issue with this approach).

packages/mco/components/discovered-application-wizard/wizard-steps/namespace-step/namespace-step.tsx
```diff
@@ -44,17 +58,121 @@ const clusterOptions = (clusters: string[]): JSX.Element[] =>
...
+      defaultValue={name}
+      popoverProps={{
+        headerContent: t('Name requirements'),
+        footerContent: `${t('Example')}: my-name`,
+      }}
+      formGroupProps={{
+        className: 'pf-v5-u-w-50',
+        label: t('Name'),
+        fieldId: 'name-input',
+        helpertext: t(
```
SanjalKatiyar, line 142:
```suggestion
        helperText: t(
```

packages/mco/components/discovered-application-wizard/wizard-steps/namespace-step/namespace-step.tsx
GowthamShanmugam: TextInputWithFieldRequirements.tsx is maintaining its own state, Even though I am setting an empty string, it won't disappear from the UI. Only validation fails.

Let me do one thing, I will store the isValid also in state so I can use the it for the validation

packages/shared/src/input-with-requirements/TextInputWithFieldRequirements.tsx
SanjalKatiyar: ` changed this in all the affected places` >> nope... `system-connection-details.tsx`, `gcp-endpoint-type.tsx` are 2 examples, there could be more...

`helperText is not part of formGroupProps in PF5` >> yeah I know, that is why there `2` method I suggested, u need to remove `helperText` from `formGroupProps` before passing it to PF5 component.

packages/mco/components/discovered-application-wizard/wizard-steps/namespace-step/namespace-step.tsx
SanjalKatiyar: `I will store the isValid also in state so I can use the it for the validation` >> why, do we need it ?? I don't think so.

packages/mco/components/discovered-application-wizard/wizard-steps/namespace-step/namespace-step.tsx
SanjalKatiyar: yup, I understood that. https://github.com/red-hat-storage/odf-console/pull/1240#discussion_r1533891467

packages/shared/src/input-with-requirements/TextInputWithFieldRequirements.tsx
GowthamShanmugam: i think you are confusing with TextInputWithFieldRequirements and FormGroupController

GowthamShanmugam: TextInputWithFieldRequirements is not used in system-connection-details.tsx, gcp-endpoint-type.tsx

packages/mco/components/discovered-application-wizard/wizard-steps/namespace-step/namespace-step.tsx
GowthamShanmugam: useEffect is to update my wizard page state

packages/shared/src/input-with-requirements/TextInputWithFieldRequirements.tsx
SanjalKatiyar: yes, my bad ! sorry about that... in that case method `1` is small as well (not much changes needed)... we can still keep it camel-case and just not pass it to https://github.com/red-hat-storage/odf-console/blob/master/packages/shared/src/input-with-requirements/TextInputWithFieldRequirements.tsx#L122, right ?

SanjalKatiyar: adding it as a `custom attribute` seems like a hack to me to resolve unit test warning...

packages/mco/components/discovered-application-wizard/wizard-steps/namespace-step/namespace-step.tsx
SanjalKatiyar: u can directly dispatch from `onChange` of `textInputProps`, instead of doing it from `useEffect` ? that will update ur wizard page...
```
onChange = {() =>  dispatch({
        type: EnrollDiscoveredApplicationStateType.SET_NAME,
        payload: newName,
      })}
```

GowthamShanmugam: Text input is displayed from TextInputWithFieldRequirements state,  if I clean my state, it will still display the text in the input box, I am taking advantage of this behavior,

When isValid is false, my state is cleaned, So validation fails, if the user clicks the next button they will see an error message, But still, the text is displayed in the input box from the TextInputWithFieldRequirements state.

packages/shared/src/input-with-requirements/TextInputWithFieldRequirements.tsx
GowthamShanmugam: ok, then I will move the helperText out of it.

packages/mco/components/discovered-application-wizard/wizard-steps/namespace-step/namespace-step.tsx
SanjalKatiyar: regarding `onChange` part: https://react.dev/learn/you-might-not-need-an-effect#how-to-remove-unnecessary-effects (`You don’t need Effects to handle user events`)...

GowthamShanmugam: I checked onChange behaviour, it return the input text but isValid is an async updated value, it is not changing at onChange, It looks  useEffect is used on purpose.

SanjalKatiyar: ack !

PR #1243, opened 2024-03-01
Failover / Relocate modal for discovered applications

GowthamShanmugam:
Issue: https://issues.redhat.com/browse/RHSTOR-4923
From this PR, I am trying to move all unit test mocks to a separate file one by one. around  5 unit test mock files i created in this PR.

PR #1245, opened 2024-03-05
Add unit tests for imperative apps list page

SanjalKatiyar:


packages/mco/components/protected-applications/protected-apps-list-page.spec.tsx
```diff
@@ -0,0 +1,377 @@
...
+let noFilteredData = false;
+let filterDRPC = '';
+
+const resetGlobals = () => {
+  noData = false;
+  noFilteredData = false;
+  filterDRPC = '';
+};
+
+const failingDRPC = {
```
GowthamShanmugam, line 29: In my PR, i am started keeping the mock object under different folder for the reuse: https://github.com/red-hat-storage/odf-console/pull/1243/files#diff-fe7b1786b63b2b92a387f1211cc4c3ba2651d3bcd63648ff728bed2e3a216f95

Shall we do the same here?

packages/mco/components/protected-applications/protected-apps-list-page.spec.tsx
```diff
@@ -0,0 +1,377 @@
...
+  metadata: {
+    name: relocatedDRPCName,
+    namespace: 'test',
+    annotations: {
+      'drplacementcontrol.ramendr.openshift.io/last-app-deployment-cluster':
+        deploymentClusterName,
+    },
+  },
+  spec: {
+    action: 'Failover',
```
GowthamShanmugam, line 69:
```suggestion
    action: 'Relocate',
```
no use with this change, just matching phase and action

GowthamShanmugam:
only optional comments PR LGTM

GowthamShanmugam (approved):

PR #1247, opened 2024-03-11
Reduce number of refreshes in tests

bipuladh:


SanjalKatiyar:
LGTM

cypress/tests/block-pool.spec.ts
```diff
@@ -102,21 +102,14 @@ describe('Tests form validations on BlockPool', () => {
...
     cy.install();
+    navigateToBlockPool();
+    cy.byTestID('item-create').click();
   });
 
   after(() => {
     cy.logout();
   });
 
-  beforeEach(() => {
```
GowthamShanmugam, line 111: try this

beforeEach(() => {
    cy.byTestID('new-pool-name-textbox').focus().clear();
  });

cypress/tests/obc-test.spec.ts
```diff
@@ -135,7 +135,6 @@ describe('Tests form validations on Object Bucket Claim', () => {
   });
 
   beforeEach(() => {
-    cy.visit('/');
     obcNavigate.navigateToOBC();
```
GowthamShanmugam, line 138: this can be moved to before : commonFlows.navigateToObjectStorage();

cypress/tests/obc-test.spec.ts
```diff
@@ -135,7 +135,6 @@ describe('Tests form validations on Object Bucket Claim', () => {
...
     cy.byTestID('item-create').click();
```
GowthamShanmugam, line 139: These two also can be moved inside before

cypress/views/block-pool.ts
```diff
@@ -107,7 +107,6 @@ export const openBlockPoolKebab = (
   cy.byLegacyTestID('item-filter').clear().type(targetPoolName);
   cy.log('Only one resource should be present after filtering');
   cy.byTestID('kebab-button').should('have.length', 1);
-  if (isDefaultPool)
-    cy.byTestID('kebab-button').first().find('button').should('be.disabled');
-  else cy.byTestID('kebab-button').first().click();
+  if (isDefaultPool) cy.byTestID('kebab-button').should('be.disabled');
```
GowthamShanmugam, line 110:
```suggestion
  if (isDefaultPool)
     cy.get(`[data-test-rows="resource-row"]`)
+        .contains(targetPoolName)
+        .parents('tr')
+        .within(() => {
+          cy.get('[data-test="kebab-button"]').should('be.disabled');
+        });
```

GowthamShanmugam: just to makesure it  never breaks

GowthamShanmugam (approved):

PR #1250, opened 2024-03-13
Unit testcase fix

GowthamShanmugam:


packages/mco/components/protected-applications/protected-apps-list-page.spec.tsx
```diff
@@ -302,7 +302,7 @@ describe('Test protected applications list page table row (ProtectedAppsTableRow
     const editConfig = 'Edit configuration';
     const failover = 'Failover';
     const relocate = 'Relocate';
-    const kebabButton = screen.getByRole('button', { name: /Actions/i });
+    const kebabButton = screen.getByRole('button', { name: /Kebab toggle/i });
```
SanjalKatiyar, line 305: why was this needed ? are test failing on latest master ?

PR #1257, opened 2024-03-15
Text changes as per the UX update

GowthamShanmugam:
Also fixing  text and some minor bugs


GowthamShanmugam:
image
image
image
image
image

packages/mco/components/modals/app-manage-policies/helper/select-policy-wizard-content.tsx
```diff
@@ -83,10 +84,8 @@ export const SelectPolicyWizardContent: React.FC<SelectPolicyWizardContentProps>
                     status: getDRPolicyStatus(policy.isValidated, t),
                   })}
               </HelperTextItem>
-              <HelperTextItem
-                variant={getValidatedProp(isValidationEnabled && !name)}
-              >
-                {t('Required')}
+              <HelperTextItem variant={getValidatedProp(isInvalidPolicy)}>
+                {isInvalidPolicy && t('Required')}
               </HelperTextItem>
```
SanjalKatiyar, line 89:
```suggestion
              isInvalidPolicy && <HelperTextItem variant={getValidatedProp(isInvalidPolicy)}>
                {t('Required')}
              </HelperTextItem>
```

plugins/mco/console-extensions.json
```diff
@@ -147,7 +147,7 @@
     "type": "acm.application/action",
     "properties": {
       "id": "appManagePolicy",
-      "title": "%plugin__acm~Manage data policy%",
+      "title": "%plugin__acm~Manage disaster recovery%",
```
SanjalKatiyar, line 150: 1. We have to create a separate `18n-comments.ts` file for these type of texts (like we do for `odf/` directory)... else these won't get translated (it will work here because we are using this same text else where, so it was added to json file via that).
2. Why are are using `plugin__acm` (I just noticed it is used at multiple places in this file) ??

GowthamShanmugam: my bad, i added Acm wrongly

packages/mco/components/modals/app-manage-policies/policy-list-view.tsx
```diff
@@ -222,9 +223,10 @@ export const PolicyListView: React.FC<PolicyListViewProps> = ({
               headingLevel="h3"
             />
             <EmptyStateBody>
-              {t(
-                "You haven't set a data policy for your application yet. To protect your application, click the 'Assign data policy' button and select a policy from the available templates."
-              )}
+              <Trans t={t}>
+                You have not enrolled this application yet. To protect your
+                application, <br /> click <strong>Enroll application.</strong>
```
SanjalKatiyar, line 228:
```suggestion
                application, click <strong>Enroll application.</strong>
```

packages/mco/components/protected-applications/components.tsx
```diff
@@ -146,14 +146,15 @@ export const EnrollApplicationButton: React.FC<{
...
+        hasAutoWidth
       >
         <Trans t={t}>
           <p className="co-break-word pf-v5-u-font-weight-bold">
             ACM discovered applications:
           </p>
           <p className="co-break-word pf-v5-u-mb-sm">
             Based on modular and microservices architecture, uses operators for
-            dynamically created kubernetes objects. Eg:{' '}
+            dynamically created <br /> kubernetes objects. Eg:{' '}
```
SanjalKatiyar, line 157:
```suggestion
            dynamically created kubernetes objects. Eg:{' '}
```

packages/mco/components/protected-applications/components.tsx
```diff
@@ -163,18 +164,18 @@ export const EnrollApplicationButton: React.FC<{
               Enrollment process:
             </span>{' '}
             Discovered applications are enrolled under disaster recovery through
-            enabling protection for their namespaces and further defining the
-            scope of this protection within namespace through recipe selection
-            or resource label.
+            enabling <br /> protection for their namespaces and further defining
```
SanjalKatiyar, line 167:
```suggestion
            enabling protection for their namespaces and further defining
```

packages/mco/components/protected-applications/components.tsx
```diff
@@ -163,18 +164,18 @@ export const EnrollApplicationButton: React.FC<{
...
+            the scope of this protection within namespace <br /> through recipe
```
SanjalKatiyar, line 168:
```suggestion
            the scope of this protection within namespace through recipe
```

packages/mco/components/protected-applications/components.tsx
```diff
@@ -163,18 +164,18 @@ export const EnrollApplicationButton: React.FC<{
...
           </p>
 
           <p className="co-break-word pf-v5-u-font-weight-bold">
             ACM managed applications:
           </p>
           <p className="co-break-word pf-v5-u-mb-sm">
             Based on subscribing to one or more Kubernetes resource repositories
-            (channel resource) that contains resources that are deployed on
-            managed clusters. Eg:{' '}
+            (channel resource) that <br /> contains resources that are deployed
```
SanjalKatiyar, line 177:
```suggestion
            (channel resource) that contains resources that are deployed
```

packages/mco/components/protected-applications/components.tsx
```diff
@@ -184,8 +185,8 @@ export const EnrollApplicationButton: React.FC<{
               Enrollment process:
             </span>{' '}
             Individually protect managed application with flexibility for
-            distinct configurations for different sub-categories of managed
-            application based on specific requirements.
+            distinct <br /> configurations for different sub-categories of
```
SanjalKatiyar, line 188:
```suggestion
            distinct configurations for different sub-categories of
```

packages/mco/components/protected-applications/components.tsx
```diff
@@ -206,15 +207,24 @@ export const NoDataMessage: React.FC = () => {
...
-        protection to your applications start by clicking on the{' '}
-        <strong>Enroll application</strong> button.
+        <p>
+          You do not have any <strong>discovered applications</strong> that are
+          protected yet. For details about your{' '}
+          <strong>protected managed applications</strong>, navigate to the{' '}
+          <strong>Applications</strong> page, as this information is not
+          maintained here.
+        </p>
+        <br />
```
SanjalKatiyar, line 223:
```suggestion
```

SanjalKatiyar (commented):
These kind of breaks are not needed IMO, we don't have to exactly follow the UX (line-by-line). We should only control the max/min width of the popover and use text alignment styles, but let things render based on user's screen size, instead of explicitly breaking the line in mid text !
It is okay to break for new paragraph etc, but not between continuous lines/texts.

GowthamShanmugam:
> These kind of breaks are not needed IMO, we don't have to exactly follow the UX (line-by-line). We should only control the max/min width of the popover and use text alignment styles, but let things render based on user's screen size, instead of explicitly breaking the line in mid text ! It is okay to break for new paragraph etc, but not between continuous lines/texts.

make sense!

packages/mco/components/protected-applications/components.tsx
GowthamShanmugam: why not? this is a line break for an empty line, not for the width

Issue #1253, opened 2024-03-14
Topology view missing functionality


SanjalKatiyar:
- "Ready" label and close button overlapping in sidebar.
- **No option to drill down the node.**
- Search bar is too long and "Expand" button is rendered in separate row.

------------------------------------------------------------------------------------------------------------------------
<img width="471" alt="Screenshot 2024-03-14 at 5 48 26 PM" src="https://github.com/red-hat-storage/odf-console/assets/39404641/c18759b6-332d-4040-a35e-bcc0e4a6911a">

------------------------------------------------------------------------------------------------------------------------

<img width="886" alt="Screenshot 2024-03-14 at 5 48 31 PM" src="https://github.com/red-hat-storage/odf-console/assets/39404641/5a0545ce-30b4-4499-be78-a7a265bd37fd">

------------------------------------------------------------------------------------------------------------------------

<img width="1405" alt="Screenshot 2024-03-14 at 5 47 00 PM" src="https://github.com/red-hat-storage/odf-console/assets/39404641/f38a5744-3437-4e94-b426-8e6dbb40625f">

PR #1259, opened 2024-03-20
Fix 'Enter node' option for Topology view

SanjalKatiyar:
https://github.com/red-hat-storage/odf-console/issues/1253

Screenshot 2024-03-20 at 7 30 49 PM

### DIFF
--- a/packages/shared/src/topology/factory/StyleNode.tsx
+++ b/packages/shared/src/topology/factory/StyleNode.tsx
@@ -7,6 +7,7 @@ import { SVGIconProps } from '@patternfly/react-icons/dist/esm/createIcon';
 import DefaultIcon from '@patternfly/react-icons/dist/esm/icons/builder-image-icon';
 import useDetailsLevel from '@patternfly/react-topology/dist/esm/hooks/useDetailsLevel';
 import classNames from 'classnames';
+import { TFunction } from 'react-i18next';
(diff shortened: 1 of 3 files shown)


packages/shared/src/topology/factory/StyleNode.tsx
```diff
@@ -7,6 +7,7 @@ import { SVGIconProps } from '@patternfly/react-icons/dist/esm/createIcon';
 import DefaultIcon from '@patternfly/react-icons/dist/esm/icons/builder-image-icon';
 import useDetailsLevel from '@patternfly/react-topology/dist/esm/hooks/useDetailsLevel';
 import classNames from 'classnames';
+import { TFunction } from 'react-i18next';
```
GowthamShanmugam, line 10:
```suggestion
import { TFunction } from 'i18next';
```

GowthamShanmugam: most of the places we are using i18next, is both are same?

SanjalKatiyar: either is fine IMO as it is just type import... but since we are react project ideally we should have used `react-i18next` everywhere... all translation related hooks are already imported from `react-i18next`, might as well the type too...

GowthamShanmugam (approved):

PR #1260, opened 2024-03-20
Displaying cleanup event for discovered app failover

GowthamShanmugam:


GowthamShanmugam:
image

packages/mco/components/discovered-application-wizard/utils/k8s-utils.ts
```diff
@@ -76,6 +76,7 @@ export const createPromise = (state: EnrollDiscoveredApplicationState) => {
   return k8sCreate({
     model: DRPlacementControlModel,
     data: getDRPCKindObj({
+      name,
```
GowthamShanmugam, line 79: While rebase, i missed this change in the custom input PR

PR #1262, opened 2024-03-21
Fetch external RHCS script from Rook CSV

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-5637


packages/odf/components/create-storage-system/external-ceph-storage/system-connection-details.tsx
```diff
@@ -65,27 +64,36 @@ export const ConnectionDetails: React.FC<ExternalComponentProps<RHCSState>> = ({
...
   const { fileName, fileData, errorMessage, isLoading } = formState;
 
-  const annotations = getAnnotations(csv);
+  const rookAnnotations = getAnnotations(rookCSV, {});
+  const ocsAnnotations = getAnnotations(ocsCSV, {});
 
-  const downloadFile = createDownloadFile(
-    annotations?.['external.features.ocs.openshift.io/export-script']
-  );
+  const baseEncodedScript = rookAnnotations?.['externalClusterScript'];
```
SanjalKatiyar, line 86: https://github.com/red-hat-storage/rook/blob/master/build/csv/csv-gen.sh#L37

packages/odf/components/create-storage-system/external-ceph-storage/system-connection-details.tsx
```diff
@@ -65,18 +64,27 @@ export const ConnectionDetails: React.FC<ExternalComponentProps<RHCSState>> = ({
...
 
   const { fileName, fileData, errorMessage, isLoading } = formState;
 
-  const annotations = getAnnotations(csv);
+  const rookAnnotations = getAnnotations(rookCSV, {});
+  const ocsAnnotations = getAnnotations(ocsCSV, {});
 
   const downloadFile = createDownloadFile(
-    annotations?.['external.features.ocs.openshift.io/export-script']
+    rookAnnotations?.['externalClusterScript']
```
SanjalKatiyar, line 87: https://github.com/red-hat-storage/rook/blob/master/build/csv/csv-gen.sh#L37

packages/odf/components/utils/common.ts
```diff
@@ -285,10 +285,12 @@ export const checkError = (
   return '';
 };
 
-export const createDownloadFile = (data: string = ''): string =>
-  `data:application/octet-stream;charset=utf-8,${encodeURIComponent(
-    Base64.decode(data)
-  )}`;
+export const createDownloadFile = (data: string): string =>
```
bipuladh, line 288: Lets keep it as it is. Right now you would get an empty file if you pass undefined or null. I think that is good enough.

SanjalKatiyar: issue it UI will still display `Download script` option because of `data:application/octet-stream;charset=utf-8` part in this util...

SanjalKatiyar: adding ternary condition is harmless IMO, I tested it...

PR #1266, opened 2024-03-25
Adds support for creation of Storage clients

bipuladh:
Screenshot from 2024-03-25 13-52-46
Screenshot from 2024-03-25 13-52-51


locales/en/plugin__odf-console.json
```diff
@@ -1,4 +1,13 @@
 {
+  "Create StorageClient": "Create StorageClient",
+  "Create a Storageclient to connect to a Data Foundation system.": "Create a Storageclient to connect to a Data Foundation system.",
```
SanjalKatiyar, line 3:
```suggestion
  "Create a StorageClient to connect to a Data Foundation system.": "Create a Storageclient to connect to a Data Foundation system.",
```

packages/client/src/components/create-client/create-client.scss
```diff
@@ -0,0 +1,15 @@
+
+.odf-create-storage-client__header {
+  border-bottom: var(--pf-v5-global--BorderWidth--sm) solid #ddd;
+  padding-right: var(--pf-v5-global--spacer--lg);
+  padding-left: var(--pf-v5-global--spacer--lg);
+  padding-bottom: var(--pf-v5-global--spacer--sm);
+  padding-top: var(--pf-v5-global--spacer--sm);
+}
+.odf-create-client__form {
+    max-width: 500px;
```
SanjalKatiyar, line 10: can we use `rem` or other relative unit, instead of `px` ?

packages/client/src/components/create-client/create-client.scss
```diff
@@ -0,0 +1,15 @@
...
+}
+
+.odf-create-client-form__ns-dropdown {
+    width: 100%;
+}
```
SanjalKatiyar, line 15: nits: prefer adding `pf-v5-u-w-100` directly to the elements wherever possible... save us from defining our own classes for setting basic styles...

packages/client/src/components/create-client/create-client.tsx
```diff
@@ -0,0 +1,170 @@
...
+        <Text component={TextVariants.h1}>{t('Create StorageClient')}</Text>
+        <Text component={TextVariants.small}>
+          {t('Create a Storageclient to connect to a Data Foundation system.')}
+        </Text>
+      </TextContent>
+    </div>
+  );
+};
+
+const isValidURL = (userInput: string) => {
```
SanjalKatiyar, line 46: nits: function is named as `isValidURL`, but its return type is not boolean...

packages/client/src/components/create-client/create-client.tsx
```diff
@@ -0,0 +1,170 @@
...
+  if (!userInput) {
+    return ValidatedOptions.default;
+  }
+  try {
+    (() => new URL(userInput))();
+    const port = userInput.split(':').pop();
+    if (port !== '' && _.isInteger(Number(port))) {
```
SanjalKatiyar, line 53: I think `_.isInteger(Number(port)` part will be automatically handled by `URL` object ?? Anyway, no harm in checking again...

packages/client/src/components/create-client/create-client.tsx
```diff
@@ -0,0 +1,170 @@
...
+        setProgress(false);
+      })
+      .catch((err) => {
+        setProgress(false);
+        setError(err);
+      });
+  };
+
+  return (
+    <div>
```
SanjalKatiyar, line 99: is `div` intentional ?
```suggestion
    <>
```

packages/client/src/components/create-client/create-client.tsx
```diff
@@ -0,0 +1,170 @@
...
+            isRequired
+          >
+            <TextArea
+              id="ticket-data"
+              value={ticket}
+              onChange={(_e, value) => setTicket(value)}
+            />
+          </FormGroup>
+          <ButtonBar inProgress={inProgress} errorMessage={error?.message}>
+            <ActionGroup className="pf-c-form">
```
SanjalKatiyar, line 157:
```suggestion
            <ActionGroup className="pf-v5-c-form">
```

packages/client/src/components/create-client/create-client.tsx
bipuladh: Port is actually mandatory. so we need to check if it actually exists without it it's an error

packages/client/src/components/create-client/create-client.tsx
bipuladh: yes

packages/client/src/components/create-client/create-client.tsx
SanjalKatiyar: aah, okay !

PR #1268, opened 2024-03-25
Dummy placement creation for discovered apps

GowthamShanmugam:
Issue: https://issues.redhat.com/browse/RHSTOR-5495

### DIFF
--- a/packages/mco/components/discovered-application-wizard/utils/k8s-utils.ts
+++ b/packages/mco/components/discovered-application-wizard/utils/k8s-utils.ts
@@ -1,7 +1,19 @@
-import { DRPlacementControlModel, DRPolicyModel } from '@odf/mco//models';
-import { RAMEN_PROTECTED_APPS_NAMESPACE } from '@odf/mco/constants';
-import { DRPlacementControlKind, DRPolicyKind } from '@odf/mco/types';
+import {
+  ACMPlacementModel,
+  DRPlacementControlModel,
+  DRPolicyModel,
+} from '@odf/mco//models';
+import {
+  DISCOVERED_APP_NS,
+  PROTECTED_APP_ANNOTATION,
+} from '@odf/mco/constants';
+import {
+  ACMPlacementKind,
+  DRPlacementControlKind,
+  DRPolicyKind,
+} from '@odf/mco/types';
 import { getName } from '@odf/shared/selectors';
+import { K8sResourceKind } from '@odf/shared/types';
 import { getAPIVersionForModel } from '@odf/shared/utils';
 import {
   MatchExpression,
@@ -23,16 +35,17 @@ export const getDRPCKindObj = (props: {
   recipeNamespace?: string;
   k8sResourceLabelExpressions?: MatchExpression[];
   pvcLabelExpressions?: MatchExpression[];
+  placementName: string;
 }): DRPlacementControlKind => ({
   apiVersion: getAPIVersionForModel(DRPlacementControlModel),
   kind: DRPlacementControlModel.kind,
   metadata: {
     name: props.name,
-    namespace: RAMEN_PROTECTED_APPS_NAMESPACE,
+    namespace: DISCOVERED_APP_NS,
   },
   spec: {
     preferredCluster: props.preferredCluster,
-    eligibleForProtectionNamespaces: props.namespaces,
+    protectedNamespace: props.namespaces,
     pvcSelector: !!props.pvcLabelExpressions.length
       ? {
           matchExpressions: props.pvcLabelExpressions,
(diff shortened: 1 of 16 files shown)


packages/mco/constants/common.ts
```diff
@@ -18,3 +18,6 @@ export const SECOND = 1000;
 export const DEFAULT_SYNC_TIME = 15;
 export const ACM_OBSERVABILITY_FLAG = 'ACM_OBSERVABILITY';
 export const ADMIN_FLAG = 'ADMIN';
+
+// Discovered application namespace
+export const DEFAULT_RAMEN_OPS_NAMESPACE = 'ramen-ops';
```
SanjalKatiyar, line 23: what is `ops` ??

SanjalKatiyar: shoudn't we just call it `DISCOVERED_APP_NS`, so that it is more intuitive ??

packages/mco/components/discovered-application-wizard/utils/k8s-utils.ts
```diff
@@ -61,31 +74,68 @@ export const getDRPCKindObj = (props: {
...
+  state: EnrollDiscoveredApplicationState
+): Promise<K8sResourceKind>[] => {
   const { namespace, configuration, replication } = state;
   const { clusterName, namespaces, name } = namespace;
   const { protectionMethod, recipe, resourceLabels } = configuration;
   const { recipeName, recipeNamespace } = recipe;
   const { k8sResourceLabelExpressions, pvcLabelExpressions } = resourceLabels;
   const { drPolicy, k8sResourceReplicationInterval } = replication;
   const namespaceList = namespaces.map(getName);
+  const placementName = `${name}-placement-1`;
```
SanjalKatiyar, line 112:
```suggestion
  const placementName = `${name}-placement`;
```

SanjalKatiyar: should we use random chars in the postfix ? Instead of `-placement` ? To be safer.

SanjalKatiyar, line 112: should we use random chars in the postfix ? Instead of `-placement` ? To be safer.

packages/mco/constants/disaster-recovery.ts
```diff
@@ -9,6 +9,8 @@ export const DRPC_NAMESPACE_ANNOTATION =
 // DRPC annotations
 export const LAST_APP_DEPLOYMENT_CLUSTER_ANNOTATION =
   'drplacementcontrol.ramendr.openshift.io/last-app-deployment-cluster';
+export const APP_NAMESPACE_ANNOTATION =
+  'drplacementcontrol.ramendr.openshift.io/app-namespace'; // ACM managed application workload namespace
```
SanjalKatiyar, line 13: nits:
```suggestion
// ACM managed application workload namespace
export const APP_NAMESPACE_ANNOTATION =
  'drplacementcontrol.ramendr.openshift.io/app-namespace';
```
not really needed, up to you if you want to make this minor change...

packages/mco/constants/disaster-recovery.ts
```diff
@@ -9,6 +9,8 @@ export const DRPC_NAMESPACE_ANNOTATION =
 // DRPC annotations
 export const LAST_APP_DEPLOYMENT_CLUSTER_ANNOTATION =
   'drplacementcontrol.ramendr.openshift.io/last-app-deployment-cluster';
+export const APP_NAMESPACE_ANNOTATION =
```
SanjalKatiyar, line 12: Name should be intuitive enough to know which app type it belongs to...
```suggestion
export const MANAGED_APP_NS_ANNOTATION =
```

packages/mco/components/discovered-application-wizard/utils/k8s-utils.ts
GowthamShanmugam: Its an in ACM format, ACM always creates placement with number, for the better readability , Using the name you can find how many placement the app is using.

GowthamShanmugam: Using the number as a postfix is safer than the random char, While Deleting and re-creating we dont create unnecessary placement everytime, even if the placement is missed to deleted, next time the user can correct it.

packages/mco/constants/disaster-recovery.ts
GowthamShanmugam: let it be, if ramen changes it is easy to update  here.

packages/mco/constants/disaster-recovery.ts
GowthamShanmugam: The idea is to introduce this annotation for subscription as well as Appset, Due to time constrain in 4.16 they may be done or may not, So lets keep this under a generic name.

packages/mco/components/discovered-application-wizard/utils/k8s-utils.ts
SanjalKatiyar: can we plz add a comment for the reason of doing this ?

packages/mco/constants/disaster-recovery.ts
SanjalKatiyar: suggestion was to shift the inline comment on top, separate line :) (not related to ramen)

packages/mco/constants/disaster-recovery.ts
SanjalKatiyar: then let's call it `MANAGED_APP_NS_ANNOTATION`...

packages/mco/constants/disaster-recovery.ts
GowthamShanmugam: ack

packages/mco/components/discovered-application-wizard/utils/k8s-utils.ts
GowthamShanmugam: sure

packages/mco/constants/disaster-recovery.ts
GowthamShanmugam: They are going to add this annotation for discovered apps too, they will make it a list of NS

packages/mco/components/discovered-application-wizard/utils/k8s-utils.ts
GowthamShanmugam: Changes done

packages/mco/constants/disaster-recovery.ts
SanjalKatiyar: ok

PR #1275, opened 2024-04-03
fix 'catastrophic backtracking' issue with text-input regex

SanjalKatiyar:
https://bugzilla.redhat.com/show_bug.cgi?id=2270446

### DIFF
--- a/packages/shared/src/utils/validation.ts
+++ b/packages/shared/src/utils/validation.ts
@@ -1,14 +1,13 @@
 /**
  * Starts and ends with a lowercase letter or number
  */
-export const startAndEndsWithAlphanumerics =
-  /^(?![\W])([a-z0-9]|[\W]*([a-z0-9]))+(?![\W])$/;
+export const startAndEndsWithAlphanumerics = /^[a-z0-9](.*[a-z0-9])?$/;
 
 /**
  * Only lowercase letters, numbers, non-consecutive periods, or hyphens
  */
 export const alphaNumericsPeriodsHyphensNonConsecutive =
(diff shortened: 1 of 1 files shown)


packages/shared/src/utils/validation.ts
```diff
@@ -1,14 +1,13 @@
 /**
  * Starts and ends with a lowercase letter or number
  */
-export const startAndEndsWithAlphanumerics =
-  /^(?![\W])([a-z0-9]|[\W]*([a-z0-9]))+(?![\W])$/;
```
SanjalKatiyar, line 5: this was causing `catastrophic backtracking` for certain format of input strings...

packages/shared/src/utils/validation.ts
```diff
@@ -1,14 +1,13 @@
...
+export const startAndEndsWithAlphanumerics = /^[a-z0-9](.*[a-z0-9])?$/;
 
 /**
  * Only lowercase letters, numbers, non-consecutive periods, or hyphens
  */
 export const alphaNumericsPeriodsHyphensNonConsecutive =
-  /^[a-z0-9]+([a-z0-9]|([-.](?![-.])))*[a-z0-9]*$/;
```
SanjalKatiyar, line 11: this is just minor enhancement, earlier it was returning false for strings prefixed with `-` or `.`, whereas it should pass for such cases and satisfy `Only lowercase letters, numbers, non-consecutive periods, or hyphens`.

alfonsomthd (approved):
It would be great to add some unit tests to ensure the behavior.

SanjalKatiyar:
> It would be great to add some unit tests to ensure the behavior.

ack, will add as part of some other PR.

PR #1277, opened 2024-04-05
Adds missing features for Provider mode

bipuladh:
- Adds more parameters that are required for provider mode deployment.
 - Adds support for key rotation modal.
 - Adds support for client deletion after 5 minutes of connectivity loss. 
 
###  Screenshots: 
 **List page updates:**
Screenshot from 2024-04-05 17-06-57

**Delete modals:**
Screenshot from 2024-04-05 17-07-06
Screenshot from 2024-04-05 17-07-11

**Rotate keys:**
Screenshot from 2024-04-05 17-14-32


packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -226,7 +250,15 @@ const StorageClientRow: React.FC<
         <DataFoudationVersion obj={obj} currentVersion={currentVersion} />
       </TableData>
       <TableData {...tableColumns[4]} activeColumnIDs={activeColumnIDs}>
-        <LastHeartBeat heartbeat={obj.status.lastHeartbeat} />
+        <LastHeartBeat heartbeat={obj?.status?.lastHeartbeat} />
+      </TableData>
+      <TableData {...tableColumns[5]} activeColumnIDs={activeColumnIDs}>
+        <Icon
+          disabled={timeDifference >= DELETE_THRESHOLD}
```
SanjalKatiyar, line 257: if `timeDifference >= DELETE_THRESHOLD`, shoudn't button be enabled ??

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -270,17 +302,31 @@ export const ClientListPage: React.FC<ClientListPageProps> = () => {
     rowFilters
   );
 
-  const onClick = (modalLauncher: typeof launchModal) => () => {
+  const launchModalOnClick = (modalComponent: ModalComponent) => () => {
     const modalComponentProps = { isOpen: true };
-    modalLauncher(ClientOnBoardingModal, modalComponentProps);
+    launchModal(modalComponent, modalComponentProps);
```
SanjalKatiyar, line 307: nits:
```suggestion
    launchModal(modalComponent, { isOpen: true });
```

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -270,17 +302,31 @@ export const ClientListPage: React.FC<ClientListPageProps> = () => {
...
+  const deleteClient = (client: StorageConsumerKind) => {
+    launchModal(RemoveClientModal, { resource: client, isOpen: true });
   };
 
   return (
     <>
       <ListPageHeader title={t('Storage clients')}>
-        <Button variant="primary" onClick={onClick(launchModal)}>
+        <Button
+          variant="primary"
```
SanjalKatiyar, line 318: `ButtonVariant.Primary` ?

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -270,17 +302,31 @@ export const ClientListPage: React.FC<ClientListPageProps> = () => {
...
+          className="pf-v5-u-mr-sm"
+          onClick={launchModalOnClick(ClientOnBoardingModal)}
+        >
           {t('Generate client onboarding token')}
         </Button>
+        <Button
+          variant="tertiary"
```
SanjalKatiyar, line 325: `ButtonVariant.Tertiary` ?

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -209,8 +225,16 @@ const DataFoudationVersion: React.FC<DataFoundationVersionProps> = ({
...
+  RowProps<
+    StorageConsumerKind,
+    {
+      currentVersion: string;
+      deleteClient: (resource: StorageConsumerKind) => void;
+    }
+  >
+> = ({ obj, activeColumnIDs, rowData: { currentVersion, deleteClient } }) => {
+  const timeDifference = getTimeDifferenceInSeconds(obj?.status?.lastHeartbeat);
+  const DELETE_THRESHOLD = 300; // wait till 5 minutes before activating the delete button
```
SanjalKatiyar, line 237: I have a doubt here:
say none of the StorageConsumer CRs have been updated in last 10 mins, this mean `useK8sWatchResource` will not rerender the list page... `timeDifference` will not get updated and "delete" button will not be enabled/disable correctly, right ?

u might want to add a setInterval on main list page, which will re-render entire list every 5mins ?? ex: https://github.com/red-hat-storage/odf-console/blob/master/packages/mco/components/protected-applications/list-page.tsx#L286

packages/odf/components/storage-consumers/remove-client-modal.tsx
```diff
@@ -0,0 +1,131 @@
...
+        <p>
+          <Trans t={t}>
+            Confirm deletion by typing&nbsp;
+            <strong className="co-break-word">
+              {{ name: resource.metadata.name }}
+            </strong>{' '}
+            below:
+          </Trans>
+        </p>
+        <input
```
SanjalKatiyar, line 91: can we not use some component from PF instead ?? plain HTML tags sometimes messes up styling in dark mode.

packages/odf/components/storage-consumers/remove-client-modal.tsx
```diff
@@ -0,0 +1,131 @@
...
+              id="confirm-action"
+              onClick={onSubmit}
+            >
+              {t('Delete')}
+            </Button>
+          </FlexItem>
+          <FlexItem>
+            <Button
+              type="button"
+              variant="secondary"
```
SanjalKatiyar, line 119: `ButtonVariant.Secondary` ??

packages/odf/components/storage-consumers/remove-client-modal.tsx
```diff
@@ -0,0 +1,131 @@
...
+            label: t(StorageConsumerModel.labelKey),
+          })}
+        />
+      </ModalBody>
+      <ModalFooter inProgress={inProgress} errorMessage={error?.message}>
+        <Flex direction={{ default: 'row' }}>
+          <FlexItem>
+            <Button
+              type="submit"
+              variant="danger"
```
SanjalKatiyar, line 107: `ButtonVariant` ??

packages/odf/components/storage-consumers/rotate-keys-modal.tsx
```diff
@@ -0,0 +1,81 @@
...
+                'Upon rotation, the existing signing key will be revoked and replaced with a new one.'
+              )}
+            </div>
+          </FlexItem>
+        </Flex>
+      </ModalBody>
+      <ModalFooter inProgress={inProgress} errorMessage={error?.message}>
+        <Flex direction={{ default: 'row' }}>
+          <FlexItem>
+            <Button variant="secondary" onClick={() => closeModal()}>
```
SanjalKatiyar, line 70: `ButtonVariant` ??

packages/odf/components/utils/common.ts
```diff
@@ -491,6 +491,21 @@ export const getOCSRequestData = ({
...
+          disableStorageClass: true,
+        },
+        cephFileSystems: {
+          disableSnapshotClass: true,
+          disableStorageClass: true,
+        },
+        cephObjectStores: {
+          hostNetwork: false,
+        },
+      });
```
SanjalKatiyar, line 508: does this mean we have to allow "BlockPools" creation as well ??

packages/odf/components/storage-consumers/client-list.tsx
SanjalKatiyar: in that interval u can calculate `timeDifference` for all the rows... so that we don't have to add `setInterval` for each row...

locales/en/plugin__odf-console.json
```diff
@@ -1108,12 +1108,21 @@
...
   "Copy to clipboard": "Copy to clipboard",
   "How to use this token?": "How to use this token?",
   "To onboard the client cluster, the provider cluster requires the onboarding token.": "To onboard the client cluster, the provider cluster requires the onboarding token.",
   "An onboarding token is needed to connect an additional OpenShift cluster to a Data Foundation deployment.": "An onboarding token is needed to connect an additional OpenShift cluster to a Data Foundation deployment.",
+  "Permanently delete storage client?": "Permanently delete storage client?",
+  "Deleting the storage client {getName(resource)} will remove all Ceph/Rook resources and erase all data associated with this client, leading to permanent deletion of the client. This action cannot be undone. It will destroy all pods, services and other objects in the namespace <4>{{name}}</4>.": "Deleting the storage client {getName(resource)} will remove all Ceph/Rook resources and erase all data associated with this client, leading to permanent deletion of the client. This action cannot be undone. It will destroy all pods, services and other objects in the namespace <4>{{name}}</4>.",
```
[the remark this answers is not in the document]
bipuladh: as per the designs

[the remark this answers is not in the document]
SanjalKatiyar: I know, but that;s not what we should follow...

packages/odf/components/storage-consumers/client-list.tsx
bipuladh: the values are string literals in pf5

packages/odf/components/storage-consumers/remove-client-modal.tsx
bipuladh: I don't think this will. This allows me to remove the state management that PF components require. NS delete modal uses the same component as well.

packages/odf/components/storage-consumers/client-list.tsx
SanjalKatiyar: but we are using it currently at many places.

packages/odf/components/utils/common.ts
bipuladh: unrelated

packages/odf/components/storage-consumers/client-list.tsx
SanjalKatiyar: https://github.com/red-hat-storage/odf-console/blob/master/packages/shared/src/alert/AlertsPanel.tsx#L65-L67 >> imports are from core...

packages/odf/components/storage-consumers/client-list.tsx
bipuladh: The CRs should get updated regularly. Even if the last heartbeat timestamp doesn't change the `resourceVersion` would definitely change.

SanjalKatiyar: `resourceVersion` will not change unless CR is updated right ?? my point is we can't assume that CRs will always get updated regularly...

SanjalKatiyar: if in any case it does not, then these icons enable/disable might have wanted behaviour...

SanjalKatiyar: hypothetically say some controller failed who was supposed to update StorageConsumer CRs... it mean no one will update these CRs and list page will not rerender... and icon will still remain disabled (unless someone refreshes by mistake)

locales/en/plugin__odf-console.json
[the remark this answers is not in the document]
SanjalKatiyar: somwhere (I need to find the link) in the OCP guildelines it was mentioned to use `kind`...

packages/odf/components/storage-consumers/client-list.tsx
bipuladh: I will test this locally creating a dummy CR and see if there are refreshes. The list page will re-render every now and then even when you scroll.

bipuladh: tested it and looks like we need to add setInterval.

SanjalKatiyar: ack !

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -209,8 +225,31 @@ const DataFoudationVersion: React.FC<DataFoundationVersionProps> = ({
...
+      );
+      if (timeDifference > DELETE_THRESHOLD && !allowDeletion) {
+        setAllowDeletion(true);
+      } else if (timeDifference < DELETE_THRESHOLD && allowDeletion) {
+        setAllowDeletion(false);
+      }
+    };
+    const id = setInterval(setter, 10000);
+    return () => clearInterval(id);
+  });
```
SanjalKatiyar, line 252: can we just add `setInterval` once on main list page instead of on every row ?
ex: https://github.com/red-hat-storage/odf-console/blob/master/packages/mco/components/protected-applications/list-page.tsx#L286

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -209,8 +225,31 @@ const DataFoudationVersion: React.FC<DataFoundationVersionProps> = ({
...
   return (
```
GowthamShanmugam, line 253: Just a optional  suggestion, This logic can be replaced by useScheduler by passing setter as a callback function to useScheduler

packages/odf/components/storage-consumers/client-list.tsx
SanjalKatiyar: if list is long, unnecessarily we will have too many counters... whereas once on top might be better, which can calculate `timeDifference` for all the rows at-once (not a heavy operation, so alright to do once for entire table)...

bipuladh: not all rows need to be updated. Here it updates the required rows only

SanjalKatiyar: no no... main counter will just calculate all time difference and we pass that to each row... then based on the value each row can decide whether to enable/disable the button or not...

SanjalKatiyar: same thing is happening current implementation as well, just we are using multiple counter to achieve so...

packages/odf/components/storage-consumers/remove-client-modal.tsx
```diff
@@ -0,0 +1,131 @@
...
+}) => {
+  const { t } = useCustomTranslation();
+  const [confirmed, setConfirmed] = React.useState(false);
+  const [inProgress, setProgress] = React.useState(false);
+  const [error, setError] = React.useState<Error>(null);
+  const MODAL_TITLE = t('Permanently delete storage client?');
+
+  const onSubmit = (event) => {
+    event.preventDefault();
+    k8sDelete({ model: StorageConsumerModel, resource })
```
GowthamShanmugam, line 43:
```suggestion
    setProgress(true);
    k8sDelete({ model: StorageConsumerModel, resource })
```

packages/odf/components/storage-consumers/remove-client-modal.tsx
```diff
@@ -0,0 +1,131 @@
...
+      <ModalBody>
+        <p>
+          <Trans t={t}>
+            Deleting the storage client {getName(resource)} will remove all
+            Ceph/Rook resources and erase all data associated with this client,
+            leading to permanent deletion of the client. This action cannot be
+            undone. It will destroy all pods, services and other objects in the
+            namespace{' '}
+            <strong className="co-break-word">
+              {{ name: resource.metadata.name }}
```
GowthamShanmugam, line 73:
```suggestion
              {{ name: getName(resource) }}
```

packages/odf/components/storage-consumers/remove-client-modal.tsx
```diff
@@ -0,0 +1,131 @@
...
+        </p>
+        <p>
+          Recommended only if the storage used by this client is no longer
+          needed, as all stored data will be erased.
+        </p>
+        <p>
+          <Trans t={t}>
+            Confirm deletion by typing&nbsp;
+            <strong className="co-break-word">
+              {{ name: resource.metadata.name }}
```
GowthamShanmugam, line 86:
```suggestion
              {{ name: getName(resource) }}
```

packages/odf/components/storage-consumers/rotate-keys-modal.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const [error, setError] = React.useState<Error>(null);
+
+  const onRotate = () => {
+    setProgress(true);
+    consoleFetch('/api/proxy/plugin/odf-console/provider-proxy/rotate-keys', {
+      method: 'post',
+    })
+      .then((response) => {
+        if (!response.ok) {
+          throw new Error('Network response is not ok!');
```
GowthamShanmugam, line 35: is it throw error or setError to display the error properly?

packages/odf/components/storage-consumers/client-list.tsx
bipuladh: currently I am not passing the value i am setting a state which gets updated only is required instead of rerendering

packages/odf/components/storage-consumers/rotate-keys-modal.tsx
bipuladh: this thrown error is caught below

packages/odf/components/storage-consumers/client-list.tsx
bipuladh: I checked the implementation of useScheduler essentially we are updating the whole list object which will inturn lead to update all the rows. Here this will re-render the row only if it actually required that.  The status being updated only if required.

bipuladh: cc @SanjalKatiyar ^

GowthamShanmugam: If your setter is callback to useSchedler, if setter only updates a particular row then useScheduler won't update whole dataset right?

useScheduler nothing but keeping common logic in a hook, here it can be used as useScheduler(setter)

Anyway its an optional change.

GowthamShanmugam:
LGTM  
cc: @SanjalKatiyar

GowthamShanmugam (approved):

PR #1294, opened 2024-04-16
Adds support for Noobaa standalone in cluster overview

bipuladh:
[Bug 2256563](https://bugzilla.redhat.com/show_bug.cgi?id=2256563)

Changes:
- Fixes StatusAndIconText support for HealthStates; previously it dint support for it.
- Adds text `Healthy` for OK State,
- Moves to a `console.dashboards/overview/health/prometheus` from `console.dashboards/overview/health/resource
`
 

#### Screenshot
Screenshot from 2024-04-16 21-25-12

### DIFF
--- a/packages/odf/components/cluster-overview-extensions/HealthOverview.tsx
+++ b/packages/odf/components/cluster-overview-extensions/HealthOverview.tsx
@@ -1,24 +1,73 @@
 import * as React from 'react';
-import { getCephsHealthState, WatchCephResources } from '@odf/ocs/utils';
-import { healthStateMessage } from '@odf/shared/dashboards/status-card/states';
+import { getNooBaaState } from '@odf/ocs/dashboards/object-service/status-card/statuses';
+import { getCephsHealthState } from '@odf/ocs/utils';
+import { healthStateMapping } from '@odf/shared/dashboards';
 import { CephClusterModel } from '@odf/shared/models';
-import { Status } from '@odf/shared/status/Status';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
 import { referenceForModel } from '@odf/shared/utils';
 import {
-  WatchK8sResults,
   StatusPopupSection,
   HealthState,
+  SubsystemHealth,
+  PrometheusHealthPopupProps,
+  FirehoseResource,
+  K8sResourceCommon,
+  PrometheusHealthHandler,
 } from '@openshift-console/dynamic-plugin-sdk';
+import * as _ from 'lodash-es';
 import { Link } from 'react-router-dom-v5-compat';
 import { Stack, StackItem } from '@patternfly/react-core';
 import '@odf/shared/popup/status-popup.scss';
 
-export const StoragePopover: React.FC<StoragePopoverProps> = ({ ceph }) => {
+export const getStorageSystemHealthState: PrometheusHealthHandler = (
+  promMetrics,
+  t,
+  ceph
+) => {
+  const isNoobaaOnly = _.isEmpty(ceph?.data);
+  if (!isNoobaaOnly) {
+    return getCephsHealthState({
+      ceph: {
+        data: ceph.data as K8sResourceCommon[],
+        loaded: ceph.loaded,
+        loadError: ceph.loadError,
+      },
+    });
+  } else {
+    return getNooBaaState(promMetrics, t, {
+      loaded: true,
+      loadError: '',
+      data: {},
+    });
+  }
+};
+
+export const StoragePopover: React.FC<PrometheusHealthPopupProps> = ({
+  responses,
+  k8sResult,
+}) => {
   const { t } = useCustomTranslation();
 
-  const health = getCephsHealthState({ ceph }, t);
-  const value = health.message || healthStateMessage(health.state, t);
+  const noobaaHealth = getNooBaaState(responses, t, {
+    loaded: true,
+    loadError: '',
+    data: {},
+  });
+  const cephData = k8sResult.data;
+  const isNoobaaOnly = _.isEmpty(k8sResult);
+  const cephHealthState: SubsystemHealth = getCephsHealthState(
+    {
+      ceph: {
+        data: cephData as K8sResourceCommon[],
+        loaded: k8sResult?.loaded,
+        loadError: k8sResult?.loadError,
+      },
+    },
+    t
+  );
+  const healthStatus: HealthState = isNoobaaOnly
+    ? noobaaHealth?.state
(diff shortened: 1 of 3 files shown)


packages/shared/src/selectors/k8s.ts
```diff
@@ -51,3 +55,16 @@ export const getOwnerReferences = <
     value,
     'metadata.ownerReferences'
   ) as K8sResourceCommon['metadata']['ownerReferences'];
+
+type KnownResourceConditions = 'Available' | 'Degraded' | 'Progressing';
+
+export const getResourceCondition = <
```
bipuladh, line 61: Not used right now but is definitely useful for future hence not removing it for now.

packages/odf/components/cluster-overview-extensions/HealthOverview.tsx
```diff
@@ -1,24 +1,76 @@
...
+) => {
+  const isNoobaaOnly = _.isEmpty(ceph?.data);
+  if (!isNoobaaOnly) {
+    return getCephsHealthState({
+      ceph: {
+        data: ceph.data as K8sResourceCommon[],
+        loaded: ceph.loaded,
+        loadError: ceph.loadError,
+      },
+    });
```
SanjalKatiyar, line 35: if it's normal deployment (noobaa + ceph both), should we include both statuses here as well ??

packages/odf/components/cluster-overview-extensions/HealthOverview.tsx
```diff
@@ -1,24 +1,76 @@
...
+      },
+    },
+    t
+  );
+  const healthStatus: HealthState = isNoobaaOnly
+    ? noobaaHealth?.state
+    : cephHealthState?.state;
+  const healthMessage: string = isNoobaaOnly
+    ? noobaaHealth?.message
+    : cephHealthState?.message;
```
SanjalKatiyar, line 73: same here, if we have both noobaa + ceph, should we reflect both ??

SanjalKatiyar (commented):
Can we try to make it as close as possible to SS overview status-card: https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/odf-dashboard/status-card/status-card.tsx#L68-L80 ??
so that there is no mismatch between status/health between 2 different dashboards (someone can raise a bug if there is a mismatch)...

packages/odf/components/cluster-overview-extensions/HealthOverview.tsx
bipuladh: currently to fix bug I am limiting to just supporting Noobaa when Ceph is absent but we need to rework on this to support multicluster scenario properly.

packages/odf/components/cluster-overview-extensions/HealthOverview.tsx
bipuladh: currently to fix bug I am limiting to just supporting Noobaa when Ceph is absent but we need to rework on this to support multicluster scenario properly.

packages/odf/components/cluster-overview-extensions/HealthOverview.tsx
```diff
@@ -48,14 +99,11 @@ export const StoragePopover: React.FC<StoragePopoverProps> = ({ ceph }) => {
...
 
-export const healthResource = {
-  ceph: {
-    kind: referenceForModel(CephClusterModel),
-    namespaced: false,
-    isList: true,
-  },
+export const healthResource: FirehoseResource = {
+  kind: referenceForModel(CephClusterModel),
+  namespaced: true,
```
SanjalKatiyar, line 106:
```suggestion
  namespaced: false,
```

packages/shared/src/status/HealthStatus.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { UnknownIcon } from '@patternfly/react-icons/dist/esm/icons/unknown-icon';
+import { useCustomTranslation } from '../useCustomTranslationHook';
+import { GreenCheckCircleIcon } from './icons';
+
+type HealthStatusWithIconProps = {
+  state: HealthState;
+  message?: string;
+};
+
+export const HealthStatusWithIcon: React.FC<HealthStatusWithIconProps> = ({
```
SanjalKatiyar, line 17: `healthStateMapping` from `shared/src/dashboards/status-card/states.tsx` will not work ?

SanjalKatiyar: I think we have many state mappings like these...

packages/shared/src/status/HealthStatus.tsx
```diff
@@ -0,0 +1,37 @@
...
+  switch (state) {
+    case HealthState.OK:
+      return (
+        <StatusIconAndText
+          icon={<GreenCheckCircleIcon />}
+          title={t('Healthy')}
+        />
+      );
+    case HealthState.PROGRESS:
+      return <StatusIconAndText icon={<HourglassHalfIcon />} title={message} />;
```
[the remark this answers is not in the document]
SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/1294/files#r1568408785

SanjalKatiyar:
/retitle Adds support for Noobaa standalone in cluster overview
(will add bug number in `release-4.16` PR)

packages/shared/src/status/HealthStatus.tsx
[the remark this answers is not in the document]
bipuladh: OCP uses that

packages/odf/components/cluster-overview-extensions/HealthOverview.tsx
bipuladh: It is a namespaced resource.

packages/shared/src/status/HealthStatus.tsx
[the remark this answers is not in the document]
bipuladh: but moving to healthstatemapping for now

packages/odf/components/cluster-overview-extensions/HealthOverview.tsx
bipuladh: I guess it is to query all the Crs from all ns's

SanjalKatiyar: doesn't we set `namespaced: true` when we are passing specific `namespace` to this object as well ??

else if we want to poll CRs from multiple namespaces we pass `namespaced: false` ?? (afaik that's how we do in other places as well)

PR #1309, opened 2024-04-18
Remove proxy call for rotation of signing keys

bipuladh:
- Removes call to proxy service for rotation of signing keys
- Directly deletes Secret from the cluster

[Bug 2275413](https://bugzilla.redhat.com/show_bug.cgi?id=2275413)

### DIFF
--- a/packages/odf/components/storage-consumers/rotate-keys-modal.tsx
+++ b/packages/odf/components/storage-consumers/rotate-keys-modal.tsx
@@ -1,8 +1,9 @@
 import * as React from 'react';
-import { ModalFooter } from '@odf/shared';
+import { ModalFooter } from '@odf/shared/generic';
 import { ModalBody, ModalTitle } from '@odf/shared/generic/ModalTitle';
+import { SecretModel } from '@odf/shared/models';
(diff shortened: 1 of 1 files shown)


packages/odf/components/storage-consumers/rotate-keys-modal.tsx
```diff
@@ -27,17 +29,18 @@ export const RotateKeysModal: RotateKeysModalProps = ({
...
   const onRotate = () => {
     setProgress(true);
-    consoleFetch('/api/proxy/plugin/odf-console/provider-proxy/rotate-keys', {
-      method: 'post',
+    k8sDelete({
+      model: SecretModel,
+      resource: {
+        metadata: {
+          name: 'onboarding-ticket-key',
+          namespace: CEPH_STORAGE_NAMESPACE,
```
SanjalKatiyar, line 37: Please use namespace hook for ODF install namespace and pass that here, we removed `CEPH_STORAGE_NAMESPACE` from everywhere with exception to `alert-action-path.tsx` file. It will be hard if we introduce it again in newer PRs.

PR #1326, opened 2024-04-23
Performance profiles: calculate min. resources taking into account OSDs

alfonsomthd:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2263148

Calculate minimum required resources based on OSD pods to be deployed. Also:
- Refactoring: avoid duplicated renders of 'ConfigurePerformance' & 'EnableTaintNodes' components.


packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -432,42 +392,91 @@ export const CapacityAndNodes: React.FC<CapacityAndNodesProps> = ({
...
+  );
+  const zones = getAllZone(nodes);
+  const flexibleScaling = isFlexibleScaling(nodes.length, zones.size);
+  const deviceSetReplica: number = getDeviceSetReplica(
+    enableArbiter,
+    flexibleScaling,
+    nodes
+  );
+
+  const deviceSetCount = getDeviceSetCount(pvBySc.length, deviceSetReplica);
```
SanjalKatiyar, line 412: everything that is `pv` related is only needed when we display `SelectedCapacityAndNodes` in deployment flow... that is when StorageClass is `isNoProvisioner`...

`deviceSetCount` is not always calculated from `pv`... for normal "internal" mode (when we use infra StorageClass for deployment), it should be `1`...
probably it used to work so far because earlier `capacityAndNodes/pvCount` was only dispatched when `SelectedCapacityAndNodes` was mounted (which was in case on no-prov SC), now it is always being dispatched... so there could be cases when PVs are existing for infra SCes as well and we end up counting them for `deviceSetCount`...

**if possible**, my suggesting will be to move all `pv` related computation inside `SelectedCapacityAndNodes`, which will make sure it only happens when needed...

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -432,42 +392,91 @@ export const CapacityAndNodes: React.FC<CapacityAndNodesProps> = ({
...
+      {(!isNoProvisioner || (pvLoaded && !pvLoadError && !!capacity)) && (
+        <>
+          <ConfigurePerformance
+            onResourceProfileChange={onProfileChange}
+            resourceProfile={resourceProfile}
+            headerText={PerformanceHeaderText}
+            profileRequirementsText={ProfileRequirementsText}
+            selectedNodes={nodes}
+            osdAmount={osdAmount}
+          />
```
SanjalKatiyar, line 470: My opinion on this is that, I don't see implementation where `ConfigurePerformance` is used both inside `SelectedCapacityAndNodes` & `SelectCapacityAndNodes`  as a repetition of code...
`ConfigurePerformance` is an individual functional unit and we can call it separately in two different FCs (`SelectedCapacityAndNodes` & `SelectCapacityAndNodes`)...

For me this implementation is logically segregated (meaning, `SelectedCapacityAndNodes` will be handling `pv` related computations to calculate `osdAmount` and passing that down to its copy of `ConfigurePerformance` and same is true for `SelectCapacityAndNodes`

SanjalKatiyar: we can save all this `useK8sWatchResource` on PVs and dispatch of `capacityAndNodes/pvCount` etc, and do it only when it is needed... that is when `SelectedCapacityAndNodes` is mounted...

SanjalKatiyar: one other way is to probably, use `useState` for `osdAmount`, let `SelectedCapacityAndNodes` & `SelectCapacityAndNodes` set its value (by passing dispatch as a prop) and use that in `ValidationMessage` & `ConfigurePerformance` (meaning `ConfigurePerformance` will stay here inside `CapacityAndNodes`, like u intended to do)...

SanjalKatiyar: there may be better ways to do it (I know older implementation was **not** flawless), but I am just sharing my concerns/opinions...

packages/odf/components/create-storage-system/footer.tsx
```diff
@@ -122,6 +129,15 @@ const canJumpToNextStep = (
...
       : true;
 
+  const zones = getAllZone(nodes);
+  const flexibleScaling = isFlexibleScaling(nodes.length, zones.size);
+  const deviceSetReplica: number = getDeviceSetReplica(
+    capacityAndNodes.enableArbiter,
+    flexibleScaling,
+    nodes
+  );
+  const osdAmount = capacityAndNodes.pvCount * deviceSetReplica;
```
SanjalKatiyar, line 139: same here, regarding `pvCount`... in case of "normal" infra SC based deployment deviceSetCount is not ur PVs count...

packages/odf/components/create-storage-system/footer.tsx
```diff
@@ -122,6 +129,15 @@ const canJumpToNextStep = (
       ? !!(publicNetwork || clusterNetwork)
       : true;
 
+  const zones = getAllZone(nodes);
+  const flexibleScaling = isFlexibleScaling(nodes.length, zones.size);
```
SanjalKatiyar, line 133: I might also add `isNoProvSC` check to, just for safety... as flexible-scaling is only supported for that scenarios...

SanjalKatiyar: same safety check for arbiter...

packages/odf/constants/common.ts
```diff
@@ -24,20 +24,33 @@ export const cephStorageLabel = (ns: string) =>
...
  */
 export const RESOURCE_PROFILE_REQUIREMENTS_MAP: ResourceProfileRequirementsMap =
   {
     [ResourceProfile.Lean]: {
       minCpu: 24,
       minMem: 72,
+      osd: {
+        cpu: 1.5,
+        mem: 3,
+      },
```
SanjalKatiyar, line 37: can we also plz add link to corresponding ocs-operator file as a comment ?? which contains these values...

SanjalKatiyar: it is also fine, if you think that is not necessary...

packages/odf/utils/ocs.ts
```diff
@@ -98,19 +98,40 @@ const getTopologyInfo = (nodes: NodeKind[]) =>
 export const isFlexibleScaling = (nodes: number, zones: number): boolean =>
   !!(nodes >= MINIMUM_NODES && zones < 3);
 
+/**
+ * Returns the minimum required resources taking into account the OSD pods.
+ * Default requirements assume 6 OSDs deployed.
+ */
```
SanjalKatiyar, line 104: do we have any absolute confirmation for this from the team who tested this (same team who came up with "Default requirements") ??
if so, can we plz add link to the official doc etc as a comment ??

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
SanjalKatiyar: there can be attached disks and that's my concern... provisioner based SC should dynamically provision the volumes for OSDs (unlike LSO mode where we need to have it pre-configured)... if there are existing PVs then that will mess up our deviceSetCount, for "normal" mode `replica = n` and `count = 1`.

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
alfonsomthd: > we can save all this `useK8sWatchResource` on PVs and dispatch of `capacityAndNodes/pvCount` etc, and do it only when it is needed... that is when `SelectedCapacityAndNodes` is mounted... also we won't have to use hard to read conditions like: `(!isNoProvisioner || (pvLoaded && !pvLoadError && !!capacity))` for displaying the component...

@SanjalKatiyar Are we really sure that when provisioner we'll NEVER have attached disks the would require pv calculation? Can we dismiss this scenario completely?

packages/odf/utils/ocs.ts
alfonsomthd: @malayparida2000 is the implementer and he confirmed this to me.
@malayparida2000 do you have any open document tha could be linked?

malayparida2000: Hi Sanjal, We haven't documented it anywhere but these initial numbers were given by me & Travis considering a default of 6 osds over a series of discussions.

SanjalKatiyar:
there is one edge-case I see as well (no one reported it in prod yet, so we might be okay... but good to fix): say I selected no-prov SC from the StorageClass dropdown and move forward with the deployment (pvCount will be set in redux)... now I decided to navigate "Back" and select prov SC now... issue is pvCount will still be older value and we **might** end up using it to calculate deviceSetCount somewhere...

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
alfonsomthd: > My opinion on this is that, I don't see implementation where `ConfigurePerformance` is used both inside `SelectedCapacityAndNodes` & `SelectCapacityAndNodes` as a repetition of code... `ConfigurePerformance` is an individual functional unit and we can call it separately in two different FCs (`SelectedCapacityAndNodes` & `SelectCapacityAndNodes`)...
> 
> For me this implementation is logically segregated (meaning, `SelectedCapacityAndNodes` will be handling `pv` related computations to calculate `osdAmount` and passing that down to its copy of `ConfigurePerformance` and same is true for `SelectCapacityAndNodes`

Regardless of the pv discussion, there is code duplication not needed regarding `<ConfigurePerformance>` & `<EnableTaintNodes` that I'm adressing here following the DRY principle (Don't Repeat Yourself).

alfonsomthd:
> there is one edge-case I see as well (no one reported it in prod yet, so we might be okay... but good to fix): say I selected no-prov SC from the StorageClass dropdown and move forward with the deployment (pvCount will be set in redux)... now I decided to navigate "Back" and select prov SC now... issue is pvCount will still be older value and we **might** end up using it to calculate deviceSetCount somewhere...

Addressed: pvCount is reset in <SelectNodesTable> to prevent this issue.

packages/odf/constants/common.ts
alfonsomthd: Done.

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
alfonsomthd: @SanjalKatiyar refactored as per your suggestions assuming no calc. is needed when there is a provisioner.

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
alfonsomthd: Refactored according to `SelectedCapacityAndNodes` suggestions BUT maintaining the decoupling of `ConfigurePerformance` & `EnableTaintNodes` from NodeTable's FCs following SRP (Single Responsibility Principle).

packages/odf/components/create-storage-system/footer.tsx
alfonsomthd: This has been addressed where we set the state for `pvCount`.

packages/odf/components/create-storage-system/footer.tsx
alfonsomthd: Refactored by adding the required to make the method complete & reliable.

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -158,13 +158,15 @@ const SelectCapacityAndNodes: React.FC<SelectCapacityAndNodesProps> = ({
...
-    (profile) => onResourceProfileChange(dispatch)(profile),
-    [dispatch]
-  );
 
   const replicas = getReplicasFromSelectedNodes(nodes);
 
+  React.useEffect(() => {
+    // Reset pvCount that could have been set by another StorageClass.
+    dispatch({ type: 'capacityAndNodes/pvCount', payload: 0 });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
bipuladh, line 167: why can't you keep dispatch in the dependency hook?

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -429,15 +407,40 @@ export const CapacityAndNodes: React.FC<CapacityAndNodesProps> = ({
...
+  const onProfileChange = React.useCallback(
+    (profile) => onResourceProfileChange(dispatch)(profile),
+    [dispatch]
+  );
+
+  React.useEffect(() => {
+    // Reset selected nodes that could have been set by another StorageClass.
+    dispatch({ type: 'wizard/setNodes', payload: [] });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
bipuladh, line 443: same as above

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
alfonsomthd: This `pvCount` reset is included to avoid [a potential issue pointed out by Sanjal](https://github.com/red-hat-storage/odf-console/pull/1326#issuecomment-2076578768) and it is my understanding that the reset should only happen once, (so even if the dispatch method is updated, provided the reset happened it should not happen again).

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
alfonsomthd: Answered above.

PR #1327, opened 2024-04-23
Fix CSS issue with RBD image state

SanjalKatiyar:
https://bugzilla.redhat.com/show_bug.cgi?id=2242832

**Before:**
Screenshot 2024-04-23 at 5 15 08 PM

**After:**
Screenshot 2024-04-23 at 5 21 45 PM
Screenshot 2024-04-23 at 5 22 08 PM

### DIFF
--- a/packages/ocs/dashboards/block-pool/mirroring-card.tsx
+++ b/packages/ocs/dashboards/block-pool/mirroring-card.tsx
@@ -137,7 +137,7 @@ const MirroringImageHealthChart: React.FC<MirroringImageHealthChartProps> = ({
           height={210}
           width={300}
           padding={{
-            right: 160,
+            right: 190,
           }}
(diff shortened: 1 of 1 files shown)


packages/ocs/dashboards/block-pool/mirroring-card.tsx
```diff
@@ -137,7 +137,7 @@ const MirroringImageHealthChart: React.FC<MirroringImageHealthChartProps> = ({
           height={210}
           width={300}
           padding={{
-            right: 160,
+            right: 190,
```
SanjalKatiyar, line 140: this should work for all cases because we are rounding off percentage to one digit after decimal before displaying and above padding renders alright upto 2 decimal places (check attached screenshots)...

PR #1341, opened 2024-04-29
Update topology notification message

TimothyAsirJeyasing:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2276438

### DIFF
--- a/packages/odf/components/topology/Topology.tsx
+++ b/packages/odf/components/topology/Topology.tsx
@@ -125,7 +125,7 @@ const MessageButton: React.FC = () => {
     <div className="odf-topology__message-button">
       <BlueInfoCircleIcon />{' '}
       {showMessage &&
-        t('This view is only supported for Internal mode cluster.')}{' '}
+        t('This view is not available for external mode cluster.')}{' '}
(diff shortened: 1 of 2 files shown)


TimothyAsirJeyasing:
image

packages/odf/components/topology/Topology.tsx
```diff
@@ -125,7 +125,7 @@ const MessageButton: React.FC = () => {
     <div className="odf-topology__message-button">
       <BlueInfoCircleIcon />{' '}
       {showMessage &&
-        t('This view is only supported for Internal mode cluster.')}{' '}
+        t('This view is not available for external mode cluster.')}{' '}
```
alfonsomthd, line 128: @TimothyAsirJeyasing Should be `external` capitalized like `Internal was before` ?

TimothyAsirJeyasing: This was given by the Doc team, Sure i will check with them once.

TimothyAsirJeyasing: @alfonsomthd I've checked with our documentation team, and they confirmed that 'external' should be in lowercase, following our current style guidelines.

PR #1346, opened 2024-04-29
Create block pool: inherit deviceClass from default pool

alfonsomthd:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2253013

### DIFF
--- a/cypress/views/block-pool.ts
+++ b/cypress/views/block-pool.ts
@@ -1,4 +1,7 @@
-import { POOL_PROGRESS } from '../constants/storage-pool-const';
+import {
+  CEPH_DEFAULT_BLOCK_POOL_NAME,
+  POOL_PROGRESS,
+} from '../constants/storage-pool-const';
 import { app } from '../support/pages/app';
 import { NS } from '../utils/consts';
 import { ODFCommon } from '../views/odf-common';
@@ -6,7 +9,6 @@ import { ODFCommon } from '../views/odf-common';
 // Pool var
 export const poolName: string = 'example.pool';
 export const replicaCount: string = '2';
-export const volumeType: string = 'ssd';
 export const scName: string = 'testing-sc';
 
 export const poolMessage: {
@@ -72,18 +74,25 @@ export const verifyFooterActions = (action: string) => {
 export const verifyBlockPoolJSON = (
   compressionEnabled: boolean = true,
   replica: string = replicaCount
-) =>
-  cy.exec(`oc get cephBlockPool ${poolName} -n  ${NS} -o json`).then((res) => {
-    const blockPool = JSON.parse(res.stdout);
-    expect(blockPool.spec?.replicated?.size).to.equal(Number(replica));
-    expect(blockPool.spec?.compressionMode).to.equal(
-      compressionEnabled ? 'aggressive' : 'none'
-    );
-    expect(blockPool.spec?.parameters?.compression_mode).to.equal(
-      compressionEnabled ? 'aggressive' : 'none'
-    );
-    expect(blockPool.spec?.deviceClass).to.equal(volumeType);
+) => {
+  cy.exec(
+    `oc get cephBlockPool ${CEPH_DEFAULT_BLOCK_POOL_NAME} -n ${NS} -o json`
+  ).then((response) => {
+    const defaultBlockPool = JSON.parse(response.stdout);
+    const defaultDeviceClass = defaultBlockPool.spec?.deviceClass;
+    cy.exec(`oc get cephBlockPool ${poolName} -n ${NS} -o json`).then((res) => {
+      const blockPool = JSON.parse(res.stdout);
+      expect(blockPool.spec?.replicated?.size).to.equal(Number(replica));
+      expect(blockPool.spec?.compressionMode).to.equal(
+        compressionEnabled ? 'aggressive' : 'none'
+      );
+      expect(blockPool.spec?.parameters?.compression_mode).to.equal(
+        compressionEnabled ? 'aggressive' : 'none'
+      );
+      expect(blockPool.spec?.deviceClass).to.equal(defaultDeviceClass);
+    });
   });
+};
 
 export const createBlockPool = () => {
   cy.byTestID('item-create').click();
--- a/packages/ocs/block-pool/CreateBlockPool.tsx
+++ b/packages/ocs/block-pool/CreateBlockPool.tsx
@@ -6,6 +6,7 @@ import { CephClusterKind } from '@odf/shared/types';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
 import { referenceForModel } from '@odf/shared/utils';
 import {
+  WatchK8sResource,
   getAPIVersionForModel,
   k8sCreate,
   useK8sWatchResource,
@@ -32,7 +33,8 @@ import './create-block-pool.scss';
 
 export const getPoolKindObj = (
   state: BlockPoolState,
-  ns: string
+  ns: string,
+  deviceClass: StoragePoolKind['spec']['deviceClass']
 ): StoragePoolKind => ({
   apiVersion: getAPIVersionForModel(CephBlockPoolModel),
   kind: CephBlockPoolModel.kind,
@@ -42,7 +44,7 @@ export const getPoolKindObj = (
   },
   spec: {
     compressionMode: state.isCompressed ? COMPRESSION_ON : 'none',
-    deviceClass: state.volumeType || '',
+    deviceClass: deviceClass,
     failureDomain: state.failureDomain,
     parameters: {
       compression_mode: state.isCompressed ? COMPRESSION_ON : 'none',
@@ -58,6 +60,17 @@ export const cephClusterResource = {
   isList: true,
 };
 
+export const poolResource = (
+  poolName: string,
+  ns: string
+): WatchK8sResource => ({
+  kind: referenceForModel(CephBlockPoolModel),
+  namespaced: true,
+  isList: false,
+  name: poolName,
+  namespace: ns,
+});
+
 const CreateBlockPool: React.FC<{}> = ({}) => {
   const { pathname: url } = useLocation();
   const params = useParams();
@@ -71,7 +84,7 @@ const CreateBlockPool: React.FC<{}> = ({}) => {
     blockPoolInitialState
   );
 
-  const [cephClusters, isLoaded, loadError] =
+  const [cephClusters, cephClustersLoaded, cephClustersLoadError] =
     useK8sWatchResource<CephClusterKind[]>(cephClusterResource);
   // only single cluster per Namespace
   const cephCluster = getCephClusterInNs(
@@ -82,6 +95,15 @@ const CreateBlockPool: React.FC<{}> = ({}) => {
   const { systemFlags, areFlagsLoaded, flagsLoadError } =
     useODFSystemFlagsSelector();
   const isExternalStorageSystem = systemFlags[poolNs]?.isExternalMode;
+
+  // Get the default deviceClass.
+  const defaultPoolName = `${systemFlags[poolNs]?.ocsClusterName}-cephblockpool`;
+  const [defaultPool, defaultPoolLoaded, defaultPoolLoadError] =
+    useK8sWatchResource<StoragePoolKind>(poolResource(defaultPoolName, poolNs));
+  const isLoaded = cephClustersLoaded && areFlagsLoaded && defaultPoolLoaded;
+  const isLoadError =
+    cephClustersLoadError || flagsLoadError || defaultPoolLoadError;
+  const defaultDeviceClass = defaultPool?.spec?.deviceClass || '';
 
   // OCS create pool page url ends with ~new, ODF create pool page ends with /create/~new
   const blockPoolPageUrl = params?.appName
@@ -95,7 +117,11 @@ const CreateBlockPool: React.FC<{}> = ({}) => {
   // Create new pool
   const createPool = () => {
     if (cephCluster?.status?.phase === POOL_STATE.READY) {
-      const poolObj: StoragePoolKind = getPoolKindObj(state, poolNs);
+      const poolObj: StoragePoolKind = getPoolKindObj(
+        state,
+        poolNs,
+        defaultDeviceClass
+      );
 
       dispatch({ type: BlockPoolActionType.SET_INPROGRESS, payload: true });
       k8sCreate({ model: CephBlockPoolModel, data: poolObj })
@@ -155,7 +181,7 @@ const CreateBlockPool: React.FC<{}> = ({}) => {
         </p>
       </div>
       <div className="ceph-create-block-pool__form">
-        {isLoaded && areFlagsLoaded && !loadError && !flagsLoadError ? (
+        {isLoaded && !isLoadError ? (
           <>
             <BlockPoolBody
               cephCluster={cephCluster}
@@ -171,8 +197,8 @@ const CreateBlockPool: React.FC<{}> = ({}) => {
           </>
         ) : (
           <StatusBox
-            loadError={loadError || flagsLoadError}
-            loaded={isLoaded && areFlagsLoaded}
+            loadError={isLoadError}
+            loaded={isLoaded}
             label={t('BlockPool Creation Form')}
           />
         )}
--- a/packages/ocs/constants/common.ts
+++ b/packages/ocs/constants/common.ts
@@ -1,7 +1,6 @@
 export const PVC_PROVISIONER_ANNOTATION =
   'volume.beta.kubernetes.io/storage-provisioner';
 export const COMPRESSION_ON = 'aggressive';
-
 export enum POOL_STATE {
   READY = 'Ready',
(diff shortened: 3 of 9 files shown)


packages/ocs/block-pool/CreateBlockPool.tsx
```diff
@@ -58,6 +64,17 @@ export const cephClusterResource = {
   isList: true,
 };
 
+export const poolResource = (
+  poolName: string,
```
GowthamShanmugam, line 68: you can directly use the default pool name inside this function, no need to pass.

packages/ocs/block-pool/CreateBlockPool.tsx
```diff
@@ -83,6 +100,16 @@ const CreateBlockPool: React.FC<{}> = ({}) => {
     useODFSystemFlagsSelector();
   const isExternalStorageSystem = systemFlags[poolNs]?.isExternalMode;
 
+  // Get the default deviceClass.
+  const [defaultPool, defaultPoolLoaded, defaultPoolLoadError] =
+    useK8sWatchResource<StoragePoolKind>(
+      poolResource(DEFAULT_BLOCK_POOL_NAME, poolNs)
+    );
+  const defaultDeviceClass =
```
GowthamShanmugam, line 108: I think we should consider this defaultPoolLoaded and defaultPoolLoadError for modal body render. In case of any  issue in default pool watch, we should not allow create.

packages/ocs/modals/block-pool/create-block-pool-modal.tsx
```diff
@@ -62,6 +54,16 @@ export const CreateBlockPoolModal = withHandlePromise(
     );
     const MODAL_TITLE = t('Create BlockPool');
 
+    // Get the default deviceClass.
+    const [defaultPool, defaultPoolLoaded, defaultPoolLoadError] =
+      useK8sWatchResource<StoragePoolKind>(
+        poolResource(DEFAULT_BLOCK_POOL_NAME, poolNs)
```
GowthamShanmugam, line 60: Same here, shall we block the pool creation up to the pool created without any error?

packages/ocs/block-pool/CreateBlockPool.tsx
alfonsomthd: > you can directly use the default pool name inside this function, no need to pass. may be we could change the function name also like defaultPoolResource

@GowthamShanmugam Could not be a use case where we need it for a pool other than default?

GowthamShanmugam: ok! i got your point

packages/ocs/constants/common.ts
```diff
@@ -1,7 +1,7 @@
 export const PVC_PROVISIONER_ANNOTATION =
   'volume.beta.kubernetes.io/storage-provisioner';
 export const COMPRESSION_ON = 'aggressive';
-
+export const DEFAULT_BLOCK_POOL_NAME = 'ocs-storagecluster-cephblockpool';
```
SanjalKatiyar, line 4: No hard coding of names, it will bite us in case of multiple StorageCluster scenarios. Please use `useODFSystemFlagsSelector` instead.
`${ocsClusterName}-cephblockpool` will be the pool name...

packages/ocs/modals/block-pool/create-block-pool-modal.tsx
SanjalKatiyar: +1

SanjalKatiyar:
`volumeType` is optional now, but we are still checking it in footer: https://github.com/red-hat-storage/odf-console/blob/master/packages/ocs/block-pool/footer.tsx#L12, might need to remove the check from here.
Same here: https://github.com/red-hat-storage/odf-console/blob/master/packages/ocs/modals/block-pool/modal-footer.tsx#L148

SanjalKatiyar:
might as well remove `volumeType` from the redux state itself, not needed anymore...

packages/ocs/block-pool/CreateBlockPool.tsx
SanjalKatiyar: yes please... we could utilise `StatusBox` for this...

packages/ocs/constants/common.ts
alfonsomthd: Done.

packages/ocs/block-pool/CreateBlockPool.tsx
alfonsomthd: Done.

alfonsomthd:
> `volumeType` is optional now, but we are still checking it in footer: https://github.com/red-hat-storage/odf-console/blob/master/packages/ocs/block-pool/footer.tsx#L12, might need to remove the check from here. Same here: https://github.com/red-hat-storage/odf-console/blob/master/packages/ocs/modals/block-pool/modal-footer.tsx#L148

Cleanup done.

packages/ocs/block-pool/CreateBlockPool.tsx
```diff
@@ -83,6 +96,18 @@ const CreateBlockPool: React.FC<{}> = ({}) => {
     useODFSystemFlagsSelector();
   const isExternalStorageSystem = systemFlags[poolNs]?.isExternalMode;
 
+  // Get the default deviceClass.
+  const defaultPoolName = `${systemFlags[poolNs]?.ocsClusterName}-cephblockpool`;
+  const [defaultPool, defaultPoolLoaded, defaultPoolLoadError] =
+    useK8sWatchResource<StoragePoolKind>(poolResource(defaultPoolName, poolNs));
+  const defaultDeviceClass =
```
GowthamShanmugam, line 103:
```suggestion
  const const isLoaded = cephClusterLoaded && areFlagsLoaded && defaultPoolLoaded;
  const isLoadError = cephClusterLoadError || flagsLoadError || defaultPoolLoadError;
  const defaultDeviceClass =
```

packages/ocs/block-pool/CreateBlockPool.tsx
```diff
@@ -83,6 +96,18 @@ const CreateBlockPool: React.FC<{}> = ({}) => {
...
+    areFlagsLoaded &&
+    !flagsLoadError &&
```
GowthamShanmugam, line 105:
```suggestion
    !!defaultPoolName &&
```

packages/ocs/block-pool/CreateBlockPool.tsx
```diff
@@ -155,7 +184,12 @@ const CreateBlockPool: React.FC<{}> = ({}) => {
         </p>
       </div>
       <div className="ceph-create-block-pool__form">
-        {isLoaded && areFlagsLoaded && !loadError && !flagsLoadError ? (
+        {isLoaded &&
+        areFlagsLoaded &&
+        defaultPoolLoaded &&
+        !loadError &&
+        !flagsLoadError &&
+        !defaultPoolLoadError ? (
```
GowthamShanmugam, line 192:
```suggestion
        {isLoaded &&
        !loadError ? (
```

packages/ocs/block-pool/CreateBlockPool.tsx
```diff
@@ -171,8 +205,8 @@ const CreateBlockPool: React.FC<{}> = ({}) => {
           </>
         ) : (
           <StatusBox
-            loadError={loadError || flagsLoadError}
-            loaded={isLoaded && areFlagsLoaded}
+            loadError={loadError || flagsLoadError || defaultPoolLoadError}
+            loaded={isLoaded && areFlagsLoaded && defaultPoolLoaded}
```
GowthamShanmugam, line 209:
```suggestion
            loadError={isLoadError}
            loaded={isLoaded}
```

packages/ocs/modals/block-pool/create-block-pool-modal.tsx
```diff
@@ -62,6 +53,20 @@ export const CreateBlockPoolModal = withHandlePromise(
...
 
+    // Get the default deviceClass.
+    const defaultPoolName = `${systemFlags[poolNs]?.ocsClusterName}-cephblockpool`;
+    const [defaultPool, defaultPoolLoaded, defaultPoolLoadError] =
+      useK8sWatchResource<StoragePoolKind>(
+        poolResource(defaultPoolName, poolNs)
+      );
+    const defaultDeviceClass =
+      areFlagsLoaded &&
+      !flagsLoadError &&
```
GowthamShanmugam, line 64:
```suggestion
      !!defaultPoolName
```

packages/ocs/block-pool/CreateBlockPool.tsx
GowthamShanmugam: this change is just to avoid using loaded and loaded error on multiples places, if we add any thing watcher in future just updating here is enough

packages/ocs/block-pool/CreateBlockPool.tsx
alfonsomthd: @GowthamShanmugam I guess you meant `defaultPool` (as `defaultPoolName` is always true).

GowthamShanmugam: ya! we can even avoid all the checks here if we only use 
!_.isEmpty(defaultPool) ? defaultPool.spec.deviceClass : '';

Becuase if defaultPool non empty means flags loaded without error, and default pool loaded without error 

it is just an optional comment,

alfonsomthd: Makes sense.

GowthamShanmugam (commented):
LGTM

GowthamShanmugam (approved):

cypress/views/block-pool.ts
```diff
@@ -82,7 +81,7 @@ export const verifyBlockPoolJSON = (
     expect(blockPool.spec?.parameters?.compression_mode).to.equal(
       compressionEnabled ? 'aggressive' : 'none'
     );
-    expect(blockPool.spec?.deviceClass).to.equal(volumeType);
+    expect(blockPool.spec?.deviceClass).to.be.undefined;
```
SanjalKatiyar, line 84: it depends, right ? this will not always be `undefined`, we are dependent upon value set in default CephBlockPool CR... IMO we should either remove this check entirely or check equality with default CR's value...

packages/ocs/block-pool/CreateBlockPool.tsx
```diff
@@ -83,6 +97,17 @@ const CreateBlockPool: React.FC<{}> = ({}) => {
...
+  // Get the default deviceClass.
+  const defaultPoolName = `${systemFlags[poolNs]?.ocsClusterName}-cephblockpool`;
+  const [defaultPool, defaultPoolLoaded, defaultPoolLoadError] =
+    useK8sWatchResource<StoragePoolKind>(poolResource(defaultPoolName, poolNs));
+  const isLoaded = cephClustersLoaded && areFlagsLoaded && defaultPoolLoaded;
+  const isLoadError =
+    cephClustersLoadError || flagsLoadError || defaultPoolLoadError;
+  const defaultDeviceClass = !_.isEmpty(defaultPool)
+    ? defaultPool.spec.deviceClass
+    : '';
```
SanjalKatiyar, line 109: nits (optional):
```suggestion
  const defaultDeviceClass = defaultPool?.spec?.deviceClass || '';
```

packages/ocs/block-pool/CreateBlockPool.tsx
```diff
@@ -171,8 +200,8 @@ const CreateBlockPool: React.FC<{}> = ({}) => {
           </>
         ) : (
           <StatusBox
-            loadError={loadError || flagsLoadError}
-            loaded={isLoaded && areFlagsLoaded}
+            loadError={isLoaded}
+            loaded={isLoadError}
```
SanjalKatiyar, line 204:
```suggestion
            loadError={isLoadError}
            loaded={isLoaded}
```

packages/ocs/modals/block-pool/create-block-pool-modal.tsx
```diff
@@ -62,6 +54,16 @@ export const CreateBlockPoolModal = withHandlePromise(
...
 
+    // Get the default deviceClass.
+    const defaultPoolName = `${systemFlags[poolNs]?.ocsClusterName}-cephblockpool`;
+    const [defaultPool, defaultPoolLoaded, defaultPoolLoadError] =
+      useK8sWatchResource<StoragePoolKind>(
+        poolResource(defaultPoolName, poolNs)
+      );
+    const defaultDeviceClass = !_.isEmpty(defaultPool)
+      ? defaultPool.spec.deviceClass
+      : '';
```
SanjalKatiyar, line 65: same here, optional change...

packages/ocs/modals/block-pool/create-block-pool-modal.tsx
```diff
@@ -121,12 +123,22 @@ export const CreateBlockPoolModal = withHandlePromise(
 
     // Create new pool
     const createPool = () => {
-      if (state.poolStatus === '') {
+      if (
+        state.poolStatus === '' &&
+        areFlagsLoaded &&
+        !flagsLoadError &&
+        defaultPoolLoaded &&
+        !defaultPoolLoadError
```
SanjalKatiyar, line 131: instead of here, IMO we can add it before ModalBody and ModalFooter is even rendered. Using `StatusBox`... similar to how we did it for `CreateBlockPool` page.

packages/ocs/modals/block-pool/create-block-pool-modal.tsx
```diff
@@ -62,6 +54,16 @@ export const CreateBlockPoolModal = withHandlePromise(
     );
     const MODAL_TITLE = t('Create BlockPool');
 
+    // Get the default deviceClass.
+    const defaultPoolName = `${systemFlags[poolNs]?.ocsClusterName}-cephblockpool`;
+    const [defaultPool, defaultPoolLoaded, defaultPoolLoadError] =
+      useK8sWatchResource<StoragePoolKind>(
+        poolResource(defaultPoolName, poolNs)
+      );
```
SanjalKatiyar, line 62: we are already polling for all CephBlockPool CRs in `PoolResourceComponent` component which call this modal. Maybe we can pass the list and filter here, instead of polling one specific pool here again.

cypress/views/block-pool.ts
alfonsomthd: The ceph block pool CR we're testing here is created without a deviceClass because the default pool doesn't have it either.
I don't see the point of complicating unnecessary the test or doing premature optimization.

SanjalKatiyar: `because the default pool doesn't have it either` >> but is it guaranteed that it won't have it in next release as well ? If ocs will make any change, it will affect our e2e CI... so better to be robust and be close to prod like flow (i.e., checking equality on default CR directly which ocs creates)...
or we can remove this `expect` condition entirely, that fine IMO as well...

packages/ocs/block-pool/CreateBlockPool.tsx
alfonsomthd: My bad: good catch!

cypress/views/block-pool.ts
alfonsomthd: And who guarantees that the current behavior won't change in the next release? :wink: 
Anyway I'll address this.

SanjalKatiyar: both are different scenarios, maybe I will try to explain what I meant a bit more...

we don't know in which case ocs adds this spec and in which case it don't (maybe based on infra, maybe type of deployment etc), so best condition is to keep it same as how we implemented it from the UI, i.e, e2e should resemble feature implementation (where spec is same as what's mentioned in default CR, not always undefined)...

if feature itself changes (which is far-fetched thing), then yes you are right we can't do anything in that, we have to update test again anyway...

packages/ocs/modals/block-pool/create-block-pool-modal.tsx
alfonsomthd: Refactored.

packages/ocs/modals/block-pool/create-block-pool-modal.tsx
alfonsomthd: After last refactoring, neither this nor StatusBox are needed.

cypress/views/block-pool.ts
SanjalKatiyar: @alfonsomthd does this comment got missed ? I can see it as resolved, but changes looks like they were before.

SanjalKatiyar:
LGTM, just https://github.com/red-hat-storage/odf-console/pull/1346#discussion_r1599374841 (let me know if we want to merge this PR asap and fix e2e in master only @alfonsomthd, but better to add `ToDo` comment in that case IMO or create a GH issue).

alfonsomthd:
> LGTM, just [#1346 (comment)](https://github.com/red-hat-storage/odf-console/pull/1346#discussion_r1599374841) (let me know if we want to merge this PR asap and fix e2e in master only @alfonsomthd, but better to add `ToDo` comment in that case IMO or create a GH issue).

It's already pushed.

PR #1354, opened 2024-05-03
Fixes storage client resource name in list page

bipuladh:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2262975

### DIFF
--- a/packages/odf/components/storage-consumers/client-list.tsx
+++ b/packages/odf/components/storage-consumers/client-list.tsx
@@ -3,7 +3,6 @@ import { Kebab } from '@odf/shared';
 import { ODF_OPERATOR } from '@odf/shared/constants/common';
 import { getTimeDifferenceInSeconds } from '@odf/shared/details-page/datetime';
 import { useFetchCsv } from '@odf/shared/hooks';
-import { getName } from '@odf/shared/selectors/k8s';
 import { StorageConsumerKind } from '@odf/shared/types';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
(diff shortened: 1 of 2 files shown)


packages/shared/src/types/storage.ts
```diff
@@ -177,6 +177,9 @@ type StorageConsumerStatus = {
 type ClientStatus = {
   platformVersion: string;
   operatorVersion: string;
+  clusterId: string;
+  clusterName: string;
```
GowthamShanmugam, line 181: clusterName is not used anywhere

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -83,7 +82,7 @@ const ClientsList: React.FC<ClientListProps> = (props) => {
     () => [
       {
         title: t('Name'),
-        sort: 'metadata.name',
+        sort: 'status.client.name',
```
GowthamShanmugam, line 85: dont we need an option symbol here like status?.client?.name? Does this won't break when the client list is empty?

bipuladh: This is used inside of lodash get.

packages/shared/src/types/storage.ts
bipuladh: However they have updated the type of ClientStatus so even if it's not used now the type definition can be useful for later.

GowthamShanmugam: ok

GowthamShanmugam (approved):

alfonsomthd:
LGTM

PR #1360, opened 2024-05-03
Removes namespace bar from Client creation flow

bipuladh:
Removes Performance configuration from the UI


GowthamShanmugam (commented):
LGTM

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -464,14 +469,16 @@ export const CapacityAndNodes: React.FC<CapacityAndNodesProps> = ({
...
         <>
-          <ConfigurePerformance
-            onResourceProfileChange={onProfileChange}
-            resourceProfile={resourceProfile}
-            headerText={PerformanceHeaderText}
-            profileRequirementsText={ProfileRequirementsText}
-            selectedNodes={nodes}
-            osdAmount={osdAmount}
-          />
+          {!isProviderMode && (
```
alfonsomthd, line 472: @bipuladh If we choose to not show the performance profile, the nwe should not include the associated validations here, right? Otherwise it could happen that the user is prevented from moving forward.
https://github.com/alfonsomthd/odf-console/blob/fix-deviceclass/packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx#L435

packages/odf/modals/configure-performance/configure-performance-modal.tsx
```diff
@@ -183,13 +189,15 @@ const ConfigurePerformanceModal: React.FC<ConfigurePerformanceModalProps> = ({
...
     >
       <ModalBody className="configure-performance-modal--overflow">
-        <ConfigurePerformance
-          onResourceProfileChange={onProfileChange}
-          resourceProfile={resourceProfile}
-          profileRequirementsText={ProfileRequirementsModalText}
-          selectedNodes={selectedNodes}
-          osdAmount={osdAmount}
-        />
+        {isProviderMode && (
```
alfonsomthd, line 192: This should be negated ?
```suggestion
        {!isProviderMode && (
```

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
bipuladh: This makes sense. We will work with balanced mode as resource profile and remove validations. As they don't support lean mode we will not default to lean mode for situations when resources are not available.

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -464,23 +471,25 @@ export const CapacityAndNodes: React.FC<CapacityAndNodesProps> = ({
...
           <EnableTaintNodes dispatch={dispatch} enableTaint={enableTaint} />
         </>
       )}
       {!!validations.length &&
         !!capacity &&
         validations.map((validation) => (
           <ValidationMessage
-            resourceProfile={resourceProfile}
-            osdAmount={osdAmount}
+            resourceProfile={isProviderMode ? undefined : resourceProfile}
```
alfonsomthd, line 491: @bipuladh This is not preventing the validation, only making the error message incomplete/misleading (missing the resource profile name before word `mode`). Tested locally:
image

As you mentioned that for Provider Mode we'll go by default with `Balanced`, I would revert this change.
From your above comment:
`We will work with balanced mode as resource profile and remove validations. As they don't support lean mode we will not default to lean mode for situations when resources are not available.`
If what you want is to ignore resource profile validations when in Provider Mode, then I suggest to add the providerMode condition [here](https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/utils/common.ts#L166):
`} else if (!isProviderMode && nodes.length && nodes.length >= MINIMUM_NODES) {`

alfonsomthd (changes requested):

bipuladh:
Updates as requested: 
Screenshots: 
Screenshot from 2024-05-10 15-35-21
Screenshot from 2024-05-10 15-35-46

PR #1378, opened 2024-05-13
Fixes provider endpoint validations

bipuladh:
[Bug 2278676](https://bugzilla.redhat.com/show_bug.cgi?id=2278676)


alfonsomthd:
@bipuladh we should add (not necessarily in this PR) some unit tests related to URL address validation to prevent future regressions.

PR #1404, opened 2024-05-23
Remove storage class prefix for client deployments

bipuladh:
https://bugzilla.redhat.com/show_bug.cgi?id=2282314


packages/ocs/dashboards/persistent-external/fallback-hook.ts
```diff
@@ -1,37 +1,13 @@
...
-
-  const provisioners = storageClasses.map((sc) => sc.provisioner);
-  const fileSystemProvisioner = provisioners.find((item) =>
-    item.includes(fsProvisionerPostFix)
-  );
-  const clusterName = fileSystemProvisioner?.split(fsProvisionerPostFix)?.[0];
-  return storageClassesLoaded && !storageClassLoadError
-    ? clusterName
-    : 'ocs-storagecluster';
+  return '';
```
SanjalKatiyar, line 10: no need for calling it a hook now (and adding a separate file for this)... we can add it as a normal function under `ocs/utils`...

packages/ocs/dashboards/persistent-external/client-fallback.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * This hooks is for ODF Client Mode use case only.
```
SanjalKatiyar, line 2: not a hook anymore :P
```suggestion
 * This is for ODF Client Mode use case only.
```

PR #1409, opened 2024-05-30
Fix dashboard queries (Inventory + Capacity + Utilization cards)

SanjalKatiyar:
Issues:
- Inventory card always showing all Ceph based PVs + PVCs (across all the Ceph clusters, this is incorrect for multiple StorageCluster scenarios).
- Capacity & Utilization card only showing data for default OCS StorageClasses and skipping the custom created ones. 

> Below cluster had:
> 2 default & 1 custom "internal" PVs + PVCs, mounted to an active Pod.
> 1 default & 1 custom "external" PVs + PVCs, mounted to an active Pod.
> 
> 1 default & 1 custom "internal" PVs + PVCs, **not** mounted to an active Pod.
> 1 default "external" PVs + PVCs, **not** mounted to an active Pod.

### Internal
Screenshot 2024-05-30 at 4 42 01 PM

### External
Screenshot 2024-05-30 at 4 42 09 PM

----------------------------------------------------------------------------------------------------------------------

### Internal
Screenshot 2024-05-31 at 1 34 31 AM

### External
Screenshot 2024-05-31 at 1 36 19 AM


packages/ocs/hooks/useStorageClassQueryFilter.ts
```diff
@@ -0,0 +1,38 @@
...
+import {
+  useK8sWatchResource,
+  WatchK8sResource,
+} from '@openshift-console/dynamic-plugin-sdk';
+import { getSpecificClusterCephSC } from '../utils';
+
+const scResource: WatchK8sResource = {
+  kind: StorageClassModel.kind,
+  isList: true,
+};
```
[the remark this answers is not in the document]
SanjalKatiyar: at this point we are doing this everywhere, instead of importing it from resources file...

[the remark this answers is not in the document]
SanjalKatiyar: at this point we are doing this everywhere, instead of importing it from resources file...

packages/ocs/hooks/useStorageClassQueryFilter.ts
```diff
@@ -0,0 +1,38 @@
...
+    useK8sWatchResource<StorageClassResourceKind[]>(scResource);
+
+  let filteredStorageClasses: StorageClassResourceKind[] | string[] = [];
+
+  if (!clusterNs) {
+    /**
+     * This is for ODF Client Mode use case only ("clusterNs" will be "undefined" when this hook will be called from client-console).
+     * ODF client mode doesn't deploy ODF StorageCluster hence no "clusterNs", moreover Route for dashboards will be differnt as well.
+     * For scenarios where Client & Provider are on same OCP, client-console will be disabled (as per design).
+     * Considering all the StorageClasses in the cluster in such case.
```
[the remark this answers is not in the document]
SanjalKatiyar: currently assumption is client is connected to only a single provider and everything is provisioned from that...
in case client/provider are on same cluster, then cleint-console will be disabled, so this UI won't render...

packages/ocs/hooks/useStorageClassQueryFilter.ts
```diff
@@ -0,0 +1,38 @@
...
+     */
+    filteredStorageClasses = ['.*'];
+  } else {
+    filteredStorageClasses = getSpecificClusterCephSC(scData, clusterNs)?.map(
```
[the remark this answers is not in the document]
SanjalKatiyar: `getCephSC` is an existing function which returns SC across all ceph clusters (in case of multiple cluster scenarios)...
on top of existing function I added `getSpecificClusterCephSC`, which will return SC corresponding to `SpecificCluster` based on namespace argument...

packages/ocs/hooks/useStorageClassQueryFilter.ts
```diff
@@ -0,0 +1,38 @@
...
+import { getSpecificClusterCephSC } from '../utils';
+
+const scResource: WatchK8sResource = {
+  kind: StorageClassModel.kind,
+  isList: true,
+};
+
+export const useStorageClassQueryFilter = (
+  clusterNs: string
+): [string, boolean, unknown] => {
```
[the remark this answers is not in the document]
SanjalKatiyar: it error from `useK8sWatchResource` hook, it's either `any` or `unknown`...

packages/ocs/dashboards/persistent-internal/utilization-card/utilization-card.tsx
```diff
@@ -32,79 +34,88 @@ export const UtilizationContent: React.FC = () => {
...
+              UTILIZATION_QUERY(ocsCluster)[
+                StorageDashboardQuery.UTILIZATION_RECOVERY_RATE_QUERY
+              ]
+            }
+            humanizeValue={humanizeDecimalBytesPerSec}
+          />
+        </>
+      )}
+      {!scQueryfilterLoaded && !scQueryfilterError && <GraphEmpty loading />}
+      {!!scQueryfilterError && <GraphEmpty />}
```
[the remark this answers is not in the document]
SanjalKatiyar: skipping... as there are more than 2 conditions required here, so ternary won't do any significant good...

packages/ocs/utils/common.ts
```diff
@@ -9,17 +9,31 @@ import {
...
 } from '@odf/shared/types';
 import { DataPoint, humanizePercentage } from '@odf/shared/utils';
 import { EventKind } from '@openshift-console/dynamic-plugin-sdk-internal/lib/api/internal-types';
 import * as _ from 'lodash-es';
+import { compose } from 'redux';
+import { OCS_STORAGECLASS_PARAMS } from '../constants';
+
+const enum Status {
+  BOUND = 'Bound',
+  AVAILABLE = 'Available',
```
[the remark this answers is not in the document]
SanjalKatiyar: moved to shared...

packages/ocs/utils/common.ts
```diff
@@ -9,17 +9,31 @@ import {
...
+}
+const isBound = (pvc: K8sResourceKind) => pvc.status.phase === Status.BOUND;
```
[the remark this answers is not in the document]
SanjalKatiyar: updated... but this is an existing function (I did not add it in this PR) and it never errored so far...

SanjalKatiyar:
cc @bipuladh @GowthamShanmugam can you take a look and tag if it's alright ??

packages/ocs/utils/common.ts
```diff
@@ -75,15 +87,27 @@ export const isPersistentStorageEvent =
...
+  scData: StorageClassResourceKind[] = []
+): StorageClassResourceKind[] =>
   scData.filter((sc) => {
     return cephStorageProvisioners.some((provisioner: string) =>
       (sc?.provisioner).includes(provisioner)
     );
   });
 
+// All Ceph based StorageClasses from a particular Ceph cluster (multiple StorageSystem/StorageCluster scenario)
+export const getSpecificClusterCephSC = compose(
```
bipuladh, line 101:
```suggestion
export const filterStorageClassByCluster = compose(
```

packages/ocs/constants/common.ts
```diff
@@ -31,9 +31,30 @@ export enum CLUSTER_STATUS {
...
+// https://github.com/rook/rook/blob/release-1.7/cluster/examples/kubernetes/ceph/flex/storageclass.yaml
+export const CLUSTER_NS = 'clusterNamespace';
+
+// can be used to determine a Ceph cluster in case of multiple StorageSystem/StorageCluster scenarios
+// we only have a single Ceph cluster per namespace
+export const OCS_STORAGECLASS_PARAMS = [
+  // represents the namespace of the Ceph cluster to provision storage from
+  CLUSTER_ID,
+  // represents the namespace of the secrets for user and/or Ceph admin credential
+  PROV_SECRET_NS,
```
bipuladh, line 53: These two can be in different namespaces?

packages/ocs/hooks/useStorageClassQueryFilter.ts
```diff
@@ -0,0 +1,38 @@
...
+  let filteredStorageClasses: StorageClassResourceKind[] | string[] = [];
+
+  if (!clusterNs) {
+    /**
+     * This is for ODF Client Mode use case only ("clusterNs" will be "undefined" when this hook will be called from client-console).
+     * ODF client mode doesn't deploy ODF StorageCluster hence no "clusterNs", moreover Route for dashboards will be differnt as well.
+     * For scenarios where Client & Provider are on same OCP, client-console will be disabled (as per design).
+     * Considering all the StorageClasses in the cluster in such case.
+     */
+    filteredStorageClasses = ['.*'];
```
bipuladh, line 30:
```suggestion
    filteredStorageClasses = ['.*'];
    return [filteredStorageClasses, true, null]
```
No need to process sc data incase of client mode.

packages/ocs/utils/common.ts
```diff
@@ -5,21 +5,33 @@ import {
...
   StorageClusterKind,
+  PersistentVolumeClaimKind,
 } from '@odf/shared/types';
 import { DataPoint, humanizePercentage } from '@odf/shared/utils';
 import { EventKind } from '@openshift-console/dynamic-plugin-sdk-internal/lib/api/internal-types';
 import * as _ from 'lodash-es';
+import { compose } from 'redux';
+import { OCS_STORAGECLASS_PARAMS } from '../constants';
+
+const isBound = (pvc: K8sResourceKind) =>
```
bipuladh, line 28:
```suggestion
const isBound = (pvc: K8sResourceKind): boolean =>
```

packages/ocs/utils/common.ts
```diff
@@ -5,21 +5,33 @@ import {
...
+  pvc?.status?.phase === PVC_STATUS.BOUND;
+const getPVStorageClass = (pv: K8sResourceKind) =>
```
bipuladh, line 30:
```suggestion
const getPVStorageClass = (pv: K8sResourceKind): string =>
```

packages/ocs/utils/common.ts
```diff
@@ -5,21 +5,33 @@ import {
...
+  pv?.spec?.storageClassName ||
+  pv?.metadata?.annotations?.[
+    'volume.kubernetes.io/provisioner-deletion-secret-namespace'
```
bipuladh, line 33: are there scenarios where `spec.storageClassName` is not defined?

packages/ocs/utils/common.ts
bipuladh, line 101: Please add type declaration to make things more readable.

packages/ocs/utils/common.ts
```diff
@@ -94,33 +118,47 @@ export const getCephNodes = (
...
   });
 
-const enum Status {
-  BOUND = 'Bound',
-  AVAILABLE = 'Available',
-}
-const isBound = (pvc: K8sResourceKind) => pvc.status.phase === Status.BOUND;
-const getPVStorageClass = (pv: K8sResourceKind) => pv?.spec?.storageClassName;
+// All Ceph based PVs from a particular Ceph cluster (multiple StorageSystem/StorageCluster scenario)
+export const getSpecificClusterCephPVs = compose(
```
bipuladh, line 136:
```suggestion
export const filterPVsByCluster = compose(
```

packages/ocs/constants/common.ts
SanjalKatiyar: it will be same as corresponding Ceph cluster namespace... and Ceph cluster itself can be in any namespace (theoretically)...

packages/ocs/utils/common.ts
SanjalKatiyar: It should not happen, but it an optional spec: `storageClassName is the name of StorageClass to which this persistent volume belongs. Empty value means that this volume does not belong to any StorageClass.`.

SanjalKatiyar: Still, I think I added it by mistake... function name is `getPVStorageClass`, but `volume.kubernetes.io/provisioner-deletion-secret-namespace` gives the namespace... so anyway this annotation is incorrect to use here. I need to check why I added it.

packages/ocs/constants/common.ts
SanjalKatiyar: maybe in upstream we support diff namespaces as well... but for ODF it should be same...

packages/ocs/utils/common.ts
SanjalKatiyar: we had existing function named `getCephSC`... `getSpecificClusterCephSC` was wrapper/extention of that, hence named that way...

SanjalKatiyar: anyway will rename as per the suggestion...

packages/ocs/constants/common.ts
SanjalKatiyar: will confirm once...

packages/ocs/utils/common.ts
SanjalKatiyar: removed it...

packages/ocs/constants/common.ts
SanjalKatiyar: it is Ceph cluster namespace...

packages/ocs/constants/common.ts
```diff
@@ -31,9 +31,30 @@ export enum CLUSTER_STATUS {
...
+  NODE_SECRET_NS,
+  CONTROLLER_SECRET_NS,
+  // represents the namespace of the Rook cluster from which to create volumes
+  CLUSTER_NS,
```
bipuladh, line 57: Do we need to keep track of all these?

SanjalKatiyar: `CLUSTER_NS` is for `ceph.rook.io/block` provisioner which is legacy, but we haven't cleaned it up from other places in our code... so better to do it once at all places in a single PR...

`PROV_SECRET_NS` `NODE_SECRET_NS` `CONTROLLER_SECRET_NS` is for ceph-rbd/fs, only reason I added it is for safety, in case one of the parameter (say `CLUSTER_ID`) is removed manually by mistake, we can check equality on other params... seems harmless to me...

but let me know in case u see any concern ??

PR #1410, opened 2024-05-30
Add separate provider for context and pass as children

TimothyAsirJeyasing:
Fixes: https://github.com/red-hat-storage/odf-console/issues/672

### DIFF
--- a/packages/mco/components/mco-dashboard/disaster-recovery/dr-dashboard.tsx
+++ b/packages/mco/components/mco-dashboard/disaster-recovery/dr-dashboard.tsx
@@ -71,7 +71,7 @@ const aggregateAppsMap = (
     return acc;
   }, {});
 
-const MonitoringDashboard: React.FC = () => {
+const CSVStatusesProvider: React.FC = ({ children }) => {
   const [csvData, csvError, csvLoading] = useCustomPrometheusPoll({
     endpoint: 'api/v1/query' as any,
     query: STATUS_QUERIES[StorageDashboard.CSV_STATUS_ALL_WHITELISTED],
@@ -79,6 +79,14 @@ const MonitoringDashboard: React.FC = () => {
     cluster: HUB_CLUSTER_NAME,
   });
 
+  return (
+    <CSVStatusesContext.Provider value={{ csvData, csvError, csvLoading }}>
+      {children}
+    </CSVStatusesContext.Provider>
+  );
+};
+
+const DRResourcesProvider: React.FC = ({ children }) => {
   const [managedClusters, managedClusterLoaded, managedClusterLoadError] =
     useK8sWatchResource<ACMManagedClusterKind[]>(
       getManagedClusterResourceObj()
@@ -122,16 +130,20 @@ const MonitoringDashboard: React.FC = () => {
     loadError,
   };
 
-  // ToDo(Sanjal): combime multiple Context together to make it scalable
-  // refer: https://javascript.plainenglish.io/how-to-combine-context-providers-for-cleaner-react-code-9ed24f20225e
   return (
-    <CSVStatusesContext.Provider value={{ csvData, csvError, csvLoading }}>
-      <DRResourcesContext.Provider value={dRResourcesContext}>
-        <UpperSection />
-      </DRResourcesContext.Provider>
-    </CSVStatusesContext.Provider>
+    <DRResourcesContext.Provider value={dRResourcesContext}>
+      {children}
+    </DRResourcesContext.Provider>
   );
 };
+
+const MonitoringDashboard: React.FC = () => (
+  <CSVStatusesProvider>
+    <DRResourcesProvider>
+      <UpperSection />
+    </DRResourcesProvider>
+  </CSVStatusesProvider>
+);
 
 const DRDashboard: React.FC = () => {
   const isMonitoringEnabled = useFlag(ACM_OBSERVABILITY_FLAG);
--- a/packages/odf/components/topology/Topology.tsx
+++ b/packages/odf/components/topology/Topology.tsx
@@ -18,9 +18,11 @@ import {
   createNode,
   defaultLayoutFactory,
   stylesComponentFactory,
+  TopologyViewLevel,
+} from '@odf/shared/topology';
+import {
   TopologyDataContext,
   TopologySearchContext,
-  TopologyViewLevel,
 } from '@odf/shared/topology';
 import {
   ClusterServiceVersionKind,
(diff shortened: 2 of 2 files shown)


packages/odf/components/topology/Topology.tsx
```diff
@@ -572,7 +592,7 @@ const Topology: React.FC = () => {
   const zones = memoizedNodes.map(getTopologyDomain);
 
   return (
-    <TopologyDataContext.Provider
+    <TopologyDataProvider
```
GowthamShanmugam, line 595: For the MCO dashboard, CSVStatusesProvider & DRResourcesProvider are not taking any arguments directly so it will optimize the rendering.  But here, we are still passing the arguments, I don't see any advantage other than the code formatting.

TimothyAsirJeyasing: Ok, I understand your concern. To optimize the rendering, I will use useMemo here for the context values to ensure that the providers only re-render when their respective values actually change.

GowthamShanmugam: usememo wont help here

GowthamShanmugam: My doubt is,
   If the argument changes, whether the provider in a different component or under the same component will render all of its children. How does this composition help?

packages/odf/components/topology/Topology.tsx
```diff
@@ -99,6 +101,24 @@ type SideBarProps = {
   isExpanded: boolean;
 };
 
+const TopologyDataProvider: React.FC<{ value: any }> = ({
```
bipuladh, line 104: add types

packages/odf/components/topology/Topology.tsx
```diff
@@ -99,6 +101,24 @@ type SideBarProps = {
...
+  children,
+  value,
+}) => (
+  <TopologyDataContext.Provider value={value}>
+    {children}
+  </TopologyDataContext.Provider>
+);
+
+const TopologySearchProvider: React.FC<{ value: any }> = ({
```
bipuladh, line 113: add types

packages/mco/components/mco-dashboard/disaster-recovery/dr-dashboard.tsx
```diff
@@ -122,17 +130,23 @@ const MonitoringDashboard: React.FC = () => {
...
-      </DRResourcesContext.Provider>
-    </CSVStatusesContext.Provider>
+    <DRResourcesContext.Provider value={dRResourcesContext}>
+      {children}
+    </DRResourcesContext.Provider>
   );
 };
 
+// ToDo(Sanjal): combime multiple Context together to make it scalable
+// refer: https://javascript.plainenglish.io/how-to-combine-context-providers-for-cleaner-react-code-9ed24f20225e
```
alfonsomthd, line 141: If this have been addressed by this PR, these comments are no longer needed, right?

packages/odf/components/topology/Topology.tsx
TimothyAsirJeyasing: When context values are memoized using useMemo, the context provider will only re-render its children when its value actually changes. This means that unrelated components won't be affected by changes in the context they don't consume. I will use useMemo, to enhance and address as follows:

const DRResourcesProvider: React.FC = ({ children }) => {
  const value = useMemo(() => dRResourcesContext, [dRResourcesContext]);

  return (
    <DRResourcesContext.Provider value={value}>
      {children}
    </DRResourcesContext.Provider>
  );
};

packages/mco/components/mco-dashboard/disaster-recovery/dr-dashboard.tsx
SanjalKatiyar: that's different from changes added by this PR... still we can remove this comment, we don't need that kind of clean right now.

packages/odf/components/topology/Topology.tsx
```diff
@@ -99,6 +101,24 @@ type SideBarProps = {
   isExpanded: boolean;
 };
 
+const TopologyDataProvider: React.FC<{ value: any }> = ({
+  children,
+  value,
```
SanjalKatiyar, line 106: refactor in a way that provider computes/passes this value directly instead of receiving it as a prop...

packages/odf/components/topology/Topology.tsx
```diff
@@ -99,6 +101,24 @@ type SideBarProps = {
...
+}) => (
+  <TopologyDataContext.Provider value={value}>
+    {children}
+  </TopologyDataContext.Provider>
+);
+
+const TopologySearchProvider: React.FC<{ value: any }> = ({
+  children,
+  value,
```
SanjalKatiyar, line 115: same here...

packages/odf/components/topology/Topology.tsx
SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/1410#discussion_r1633479899

packages/mco/components/mco-dashboard/disaster-recovery/dr-dashboard.tsx
TimothyAsirJeyasing: Thank you for your feedback! I’ll address this in a separate PR later.

packages/odf/components/topology/Topology.tsx
SanjalKatiyar: reason for doing so: each provider can deal with its individual/independent functionality, instead of doing everything in main `Topology` FC.

TimothyAsirJeyasing: Ack

packages/odf/components/topology/Topology.tsx
```diff
@@ -458,81 +603,11 @@ const TopologyViewComponent: React.FC = () => {
...
-    useSafeK8sWatchResource<K8sResourceCommon[]>(odfStatefulSetResource);
-
-  const [replicaSets, replicaSetsLoaded, replicaSetsError] =
-    useSafeK8sWatchResource<K8sResourceCommon[]>(odfReplicaSetResource);
-
-  const [daemonSets, daemonSetsLoaded, daemonSetError] =
-    useSafeK8sWatchResource<K8sResourceCommon[]>(odfDaemonSetResource);
-
-  // ToDo (epic 4422): This will work as Internal mode cluster will only be created in ODF install namespace.
-  // Still, make this generic so that this works even if it gets created in a different namespace.
```
SanjalKatiyar, line 498: when u r copying and pasting make sure to not miss anything, even something as simple as comment... this is an important comment, plz add it back to the new provider component...

packages/odf/components/topology/Topology.tsx
```diff
@@ -99,6 +101,149 @@ type SideBarProps = {
...
+    deployments: memoizedDeployments,
+    visualizationLevel: visualizationLevel,
+    activeNode,
+    setActiveNode,
+    nodeDeploymentMap,
+    selectedElement,
+    setSelectedElement,
+  };
+
+  const Error = ({ error }) => <>{error}</>;
```
SanjalKatiyar, line 182: was this nested earlier like this ??

SanjalKatiyar:
plz test the PR properly and attach screenshots...

packages/odf/components/topology/Topology.tsx
TimothyAsirJeyasing: Earlier,

<TopologyDataContext.Provider>
  <TopologySearchContext.Provider>
    <VisualizationProvider>
      <HandleErrorAndLoading>
        <TopologyViewComponent />
      </HandleErrorAndLoading>
    </VisualizationProvider>
  </TopologySearchContext.Provider>
</TopologyDataContext.Provider>

The TopologySearchContext.Provider is nested within the TopologyDataContext.Provider, and the VisualizationProvider contains the child components: TopologyTopBar, HandleErrorAndLoading, and TopologyViewComponent. I have localized all the data and error states related to TopologyDataProvider inside, along with HandleErrorAndLoading, as it specifically/closely deals with TopologyDataProvider.

SanjalKatiyar: don't nest FCs...

TimothyAsirJeyasing:
image
image

packages/odf/components/topology/Topology.tsx
TimothyAsirJeyasing: const Error = ({ error }) => <>{error}</>;
The Error component used within HandleErrorAndLoading appears to be a simple functional component rendered inline was used like this only before.

packages/odf/components/topology/Topology.tsx
```diff
@@ -99,6 +101,151 @@ type SideBarProps = {
...
+}) => {
+  const [activeItemsUID, setActiveItemsUID] = React.useState<string[]>([]);
+  const [activeItem, setActiveItem] = React.useState<string>('');
+
+  const value = {
+    activeItemsUID,
+    setActiveItemsUID,
+    activeItem,
+    setActiveItem,
+  };
```
GowthamShanmugam, line 240: won't this create a new object after every render? due to this TopologySearchContext.Provider will get new params after every render right?

GowthamShanmugam:
I still have doubts about this PR, Whether the provider is a separate component does not matter, when the provider props change then It will re-render all of its children. It does not matter whether the child component is a consumer of the provider's property or not. 

All we can do is pass the parent state to the provider instead of passing the direct value to avoid the request re-rendering.

GowthamShanmugam:
> > I still have doubts about this PR, Whether the provider is a separate component does not matter, when the provider props change then It will re-render all of its children. It does not matter whether the child component is a consumer of the provider's property or not.
> > All we can do is pass the parent state to the provider instead of passing the direct value to avoid the request re-rendering.
> 
> `{children}` part of this snippet should be the one making the difference:
> 
> ```
> <TopologyDataContext.Provider value={value}>
>     {children}
> </TopologyDataContext.Provider>
> ```
> 
> ^ this is something which can be tested and cross-verified by updating the state continuously and adding a logger in any of the consumer component.

ack

GowthamShanmugam:
LGTM

packages/odf/components/topology/Topology.tsx
```diff
@@ -99,6 +101,151 @@ type SideBarProps = {
...
 };
 
+const Error = ({ error }) => <>{error}</>;
+
+const TopologyDataProvider: React.FC<{
+  children: React.ReactNode;
+  visualizationLevel: TopologyViewLevel;
+  activeNode: string;
+  setActiveNode: React.Dispatch<React.SetStateAction<string>>;
+}> = ({ children, visualizationLevel, activeNode, setActiveNode }) => {
```
bipuladh, line 111: Use types

bipuladh (commented):
@TimothyAsirJeyasing can we test this PR out and see if there is a difference in re-render amounts significantly?

SanjalKatiyar:
In case there is any confusion on why this change was needed (or why GH issue was created), plz check:
https://codesandbox.io/p/sandbox/without-children-h3kmm5
https://codesandbox.io/p/sandbox/with-children-qn3tcr

^ above was the idea for requesting this change, it's more of a good practice to added context as shown in the shared sandbix.

SanjalKatiyar:
In case there is any confusion on why this change was needed (or why GH issue was created), plz check:
https://codesandbox.io/p/sandbox/without-children-h3kmm5
https://codesandbox.io/p/sandbox/with-children-qn3tcr

^ above was the idea for requesting this change, it's more of a good practice to add context following the pattern shown in the shared sandbox.

alfonsomthd:
> In case there is any confusion on why this change was needed (or why GH issue was created), plz check: https://codesandbox.io/p/sandbox/without-children-h3kmm5 https://codesandbox.io/p/sandbox/with-children-qn3tcr
> 
> ^ above was the idea for requesting this change, it's more of a good-practice/correct-way to add context following the pattern shown in the shared sandbox.

@TimothyAsirJeyasing @SanjalKatiyar are we moving forward with this?

alfonsomthd:
> > @TimothyAsirJeyasing @SanjalKatiyar are we moving forward with this?
> 
> We need this yes. @TimothyAsirJeyasing PTAL.

@TimothyAsirJeyasing you need to rebase.

PR #1412, opened 2024-06-03
Fixes deletion operation on client list page

bipuladh:
[Bug 2283621](https://bugzilla.redhat.com/show_bug.cgi?id=2283621)
##### Before
Screenshot from 2024-06-03 18-42-23

#####  After
Screenshot from 2024-06-03 18-39-26

#### Changes

- Disables deletion icon until the heartbeat reaches 5 minutes
- Shows dashes when the client is disconnected as most of the information in our table is pulled from the status.

### DIFF
--- a/packages/odf/components/storage-consumers/client-list.tsx
+++ b/packages/odf/components/storage-consumers/client-list.tsx
@@ -25,7 +25,13 @@ import {
   useModal,
 } from '@openshift-console/dynamic-plugin-sdk';
 import { ModalComponent } from '@openshift-console/dynamic-plugin-sdk/lib/app/modal-support/ModalProvider';
-import { Button, Icon, Popover, PopoverPosition } from '@patternfly/react-core';
+import * as _ from 'lodash-es';
+import {
+  Button,
+  ButtonVariant,
+  Popover,
+  PopoverPosition,
+} from '@patternfly/react-core';
 import { TrashIcon } from '@patternfly/react-icons';
 import { sortable } from '@patternfly/react-table';
 import { StorageConsumerModel } from '../../models';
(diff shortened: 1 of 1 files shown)


packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -247,9 +249,10 @@ const StorageClientRow: React.FC<
         setAllowDeletion(false);
       }
     };
+    setter();
     const id = setInterval(setter, 10000);
     return () => clearInterval(id);
-  });
+  }, [setInterval, allowDeletion]);
```
alfonsomthd, line 255: @bipuladh Is this intentional?
`React Hook React.useEffect has a missing dependency: 'obj?.status?.lastHeartbeat'.`
If yes, can you suppress the warning?

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -268,8 +271,16 @@ const StorageClientRow: React.FC<
         <LastHeartBeat heartbeat={obj?.status?.lastHeartbeat} />
       </TableData>
       <TableData {...tableColumns[5]} activeColumnIDs={activeColumnIDs}>
-        <Icon disabled={allowDeletion} onClick={() => deleteClient(obj)}>
-          <TrashIcon />
+        <Icon
+          disabled={!allowDeletion}
+          onClick={() => (allowDeletion ? deleteClient(obj) : {})}
```
GowthamShanmugam, line 276: is on click work for disabled icon?

bipuladh: yes. that was the original bug.

GowthamShanmugam (commented):
LGTM

PR #1417, opened 2024-06-04
dashboard: enable capacity-trend card

weirdwiz:
this commit enables capacity-trend card in the storage system dashboard.
addition to that change the query to calculate daysUp with something in
the prometheus.


alfonsomthd:
@weirdwiz Can you add a screenshot?

bipuladh (approved):
/lgtm

PR #1419, opened 2024-06-04
Fixes Create LVS re-render issue

bipuladh:
[Bug 2284430](https://bugzilla.redhat.com/show_bug.cgi?id=2284430)
Memoizes CSV
Updates dependency array for initDiskRecovery parent useEffect hook

### DIFF
--- a/packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/create-local-volume-set-step.tsx
+++ b/packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/create-local-volume-set-step.tsx
@@ -22,6 +22,7 @@ import {
   createLocalVolumeDiscovery,
   updateLocalVolumeDiscovery,
 } from '@odf/core/utils';
+import { getNamespace } from '@odf/shared';
 import { ErrorAlert } from '@odf/shared/generic/Error';
 import { LoadingInline } from '@odf/shared/generic/Loading';
 import { useFetchCsv } from '@odf/shared/hooks/use-fetch-csv';
@@ -324,16 +325,20 @@ export const CreateLocalVolumeSet: React.FC<CreateLocalVolumeSetProps> = ({
     allNodes.current = createWizardNodeState(nonTaintedNodes);
   }, [rawNodes]);
 
+  const shouldStartDiscovery =
+    !csvLoadError && csvLoaded && allNodes.current.length;
+  const csvNamespace = getNamespace(csv);
+
   React.useEffect(() => {
-    if (!csvLoadError && csvLoaded && allNodes.current.length) {
+    if (shouldStartDiscovery) {
       initDiskDiscovery(
         allNodes.current,
-        csv?.metadata.namespace,
+        csvNamespace,
         setLvdError,
         setLvdInProgress
       );
     }
-  }, [csv, csvLoadError, csvLoaded, rawNodes]);
+  }, [csvNamespace, shouldStartDiscovery]);
 
   const discoveriesLoaded =
     csvLoaded &&


packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/create-local-volume-set-step.tsx
```diff
@@ -324,6 +325,8 @@ export const CreateLocalVolumeSet: React.FC<CreateLocalVolumeSetProps> = ({
     allNodes.current = createWizardNodeState(nonTaintedNodes);
   }, [rawNodes]);
 
+  const memoizedCSV = useDeepCompareMemoize(csv, true);
+
```
GowthamShanmugam, line 329: const canInitalizeDisk = !csvLoadError && csvLoaded && allNodes.current.length;
const namespace = getNamespace(csv);

Passing only "canInitalizeDisk" & "namespace" inside useEffect will reduce the risk.

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/create-local-volume-set-step.tsx
```diff
@@ -333,7 +336,12 @@ export const CreateLocalVolumeSet: React.FC<CreateLocalVolumeSetProps> = ({
         setLvdInProgress
       );
     }
-  }, [csv, csvLoadError, csvLoaded, rawNodes]);
+  }, [
```
SanjalKatiyar, line 339: we are passing `memoizedCSV`, but still using value of `csv` inside callback...

SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/1419/files#r1625863722 >> +1

SanjalKatiyar: also it's strange that this bug never happened before !!

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/create-local-volume-set-step.tsx
bipuladh: makes sense. With isolation of namespace we can remove the deep compare memoize as well.

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/create-local-volume-set-step.tsx
bipuladh: It might be happening all the time. You have to wait on that page for a while for it to happen. Prolly no one waited that long. I will test it with other releases and do the needful

packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/create-local-volume-set-step.tsx
```diff
@@ -324,16 +325,20 @@ export const CreateLocalVolumeSet: React.FC<CreateLocalVolumeSetProps> = ({
...
+    if (shouldStartDiscovery) {
       initDiskDiscovery(
         allNodes.current,
-        csv?.metadata.namespace,
+        csvNamespace,
         setLvdError,
         setLvdInProgress
       );
     }
-  }, [csv, csvLoadError, csvLoaded, rawNodes]);
```
SanjalKatiyar, line 336: need to add `rawNodes` back... it was originally added here so that if new nodes are added dynamically (we are keeping watch on nodes), this page should initiate discovery on that as well...
removing it from here means we are moving away from existing functionality of this page...

SanjalKatiyar: better to move https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/create-storage-system/create-storage-system-steps/create-local-volume-set-step/create-local-volume-set-step.tsx#L324-L325 inside this effect itself...

SanjalKatiyar: I just checked, looks like `rawNodes` itself is the one which was causing the issue... maybe we can add `allNodes.current?.length` to the dep list instead ??

SanjalKatiyar: it is happening because of `useNodesData` hook: https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/hooks/useNodesData.ts#L21, Prom response changes every few seconds causing `rawNodes` ref to change as well...
no issue in the hook though, hook's consumer needs to be resilient on its usage though...

PR #1452, opened 2024-06-25
Add support for MCG in client console

bipuladh:
Screenshot from 2024-06-25 12-45-11

### DIFF
--- a/packages/odf/redux/provider-hooks/useODFNamespace.ts
+++ b/packages/odf/redux/provider-hooks/useODFNamespace.ts
@@ -14,7 +14,10 @@ import {
 import { useODFNamespaceDispatch } from '../dispatchers';
 
 const FALLBACK_NAMESPACE = 'openshift-storage';
-const SPEC_NAME = 'odf-operator';
+const ODF_SUBSCRIPTION_NAME = 'odf-operator';
+const CLIENT_SUBSCRIPTION_NAME = 'ocs-client-operator';
+
+const getSpecName = (resource: SubscriptionKind) => resource?.spec?.name;
 
 export const FDF_FLAG = 'FDF_FLAG'; // Based on whether installed operator is ODF or FDF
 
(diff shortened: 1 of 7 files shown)


plugins/client/console-extensions.json
```diff
@@ -32,6 +32,182 @@
...
+  {
+    "type": "console.flag/hookProvider",
+    "properties": {
+      "handler": {
+        "$codeRef": "features.useODFSystemFlags"
+      }
+    },
+    "flags": {
+      "disallowed": ["PROVIDER_MODE"]
+    }
```
SanjalKatiyar, line 95: It seems very odd to add these ODF specific hooks to client console, but I can understand we don't have a way to tell if it's client-only cluster or not.

SanjalKatiyar: Did u test this on a client-only cluster though (no ODF) ?? cause these hooks will error out in that case, I hope it won't affect NooBaa pages in that case.

plugins/client/console-extensions.json
```diff
@@ -32,6 +32,182 @@
...
+  },
+  {
+    "type": "console.flag/hookProvider",
+    "properties": {
+      "handler": {
+        "$codeRef": "features.useODFSystemFlags"
```
SanjalKatiyar, line 90: do we need this ?? I don't see it being used anywhere in NooBaa resources pages... only used in NooBaa dashboards...

plugins/client/console-extensions.json
bipuladh: I have updated the hook to fallback to client operator namespace

packages/odf/redux/provider-hooks/useODFNamespace.ts
```diff
@@ -25,11 +26,18 @@ const namespaceDetector = async (maxAttempt = 5): Promise<string> => {
         model: SubscriptionModel,
         queryParams: { ns: null },
       })) as SubscriptionKind[];
-      ns = getNamespace(
-        subscriptions.find((subscription) =>
-          subscription.spec.name.includes(SPEC_NAME)
-        )
+      const odfSubscription = subscriptions.find((subscription) =>
+        getName(subscription).includes(ODF_SUBSCRIPTION_NAME)
```
SanjalKatiyar, line 30: isn't it better to keep relying on `subscription.spec.name.includes(SPEC_NAME)` as we have been doing so for a long time (`useFetchCsv` hook) and it always worked...

though `getName(subscription).includes(ODF_SUBSCRIPTION_NAME)` should work as well, but better to be safe than updating the check ??

packages/odf/redux/provider-hooks/useODFNamespace.ts
```diff
@@ -25,11 +26,18 @@ const namespaceDetector = async (maxAttempt = 5): Promise<string> => {
...
       );
+      const isODFPresent = odfSubscription !== undefined;
+      if (isODFPresent) {
+        ns = getNamespace(odfSubscription);
+      } else {
```
SanjalKatiyar, line 35: should we add it ?? just for safety... this will ensure mostly all of existing functionality of this hook will remain intact..
```suggestion
      } else if (!isProviderMode) {
```

packages/odf/redux/provider-hooks/useODFNamespace.ts
bipuladh: sure changing the logic

packages/odf/redux/provider-hooks/useODFNamespace.ts
SanjalKatiyar: my bad... my intension was to not run client fallback **at all** if it's normal ODF mode deployment (to make it how existing flow works)... but I guess we can't determine that !!

PR #1457, opened 2024-07-01
Hide 'Provider Mode' deployment for ODF

SanjalKatiyar:
https://bugzilla.redhat.com/show_bug.cgi?id=2294383

Screenshot 2024-07-01 at 12 23 22 PM

### DIFF
--- a/packages/odf/redux/provider-hooks/useODFNamespace.ts
+++ b/packages/odf/redux/provider-hooks/useODFNamespace.ts
@@ -1,68 +1,92 @@
 import * as React from 'react';
-import { SubscriptionModel } from '@odf/shared/models';
+import {
+  SubscriptionModel,
+  ClusterServiceVersionModel,
+} from '@odf/shared/models';
 import { getNamespace } from '@odf/shared/selectors';
-import { SubscriptionKind } from '@odf/shared/types';
+import { SubscriptionKind, ClusterServiceVersionKind } from '@odf/shared/types';
 import { isAbortError } from '@odf/shared/utils';
-import { k8sList } from '@openshift-console/dynamic-plugin-sdk';
+import {
+  k8sList,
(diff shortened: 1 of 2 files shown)


bipuladh (approved):
/lgtm

packages/odf/redux/provider-hooks/useODFNamespace.ts
```diff
@@ -1,68 +1,92 @@
...
       if (!ns) throw new Error('ODF install namespace not found');
+
+      // ToDo: Remove in z-stream (https://bugzilla.redhat.com/show_bug.cgi?id=2294383)
+      // eslint-disable-next-line no-await-in-loop
+      const csv: ClusterServiceVersionKind = await k8sGet({
+        model: ClusterServiceVersionModel,
+        name: subscription?.status?.installedCSV,
+        ns,
+      });
+      isFDF = !['redhat', 'red hat'].includes(
```
alfonsomthd, line 53: @SanjalKatiyar Just for safety, Should we check for substrings (and not only exact match) ? Just in case that text is extended in future releases...

SanjalKatiyar: we should be fine IMO, it has always be constant name: https://github.com/red-hat-storage/odf-operator/blob/main/config/manifests/bases/odf-operator.clusterserviceversion.yaml#L82

also, we are going to remove it in next z-stream anyway...

SanjalKatiyar:
CI passed already once: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/red-hat-storage_odf-console/1457/pull-ci-red-hat-storage-odf-console-master-odf-console-e2e-aws/1807701173679951872, blocking the re-testing as no major changes occurred after that...

PR #1475, opened 2024-07-18
Bug 2297295: Hide builtin-mgr CephBlockPool

alfonsomthd:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2297295

- Storage pool list
Before:
pool-list-before

After:
pool-list-after

- StorageClass creation Form:
Before:
sc-form-before

After:
sc-form-after

### DIFF
--- a/packages/odf/resources.ts
+++ b/packages/odf/resources.ts
@@ -57,13 +57,16 @@ export const subscriptionResource: WatchK8sResource = {
   kind: referenceForModel(SubscriptionModel),
   namespaced: false,
 };
-
-export const cephBlockPoolResource: K8sResourceObj = (ns) => ({
+/**
+ * Retrieve all the CRs except those not meant to be
+ * exposed to the users. See:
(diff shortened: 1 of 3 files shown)


packages/odf/resources.ts
```diff
@@ -58,12 +58,11 @@ export const subscriptionResource: WatchK8sResource = {
...
 };
 
-export const cephBlockPoolResource: K8sResourceObj = (ns) => ({
+export const cephBlockPoolResource: WatchK8sResource = {
   kind: referenceForModel(CephBlockPoolModel),
-  namespaced: true,
   isList: true,
-  namespace: ns,
-});
+  fieldSelector: 'metadata.name!=builtin-mgr',
```
SanjalKatiyar, line 64: can u plz add a comment here explaining why we are filtering this name ??

alfonsomthd: Done.

PR #1476, opened 2024-07-18
Annotate RBD storageclasses for keyrotation

black-dragon74:
https://issues.redhat.com/browse/RHSTOR-5882

### DIFF
--- a/packages/ocs/storage-class/mutators.ts
+++ b/packages/ocs/storage-class/mutators.ts
@@ -44,8 +44,25 @@ export const addReclaimSpaceAnnotation = (sc: StorageClass): StorageClass => {
   return sc;
 };
 
+export const addEncryptionKeyRotationAnnotation = (
+  sc: StorageClass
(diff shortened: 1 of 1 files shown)


SanjalKatiyar:
@black-dragon74 why do we need it now, when it was confirmed earlier: https://issues.redhat.com/browse/RHSTOR-3922?focusedId=24758628&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-24758628 that we don't wanna do it ??

black-dragon74:
@SanjalKatiyar that discussion pertained to adding new UI elements wrt key rotation. We do have to add the annotation in cases where sc is created from the UI. The labels we add in turn trigger the automation. 

So no literal UI (element) changes, but this change. I get the similarity between the terms that lead to this ambiguity. 

Regards

SanjalKatiyar (approved):

PR #1482, opened 2024-07-22
Cleanup: remove duplicated resource models

alfonsomthd:
- Keep `CephClusterModel` & `OCSStorageCLusterModel` in `shared` library as the library is being used by other projects.

### DIFF
--- a/packages/ocs/hooks/useOcsHealth.ts
+++ b/packages/ocs/hooks/useOcsHealth.ts
@@ -3,6 +3,7 @@ import {
   useCustomPrometheusPoll,
   usePrometheusBasePath,
 } from '@odf/shared/hooks/custom-prometheus-poll';
+import { CephClusterModel } from '@odf/shared/models';
 import { getName, getNamespace } from '@odf/shared/selectors';
 import { K8sResourceKind, StorageSystemKind } from '@odf/shared/types';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
@@ -13,11 +14,7 @@ import {
   WatchK8sResource,
 } from '@openshift-console/dynamic-plugin-sdk';
 import * as _ from 'lodash-es';
-import {
-  CephClusterModel,
-  CephObjectStoreModel,
-  NooBaaSystemModel,
-} from '../models/core';
+import { CephObjectStoreModel, NooBaaSystemModel } from '../models/core';
 import { Health, HEALTH_QUERY } from '../queries';
 import {
   getCephHealthState,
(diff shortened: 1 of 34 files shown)


TimothyAsirJeyasing:
Hope we may need to update the test case also ...

packages/ocs/hooks/useOcsHealth.ts
```diff
@@ -13,11 +14,7 @@ import {
   WatchK8sResource,
 } from '@openshift-console/dynamic-plugin-sdk';
 import * as _ from 'lodash-es';
-import {
-  CephClusterModel,
-  CephObjectStoreModel,
-  NooBaaSystemModel,
-} from '../models/core';
+import { CephObjectStoreModel, NooBaaSystemModel } from '../models/core';
```
GowthamShanmugam, line 17: is that ok to keep some OCS and ODF models under non-shared repo? In future other plugins also may use these types right? What about moving all ODF/OCS related types to shared?

packages/ocs/storage-pool/body.tsx
```diff
@@ -118,7 +118,11 @@ export const StoragePoolBody: React.FC<StoragePoolBodyProps> = ({
   const poolNs = getNamespace(cephCluster);
 
   const [storageCluster, storageClusterLoaded, storageClusterLoadError] =
-    useK8sGet<ListKind<StorageClusterKind>>(StorageClusterModel, null, poolNs);
+    useK8sGet<ListKind<StorageClusterKind>>(
```
GowthamShanmugam, line 121:
```suggestion
    useK8sGet<ListKind<OCSStorageClusterKind>>(
```

Optional comment

packages/ocs/hooks/useOcsHealth.ts
alfonsomthd: It's all about team consensus: if we agree on moving ODF/OCS types & models to `shared` then we can create a technical issue to address that in another PR: here I'm only removing duplicates that add confusion when deciding where to import from.

packages/ocs/storage-pool/body.tsx
alfonsomthd: @GowthamShanmugam Although I understand this suggestion for naming consistency, it'll have side effects (breaking changes) in other projects (when there is an upgrade of the `shared` library version).
Waiting for @SanjalKatiyar feedback in order to procced or not with this suggestion.

SanjalKatiyar: we don't use `OCSStorageClusterModel` or `StorageClusterKind` in Fusion, only place it's used in this repo...

we can either remove `OCS` part from `OCSStorageClusterModel`, or add `OCS` to `StorageClusterKind`... either is alright...

packages/ocs/hooks/useOcsHealth.ts
GowthamShanmugam: ack, it could be a separate topic to discuss

packages/ocs/storage-pool/body.tsx
alfonsomthd: @GowthamShanmugam @SanjalKatiyar then I'll remove `OCS` prefix so the name it's more aligned with the CR (`StorageCluster`).

PR #1483, opened 2024-07-22
tsconfig: set '@odf' path mapping correctly

alfonsomthd:
Imports should start with `'@odf`  instead of `'packages/...`.

Before (VSCode intellisense suggestions):
Screenshot from 2024-07-22 13-45-42

After:
Screenshot from 2024-07-22 13-46-15


alfonsomthd:
@bipuladh @SanjalKatiyar ping.

SanjalKatiyar (approved):

PR #1487, opened 2024-07-23
Fix Tooltip Rendering Issue by Upgrading react-charts

TimothyAsirJeyasing:
This PR addresses the tooltip rendering issues
encountered in react-charts@7.1.0.
The issue has been resolved in react-charts@7.3.0.

https://bugzilla.redhat.com/show_bug.cgi?id=2281587

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "@openshift-console/dynamic-plugin-sdk-webpack": "1.1.0",
     "@openshift-console/plugin-shared": "^0.0.1",
     "@patternfly/patternfly": "5.0.2",
-    "@patternfly/react-charts": "7.1.0",
+    "@patternfly/react-charts": "7.3.0",
     "@patternfly/react-core": "5.1.0",
     "@patternfly/react-icons": "5.0.1",
     "@patternfly/react-table": "5.1.0",
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -21,7 +21,7 @@
   },
   "dependencies": {
     "@patternfly/patternfly": "5.0.2",
-    "@patternfly/react-charts": "7.1.0",
+    "@patternfly/react-charts": "7.3.0",
     "@patternfly/react-core": "5.1.0",
     "@patternfly/react-icons": "5.0.1",
     "@patternfly/react-table": "5.1.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1076,7 +1076,7 @@
   version "0.0.0-beta15"
   dependencies:
     "@patternfly/patternfly" "5.0.2"
-    "@patternfly/react-charts" "7.1.0"
+    "@patternfly/react-charts" "7.3.0"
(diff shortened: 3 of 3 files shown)


GowthamShanmugam:
Please check the build failure

SanjalKatiyar:
> update `shared` dependencies as well...

this is still missing...

SanjalKatiyar:
why do we need changes in analyse file ??

TimothyAsirJeyasing:
> why do we need changes in analyse file


We updated the analyzeTest.ts file to fix two key issues: First, after upgrading react-charts, the script needed adjustments to accurately analyze the new build artifacts and avoid misinterpretations during the build process. Second, it was encountering a memory limit error ("Cannot create a string longer than 0x1fffffe8 characters") when processing large files, which we've optimized.

SanjalKatiyar:
> We updated the analyzeTest.ts file to fix two key issues: First, after upgrading react-charts, the script needed adjustments to accurately analyze the new build artifacts and avoid misinterpretations during the build process. Second, it was encountering a memory limit error ("Cannot create a string longer than 0x1fffffe8 characters") when processing large files, which we've optimized.

You have not "optimized" anything, rather just increases the base size to make sure there are no errors while running the script. I don't think all the changes in the analyse file is necessary, plz check again.
https://github.com/red-hat-storage/odf-console/pull/1520 should do the required optimisation.

SanjalKatiyar:
fix ur commits in this PR... also check why tests are failing...

package.json
```diff
@@ -89,6 +89,7 @@
     "js-base64": "^2.1.9",
     "js-yaml": "^3.13.1",
     "lodash-es": "^4.17.21",
+    "mocha": "^10.7.3",
```
alfonsomthd, line 92: @TimothyAsirJeyasing Is this dependency required?

alfonsomthd:
@TimothyAsirJeyasing After some investigation about the E2E errors I found that the chart was crashing (at least when you create a storage system and there is no total/available/used capacity yet).
The cause was the tooltip that depends on `victory-tooltip` dep.
Adding this under the `resolutions` section of `package.json` solved the issue in my environment:
`"@patternfly/react-charts/victory-tooltip": "^37.1.1"`
Probably you'll need to also add this in the shared library.
Please test that all charts affected by this PR work as expected.
@SanjalKatiyar FYI.

SanjalKatiyar:
If version upgrade is causing multiple issues, can't we try to fix tooltip based on some CSS property ?? (like z-index or something similar)

TimothyAsirJeyasing:
The react-charts version updated in https://github.com/red-hat-storage/odf-console/pull/1650

alfonsomthd:
> The react-charts version updated in #1650

@SanjalKatiyar this PR can be closed once @TimothyAsirJeyasing confirms that the issue is fixed in master.
The bug should point at https://github.com/red-hat-storage/odf-console/pull/1650 as its fix.

PR #1489, opened 2024-07-24
Provider Mode: Object Dashboard: fix RGW performance card

alfonsomthd:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2276393

- Also fixed: occasional crashes related to wrong date values in metric data points.

BEFORE: page crash when selecting `Object Gateway (RGW)` on Performance Card (Object Dashboard)
AFTER:
Screenshot from 2024-07-24 12-37-38
image

### DIFF
--- a/packages/ocs/dashboards/object-service/data-consumption-card/data-consumption-card.tsx
+++ b/packages/ocs/dashboards/object-service/data-consumption-card/data-consumption-card.tsx
@@ -219,7 +219,7 @@ const ServiceTypeRGW: React.FC<ServiceTypeProps> = ({ queries, metric }) => {
   const loading = getLoading || putLoading;
   const error = !!getError || !!putError;
   const data = !!get && !!put;
-  const response: Response = React.useMemo(() => {
+  const response: DataPoint<Date>[][] = React.useMemo(() => {
     return !loading && !error && data
       ? [...getRangeVectorStats(get), ...getRangeVectorStats(put)]
       : [];
@@ -229,7 +229,7 @@ const ServiceTypeRGW: React.FC<ServiceTypeProps> = ({ queries, metric }) => {
     <PerformanceGraph
       loading={loading}
       loadError={error}
-      dataPoints={response as DataPoint[][][]}
+      dataPoints={response}
       metricType={metric}
     />
   );
--- a/packages/shared/src/details-page/datetime.ts
+++ b/packages/shared/src/details-page/datetime.ts
@@ -220,6 +220,9 @@ export const twentyFourHourTime = (
   date: Date,
   showSeconds?: boolean
 ): string => {
+  if (!_.isDate(date)) {
+    date = new Date(0);
+  }
   const hours = zeroPad(date.getHours() ?? 0);
   const minutes = `:${zeroPad(date.getMinutes() ?? 0)}`;
   const seconds = showSeconds ? `:${zeroPad(date.getSeconds() ?? 0)}` : '';
--- a/packages/ocs/dashboards/object-service/data-consumption-card/performance-graph.tsx
+++ b/packages/ocs/dashboards/object-service/data-consumption-card/performance-graph.tsx
@@ -22,7 +22,7 @@ import { convertNaNToNull, getLatestValue } from '../../../utils';
 import './data-consumption-card.scss';
 
 type PerformanceGraphProps = {
-  dataPoints: DataPoint[][][];
+  dataPoints: DataPoint[][];
   loading: boolean;
   loadError: boolean;
   metricType: string;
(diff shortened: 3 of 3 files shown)


packages/shared/src/details-page/datetime.ts
```diff
@@ -217,12 +217,13 @@ export const toISODateString = (date: Date): string =>
...
 
 export const twentyFourHourTime = (
-  date: Date,
+  dateString: string,
   showSeconds?: boolean
 ): string => {
-  const hours = zeroPad(date.getHours() ?? 0);
-  const minutes = `:${zeroPad(date.getMinutes() ?? 0)}`;
-  const seconds = showSeconds ? `:${zeroPad(date.getSeconds() ?? 0)}` : '';
+  const date = dateString ? new Date(dateString) : null;
```
SanjalKatiyar, line 223: This is not used by Fusion right now, so maybe safe to update.

But, this is a generic util... ideally we should make exceptions and update things under `shared` like this very rarely... and make it a practice not to do it under general cases.
```suggestion
  const date = typeof date === 'string' ? new Date(dateString) : date || null;
```

packages/shared/src/details-page/datetime.ts
```diff
@@ -217,12 +217,13 @@ export const toISODateString = (date: Date): string =>
   )}`;
 
 export const twentyFourHourTime = (
-  date: Date,
+  dateString: string,
```
SanjalKatiyar, line 220:
```suggestion
  date: Date | string,
```

packages/shared/src/details-page/datetime.ts
SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/1489/files#r1689748777

alfonsomthd: Currently this util is used only in this component so it's safe to update.

SanjalKatiyar: it is still exported from the file, which means anyone using the `odf-shared` library can import it...
of course, we can always check Fusion repo and verify it's safe or not (currently Fusion is not using it), but it's matter of practice that only make breaking changes in `odf-shared` when really needed or in case of rare exceptions (which we can always discuss offline and decide).

One case of "rare exceptions" IMO was latest `OCSStorageClusterModel` to `StorageClusterModel` change, StorageCluster is ODF specific entity, so it's highly unlikely any other repo will have need to import it.

alfonsomthd: @SanjalKatiyar I detected the page was crashing  sometimes (when passed a non-date value), so it's justified.

alfonsomthd: After debugging a bit more, I found that actually this method is receiving either date object (that was being coerced to string when printing) or sometimes a number (thus crashing the page).
So I simplified the fix using `lodash`. PTAL.

SanjalKatiyar: the alternative I suggested should work well too (it's not a hack)...

but I am okay if we don't want to update it or if there are any concerns with the suggested change...

SanjalKatiyar: ^ my above comment was reply to https://github.com/red-hat-storage/odf-console/pull/1489#discussion_r1689825929

packages/shared/src/details-page/datetime.ts
alfonsomthd: Refactored according to [this comment](https://github.com/red-hat-storage/odf-console/pull/1489#discussion_r1689832018).

packages/ocs/dashboards/object-service/data-consumption-card/data-consumption-card.tsx
```diff
@@ -219,7 +219,7 @@ const ServiceTypeRGW: React.FC<ServiceTypeProps> = ({ queries, metric }) => {
   const loading = getLoading || putLoading;
   const error = !!getError || !!putError;
   const data = !!get && !!put;
-  const response: Response = React.useMemo(() => {
+  const response: DataPoint<Date>[][] = React.useMemo(() => {
```
bipuladh, line 222: We need to test this with multiple versions before backporting for now this is working but the reason this chart is failing is because the data structure suddenly changed in the metrics.

bipuladh (approved):

packages/ocs/dashboards/object-service/data-consumption-card/data-consumption-card.tsx
alfonsomthd: @bipuladh This has been tested with a 4.16 BM cluster (the same version as in the reported issue)  so we're safe backporting to 4.16.
Do we need to backport this to 4.15?

PR #1490, opened 2024-07-24
Optimization1: Lazy loading of the main pages under the navigation - WIP

GowthamShanmugam:
Issue: https://issues.redhat.com/browse/RHSTOR-6135
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2290675

1) Importing each main page under the navigation  (Data Services -> Disaster recovery) at runtime. It will reduce the chunk size and optimize the page loading time.

2) Dashboard enabled only when ACM observability is enabled, We don't have to keep the dashboard-related imports on chunk all the time.

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -20,7 +20,7 @@ const NODE_ENV = (process.env.NODE_ENV ||
   'development') as webpack.Configuration['mode'];
 const PLUGIN = process.env.PLUGIN;
 const OPENSHIFT_CI = process.env.OPENSHIFT_CI;
-const isProduction = NODE_ENV === 'production';
+const IS_PRODUCTION = NODE_ENV === 'production';
 
 if (PLUGIN === undefined) {
   process.exit(1);
@@ -87,7 +87,7 @@ const config: webpack.Configuration & DevServerConfiguration = {
           {
             loader: 'thread-loader',
             options: {
-              ...(!isProduction
+              ...(!IS_PRODUCTION
                 ? { poolTimeout: Infinity, poolRespawn: false }
                 : OPENSHIFT_CI
                 ? {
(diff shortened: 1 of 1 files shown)


SanjalKatiyar:
better to wrap `<HorizontalNav pages={pages} />` under `Suspense` for safety...

packages/mco/components/disaster-recovery-page/disaster-recovery.tsx
```diff
@@ -28,12 +26,16 @@ const DisasterRecovery: React.FC = () => {
...
+          () => import('../drpolicy-list-page/drpolicy-list-page')
+        ),
       },
       {
         href: 'protected-applications',
         name: t('Protected applications'),
-        component: ProtectedApplicationsListPage,
+        component: React.lazy(
+          () => import('../protected-applications/list-page')
+        ),
```
SanjalKatiyar, line 38: IIRC we have disabled caching in nginx... does it means everytime we change a tab `Overview -> Policies -> Overview -> Policies`, chuck will get fetched again and again ??

SanjalKatiyar: what's the behaviour ??

GowthamShanmugam: I am afraid it is not just with lazy loading if the cache is disabled means it will fetch chunks for static import also, imagine how big chink it is trying to download every time.

Anyway I will test this behaviour and confirm.

SanjalKatiyar: I agree... that will happen when user navigate to `Data services`... but within tabs, it's a one big chuck, so no additional fetching for subsequent tabs or while navigating between different tabs...

fix here will be to investigate and enable the caching again.

PR #1375, opened 2024-05-10
Updated the clean up application resource message

TimothyAsirJeyasing:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2280308
   BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2277125
   BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2294234

PR #1498, opened 2024-07-26
Update Discovered App Status Panel for Failover Instructions

TimothyAsirJeyasing:
When Failover is initiated and the DRPC progression is stuck at "WaitForReadiness" instead of "WaitOnUserToCleanUp," the Discovered App Status panel now instructs users to delete the application on the cluster they are
failing over from. This update provides clearer guidance during the failover process.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2294234

### DIFF
--- a/packages/mco/components/protected-applications/utils.tsx
+++ b/packages/mco/components/protected-applications/utils.tsx
@@ -28,11 +28,7 @@ import {
 } from '../../constants';
 import { DRPlacementControlModel } from '../../models';
 import { DRPlacementControlKind } from '../../types';
-import {
-  getVolumeReplicationHealth,
-  isDRPCAvailable,
-  isPeerReady,
-} from '../../utils';
+import { getVolumeReplicationHealth } from '../../utils';
 import { DiscoveredApplicationParser as DiscoveredApplicationModal } from '../modals/app-failover-relocate/parser/discovered-application-parser';
 import RemoveDisasterRecoveryModal from '../modals/remove-disaster-recovery/remove-disaster-recovery';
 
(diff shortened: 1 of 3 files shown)


TimothyAsirJeyasing:
Screenshot from 2024-07-26 17-23-29

packages/mco/components/protected-applications/utils.tsx
```diff
@@ -30,8 +30,8 @@ import { DRPlacementControlModel } from '../../models';
 import { DRPlacementControlKind } from '../../types';
 import {
   getVolumeReplicationHealth,
-  isDRPCAvailable,
   isPeerReady,
+  isDRPCAvailable,
```
alfonsomthd, line 34: This change seems unnecessary...

packages/mco/utils/disaster-recovery.tsx
```diff
@@ -270,10 +270,8 @@ export const isPeerReady = (drpc: DRPlacementControlKind) =>
   );
 
 export const isDRPCAvailable = (drpc: DRPlacementControlKind) =>
-  !!drpc?.status?.conditions?.some(
-    (condition) =>
-      condition?.type === 'Available' && condition?.status === 'True'
-  );
+  drpc?.status?.progression === 'WaitOnUserToCleanUp' &&
+  !!drpc?.status?.conditions?.some((condition) => condition?.status === 'True');
```
alfonsomthd, line 274: Is `condition?.type === 'Available'` no longer necessary? Provided the method name contains the word `Available`, maybe we should rename the method to better reflect what we're checking here (unless we agree is still correct).

TimothyAsirJeyasing:
@GowthamShanmugam @SanjalKatiyar Please review

packages/mco/utils/disaster-recovery.tsx
TimothyAsirJeyasing: yes its not necessary, i will revert the changes and update isCleanupPending method.

TimothyAsirJeyasing:
I am closing this PR as its already incorporated with https://github.com/red-hat-storage/odf-console/pull/1375

PR #1519, opened 2024-08-06
Fix nodes' label for ROSA clusters

SanjalKatiyar:
https://bugzilla.redhat.com/show_bug.cgi?id=2302235

### DIFF
--- a/packages/odf/modals/add-capacity/add-capacity-modal.tsx
+++ b/packages/odf/modals/add-capacity/add-capacity-modal.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import NamespaceSafetyBox from '@odf/core/components/utils/safety-box';
 import { useNodesData } from '@odf/core/hooks';
-import { useODFNamespaceSelector } from '@odf/core/redux';
 import { NodeData } from '@odf/core/types';
 import { getStorageClassDescription } from '@odf/core/utils';
 import { getCephNodes } from '@odf/ocs/utils/common';
@@ -17,7 +16,7 @@ import { CommonModalProps } from '@odf/shared/modals/common';
 import { ModalBody, ModalFooter, ModalHeader } from '@odf/shared/modals/Modal';
 import { StorageClusterModel } from '@odf/shared/models';
 import { PersistentVolumeModel, StorageClassModel } from '@odf/shared/models';
-import { getName } from '@odf/shared/selectors';
+import { getName, getNamespace } from '@odf/shared/selectors';
 import {
   StorageClassResourceKind,
   StorageClusterKind,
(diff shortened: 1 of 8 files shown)


packages/shared/src/constants/common.ts
```diff
@@ -1,6 +1,6 @@
 export const DASH = '-';
 export const AVAILABLE = 'Available';
-export const CEPH_STORAGE_NAMESPACE = 'openshift-storage';
+export const DEFAULT_STORAGE_NAMESPACE = 'openshift-storage'; // This hardcoded namespace is only for specific cases, do not use it otherwise.
```
bipuladh, line 3: This is going to cause confusion just use this value directly in that function. Else people are going to use it in multiple places.

SanjalKatiyar: we anyway need it here: https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/alerts/alert-action-path.tsx#L24, so I can't remove this variable entirely... and using directly in function would mean to repeating `openshift-storage` at multiple places (had it been only single or twice instances, I would have opted for direct value)...

SanjalKatiyar: now we are only keeping `openshift-storage` at a single place...

packages/odf/modals/add-capacity/add-capacity-modal.tsx
```diff
@@ -251,7 +250,7 @@ export const AddCapacityModal: React.FC<AddCapacityModalProps> = ({
   const replica = getDeviceSetReplica(
     isArbiterEnabled,
     hasFlexibleScaling,
-    createWizardNodeState(getCephNodes(nodesData, odfNamespace) as NodeData[])
+    createWizardNodeState(getCephNodes(nodesData, ocsClusterNs) as NodeData[])
```
bipuladh, line 253: Just use the variable directly here. Instead of relying on `ocsClusterNs`

SanjalKatiyar: `getCephNodes` will work even if I pass empty string here, I passed `ocsClusterNs` to keep it aligned with dashboards and other places.

Originally (in 4.15, when label was dynamic) I should have passed cluster namespace here (i.e., ocsClusterNs), but I missed earlier.

Anyway, `getCephNodes`, `cephStorageLabel` & `labelNodes` will internally use hardcoded `openshift-storage`... and when we support multiple internal clusters we can just update these functions and all other places (where these functions are called will automatically be fixed).
I have added related comments on top of these functions as well.

SanjalKatiyar: Idea is to not update the code where these functions are called (as much as possible), rather update/hack the functions themselves.

Other way is: I go to each file and pass `DEFAULT_STORAGE_NAMESPACE` to the function and do not update functions' code itself... but that probably will be more number of changes than the current fix and will need more updates later when/if we support multiple internal clusters.





[4:13](https://ibm-systems-storage.slack.com/archives/D04SUE8L3HS/p1722940995343479)
I can do this other way as well, if that make more sense... but probably that will be slightly more number of changes than current fi

SanjalKatiyar: Let me know which one you prefer.

bipuladh (approved):
/lgtm

PR #1520, opened 2024-08-06
Optimizing the chunk size using devtool

GowthamShanmugam:
Issues: https://issues.redhat.com/browse/RHSTOR-6134
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2290675

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -20,6 +20,7 @@ const NODE_ENV = (process.env.NODE_ENV ||
   'development') as webpack.Configuration['mode'];
 const PLUGIN = process.env.PLUGIN;
 const OPENSHIFT_CI = process.env.OPENSHIFT_CI;
+const isProduction = NODE_ENV === 'production';
 
 if (PLUGIN === undefined) {
   process.exit(1);
@@ -86,7 +87,7 @@ const config: webpack.Configuration & DevServerConfiguration = {
           {
             loader: 'thread-loader',
             options: {
-              ...(NODE_ENV === 'development'
+              ...(!isProduction
                 ? { poolTimeout: Infinity, poolRespawn: false }
                 : OPENSHIFT_CI
                 ? {
@@ -153,13 +154,15 @@ const config: webpack.Configuration & DevServerConfiguration = {
       cwd: process.cwd(),
     }),
   ],
-  devtool: 'eval-cheap-module-source-map',
+  // 'source-map' is recommended choice for production builds, A full SourceMap is emitted as a separate file.
+  // 'eval-source-map' is recommended for development but 'eval-cheap-module-source-map' is faster and gives better result.
+  devtool: isProduction ? 'source-map' : 'eval-cheap-module-source-map',
   optimization: {
     chunkIds: 'named',
   },
 };
 
-if (NODE_ENV === 'production' || process.env.DEV_NO_TYPE_CHECK !== 'true') {
+if (isProduction || process.env.DEV_NO_TYPE_CHECK !== 'true') {
   config.plugins?.push(
     new ForkTsCheckerWebpackPlugin({
       issue: {


webpack.config.ts
```diff
@@ -153,13 +144,34 @@ const config: webpack.Configuration & DevServerConfiguration = {
...
+  // 'source-map' is recommended choice for production builds, A full SourceMap is emitted as a separate file.
+  // 'eval-source-map' is recommended for development but 'eval-cheap-module-source-map' is faster and gives better result.
+  devtool: isProduction ? 'source-map' : 'eval-cheap-module-source-map',
   optimization: {
-    chunkIds: 'named',
+    // 'deterministic' Good for long term caching.
+    // 'named' Readable ids for better debugging.
+    chunkIds: isProduction ? 'deterministic' : 'named',
+  },
+  devServer: {
```
GowthamShanmugam, line 155: Moved devServer at the bottom for better readability

package.json
```diff
@@ -9,10 +9,6 @@
   ],
   "repository": "git@github.com:red-hat-storage/odf-console.git",
   "scripts": {
-    "clean": "yarn clean-mco && yarn clean-odf",
-    "clean-mco": "cd plugins/mco && rm -rf ./dist",
-    "clean-odf": "cd plugins/odf && rm -rf ./dist",
-    "clean-client": "cd plugins/client && rm -rf ./dist",
```
SanjalKatiyar, line 15: why we removing this ?? We use them on local.

GowthamShanmugam: webpack config: clean: true, will takecare of this.

webpack.config.ts
```diff
@@ -9,53 +9,44 @@ import * as webpack from 'webpack';
...
   mode: NODE_ENV,
   entry: {},
   output: {
     path: path.resolve('./dist'),
-    filename: '[name]-bundle.js',
-    chunkFilename: '[name]-chunk.js',
+    filename: '[name].bundle.[contenthash:8].js',
+    chunkFilename: '[name].[contenthash:8].js',
+    // Clean the output directory(dist) before emit
+    clean: true,
```
SanjalKatiyar, line 44: this might be not enough to enable caching... manifest file (plugin metadata) has a constant name which can get cached, IIRC console rely on `version` field to prevent that.
We need to confirm with OCP team though once, whether that fix will work or not, then update our version accordingly for main and compatibility builds (if not already).

webpack.config.ts
```diff
@@ -153,13 +144,34 @@ const config: webpack.Configuration & DevServerConfiguration = {
       cwd: process.cwd(),
     }),
   ],
-  devtool: 'eval-cheap-module-source-map',
+  // 'source-map' is recommended choice for production builds, A full SourceMap is emitted as a separate file.
```
SanjalKatiyar, line 147: we don't need source for production, it will unnecessary increase the bundle size...

SanjalKatiyar: maybe I am missing something here ??

webpack.config.ts
GowthamShanmugam: ack

webpack.config.ts
GowthamShanmugam: I thought the same but one corner case here is, in case of any bug in production we don't get any stack trace for debugging the issue.

webpack.config.ts
GowthamShanmugam: Validation on the odf-console side:
  After the base path is changed the version is updated on the manifest json,
  For the main branch: "version": "4.17.0",
  For the compatibility branch: "version": "4.17.0-compatibility",
  
  So it's a good sign for further testing.

package.json
GowthamShanmugam: The clean config is not working always, I am also facing same issue:https://github.com/webpack/webpack-dev-middleware/issues/861 

i will revert this change

SanjalKatiyar:
@GowthamShanmugam if we are planning to backport this PR to 4.16.z can we remove hashing and other unrelated changes from this PR and add them to https://github.com/red-hat-storage/odf-console/pull/1525 instead ?? Let's keep this PR short with only what's needed (`devTool` part), wdyt ??

In case this is 4.17 onwards fix then we are good.

GowthamShanmugam:
> @GowthamShanmugam if we are planning to backport this PR to 4.16.z can we remove hashing and other unrelated changes from this PR and add them to #1525 instead ?? Let's keep this PR short with only what's needed (`devTool` part), wdyt ??
> 
> In case this is 4.17 onwards fix then we are good.

ack, that's a good suggestion

PR #1525, opened 2024-08-06
Openshift-SDK and PF dependency version upgrade

GowthamShanmugam:
On top of: https://github.com/red-hat-storage/odf-console/pull/1520
Console PR ref: https://github.com/openshift/console/pull/13992
Slack Link: https://redhat-internal.slack.com/archives/C011BL0FEKZ/p1718906302730019

After the SDK version update, the stats.json file size for the bundle analysis became around 992MB, Node has a string size limit of 500MB and the current s.readFileSync(...) logic to read stats.json files fails for files >500MB. So I switched to streaming. Ref: https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/423

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,17 +58,17 @@
     "dev:c": "yarn ocp-console & PLUGIN=${PLUGIN} I8N_NS=${I8N_NS} yarn server:plugin & wait"
   },
   "dependencies": {
-    "@openshift-console/dynamic-plugin-sdk": "1.3.0",
+    "@openshift-console/dynamic-plugin-sdk": "1.5.0",
     "@openshift-console/dynamic-plugin-sdk-internal": "1.0.0",
-    "@openshift-console/dynamic-plugin-sdk-webpack": "1.1.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -58,17 +54,17 @@
...
-    "@patternfly/react-table": "5.1.0",
-    "@patternfly/react-tokens": "5.1.0",
-    "@patternfly/react-topology": "5.0.0",
+    "@patternfly/patternfly": "5.0.4",
+    "@patternfly/react-charts": "7.1.1",
+    "@patternfly/react-core": "5.1.1",
+    "@patternfly/react-icons": "5.1.1",
+    "@patternfly/react-table": "5.1.1",
+    "@patternfly/react-tokens": "5.1.1",
+    "@patternfly/react-topology": "5.1.0",
```
SanjalKatiyar, line 67: we have package.json under `shared` as well...

GowthamShanmugam: ack, will update both

PR #1529, opened 2024-08-14
Fix custom storage quota (value input)

SanjalKatiyar:
https://bugzilla.redhat.com/show_bug.cgi?id=2304305

Quota value was being set as "string" in certain cases, hence API was returning error.

### DIFF
--- a/packages/shared/src/utils/RequestSizeInput.tsx
+++ b/packages/shared/src/utils/RequestSizeInput.tsx
@@ -53,7 +53,7 @@ export const RequestSizeInput: React.FC<RequestSizeInputProps> = ({
   testID,
   unitText,
 }) => {
-  const parsedRequestSizeValue = parseInt(defaultRequestSizeValue, 10);
+  const parsedRequestSizeValue = parseInt(String(defaultRequestSizeValue), 10);
   const defaultValue = Number.isFinite(parsedRequestSizeValue)
     ? parsedRequestSizeValue
     : null;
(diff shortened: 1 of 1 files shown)


packages/shared/src/utils/RequestSizeInput.tsx
```diff
@@ -53,7 +53,10 @@ export const RequestSizeInput: React.FC<RequestSizeInputProps> = ({
   testID,
   unitText,
 }) => {
-  const parsedRequestSizeValue = parseInt(defaultRequestSizeValue, 10);
+  const parsedRequestSizeValue = parseInt(
+    defaultRequestSizeValue as string,
```
SanjalKatiyar, line 57: using `as string` just to make sure there are no type errors as `parseInt` takes first arg as `string`...

but in reality we are passing `defaultRequestSizeValue` as integer in many cases as well (during re-renders)...

SanjalKatiyar: no change functionality wise, just adding types...

GowthamShanmugam: it is better to add this behavior as a comment in the code.

SanjalKatiyar: I can add it, if really needed... but it doesn't look like a big deal to me... I mean it's not affecting any functionality or runtime behaviour...

bipuladh, line 57:
```suggestion
    String(defaultRequestSizeValue)
```
This would guarantee you string everytime

SanjalKatiyar: ack, will update...
though passing `number` doesn't affect the functionality...

PR #1540, opened 2024-08-14
[release-4.16-compatibility] Bug 2290675: Optimize chunk size by changing devtool

GowthamShanmugam:


bipuladh:
Please apply the same changes as per https://github.com/red-hat-storage/odf-console/pull/1534#pullrequestreview-2244648402

PR #1548, opened 2024-08-22
Bug 2304905: [release-4.16] Add cluster name and id on clients' list page

SanjalKatiyar:
Screenshot 2024-08-22 at 6 22 50 PM

### DIFF
--- a/packages/odf/components/storage-consumers/client-list.tsx
+++ b/packages/odf/components/storage-consumers/client-list.tsx
@@ -51,8 +51,8 @@ const tableColumns = [
     id: 'name',
   },
   {
-    className: '',
-    id: 'clusterID',
+    className: 'pf-m-width-20',
+    id: 'clusterName',
   },
   {
     className: '',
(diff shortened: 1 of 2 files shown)


GowthamShanmugam (commented):
LGTM

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -268,7 +268,9 @@ const StorageClientRow: React.FC<
         {obj?.status?.client?.name || '-'}
       </TableData>
       <TableData {...tableColumns[1]} activeColumnIDs={activeColumnIDs}>
-        {obj?.status?.client?.clusterId || '-'}
+        {`${obj?.status?.client?.clusterName || '-'} (${
+          obj?.status?.client?.clusterId || '-'
```
bipuladh, line 272: Why are we falling back to clusterId?

SanjalKatiyar: We are not falling back, we are displaying both together: `ClusterName (ClusterId)`.

It's also same as what's happening in the master.

PR #1569, opened 2024-09-10
Add initial setup for AWS S3 SDK

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-6201

### DIFF
--- a/packages/odf/components/s3-browser/noobaa-context.tsx
+++ b/packages/odf/components/s3-browser/noobaa-context.tsx
@@ -1,0 +1,81 @@
+import * as React from 'react';
+import { useSafeK8sGet } from '@odf/core/hooks';
+import { useODFNamespaceSelector } from '@odf/core/redux';
+import { StatusBox } from '@odf/shared/generic/status-box';
+import { SecretModel, RouteModel } from '@odf/shared/models';
+import { S3Commands } from '@odf/shared/s3';
+import { SecretKind, K8sResourceKind } from '@odf/shared/types';
+import * as _ from 'lodash-es';
+import {
+  NOOBAA_ADMIN_SECRET,
+  NOOBAA_S3_ROUTE,
+  NOOBAA_ACCESS_KEY_ID,
+  NOOBAA_SECRET_ACCESS_KEY,
+} from '../../constants';
+
+type NoobaaS3ContextType = {
+  noobaaS3: S3Commands;
+};
+
+type NoobaaS3ProviderType = {
+  loading?: boolean;
+  error?: unknown;
+};
+
+export const NoobaaS3Context = React.createContext<NoobaaS3ContextType>(
+  {} as NoobaaS3ContextType
+);
+
+// ToDo: In case this provider is needed at too many places, consider applying it to the console's root or use redux instead
+export const NoobaaS3Provider: React.FC<NoobaaS3ProviderType> = ({
+  children,
+  loading,
+  error,
+}) => {
+  const { isODFNsLoaded, odfNsLoadError } = useODFNamespaceSelector();
+
+  const [secretData, secretLoaded, secretError] = useSafeK8sGet<SecretKind>(
+    SecretModel,
+    NOOBAA_ADMIN_SECRET
+  );
+  // ToDo: Configure ConsolePlugin proxy instead of using Route
+  const [routeData, routeLoaded, routeError] = useSafeK8sGet<K8sResourceKind>(
+    RouteModel,
+    NOOBAA_S3_ROUTE
+  );
+
+  const [noobaaS3, noobaaS3Error]: [S3Commands, unknown] = React.useMemo(() => {
+    if (!_.isEmpty(secretData) && !_.isEmpty(routeData)) {
+      try {
+        // ToDo: Remove this once ConsolePlugin proxy is configured
+        const endpoint = `http://${routeData.spec.host}`;
+        const accessKeyId = atob(secretData.data?.[NOOBAA_ACCESS_KEY_ID]);
+        const secretAccessKey = atob(
+          secretData.data?.[NOOBAA_SECRET_ACCESS_KEY]
+        );
+
+        return [new S3Commands(endpoint, accessKeyId, secretAccessKey), null];
+      } catch (err) {
+        return [{} as S3Commands, err];
+      }
+    }
+    return [{} as S3Commands, null];
+  }, [secretData, routeData]);
+
+  const allLoaded =
+    isODFNsLoaded &&
+    secretLoaded &&
+    routeLoaded &&
+    !loading &&
+    !_.isEmpty(noobaaS3);
+  const anyError =
+    odfNsLoadError || secretError || routeError || noobaaS3Error || error;
+
+  return allLoaded && !anyError ? (
+    <NoobaaS3Context.Provider value={{ noobaaS3 }}>
+      {children}
+    </NoobaaS3Context.Provider>
+  ) : (
+    <StatusBox loaded={allLoaded} loadError={anyError} />
+  );
+};
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "dev:c": "yarn ocp-console & PLUGIN=${PLUGIN} I8N_NS=${I8N_NS} yarn server:plugin & wait"
   },
   "dependencies": {
+    "@aws-sdk/client-s3": "3.614.0",
     "@openshift-console/dynamic-plugin-sdk": "1.3.0",
     "@openshift-console/dynamic-plugin-sdk-internal": "1.0.0",
     "@openshift-console/dynamic-plugin-sdk-webpack": "1.1.0",
--- a/packages/odf/constants/index.ts
+++ b/packages/odf/constants/index.ts
@@ -6,3 +6,4 @@ export * from './tooltips';
 export * from './mcg';
 export * from './providerSchema';
 export * from './data-protection';
+export * from './s3-browser';
--- a/packages/odf/constants/s3-browser.ts
+++ b/packages/odf/constants/s3-browser.ts
@@ -1,0 +1,4 @@
+export const NOOBAA_ADMIN_SECRET = 'noobaa-admin';
+export const NOOBAA_S3_ROUTE = 's3';
+export const NOOBAA_ACCESS_KEY_ID = 'AWS_ACCESS_KEY_ID';
+export const NOOBAA_SECRET_ACCESS_KEY = 'AWS_SECRET_ACCESS_KEY';
--- a/packages/shared/src/models/console-models.ts
+++ b/packages/shared/src/models/console-models.ts
@@ -62,3 +62,17 @@ export const ProjectModel: K8sKind = {
   // t('Projects')
   labelPluralKey: 'Projects',
 };
+
+export const RouteModel: K8sKind = {
+  label: 'Route',
+  labelKey: 'Route',
+  labelPlural: 'Routes',
+  labelPluralKey: 'Routes',
+  apiGroup: 'route.openshift.io',
+  apiVersion: 'v1',
+  plural: 'routes',
+  abbr: 'RT',
+  namespaced: true,
+  kind: 'Route',
+  id: 'route',
+};
--- a/packages/shared/src/s3/commands.ts
+++ b/packages/shared/src/s3/commands.ts
@@ -1,0 +1,31 @@
+import {
+  S3Client,
+  ListBucketsCommand,
+  ListObjectsCommand,
+} from '@aws-sdk/client-s3';
+import { ListBuckets, ListObjects } from './types';
+
+export class S3Commands {
+  private s3Client: S3Client;
+
+  constructor(endpoint: string, accessKeyId: string, secretAccessKey: string) {
+    this.s3Client = new S3Client({
+      // "region" is a required parameter for the SDK, using "none" as a workaround
+      region: 'none',
+      endpoint,
+      credentials: {
+        accessKeyId,
+        secretAccessKey,
+      },
+      forcePathStyle: true,
+    });
+  }
+
+  // Bucket command members
+  listBuckets: ListBuckets = (input) =>
+    this.s3Client.send(new ListBucketsCommand(input));
+
+  // Object command members
+  listObjects: ListObjects = (input) =>
+    this.s3Client.send(new ListObjectsCommand(input));
+}
--- a/packages/shared/src/s3/index.ts
+++ b/packages/shared/src/s3/index.ts
@@ -1,0 +1,2 @@
+export * from './commands';
+export * from './types';
--- a/packages/shared/src/s3/types.ts
+++ b/packages/shared/src/s3/types.ts
@@ -1,0 +1,16 @@
+import {
+  ListBucketsCommandInput,
+  ListBucketsCommandOutput,
+  ListObjectsCommandInput,
+  ListObjectsCommandOutput,
+} from '@aws-sdk/client-s3';
+
+// Bucket command types
+export type ListBuckets = (
+  input?: ListBucketsCommandInput
+) => Promise<ListBucketsCommandOutput>;
+
+// Object command types
+export type ListObjects = (
+  input: ListObjectsCommandInput
+) => Promise<ListObjectsCommandOutput>;
--- a/yarn.lock
+++ b/yarn.lock
@@ -15,6 +15,598 @@
     "@jridgewell/gen-mapping" "^0.1.0"
     "@jridgewell/trace-mapping" "^0.3.9"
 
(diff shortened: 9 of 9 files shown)


packages/s3-browser/client-context.tsx
```diff
@@ -0,0 +1,87 @@
...
+    }));
+
+  const s3Client: S3Client = React.useMemo(() => {
+    if (!_.isEmpty(secretData) && !_.isEmpty(routeData)) {
+      try {
+        return new S3Client({
+          // "region" is a required parameter for the SDK, using "none" as a workaround
+          region: 'none',
+          // ToDo: Enable TLS (https)
+          endpoint: `http://${routeData.spec.host}`,
```
SanjalKatiyar, line 58: have a separate story for this: https://issues.redhat.com/browse/RHSTOR-6222

packages/s3-browser/hooks/bucket-commands.tsx
```diff
@@ -0,0 +1,20 @@
...
+) => Promise<ListBucketsCommandOutput>;
+
+export const useListBuckets: UseListBuckets = () => {
+  const { s3Client } = React.useContext(NoobaaS3ClientContext);
+
+  return React.useCallback(
+    (input) => s3Client.send(new ListBucketsCommand(input)),
+    [s3Client]
+  );
+};
```
SanjalKatiyar, line 20: adding these commands as a new hook so that we don't have to manually pass `s3Client` everytime...

packages/s3-browser/client-context.tsx
```diff
@@ -0,0 +1,87 @@
+import * as React from 'react';
+import { S3Client } from '@aws-sdk/client-s3';
+import { useSafeK8sWatchResource } from '@odf/core/hooks';
+import { useODFNamespaceSelector } from '@odf/core/redux';
+import { StatusBox } from '@odf/shared/generic/status-box';
+import { SecretModel, RouteModel } from '@odf/shared/models';
+import { SecretKind, K8sResourceKind } from '@odf/shared/types';
```
GowthamShanmugam, line 7: Just a suggestion:
This looks like a library, but importing components from odf-core and odf-core will import for the s3 CRUD operations feels like a cycle dependency. 

In case we need to move this with the client console, then it will add 3rd dependency.  

What about implementing this s3 browser as a pure library like shared? so it can be re-used between ODF and client easily?

packages/s3-browser/hooks/bucket-commands.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+type UseListBuckets = () => (
+  input?: ListBucketsCommandInput
+) => Promise<ListBucketsCommandOutput>;
+
+export const useListBuckets: UseListBuckets = () => {
+  const { s3Client } = React.useContext(NoobaaS3ClientContext);
+
+  return React.useCallback(
+    (input) => s3Client.send(new ListBucketsCommand(input)),
```
alfonsomthd, line 17: @SanjalKatiyar We should add [error handling](https://aws.amazon.com/blogs/developer/service-error-handling-modular-aws-sdk-js/) here.

SanjalKatiyar: that's the job of the consumer of this callback (entity who will call this hook) IMHO... consumer can set local state on error/success or maybe decide to perform any other use case specific action based on API response... hence callback returns a promise...

ex:
```
const listBuckets  = useListBuckets();
listBuckets(input).then((res) => {
    // do something here
}).catch((err) => {
    // do something here
});
```

packages/s3-browser/client-context.tsx
```diff
@@ -0,0 +1,87 @@
...
+  React.createContext<NoobaaS3ClientContextType>(
+    {} as NoobaaS3ClientContextType
+  );
+
+export const NoobaaS3ClientProvider: React.FC<NoobaaS3ClientProviderType> = ({
+  children,
+  loading,
+  error,
+}) => {
+  const { isODFNsLoaded, odfNsLoadError } = useODFNamespaceSelector();
```
GowthamShanmugam, line 32: isODFNsLoaded and odfNsLoadError also can be a properties to avoid odf-core dependency

packages/s3-browser/client-context.tsx
SanjalKatiyar: It should be re-used between ODF and Client, I agree. But, I did not get the cyclic dependency part ??

Client itself depends on odf/core and odf/ocs: https://github.com/red-hat-storage/odf-console/blob/master/packages/client/package.json#L6-L7

packages/s3-browser/client-context.tsx
```diff
@@ -0,0 +1,87 @@
...
+      name: NOOBAA_ADMIN_SECRET,
+    }));
+  const [routeData, routeLoaded, routeError] =
+    useSafeK8sWatchResource<K8sResourceKind>((ns) => ({
+      isList: false,
+      kind: referenceForModel(RouteModel),
+      namespaced: true,
+      namespace: ns,
+      name: NOOBAA_S3_ROUTE,
+    }));
```
GowthamShanmugam, line 49: Same here, either these can be a parameter or by moving useSafeK8sWatchResource to shared we can avoid odf-core dependency.

packages/s3-browser/client-context.tsx
GowthamShanmugam: again these are all optional comments,  consider this only if we are ready to make this as a library.

packages/s3-browser/client-context.tsx
```diff
@@ -0,0 +1,87 @@
...
+    if (!_.isEmpty(secretData) && !_.isEmpty(routeData)) {
+      try {
+        return new S3Client({
+          // "region" is a required parameter for the SDK, using "none" as a workaround
+          region: 'none',
+          // ToDo: Enable TLS (https)
+          endpoint: `http://${routeData.spec.host}`,
+          credentials: {
+            accessKeyId: atob(secretData.data?.['AWS_ACCESS_KEY_ID']),
+            secretAccessKey: atob(secretData.data?.['AWS_SECRET_ACCESS_KEY']),
```
GowthamShanmugam, line 61: these two keys can be a local constants

packages/s3-browser/client-context.tsx
```diff
@@ -0,0 +1,87 @@
...
+          },
+          forcePathStyle: true,
+        });
+      } catch {
+        return {} as S3Client;
```
GowthamShanmugam, line 66: any todo needed here for handling exceptions?

packages/s3-browser/client-context.tsx
GowthamShanmugam: I am not sure why we added core to the client, I don't think it is a good design. Then there is no point in keeping a shared library. 

We import some components from odf-core into s3-browser. To display the s3 browser-related components, we invoke the context and command hooks into odf-core. But this s3 and client are out of the odf-core folder. These imports look very weird.

SanjalKatiyar: Reason why there are imports from odf/core because of redux and namespace hooks (these are shared by both odf and client), we can move redux to `shared` but that's illogical as well (ODF/Client specific namespace hooks and redux has nothing to do with the common shared).
Also, most of UI pieces in client is exactly same as ODF (like OB/OBC/NS/BS/BC and few dashboard cards), this imported from ODF directly.

`To display the s3 browser-related components, we invoke the context and command hooks into odf-core` >> I was not planning to import the context back in `odf/core`. If that will be the case we can move this folder inside `core` package directly, I agree.
All object browser related code will be inside this new package, we are not importing anything to odf/core.

SanjalKatiyar: but let me know, if needed I can move `s3-browser` right away under core, instead of a new separate package...
cc @GowthamShanmugam @bipuladh

packages/s3-browser/client-context.tsx
```diff
@@ -0,0 +1,87 @@
...
+import * as _ from 'lodash-es';
+import { NOOBAA_ADMIN_SECRET, NOOBAA_S3_ROUTE } from './constants';
+
+type NoobaaS3ClientContextType = {
+  s3Client: S3Client;
+  test?: string;
+};
+
+type NoobaaS3ClientProviderType = {
+  loading?: boolean;
```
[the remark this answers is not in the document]
SanjalKatiyar: nope, it should not be a mandatory field... it's intentionally optional...

packages/s3-browser/client-context.tsx
SanjalKatiyar: I don't think so...

SanjalKatiyar: we could probably, yeah !!

packages/s3-browser/client-context.tsx
SanjalKatiyar: `useSafeK8sWatchResource` is ODF specific hook, should not be a part of common `shared`...

packages/s3-browser/client-context.tsx
GowthamShanmugam: ya, i am confused, The Client has a separate console-extension.json, so it adds of-core as a dependency and expenses its own client-related components(Still I have a concern about adding odf-core as as dependency).  

But here, if it is not a library or package, how are we going to expose the s3-browser components? Is it like adding odf-core as a dependency and also updating the console-extension.json of odf-core to expose the s3-browser components?

SanjalKatiyar: `console-extension.json` is not part of `core` package, nor it has any package.json of it's own... to me it's totally fine to import from `odf/s3` directly in ODF `console-plugin.json` file and same when we add it for Client (import in it's console-plugin.json file).

But yeah, if that's a concern I can move `s3-browser` under odf/core directly.

packages/s3-browser/hooks/bucket-commands.tsx
alfonsomthd: If we want this hook to be only a thin wrapper then it's OK, but I was thinking more about handling generic errors like `InvalidSignatureException` or `ResourceNotFoundException` in every place. We'll probably have to add the same checks in almost every call (as certain errors will be meaningful).
If we see the need in the future, we could handle some cases automatically so we don't have to add repetitive code checks everywhere.

packages/s3-browser/client-context.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { StatusBox } from '@odf/shared/generic/status-box';
+import { SecretModel, RouteModel } from '@odf/shared/models';
+import { SecretKind, K8sResourceKind } from '@odf/shared/types';
+import { referenceForModel } from '@odf/shared/utils';
+import * as _ from 'lodash-es';
+import { NOOBAA_ADMIN_SECRET, NOOBAA_S3_ROUTE } from './constants';
+
+type NoobaaS3ClientContextType = {
+  s3Client: S3Client;
+  test?: string;
```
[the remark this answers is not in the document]
SanjalKatiyar: done

packages/s3-browser/client-context.tsx
SanjalKatiyar: done

packages/s3-browser/hooks/bucket-commands.tsx
SanjalKatiyar: is this what u r suggesting ??

```
  return React.useCallback(
    (input, onSuccess, onError, onInvalidSignatureError, onResourceNotFoundError ...) => s3Client.send(new ListBucketsCommand(input)).then((res) => {
        onSuccess();
    }).catch((err) => {
        onError();
        if (e instanceof InvalidSignatureException) {
            onInvalidSignatureError?.();
        } else if (e instanceof ResourceNotFoundException) {
            onResourceNotFoundError?.();
        } else if (e instanceof FooServiceException) {
            // Handle all other server-side exceptions from Foo service
        } 
        .
        .
        .
        // similarly if/else for all type of exceptions that SDK supports 
    }),
    [s3Client]
  );
```

SanjalKatiyar: If no, then can u plz add the snippet ??

If yes, then:
1. We will have to handle each type of exception as we never know what type the consumer will be interested in.
2. We will have to pass too many arguments.
3. I really doubt that handling particular type of exception will be a common case. Might be at few places, but not too many IMO.
4. How we will use `Promise.all`, if we don't return promise ? (this is much more common case).

alfonsomthd: No, I meant cases like handling a `ResourceNotFoundException` by returning an empty list/object (or null or what we agree upon) instead of having to check on every `catch` if the error is a `ResourceNotFoundException`.
Basically saving boilerplate code.

I'm OK with the current implementation if we interpret my suggestion as a premature optimization.

packages/s3-browser/hooks/object-commands.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
```
bipuladh, line 1: Rename the file to `ts`

packages/s3-browser/client-context.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [noobaaS3Client, noobaaS3ClientError]: [S3Client, unknown] =
+    React.useMemo(() => {
+      if (!_.isEmpty(secretData) && !_.isEmpty(routeData)) {
+        try {
+          return [
+            new S3Client({
+              // "region" is a required parameter for the SDK, using "none" as a workaround
+              region: 'none',
+              // ToDo: Enable TLS (https)
+              endpoint: `http://${routeData.spec.host}`,
```
bipuladh, line 59: Why does this need to be an external endpoint? Can't we bridge it and use the internal noobaa service?

packages/s3-browser/hooks/bucket-commands.tsx
```diff
@@ -0,0 +1,20 @@
...
+  ListBucketsCommandInput,
+  ListBucketsCommandOutput,
+} from '@aws-sdk/client-s3';
+import { NoobaaS3ClientContext } from '../client-context';
+
+type UseListBuckets = () => (
+  input?: ListBucketsCommandInput
+) => Promise<ListBucketsCommandOutput>;
+
+export const useListBuckets: UseListBuckets = () => {
```
bipuladh, line 13: I would find a class with methods that we need for use cases much more useful than creating these many hooks.

packages/s3-browser/client-context.tsx
```diff
@@ -0,0 +1,93 @@
...
+  loading?: boolean;
+  error?: unknown;
+};
+
+export const NoobaaS3ClientContext =
+  React.createContext<NoobaaS3ClientContextType>(
+    {} as NoobaaS3ClientContextType
+  );
+
+export const NoobaaS3ClientProvider: React.FC<NoobaaS3ClientProviderType> = ({
```
bipuladh, line 26: Why do we need a provider? I think a provider here might just be a overkill.

packages/s3-browser/client-context.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [secretData, secretLoaded, secretError] =
+    useSafeK8sWatchResource<SecretKind>((ns) => ({
+      isList: false,
+      kind: referenceForModel(SecretModel),
+      namespaced: true,
+      namespace: ns,
+      name: NOOBAA_ADMIN_SECRET,
+    }));
+  const [routeData, routeLoaded, routeError] =
+    useSafeK8sWatchResource<K8sResourceKind>((ns) => ({
```
bipuladh, line 42: I don't think we need to watch a get would suffice.

bipuladh (changes requested):

bipuladh:
I don't think we should create a seperate package for this. Most of it can reside inside of shared and core.

packages/odf/components/s3-browser/noobaa-context.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  const [secretData, secretLoaded, secretError] = useSafeK8sGet<SecretKind>(
+    SecretModel,
+    NOOBAA_ADMIN_SECRET
+  );
+  // ToDo: Configure ConsolePlugin proxy instead of using Route
+  const [routeData, routeLoaded, routeError] = useSafeK8sGet<K8sResourceKind>(
+    RouteModel,
+    NOOBAA_S3_ROUTE
+  );
```
SanjalKatiyar, line 45: will do it as part of: https://github.com/red-hat-storage/odf-console/pull/1569#discussion_r1751670603

packages/s3-browser/client-context.tsx
SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/1569#discussion_r1758224023

PR #1583, opened 2024-09-17
Add optional chaining for owner references

bipuladh:
Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2245068

### DIFF
--- a/packages/shared/src/topology/utils/resource.ts
+++ b/packages/shared/src/topology/utils/resource.ts
@@ -11,7 +11,7 @@ export const resolveResourceUntilDeployment = (
   if (!owner) {
     return null;
   }
-  if (owner.kind === DeploymentModel.kind) {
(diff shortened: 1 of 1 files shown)


packages/shared/src/topology/utils/resource.ts
```diff
@@ -11,7 +11,7 @@ export const resolveResourceUntilDeployment = (
   if (!owner) {
     return null;
   }
-  if (owner.kind === DeploymentModel.kind) {
+  if (owner?.kind === DeploymentModel.kind) {
```
SanjalKatiyar, line 14: We are checking `if (!owner)` just before this line, do we need optional chain again ?

SanjalKatiyar:
@bipuladh closing this PR as BZ is already ON_QA.

PR #1588, opened 2024-09-18
Fix UI alignment on Connect KMS form

TimothyAsirJeyasing:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2310160

### DIFF
--- a/packages/odf/components/kms-config/kms-config.tsx
+++ b/packages/odf/components/kms-config/kms-config.tsx
@@ -103,7 +103,7 @@ export const KMSConfigure: React.FC<KMSConfigureProps> = ({
   );
 
   return (
-    <div className="odf-m-pane__form">
+    <div className="ocs-storage-class-encryption__form-dropdown--padding">
       {!isWizardFlow && (
         <h3 className="ocs-install-kms__heading">
           {t('Connect to a Key Management Service')}


TimothyAsirJeyasing:
Screenshot from 2024-09-18 13-53-31
Screenshot from 2024-09-18 13-53-13

packages/ocs/storage-class/sc-form.scss
```diff
@@ -17,8 +17,12 @@
   }
 
   &__card {
+    box-sizing: border-box; // Include padding and border in element's total width
+    display: flex;
+    flex-direction: column; // Ensure contents stack vertically
     margin: var(--pf-v5-global--spacer--md) !important; // To override patternfly card class
     padding: var(--pf-v5-global--spacer--md) !important; // To override patternfly card class
+    width: 100%; // Allow card to take full width of its container
```
SanjalKatiyar, line 25: @TimothyAsirJeyasing where is this class used ?? how it's solving this bug ??

TimothyAsirJeyasing: Sorry i have missed to add the class

SanjalKatiyar: but in the previous screenshots attached to the PR, it works fine without the class as well... why is that happening ??

TimothyAsirJeyasing:
before:
https://github.com/user-attachments/assets/b70aaf92-ec28-4ab2-b12d-c5768eaef830


after:
https://github.com/user-attachments/assets/53359e7f-48ad-4025-b65d-e4ea4c545afd

packages/odf/components/kms-config/kms-config.tsx
```diff
@@ -103,7 +103,7 @@ export const KMSConfigure: React.FC<KMSConfigureProps> = ({
   );
 
   return (
-    <div className="odf-m-pane__form">
+    <div className="ocs-storage-class-encryption__form-dropdown--padding">
```
SanjalKatiyar, line 106: u are adding `width: 100%;` to `&__card` and using `&form-dropdown--padding` here... what's the fix for this bug ??

packages/odf/components/kms-config/kms-config.tsx
```diff
@@ -103,7 +103,7 @@ export const KMSConfigure: React.FC<KMSConfigureProps> = ({
   );
 
   return (
-    <div className="odf-m-pane__form">
```
SanjalKatiyar, line 106: why are u removing this class ??

packages/odf/components/kms-config/kms-config.tsx
TimothyAsirJeyasing: The issue arises due to odf-m-pane__form having restrictive min-width and max-width rules, causing content overflow during zooming or resizing. Adding width: 100% to &__card ensures the card adapts dynamically to its parent container, preventing overflow. The &form-dropdown--padding ensures consistent spacing within the form for better alignment.

packages/odf/components/kms-config/kms-config.tsx
TimothyAsirJeyasing: I removed odf-m-pane__form because its fixed width properties were the root cause of the bug.

packages/odf/components/kms-config/kms-config.tsx
bipuladh: Where is `&__card` being used in this form? I am getting a bit confused.

TimothyAsirJeyasing: Correct, The &__card class is not used in this form. After reviewing the code, I realized that adding width: 100% to &__card is not directly relevant to resolving the alignment issue in this form. The primary fix for the issue comes from only removing the odf-m-pane__form class, which had restrictive width rules causing overflow during zooming or resizing. I added using ocs-storage-class-encryption__form-dropdown--padding for consistent padding and better alignment just like we are doing for sc-form.tsx. I will remove &__card changes and will test and update the PR.

PR #1589, opened 2024-09-19
RHSTOR-6347: Enable recipe-based DR enrollment for discovered applications

GowthamShanmugam:


packages/shared/src/utils/AsyncLoader.tsx
```diff
@@ -47,13 +47,18 @@ const useAsynchronousLoading: UseAsynchronousLoading = (
...
+  React.useEffect(() => () => (mounted.current = false), []);
 
   const prevLoader = React.useRef<PromiseComponent>(null);
 
   const setComponent = React.useCallback(
     (value) => {
       Component.current = value;
-      setLoaded(true);
+      if (mounted.current) {
+        setLoaded(true);
```
GowthamShanmugam, line 60: To fix unit test case waring, React state update on an unmounted component

packages/mco/components/modals/app-manage-policies/helper/pvc-details-wizard-content.tsx
```diff
@@ -295,23 +294,23 @@ export const PVCDetailsWizardContent: React.FC<PVCDetailsWizardContentProps> =
     return (
       <Form>
         <FormGroup>
-          <Text>
+          <span>
```
GowthamShanmugam, line 297: To  fix unit test case waring, div cannot appear as a descendant of p

GowthamShanmugam:
Screenshot 2024-10-07 at 4 20 28 PM

packages/shared/src/table/selectable-table.tsx
```diff
@@ -152,7 +156,7 @@ export const SelectableTable: SelectableTableProps = <
                         onSelect: onSelect,
                         isSelected: isRowSelected(getUID(row), selectedRows),
                         isDisabled:
-                          !isRowSelectable?.(row) ||
+                          (!!isRowSelectable && !isRowSelectable?.(row)) ||
```
SanjalKatiyar, line 159: this is not needed... `!isRowSelectable?.(row)` syntax will only execute when function exists, else not...
```suggestion
                          (!isRowSelectable?.(row) ||
```

SanjalKatiyar: original line was correct...

packages/shared/src/utils/AsyncLoader.tsx
```diff
@@ -47,13 +47,18 @@ const useAsynchronousLoading: UseAsynchronousLoading = (
   const Component = React.useRef<React.ComponentType>(null);
   const [loadingStarted, setLoadingStarted] = React.useState(false);
   const [loaded, setLoaded] = React.useState(false);
+  // Mount status for safty state updates
```
SanjalKatiyar, line 50:
```suggestion
  // Mount status for safely state updates
```

SanjalKatiyar: we can remove the comment as well... it's kind of self-explanatory from variable name itself...

packages/shared/src/utils/AsyncLoader.tsx
SanjalKatiyar: Just curious: there are other state variables too, why jest is complaining only about this one in particular ??
Also, why doesn't it do so for other components that we have in this repo, why only this ??

SanjalKatiyar: are we sure there isn't some other reason for this warning ??

packages/shared/src/table/selectable-table.tsx
GowthamShanmugam: The issue in original condition is, 
  isRowSelectable?: IsRowSelectable; is mentioned as optional. But if the component is not passing isRowSelectable, then !isRowSelectable?.(row)always returns True, and all the rows are disabled by default. It's not correct.

SanjalKatiyar: got it !!
then probably we can remove `?.` part entirely... it's not needed when we are using condition like `!!isRowSelectable && ...`

GowthamShanmugam: ack, is it won't affect fusion UI? i think fusion is using this table

SanjalKatiyar: it should not affect in this case...

packages/shared/src/utils/AsyncLoader.tsx
GowthamShanmugam: I verified in the console warning that this error is only being raised in jest, and I am not receiving it.  The async function's state update, which was carried out after the component unmounts, is the cause of this error. This is occurring as a result of the jest test case being executed more quickly. 

adding act() function has resolved this issue.

packages/shared/src/table/selectable-table.tsx
SanjalKatiyar: I meant using `!!isRowSelectable && !isRowSelectable(row))` instead of `!!isRowSelectable && !isRowSelectable?.(row))`.

SanjalKatiyar: `?.`, when we optionally call `isRowSelectable`.

GowthamShanmugam: ack, changes done

packages/shared/src/table/selectable-table.tsx
```diff
@@ -76,7 +76,11 @@ export const SelectableTable: SelectableTableProps = <
 
   const [selectableRows, rowIds] = React.useMemo(() => {
     const selectableRows =
-      sortedRows?.filter(isRowSelectable || hasNoDeletionTimestamp) || [];
+      sortedRows?.filter((selectedRow) =>
+        !!isRowSelectable
+          ? isRowSelectable(selectedRow)
+          : true && hasNoDeletionTimestamp(selectedRow)
+      ) || [];
```
SanjalKatiyar, line 83: why do we need this ??
`sortedRows?.filter(isRowSelectable || hasNoDeletionTimestamp) || [];` format is correct, `filter` takes a callback as it's argument.

SanjalKatiyar: even if `isRowSelectable` is undefined here that should work as expected...

packages/mco/components/discovered-application-wizard/enroll-discovered-application.spec.tsx
```diff
@@ -284,16 +290,17 @@ const moveToStep = async (step: number) => {
...
+    });
   }
 
-  // Commented out recipe-based steps
   if (step > 2) {
     // Select recipe
-    // fireEvent.click(screen.getByText('Recipe'));
-    // fireEvent.click(screen.getByText('Select a recipe'));
-    // fireEvent.click(screen.getByText('mock-recipe-1'));
+    fireEvent.click(screen.getByText('Recipe'));
```
alfonsomthd, line 301: Please don't use [fireEvent](https://github.com/testing-library/user-event?tab=readme-ov-file#the-problem): use `user-event` instead.

alfonsomthd: You have n example of `user-event` usage [here](packages/odf/components/namespace-store/namespace-store-form.spec.tsx).

packages/shared/src/table/selectable-table.tsx
GowthamShanmugam: right! my bad

PR #1591, opened 2024-09-20
Add buckets overview page (header, details/objects list tabs, actions) layout

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-6196
https://issues.redhat.com/browse/RHSTOR-6199

Screenshot 2024-09-25 at 6 59 04 PM
Screenshot 2024-09-25 at 7 12 30 PM
Screenshot 2024-09-25 at 7 46 01 PM
Screenshot 2024-09-25 at 7 53 29 PM


packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
```diff
@@ -0,0 +1,213 @@
...
+  navigate: NavigateFunction,
+  bucketName: string,
+  isCreatedByOBC: boolean,
+  noobaaObjectBucket: K8sResourceKind
+): IAction[] => [
+  // ToDo: add empty/delete bucket actions
+  {
+    title: t('Empty bucket'),
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    onClick: () => {},
```
alfonsomthd, line 44: You don't need the disable comment if you return `undefined` (according to the TypeScript documentation, a void return type is equivalent to a return type of `undefined`):
```suggestion
    onClick: () => undefined,
```

packages/odf/utils/s3-browser.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  const initialBreadcrumb = [
+    ...bucketsListPage,
+    { name: bucketName, path: `${BUCKETS_BASE_ROUTE}/${bucketName}` },
+  ];
+  let encodedPathTofolder = '';
+  return {
+    breadcrumbs: folders.reduce((acc, folderName) => {
+      const folderNameWDelimiter = folderName + DELIMITER;
+      encodedPathTofolder = !encodedPathTofolder
```
alfonsomthd, line 35:
```suggestion
      encodedPathToFolder = !encodedPathToFolder
```

alfonsomthd:
@SanjalKatiyar you should rebase in order to fix conflicts in`locales/en/plugin__odf-console.json`.

SanjalKatiyar:
> @SanjalKatiyar you should rebase in order to fix conflicts in`locales/en/plugin__odf-console.json`.

Ack, will do once PR is reviewed and fix it along with other review comments.

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
```diff
@@ -0,0 +1,213 @@
...
+        },
+      ]
+    : []),
+  ...(isCreatedByOBC
+    ? [
+        {
+          title: t('Edit bucket'),
+          onClick: () => navigate(`${BUCKETS_BASE_ROUTE}/${bucketName}/yaml`),
+        },
+      ]
```
GowthamShanmugam, line 87: is there any reason the same check has been repeated? we can move three options from the same list, right?

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
```diff
@@ -0,0 +1,213 @@
...
+      ? [
+          {
+            href: 'details',
+            name: t('Details'),
+            // ToDo: add bucket details page
+            component: () => <>Bucket details page</>,
+          },
+        ]
+      : []),
+    ...(isCreatedByOBC
```
GowthamShanmugam, line 142: is YAML is visible for folder view?

GowthamShanmugam: ok! isCreatedByOBC will be false for folder view, got it
I prefer to move comment (will be false if we are inside folder view): https://github.com/red-hat-storage/odf-console/pull/1591/files#diff-1f39da47c37a872f0a045b71647ab27ae7d8e5527e3c38457b910e79538ef615R118 here

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
```diff
@@ -0,0 +1,213 @@
...
+        {
+          kind: referenceForModel(NooBaaObjectBucketModel),
+          namespaced: false,
+          isList: true,
+        },
+        !foldersPath
+      )
+    );
+  const noobaaObjectBucket: K8sResourceKind = objectBuckets?.find(
+    (ob) => ob.spec?.endpoint?.bucketName === bucketName
```
GowthamShanmugam, line 116: Doub is do we need to loop all object buckets? we can use the bucketName in useK8sWatchResource to look for the specific object right? if the bucket name found then it is obc then not found then s3.

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
```diff
@@ -0,0 +1,213 @@
...
+  );
+  // denotes whether bucket is created via OBC or S3 endpoint (will be false if we are inside folder view)
+  const isCreatedByOBC = !!noobaaObjectBucket;
```
GowthamShanmugam, line 119: For s3 bucket, are we not triggering any API call to read further information?

packages/odf/components/s3-browser/bucket-overview/PageTitle.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+const CreatedOnSkeleton: React.FC<{}> = () => (
+  <Skeleton width="25%" height="15%" />
+);
+
+const CreatedOn: React.FC<{ bucketName: string }> = ({ bucketName }) => {
+  const { t } = useCustomTranslation();
+
+  const { noobaaS3 } = React.useContext(NoobaaS3Context);
+  const { data, error, isLoading } = useSWR(LIST_BUCKET, () =>
```
GowthamShanmugam, line 32: ok! there is no API endpoint to read a specific bucket?

packages/odf/components/s3-browser/bucket-overview/PageTitle.tsx
```diff
@@ -0,0 +1,120 @@
...
+    !isLoading && !error
+      ? data?.Buckets?.find((bucket) => bucket?.Name === bucketName)
+          ?.CreationDate
+      : null;
+
+  return isLoading ? (
+    <CreatedOnSkeleton />
+  ) : (
+    <h4 className="text-muted">
+      {t('Created on: ') + bucketCreatedOn?.toString()}
```
GowthamShanmugam, line 46: OCP console follows UTC date-time format, we need to makesure it display on a same format

packages/odf/components/s3-browser/objects-list/ObjectsList.tsx
```diff
@@ -0,0 +1,330 @@
...
+  MenuToggle,
+} from '@patternfly/react-core';
+import { AngleLeftIcon, AngleRightIcon } from '@patternfly/react-icons';
+import {
+  ActionsColumn,
+  IAction,
+  CustomActionsToggleProps,
+} from '@patternfly/react-table';
+import { LIST_OBJECTS, DELIMITER, MAX_KEYS, PREFIX } from '../../../constants';
+import { ObjectCrFormat } from '../../../types';
```
GowthamShanmugam, line 29:
```suggestion
import { ObjectCRFormat } from '../../../types';
```
thoughts?

packages/odf/components/s3-browser/objects-list/ObjectsList.tsx
```diff
@@ -0,0 +1,330 @@
...
+import { getPath, convertObjectsDataToCrFormat } from '../../../utils';
```
GowthamShanmugam, line 30:
```suggestion
import { getPath, convertObjectsDataToCRFormat } from '../../../utils';
```

same here

packages/odf/components/s3-browser/objects-list/ObjectsList.tsx
```diff
@@ -0,0 +1,330 @@
...
+) => {
+  setContinuationTokens((oldTokens) => {
+    const newTokens = _.cloneDeep(oldTokens);
+    if (isNext) {
+      newTokens.previous.push(newTokens.current);
+      newTokens.current = newTokens.next;
+    } else {
+      newTokens.current = newTokens.previous.pop();
+    }
+    newTokens.next = response.NextContinuationToken;
```
GowthamShanmugam, line 81:
```suggestion
    if (isNext) {
      newTokens.previous.push(newTokens.current);
      newTokens.current = newTokens.next;
      newTokens.next = response.NextContinuationToken;
    } else {
      // I am assuming this is for back
       newTokens.next = newTokens.current
      newTokens.current = newTokens.previous.pop();
    }
```

because back noting to do with response.NextContinuationToken I think

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
SanjalKatiyar: "ObjectBucket" CR name is different from actual underlying "Bucket" name... CR's spec will have the correct name...

packages/odf/components/s3-browser/bucket-overview/PageTitle.tsx
SanjalKatiyar: yeah, I did not find any...

packages/odf/components/s3-browser/bucket-overview/PageTitle.tsx
```diff
@@ -0,0 +1,120 @@
...
+            </Label>
+            {resourceStatus && (
+              <BucketResourceStatus resourceStatus={resourceStatus} />
+            )}
+          </>
+        )}
+      </div>
+      {!foldersPath &&
+        (fresh ? <CreatedOn bucketName={bucketName} /> : <CreatedOnSkeleton />)}
+      <h4>
```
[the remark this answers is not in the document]
SanjalKatiyar: yeah, no issues with that...

packages/odf/components/s3-browser/bucket-overview/PageTitle.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const objectPath = getPath(bucketName, foldersPath);
+  const resourceStatus = isCreatedByOBC
+    ? getResourceStatus(noobaaObjectBucket)
+    : null;
+  const createdBy = isCreatedByOBC ? t('Created via OBC') : t('Created via S3');
+
+  return (
+    <div className="pf-v5-u-display-flex pf-v5-u-flex-direction-column">
+      <div className="pf-v5-u-display-flex pf-v5-u-flex-direction-row">
+        {!foldersPath ? bucketName : currentFolder}{' '}
```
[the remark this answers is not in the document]
SanjalKatiyar: I think it was added by linter, will check again though...

packages/odf/components/s3-browser/objects-list/ObjectsList.tsx
SanjalKatiyar: current conditions in the PR are tested and works well...

GowthamShanmugam: Ok! it works becuase in esponse.NextContinuationToken and newTokens.current may have same values.

GowthamShanmugam: But i agree it is safer to use in response.NextContinuationToken for newTokens.next. because,if any object is deleted, then newTokens.current is not the right value to use.

packages/odf/components/s3-browser/objects-list/ObjectsList.tsx
SanjalKatiyar: either is fine IMO, we don't follow any specific rule in this repo... I have seen both naming formats...

packages/odf/components/s3-browser/bucket-overview/PageTitle.tsx
SanjalKatiyar: ack !

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
```diff
@@ -0,0 +1,213 @@
...
+  const [fresh, triggerRefresh] = useRefresh();
+
+  const launcher = useModal();
+  const navigate = useNavigate();
+
+  const { bucketName } = useParams();
+  const [searchParams] = useSearchParams();
+
+  // if non-empty means we are inside particular folder(s) of a bucket, else just inside a bucket (top-level)
+  const foldersPath = searchParams.get(PREFIX);
```
bipuladh, line 102: The constant name `PREFIX` is very vague.

packages/odf/components/s3-browser/bucket-overview/PageTitle.tsx
```diff
@@ -0,0 +1,120 @@
...
+  fresh: boolean;
+  isCreatedByOBC: boolean;
+  noobaaObjectBucket: K8sResourceKind;
+};
+
+const CreatedOnSkeleton: React.FC<{}> = () => (
+  <Skeleton width="25%" height="15%" />
+);
+
+const CreatedOn: React.FC<{ bucketName: string }> = ({ bucketName }) => {
```
bipuladh, line 28: Use types.

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
SanjalKatiyar: this is exact same name as API's params...
```
Prefix — (String)
Limits the response to keys that begin with the specified prefix.
```

packages/odf/components/s3-browser/objects-list/table-components.tsx
```diff
@@ -0,0 +1,152 @@
...
+  extraProps,
+}) => {
+  const { t } = useCustomTranslation();
+
+  const { launcher, bucketName, foldersPath } = extraProps;
+  const isFolder = object.isFolder;
+  const name = object.metadata.name.replace(foldersPath, '');
+  const prefix = !!foldersPath
+    ? encodeURIComponent(foldersPath + name)
+    : encodeURIComponent(name);
```
GowthamShanmugam, line 98: prefix is required for non-foldersPath also? means, when the foldersPath is empty then encodeURIComponent for prefix and prefix used under  {isFolder ? (

os isFolder can be true when foldersPath is empty ?

packages/odf/components/s3-browser/objects-list/table-components.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+export const TableRow: React.FC<RowComponentType<ObjectCrFormat>> = ({
+  row: object,
+  extraProps,
+}) => {
+  const { t } = useCustomTranslation();
+
+  const { launcher, bucketName, foldersPath } = extraProps;
+  const isFolder = object.isFolder;
+  const name = object.metadata.name.replace(foldersPath, '');
```
GowthamShanmugam, line 95:
```suggestion
  const name = getName(object).replace(foldersPath, '');
```
optional

packages/odf/components/s3-browser/bucket-overview/PageTitle.tsx
SanjalKatiyar: I didn't get it, I am using type, right ?? just I am not storing it in a separate variable as it's only a single prop...

packages/odf/components/s3-browser/objects-list/table-components.tsx
```diff
@@ -0,0 +1,152 @@
...
+      <Td translate={null} dataLabel={columnNames[1]}>
+        {object.apiResponse.size}
+      </Td>
+      <Td translate={null} dataLabel={columnNames[2]}>
+        {object.type}
+      </Td>
+      <Td translate={null} dataLabel={columnNames[3]}>
+        {object.apiResponse.lastModified}
+      </Td>
+      <Td translate={null} dataLabel={columnNames[3]} isActionCell>
```
GowthamShanmugam, line 122: columnNames[3] is used twice, is this expected?

packages/odf/components/s3-browser/objects-list/table-components.tsx
SanjalKatiyar: this is used for creating a URL for the folder view, each folder view will have a prefix in the URL...

initial buckets page will not have `foldersPath`, but when we drill down prefix will be folder name...

packages/odf/components/s3-browser/objects-list/table-components.tsx
alfonsomthd: @GowthamShanmugam @SanjalKatiyar this is why I would prefer for all our tables to not depend on hardcoded column numbers but iterating and using the iterator index so adding/removing column should be error-free.
Here is how I did it for [storage client list](https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/storage-consumers/client-list.tsx#L160).

alfonsomthd: @SanjalKatiyar not saying that my suggestion should be applied here in this PR: just something that we should discuss (also better alternatives) for future implementations.

GowthamShanmugam: From the link above, what I understood is, that you are suggesting to use  name as a column id.

packages/shared/src/hooks/useRefresh.ts
```diff
@@ -0,0 +1,33 @@
...
+  shouldUsePromise = false
+): [boolean, () => void] => {
+  const [fresh, setFresh] = React.useState(true);
+  const id = React.useRef<NodeJS.Timeout>();
+
+  const triggerRefresh = React.useCallback(() => {
+    setFresh(false);
+
+    if (shouldUsePromise) {
+      // queuing for immediate re-toggle
```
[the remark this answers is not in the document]
SanjalKatiyar: both spellings are correct AFAIU :)

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
SanjalKatiyar: will update...

packages/odf/components/s3-browser/objects-list/table-components.tsx
SanjalKatiyar: ack !

packages/odf/components/s3-browser/bucket-overview/PageTitle.tsx
SanjalKatiyar: I am keeping it as it is for now... ODF shows UTC only for tooltip, we don't have tooltip right now... other issue is `Timestamp` FC display time as eg: `7 mins ago` etc which works well for websocket connections are time get continuously updated, but here we have manual refresh...

SanjalKatiyar: IIUC displayed format here will be dependent on local timezone... we can confirm on @alfonsomthd system once this PR is merged...

packages/odf/components/s3-browser/bucket-overview/PageTitle.tsx
alfonsomthd: @SanjalKatiyar have you cheked [`HeadBucket`](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadBucket.html) ?

packages/odf/components/s3-browser/bucket-overview/PageTitle.tsx
alfonsomthd: @SanjalKatiyar the UK dev cluster has set the London timezone, so we can test it there.

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
```diff
@@ -0,0 +1,203 @@
...
+            href: 'yaml',
+            name: t('YAML'),
+            component: YAMLEditorWrapped,
+          },
+        ]
+      : []),
+  ];
+
+  const actions = () => {
+    return (
```
bipuladh, line 144:
```suggestion
  const actions = () => (
```
nit

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
```diff
@@ -0,0 +1,203 @@
...
+      <>
+        <Button
+          className="pf-v5-u-mr-md pf-v5-u-mb-xs"
+          variant={ButtonVariant.link}
+          icon={<BlueSyncIcon />}
+          onClick={triggerRefresh}
+          isDisabled={!fresh}
```
bipuladh, line 151: Does `fresh` mean being refreshed? If that is the case let's rename it to `isRefreshing`

packages/odf/components/s3-browser/bucket-overview/PageTitle.tsx
bipuladh: We always use a different variable in our codebase. So it would be nice to stick to the conventions.

packages/odf/components/s3-browser/bucket-overview/PageTitle.tsx
```diff
@@ -0,0 +1,120 @@
...
+  return isLoading ? (
+    <CreatedOnSkeleton />
+  ) : (
+    <h4 className="text-muted">
+      {t('Created on: ') + bucketCreatedOn?.toString()}
+    </h4>
+  );
+};
+
+const BucketResourceStatus: React.FC<{ resourceStatus: string }> = ({
```
bipuladh, line 51: Same as above. Let's stick to our conventions.

packages/odf/components/s3-browser/bucket-overview/bucket-overview.scss
```diff
@@ -0,0 +1,7 @@
+.bucket-label--height {
+    height: 1rem;
+}
+
+.bucket-label--margin-top {
+    margin-top: calc(0.3rem * -1);
```
bipuladh, line 6: why are we using negative margin to pull things? There should be a better way to do this.

bipuladh:
/lgtm
We can take up the minor comments in the follow up PRs

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
SanjalKatiyar: `fresh = true` means it's latest... `fresh = false` means it's refreshing (mounting/un-mounting)...

packages/odf/components/s3-browser/bucket-overview/bucket-overview.scss
SanjalKatiyar: Just curious, is something wrong with -ve values ??
Even MDN describe it as `A positive value places it farther from its neighbors, while a negative value places it closer.` in it's very first paragraph :/ (without mentioning it as a bad practice or convention).

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
SanjalKatiyar: ack !

PR #1594, opened 2024-09-23
Create Bucket page: display CreateOBC form.

alfonsomthd:
- Add CreateBucket page.
- Refactor CreateOBC form to show the namespace selector when loaded in the CreateBucket page.
- TODO (follow-up): implement S3 form.
image


packages/shared/src/dropdown/ResourceDropdown.tsx
```diff
@@ -184,6 +184,7 @@ const ResourceDropdown: ResourceDropdown = <T extends unknown>({
             filterName(searchText, getName(res)) &&
             (filterResource ? filterResource(res) : true)
         )
+        .sort((a, b) => getName(a).localeCompare(getName(b)))
```
alfonsomthd, line 187: Sort dropdown items alphabetically.

locales/en/plugin__odf-console.json
```diff
@@ -1096,6 +1097,14 @@
   "NamespaceStore details": "NamespaceStore details",
   "Target Blob Container": "Target Blob Container",
   "Num Volumes": "Num Volumes",
+  "Create Bucket": "Create Bucket",
+  "An object bucket is a cloud storage container that organizes and manages files (objects), allowing users to store, retrieve and control access to data efficiently.": "An object bucket is a cloud storage container that organizes and manages files (objects), allowing users to store, retrieve and control access to data efficiently.",
+  "Select bucket creation method": "Select bucket creation method",
+  "Create via S3 API": "Create via S3 API",
+  "Ideal for applications and systems that need to interact directly with S3 storage, especially when integrating with AWS-native services and features.": "Ideal for applications and systems that need to interact directly with S3 storage, especially when integrating with AWS-native services and features.",
```
SanjalKatiyar, line 1104: Didn't notice it in the mockups, but this has nothing to do with "AWS-native". Added a comment on designs as well.
```suggestion
  "Ideal for applications and systems that need to interact directly with S3 storage, especially when integrating with AWS-native services and features.": "Ideal for applications and systems that need to interact directly with S3 storage.",
```

packages/odf/components/s3-browser/create-bucket/CreateBucket.tsx
```diff
@@ -0,0 +1,71 @@
...
+          </FormGroup>
+        </div>
+        <Alert
+          variant="info"
+          isInline
+          title={t(
+            'OBC references a StorageClass that uses a provisioner to interact with the S3 API and create the bucket. Kubernetes binds the OBC to the bucket, making it accessible to applications.'
+          )}
+          className="pf-v5-u-mb-md"
+        />
```
SanjalKatiyar, line 61: this info should only be present when "Create via OBC" tile is selected IMO, not both.

packages/shared/src/dropdown/ResourceDropdown.tsx
SanjalKatiyar: why do we need it ?? I mean, maybe good to have, but so far we are using it without it everywhere, if really needed can we add an optional parameter for sorting else let's not do it right now ??

packages/odf/components/s3-browser/create-bucket/CreateBucket.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from 'react';
+import { CreateOBC } from '@odf/core/components/mcg/CreateObjectBucketClaim';
+import { useCustomTranslation } from '@odf/shared';
+import { Alert, FormGroup, Tile } from '@patternfly/react-core';
+
+enum CreationMethod {
+  OBC = 'obc',
+  S3 = 's3',
+}
```
SanjalKatiyar, line 9: nits: should go under `packages/odf/constants/s3-browser.ts`

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -142,10 +145,20 @@ const generateAdditionalReplicationResources = (
...
+    showNamespaceSelector = false,
+    control,
+    fieldRequirements,
+  } = props;
   const isNoobaa = state.scProvisioner?.includes(NOOBAA_PROVISIONER);
 
   const { odfNamespace } = useODFNamespaceSelector();
+  const [namespace, setNamespace] = React.useState<string>(
+    showNamespaceSelector ? odfNamespace : ns
+  );
```
SanjalKatiyar, line 161: I doubt this will work this way... everything happening inside `CreateOBC` is tied to OCP's project dropdown (this line: `const [namespace, setNamespace] = useActiveNamespace();`)...

whereas everything happening inside this `CreateOBCForm` is tied to local `useState` (this: `const [namespace, setNamespace] = React.useState<string>(
    showNamespaceSelector ? odfNamespace : ns
  );`)

SanjalKatiyar: both namespaces can be different at the same time...

SanjalKatiyar: we need to refactor is such a way that a parent component will pass this namespace to `CreateOBC`, rest everything should remain same...
that parent component will decide whether the namespace should be `useActiveNamespace` or `useState`.

SanjalKatiyar: let me know if I misunderstood something...

packages/shared/src/dropdown/ResourceDropdown.tsx
alfonsomthd: @SanjalKatiyar Without sorting you have a very confusing list that IMHO is bad UX experience:
image
We can add an optional parameter, but what should be the use case for NOT sorting?

locales/en/plugin__odf-console.json
alfonsomthd: Until further discussion, we'll go with this text: `We'll go with this text: 'Ideal for applications and systems that need to interact directly with S3-compatible  storage.'`

packages/odf/components/s3-browser/create-bucket/CreateBucket.tsx
alfonsomthd: > this info should only be present when "Create via OBC" tile is selected IMO, not both.

I see you point.
I'll confirm with Chandan, as I understood that the info was to help the user when choosing between the 2 options, hence the reason why I made it always visible...

packages/odf/components/s3-browser/create-bucket/CreateBucket.tsx
alfonsomthd: This only makes sense in the context of this form: it's not a generic constant.

Is it not the rule to put under constants file when it makes sense (required by several files)?

packages/shared/src/dropdown/ResourceDropdown.tsx
SanjalKatiyar: no particular use case, just keeping the things as it is unless anyone complaints, suggest us otherwise or there's a need...
however sorting is not a big change, I agree... if there is a strong UX impact we can sort as well...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
alfonsomthd: The original behavior hasn't changed and the new one it's tested: let's have a call on this.
The only thing it was wrong after inspecting is that I accidnetally removed the namespace prop here:
https://github.com/red-hat-storage/odf-console/pull/1594/files#diff-2a373a2882ddcfb4402ff2f81856b915b2e91142e87b75c305215a4cc61bde7bL495

packages/shared/src/dropdown/ResourceDropdown.tsx
alfonsomthd: @SanjalKatiyar provided `<ResourcesDropdown>` is only used here in `CreateObjectBucketClaim` and the sorting is not breaking any implementation, can we agree on adding sorting as the default behavior?

packages/odf/components/s3-browser/create-bucket/CreateBucket.tsx
alfonsomthd: Now I see in the design that it's not present when the user has selected the other option, I'll address this.

packages/shared/src/dropdown/ResourceDropdown.tsx
SanjalKatiyar: `ResourcesDropdown` is used thought the repo, not just `CreateObjectBucketClaim`.
I am okay with sorting (as it's not a major change), but just my intuition will be not to do it unless needed.

packages/odf/components/s3-browser/create-bucket/CreateBucket.tsx
SanjalKatiyar: either place is alright...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
SanjalKatiyar: existing OBC page will work correctly, that's true...

this new OBC page (under "Create Bucket") will not work for `additionalResources` that `CreateOBC` component creates, that's still bound to the namespace from the Project dropdown (which is a global state, across OCP + ODF)...

alfonsomthd: Addressed.

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -477,6 +488,80 @@ export const CreateOBCPage: React.FC<{}> = () => {
...
+              helperText: t(
+                'A namespace controls access to the OBC and ties the buckets to a specific project.'
+              ),
+              isRequired: true,
+            }}
+            render={({ onBlur }) => (
+              <ResourcesDropdown<K8sResourceCommon>
+                resourceModel={NamespaceModel}
+                onSelect={(selectedNamespace) =>
+                  setNamespace(getName(selectedNamespace))
```
SanjalKatiyar, line 510: `setNamespace` is a global setter, it sets value for entire OCP "Project" dropdown... using it for our local namespace dropdown might not be a good thing to do (it's just a good hack)...

my suggestion would be to create (FC names are just for reference, not a naming suggestion):
```
const OBCWrapperWithProjectDropdownSetter = () => {
  const [namespace, setNamespace] = useActiveNamespace();
  // all code related to setting OCP Project's dropdown moved here
  .
  .
  return <CreateOBC namespace={namespace} />;
}

const CreateOBC = () => {
  // this local state only be used when showNamespaceSelector is true
  const [namespace, setNamespace] = useState();
  .
  .
  return (...);
}
```

SanjalKatiyar: or if u don't wanna do above refactoring then maybe add a `const [namespace, setNamespace] = useState();` in Current_CreateOBC_FC_Introduced_In_This_PR itself directly... with moving ProjectDropdown_Setter code outside... it will make FC a bit messy though...

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -407,12 +415,15 @@ export const CreateOBCPage: React.FC<{}> = () => {
...
+      !showNamespaceSelector
+    ) {
       navigate(
         `/odf/object-storage/${referenceForModel(NooBaaObjectBucketClaimModel)}`
       );
     }
-  }, [navigate, namespace, setNamespace]);
+  }, [navigate, namespace, setNamespace, showNamespaceSelector]);
 
   const save = (
```
SanjalKatiyar, line 428: u also need to update this `save`... currently it will redirect to OBC details page, where it should redirect to a bucket's list page now as create form is called "Create Bucket" in new use case.

Route should be something like this: `/odf/object-storage/buckets/:bucketName` (bucketName will be part of newly created OBC CR)

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
alfonsomthd: OK, provided we don't want to change the project ns from the bucket ns selector and taking into account that the project namespace is only required for setting the default ns (when showing the bucket ns selector), IMHO this would  be enough without the need of another wrapper:
```
  const [projectNs] = useActiveNamespace();
  const [namespace, setNamespace] = React.useState(showNamespaceSelector ? DEFAULT_STORAGE_NAMESPACE : projectNs);
```

SanjalKatiyar: When we are calling OBC page standalone (like today), then we need `const [namespace, setNamespace] = useActiveNamespace();` (both setter and the value) because here we want global state to get set (as we are using OCP Project dropdown component here from the sdk).

When we are calling OBC page from "Create Bucket" page then we don't need to mess with the global state at all which OCP's Project dropdown maintains (it won't cause any bug, but is logically incorrect)... that's why initial suggestion was to create a wrapper for `const [namespace, setNamespace] = useActiveNamespace()` and all it's related hooks and only call this wrapper in case of standalone OBC page.
Other option in case we don't want this wrapper based refactoring would be to keep both in same component (bit messy, but maybe okay for now):
```
// this namespace is used when we display OCP's "Project" dropdown on the page header (used in Storage > Object Storage > Object Bucket Claims)
const [namespace, setNamespace] = useActiveNamespace();

// this namespace is used when we have custom namespace dropdown (used in Storage > Object  Storage > Create Bucket)
const [selectorNamespace, setSelectorNamespace] = React.useState(odfNamespace);
```
We also need to make sure this hook: https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/mcg/CreateObjectBucketClaim.tsx#L405-L415 only runs when `showNamespaceSelector === false` as it sets global state as well.

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
alfonsomthd: Added a @TODO comment until the route is confirmed.

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
alfonsomthd: My suggestion is aligned with your last comment.

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -364,10 +370,19 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+
+export const CreateOBC: React.FC<CreateOBCProps> = ({
+  showNamespaceSelector = false,
+}) => {
   const { t } = useCustomTranslation();
   const [state, dispatch] = React.useReducer(commonReducer, defaultState);
-  const [namespace, setNamespace] = useActiveNamespace();
+  const [projectNs, setProjectNs] = useActiveNamespace();
+  const [namespace, setNamespace] = React.useState(
+    showNamespaceSelector ? DEFAULT_STORAGE_NAMESPACE : projectNs
```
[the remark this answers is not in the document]
alfonsomthd: DOne.

packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -364,10 +366,20 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
...
+  showNamespaceSelector = false,
+}) => {
   const { t } = useCustomTranslation();
   const [state, dispatch] = React.useReducer(commonReducer, defaultState);
-  const [namespace, setNamespace] = useActiveNamespace();
+  const [projectNs, setProjectNs] = useActiveNamespace();
+  const { odfNamespace } = useODFNamespaceSelector();
+  const [namespace, setNamespace] = React.useState(
+    showNamespaceSelector ? odfNamespace : projectNs
+  );
```
SanjalKatiyar, line 382: Consider this scenario:
1. Go to "Storage > Object Storage > Object Bucket Claims".
2. Make sure "All namespaces" is selected from the project dropdown.
3. Click "Create" OBC.
4. U will reach current OBC form with "default" pre-selected in the dropdown.
5. Check value of variable `projectNs` (it should be `default` as `useEffect` will update this and re-render)... but check value of variable `namespace` (is it `ALL_NAMESPACES_KEY` ??).
6. If value of `namespace` is `ALL_NAMESPACES_KEY` then it's an issue as we r passing this `namespace` to all other components & resources that this form is creating.

SanjalKatiyar: basically we need to use `namespace` if `showNamespaceSelector === true` and need to use `projectNs` when `showNamespaceSelector === false`.

```
  const [namespace, setNamespace] = React.useState(
    showNamespaceSelector ? odfNamespace : projectNs
  );
```
this snippet will set `namespace = projectNs` only during first render... when `useEffect` will update `projectNs` to `default` and re-render, then `namespace` will not get changed.

alfonsomthd: Ensured they keep in sync when the namespace form selector is not shown.

PR #1596, opened 2024-09-26
Fix security issue: Backtracking regular expressions

alfonsomthd:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2311004

Update `path-to-regexp` to patched versions:
- Update `express` to a version that requires a patched version of `path-to-regexp`.
- Remove unnecessary dependency `react-router` v5 that has `path-to-regexp` as dependency.
`react-router-dom-v5-compat` can be replaced in a follow-up PR with `react-router` v6.
- Ensure that `path-to-regexp` version installed by `@openshift-console/dynamic-plugin-sdk-internal`
is a [patched version](https://github.com/pillarjs/path-to-regexp/releases?q=1.9.0&expanded=true).
```
odf-plugin@0.0.0
├─┬ @openshift-console/dynamic-plugin-sdk-internal@1.0.0
│ └─┬ react-router@5.3.4
│   └── path-to-regexp@1.9.0
└─┬ webpack-dev-server@4.15.2
  └─┬ express@4.21.0
    └── path-to-regexp@0.1.10
```


packages/shared/package.json
```diff
@@ -61,7 +61,6 @@
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "react-i18next": "^11.11.4",
-    "react-router": "5.3.x",
```
alfonsomthd, line 64: @SanjalKatiyar in the external plugin that uses the shared library we're using `react-router-dom-v5-compat`.

aruniiird:
cc @aruniiird

GowthamShanmugam:
I think this express package update will solve all other CVEs, almost 90% CVEs that we got have problems with this express dependency only.

TimothyAsirJeyasing:
LGTM

SanjalKatiyar (commented):
`react-router-dom-v5-compat can be replaced with react-router v6 in a follow-up PR` >> First OCP console need to upgrade, then only we should do so.

PR #1598, opened 2024-09-26
RHSTOR-6195: Bucket - list page

GowthamShanmugam:
https://issues.redhat.com/browse/RHSTOR-6195
ToDo bucket sync button after merging: https://github.com/red-hat-storage/odf-console/pull/1591

To solve the bundle analyzer issue: https://github.com/red-hat-storage/odf-console/pull/1626


GowthamShanmugam:
Screenshot 2024-09-26 at 7 34 07 PM

packages/odf/components/s3-browser/bucket/buckets-list-page/bucketsListPage.tsx
```diff
@@ -0,0 +1,250 @@
...
+const BucketsListPage: React.FC = () => {
+  const { t } = useCustomTranslation();
+  const { noobaaS3 } = React.useContext(NoobaaS3Context);
+  const { data, error, isLoading } = useSWR(LIST_BUCKET, () =>
+    noobaaS3.listBuckets()
+  );
+  const [buckets, filteredData, onFilterChange] = useListPageFilter(
+    convertS3ListBucketOutputToK8sCompatible(data)
+  );
+  const [favorites, setFavorites] = useUserSettings<BucketFavoritesType>(
```
SanjalKatiyar, line 186: `useUserSettings` first stores items in a ConfigMap then as a fallback to a local-storage...

for our use case there can be 100s of buckets (probably 1000), so might be too much info for a single CM ??

should we use `useUserSettingsLocalStorage` only ??

cc @GowthamShanmugam @bipuladh @alfonsomthd wdyt ?

packages/odf/components/s3-browser/bucket/buckets-list-page/bucketsListPage.tsx
```diff
@@ -0,0 +1,250 @@
...
+};
+
+const BucketsListPage: React.FC = () => {
+  const { t } = useCustomTranslation();
+  const { noobaaS3 } = React.useContext(NoobaaS3Context);
+  const { data, error, isLoading } = useSWR(LIST_BUCKET, () =>
+    noobaaS3.listBuckets()
+  );
+  const [buckets, filteredData, onFilterChange] = useListPageFilter(
+    convertS3ListBucketOutputToK8sCompatible(data)
```
SanjalKatiyar, line 184: nits: name too long:P

packages/odf/components/s3-browser/bucket/buckets-list-page/bucketsListPage.tsx
```diff
@@ -0,0 +1,250 @@
...
+  return (
+    <NoobaaS3Provider loading={false}>
+      <BucketsListPage />
+    </NoobaaS3Provider>
+  );
+};
+
+type BucketFavoritesType = {
+  [key in string]: boolean;
+};
```
SanjalKatiyar, line 243: better to have a `Set` with all the buckets which are marked as favourite...

no need to store the once which are false...

packages/odf/components/s3-browser/bucket/buckets-list-page/bucketsListPage.tsx
```diff
@@ -0,0 +1,250 @@
...
+  const [buckets, filteredData, onFilterChange] = useListPageFilter(
+    convertS3ListBucketOutputToK8sCompatible(data)
+  );
+  const [favorites, setFavorites] = useUserSettings<BucketFavoritesType>(
+    BUCKET_BOOKMARKS_USER_SETTINGS_KEY,
+    undefined,
+    true
+  );
+
+  const isLoadeddWOAnyError = !isLoading && !error;
```
SanjalKatiyar, line 192:
```suggestion
  const isLoadedWOAnyError = !isLoading && !error;
```

packages/odf/components/s3-browser/bucket/buckets-list-page/bucketsListPage.tsx
alfonsomthd: The data stored in a ConfigMap cannot exceed 1 MiB. provided that even 1000 buckets do not reach that threshold, then it's a matter of consensus. Otherwise we should definitely not store it in a ConfigMap.

packages/shared/src/table/composable-table.tsx
```diff
@@ -67,6 +71,7 @@ export const ComposableTable: ComposableTableProps = <
         translate={null}
         aria-label="Composable table"
         className="pf-v5-u-mt-md"
+        variant={isCompact ? TableVariant.compact : null}
```
SanjalKatiyar, line 74: totally optional comment:
```suggestion
        {...(isCompact ? {variant: ableVariant.compact} : {})}
```

plugins/odf/console-extensions.json
```diff
@@ -509,9 +509,6 @@
       "section": "storage",
       "name": "Object Storage",
       "href": "/odf/object-storage"
-    },
-    "flags": {
-      "required": ["MCG"]
```
SanjalKatiyar, line 514: no need to remove this IMO...
why are we removing this ??

plugins/odf/console-extensions.json
```diff
@@ -587,22 +600,6 @@
...
-      "after": "namespace-store",
-      "href": "objectbucket.io~v1alpha1~ObjectBucket",
-      "component": {
-        "$codeRef": "ob.ObjectBucketListPage"
-      }
-    },
-    "flags": {
-      "required": ["MCG"]
-    }
-  },
```
SanjalKatiyar, line 605: we still need it for non-mcg-standalone deployments...

SanjalKatiyar: only time we don't need is for mcg-standalone specifically...

packages/odf/components/s3-browser/bucket/buckets-list-page/bucketsListPage.tsx
SanjalKatiyar: it's a global CM (shared by OCP and all its dynamic plugins), lot other things are stored there...

SanjalKatiyar: also, in future when we support more accounts we definitely will have to revert to local storage way only... because CM is still global...

packages/shared/src/list-page/paginated-list-page.tsx
```diff
@@ -20,21 +24,66 @@ const COUNT_PER_PAGE_NUMBER = 10;
 export type PaginatedListPageProps = {
   countPerPage?: number;
   filteredData: K8sResourceCommon[];
-  CreateButton: React.FC<unknown>;
```
SanjalKatiyar, line 23: it's a `shared` component, do we really need to change the name ??
need to check the impact (if any) on Fusion side, if we are planning a rename...

packages/odf/components/s3-browser/bucket/buckets-list-page/bucketsListPage.tsx
alfonsomthd: > it's a global CM (shared by OCP and all its dynamic plugins), lot other things are stored there...

Then definitely it's not a safe place to store that data.

locales/en/plugin__odf-console.json
```diff
@@ -1097,6 +1093,16 @@
   "NamespaceStore details": "NamespaceStore details",
   "Target Blob Container": "Target Blob Container",
   "Num Volumes": "Num Volumes",
+  "Storage endpoint": "Storage endpoint",
+  "Create on": "Create on",
+  "No buckets found": "No buckets found",
+  "Create bucket": "Create bucket",
+  "Create and manage your buckets": "Create and manage your buckets",
+  "Navigate through your buckets effortlessly. View the contents of your S3-managed and Openshift-managed buckets, making it easy to locate and inspect objects. Track and manage object versions and deleted items. View and restore previous versions of objects and access a history of deleted items, ensuring you can recover or review data as needed.": "Navigate through your buckets effortlessly. View the contents of your S3-managed and Openshift-managed buckets, making it easy to locate and inspect objects. Track and manage object versions and deleted items. View and restore previous versions of objects and access a history of deleted items, ensuring you can recover or review data as needed.",
```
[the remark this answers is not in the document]
GowthamShanmugam: ack

packages/odf/components/s3-browser/bucket/buckets-list-page/bucketsListPage.tsx
GowthamShanmugam: Ok! I thought the bucket count wouldn't cross more than 200, so I decided to use the same configMap. And the user won't select all 200 as fav.

packages/odf/components/s3-browser/bucket/buckets-list-page/bucketsListPage.tsx
GowthamShanmugam: ya

packages/shared/src/list-page/paginated-list-page.tsx
GowthamShanmugam: No problem. for now, I will keep the name CreateButton. I renamed this to make this paginated component generic to create any kind of list page between MCO and ODF.

plugins/odf/console-extensions.json
GowthamShanmugam: ok! i accidently removed this

packages/odf/components/s3-browser/bucket/buckets-list-page/bucketsListPage.tsx
GowthamShanmugam: But local storage will be a terrible user experience, The user will see different fav in different browsers. And what is they selected 500 buckets as fav and other browser displays nothing

GowthamShanmugam: I think this marking favorite is not a good idea here cc: @Chandan141996

SanjalKatiyar: make sense !!
let's confirm with OCP team on what they think.

Also, @GowthamShanmugam just a reminder that on a bucket deletion, we need to update this CM/LocalStorage as well.

GowthamShanmugam:
Screenshot 2024-10-03 at 9 25 38 PM

alfonsomthd:
> no need for extra "bucket" directory here: `/s3-browser/bucket/buckets-list-page/bucketsListPage.tsx`... we can directly do `packages/odf/components/s3-browser/buckets-list-page/bucketsListPage.tsx`
> 
> Also, `bucketsListPage.tsx` should be `BucketsListPage.tsx`...
> 
> IIRC that's what we decided, cc @alfonsomthd to keep me honest.

@SanjalKatiyar yes, we finally agreed  on this:
```
    └── s3-browser/
        └── create-bucket/    (CRUD-action (use case) folders)
         ...
```

packages/shared/src/table/composable-table.tsx
```diff
@@ -67,6 +71,7 @@ export const ComposableTable: ComposableTableProps = <
         translate={null}
         aria-label="Composable table"
```
alfonsomthd, line 72:
```suggestion
        aria-label={t('Composable table')}
```

GowthamShanmugam: ack

packages/shared/src/hooks/useUserSettingsLocalStorage.ts
```diff
@@ -0,0 +1,187 @@
...
+  // ODF settings base storage path
+  storageKey: string = DEFAULT_ODF_CONSOLE_USER_SETTINGS_KEY
+): [T, React.Dispatch<React.SetStateAction<T>>] => {
+  // Mount status for safty state updates
+  const mounted = React.useRef(true);
+  React.useEffect(() => () => (mounted.current = false), []);
+  const keyRef = React.useRef<string>(
+    userSettingsKey?.replace(/[^-._a-zA-Z0-9]/g, '_')
+  );
+  const defaultValueRef = React.useRef(defaultValue);
```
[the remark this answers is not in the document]
SanjalKatiyar: but maybe okay if this is copied exactly from console...

packages/shared/src/hooks/useUserSettingsLocalStorage.ts
```diff
@@ -0,0 +1,187 @@
...
+        dataRef.current,
+        keyRef.current,
+        setData
+      ),
+    [storageKey]
+  );
+
+  React.useEffect(() => {
+    if (sync) {
+      window.addEventListener('storage', syncLatestData);
```
SanjalKatiyar, line 154: nits: const STORAGE_EVENT = 'storage';
```suggestion
      window.addEventListener(STORAGE_EVENT, syncLatestData);
```

packages/odf/components/s3-browser/buckets-list-page/bucketsListPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+          {t('Create bucket')}
+        </ListPageCreateLink>
+      </ListPageHeader>
+      <div className="pf-v5-u-ml-lg pf-v5-u-mr-lg text-muted">
+        <Trans t={t}>
+          Browse, upload, and manage objects in buckets.
+          <ViewDocumentation
+            doclink={odfDocLinks(odfDocVersion).S3_BUCKET}
+            text={t('Help materials')}
+          />
```
SanjalKatiyar, line 93: we can skip it for now with a ToDo comment... if doc team will add something we can add it then... else we will get bugs that link is not working etc... wdyt ?? (this is an optional comment)
```suggestion
```

packages/odf/components/s3-browser/buckets-list-page/bucketsListPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+    <ListPageBody>
+      <Grid>
+        <GridItem md={11} sm={12} className="pf-v5-u-mt-md">
+          <div className="pf-v5-u-display-flex pf-v5-u-flex-direction-column pf-v5-u-flex-direction-row-on-md">
+            <ListPageFilter
+              loaded={true}
+              hideLabelFilter={true}
+              nameFilterPlaceholder={t('Search a bucket by name')}
+              data={getValidFilteredData(allBuckets)}
+              onFilterChange={onFilterChange}
```
SanjalKatiyar, line 43: if I type something in this search bar and click refresh, will that text refresh as well ?? It should...

packages/odf/components/s3-browser/buckets-list-page/bucketsListPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+      {fresh ? (
+        <BucketsListTable
+          allBuckets={allBuckets}
+          filteredBuckets={filteredBuckets}
+          loaded={loaded}
+          error={loadError}
+        />
+      ) : (
+        <LoadingBox />
+      )}
```
SanjalKatiyar, line 72: nits (optional comment): `StatusBox` inside `BucketsListTable` should handle the loading state in form of table skeleton...
```suggestion
      {fresh && (
        <BucketsListTable
          allBuckets={allBuckets}
          filteredBuckets={filteredBuckets}
          loaded={loaded}
          error={loadError}
        />
      )}
```

packages/odf/components/s3-browser/buckets-list-page/bucketsListPage.tsx
SanjalKatiyar, line 43: is there some prop to disable this search bar when it's `!fresh` ??

packages/odf/components/s3-browser/buckets-list-page/bucketListTable.tsx
```diff
@@ -0,0 +1,199 @@
...
+      />
+      <Td translate={null} dataLabel={columnNames[1]}>
+        <Link to={`${BUCKET_DETAILS_PAGE_PATH}/${name}`}>{name}</Link>
+      </Td>
+      <Td translate={null} dataLabel={columnNames[2]}>
+        {/* ToDo: Currently we only support MCG, make is configurable once RGW is supported as well */}
+        <Label color="gold">{t('MCG')}</Label>
+      </Td>
+      <Td translate={null} dataLabel={columnNames[3]}>
+        {<Timestamp timestamp={creationTimestamp} />}
```
SanjalKatiyar, line 148: I will suggest not to use `Timestamp` component here... this component displays data in `fromNow` format, ex: `1 minute ago`
This make sense when we have continuous polling or websockets... but here we have manual refresh, so this FC will display stale/incorrect data meanwhile...

SanjalKatiyar: plz just display simple string timestamp, similar to objects list page...

packages/odf/components/s3-browser/buckets-list-page/bucketListTable.tsx
```diff
@@ -0,0 +1,199 @@
...
+          {t('Erase the contents of your bucket')}
+        </p>
+      </>
+    ),
+    onClick: () => undefined,
+  },
+  {
+    title: t('Delete bucket'),
+    onClick: () => undefined,
+  },
```
SanjalKatiyar, line 37: Just adding for reference, good to add a ToDo comment... we need to clear corresponding localStorage when a bucket is deleted...

packages/odf/components/s3-browser/buckets-list-page/bucketListTable.tsx
```diff
@@ -0,0 +1,199 @@
...
+      ButtonComponent={() => <></>}
+      title={t('Create and manage your buckets')}
+      isLoaded
+      canAccess
+    >
+      <Trans t={t}>
+        Navigate through your buckets effortlessly. View the contents of your
+        S3-managed and Openshift-managed buckets, making it easy to locate and
+        inspect objects.
+      </Trans>
```
SanjalKatiyar, line 99: any reason why we need `Trans` here ??

packages/odf/components/s3-browser/buckets-list-page/bucketPagination.tsx
```diff
@@ -0,0 +1,87 @@
...
+    React.useState<ContinuationTokens>({
+      previous: [],
+      current: '',
+      next: '',
+    });
+
+  React.useEffect(() => {
+    setBucketInfo([
+      convertBucketDataToCrFormat(data),
+      !isMutating && !error,
```
SanjalKatiyar, line 37:
```suggestion
      !isMutating,
```

SanjalKatiyar:
> Screenshot 2024-10-03 at 9 25 38 PM

probably we can add a bit of margin to pagination icons, to make it aligned with search bar and refresh icon, wdyt ??

packages/odf/constants/s3-browser.ts
```diff
@@ -6,9 +6,15 @@ export const NOOBAA_SECRET_ACCESS_KEY = 'AWS_SECRET_ACCESS_KEY';
 export const DELIMITER = '/';
 export const PREFIX = 'prefix';
 export const MAX_KEYS = 300;
+export const MAX_BUCKETS = 300;
```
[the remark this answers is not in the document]
SanjalKatiyar: fell free to skip this comment for now maybe...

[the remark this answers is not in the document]
SanjalKatiyar: feel free to skip this comment for now...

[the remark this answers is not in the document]
GowthamShanmugam: ack, i just copy pasted the above number and forgot to change

packages/shared/src/table/composable-table.tsx
```diff
@@ -36,6 +37,8 @@ export const ComposableTable: ComposableTableProps = <
   unfilteredData,
   noDataMsg,
   emptyRowMessage,
+  isFavorites,
+  isCompact,
```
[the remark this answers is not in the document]
GowthamShanmugam: makesense

plugins/odf/console-extensions.json
GowthamShanmugam: Do we have a flag to check MCG only? i though this MCG flag is to decide MCG only

SanjalKatiyar: it is for MCG only, that correct...
there are other deployments like Full (internal/external) & Provider, ObjectBuckets tab should be visible for all... only mode which needs to hide it is MCG only mode...

SanjalKatiyar: all these other modes can/can-not have RGW...

packages/shared/src/hooks/useUserSettingsLocalStorage.ts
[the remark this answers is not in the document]
GowthamShanmugam: I think it is ok, may passing array sometimes trigger re-render. this will makesure, to read data from local storage only when it is required and no unnecessary re-rendering.

packages/odf/components/s3-browser/buckets-list-page/bucketsListPage.tsx
GowthamShanmugam: ok for me to remove

packages/odf/components/s3-browser/buckets-list-page/bucketsListPage.tsx
GowthamShanmugam: Filter and refresh are different, they can add/remove filters anytime. force removing filter every refresh is not a good user experience. i check AWS console also.

GowthamShanmugam: unfortunately there is no disable prop in ListPageFilter, all we can do is completely hide

packages/odf/components/s3-browser/buckets-list-page/bucketListTable.tsx
GowthamShanmugam: I think i added to reduce the length of the line in file.

packages/shared/src/hooks/useUserSettingsLocalStorage.ts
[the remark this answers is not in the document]
SanjalKatiyar: that maybe true for other fields... but we are not using `defaultValueRef` anywhere in useEffect/useMemo... but okay to keep it as it is if it's copied from console...

packages/odf/components/s3-browser/buckets-list-page/bucketsListPage.tsx
SanjalKatiyar: I am not asking for removing and re-adding filter... I meant contents of the search bar should reset as well, right ??
I type something and refresh, what will happen to the typed text ??

packages/odf/components/s3-browser/buckets-list-page/bucketListTable.tsx
SanjalKatiyar: okay, that's strange use of this component:)

package.json
```diff
@@ -59,7 +59,7 @@
     "dev:c": "yarn ocp-console & PLUGIN=${PLUGIN} I8N_NS=${I8N_NS} yarn server:plugin & wait"
   },
   "dependencies": {
-    "@aws-sdk/client-s3": "3.614.0",
+    "@aws-sdk/client-s3": "3.663.0",
```
GowthamShanmugam, line 62: bucket listing and object listing is working fine with this version

packages/odf/components/s3-browser/buckets-list-page/bucketsListPage.tsx
GowthamShanmugam: The filtering text will stay as it is.  After clicking the refresh button, the new list will be auto-filtered based on the filtered text. It's a correct behavior.

GowthamShanmugam: it's ok to keep it as it is, even though ListPageFilter is enabled, the table is anyway remounted. and results are auto-filtered. very similar behaviour to AWS console.

SanjalKatiyar: got it !!
I have another doubt here, `ListPageFilter` will by default add name and label filter dropdown... we need to support filter by name, but label here doesn't make sense... what are we doing regarding label option ??

SanjalKatiyar: got it !!

packages/shared/src/hooks/useUserSettingsLocalStorage.ts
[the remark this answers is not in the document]
GowthamShanmugam: right, i will remove this

packages/odf/components/s3-browser/buckets-list-page/bucketListTable.tsx
GowthamShanmugam: i replaced this with t()

GowthamShanmugam:
Screenshot 2024-10-08 at 4 31 00 PM

packages/shared/src/hooks/useUserSettingsLocalStorage.ts
```diff
@@ -0,0 +1,181 @@
+import * as React from 'react';
+
+// Default ODF user settings storage path
+const DEFAULT_ODF_CONSOLE_USER_SETTINGS_KEY = 'odf-console-user-settings';
+const STORAGE_EVENT = 'storage';
+
+export const deseralizeData = (data: string | null) => {
```
alfonsomthd, line 7:
```suggestion
export const deserializeData = (data: string | null) => {
```

packages/shared/src/hooks/useUserSettingsLocalStorage.ts
```diff
@@ -0,0 +1,181 @@
...
+    return data;
+  }
+  try {
+    return JSON.parse(data);
+  } catch {
+    return data;
+  }
+};
+
+export const seralizeData = <T>(data: T) => {
```
alfonsomthd, line 18:
```suggestion
export const serializeData = <T>(data: T) => {
```

packages/shared/src/hooks/useUserSettingsLocalStorage.ts
```diff
@@ -0,0 +1,181 @@
...
+  if (typeof data === 'string') {
+    return data;
+  }
+  return JSON.stringify(data);
```
alfonsomthd, line 22: You should a try-catch as [JSON.stringify can throw errors](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#exceptions) that we need to handle.

packages/shared/src/hooks/useUserSettingsLocalStorage.ts
```diff
@@ -0,0 +1,181 @@
...
+};
+
+// Get user settings from local storage
+const getDataFromLocalStorage = (storageKey: string) =>
+  deseralizeData(localStorage.getItem(storageKey)) ?? {};
+
+// Initialize state
+const intializeDataFromLocalStorage = <T>(
```
SanjalKatiyar, line 30:
```suggestion
const initializeDataFromLocalStorage = <T>(
```

packages/shared/src/hooks/useUserSettingsLocalStorage.ts
```diff
@@ -0,0 +1,188 @@
...
+    // eslint-disable-next-line no-console
+    console.error(
+      `Error while updating local storage for key ${storageKey}`,
+      err
+    );
+  }
+};
+
+// Generate storage event to sync latest settings on other tabs
+const generateStorageEventToUpdate = (storageKey, newValue: string) => {
```
SanjalKatiyar, line 100: IIRC this is not for other tabs, but for current itself, right ??
```suggestion
const generateStorageEventToUpdate = (storageKey, newValue: string) => {
```

SanjalKatiyar: in which case this comment is wrong...

SanjalKatiyar: no need to block the PR on this, but we can do it in next follow-up PR...

GowthamShanmugam: check note: https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event

SanjalKatiyar: got it !

SanjalKatiyar: Just FYI: https://github.com/red-hat-storage/odf-console/issues/1629

packages/shared/src/list-page/paginated-list-page.tsx
GowthamShanmugam: reverted back all paginated-list-page.tsx changes

PR #1608, opened 2024-10-03
Add 'Create Bucket' form (via S3-compat API)

alfonsomthd:
Screenshot from 2024-10-03 09-26-09
Screenshot from 2024-10-03 09-26-26


packages/odf/components/mcg/CreateObjectBucketClaim.tsx
```diff
@@ -367,10 +367,12 @@ export const CreateOBCForm: React.FC<CreateOBCFormProps> = (props) => {
 };
 
 type CreateOBCProps = {
+  formClassName?: string;
   showNamespaceSelector?: boolean;
 };
 
 export const CreateOBC: React.FC<CreateOBCProps> = ({
+  formClassName = '',
```
SanjalKatiyar, line 375: just an optional comment, why can't we just pass `className` itself ?? that way we don't have to be aware of a custom prop name `formClassName`.
```suggestion
  className = '',
```

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
```diff
@@ -0,0 +1,173 @@
...
+const CreateBucketForm: React.FC<{}> = () => {
+  const { t } = useCustomTranslation();
+  const navigate = useNavigate();
+  const [inProgress, setInProgress] = React.useState(false);
+  const [errorMessage, setErrorMessage] = React.useState('');
+  const [tagsData, setTagsData] = React.useState<string[][]>([]);
+  const { noobaaS3 } = React.useContext(NoobaaS3Context);
+  const { data, error, isLoading } = useSWR(LIST_BUCKET, () =>
+    noobaaS3.listBuckets()
+  );
```
SanjalKatiyar, line 42: I am not sure if it's a good idea to use this anymore. It might work for some of the cases, but definitely not all.

I agree to your older suggestion, we should let backend take care of the naming validation (already does), and UI will just display the error message.

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
```diff
@@ -0,0 +1,173 @@
...
+            .filter((pair: string[]) => !_.isEmpty(pair[0]))
+            .map((pair: string[]) => ({ Key: pair[0], Value: pair[1] }));
+        if (!_.isEmpty(tagSet)) {
+          await noobaaS3.updateBucketTags({
+            Bucket: bucketName,
+            Tagging: { TagSet: tagSet },
+          });
+        }
+      } catch ({ name, message }) {
+        // @TODO: add a toast warning that is visible after redirection to bucket details.
```
SanjalKatiyar, line 80: We are not implementing toast pop-up this release for any epic.

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
```diff
@@ -0,0 +1,173 @@
...
+          tagsData
+            .filter((pair: string[]) => !_.isEmpty(pair[0]))
+            .map((pair: string[]) => ({ Key: pair[0], Value: pair[1] }));
+        if (!_.isEmpty(tagSet)) {
+          await noobaaS3.updateBucketTags({
+            Bucket: bucketName,
+            Tagging: { TagSet: tagSet },
+          });
+        }
+      } catch ({ name, message }) {
```
SanjalKatiyar, line 79: I don't think we need another try/catch or an if/else on `errorMessage`.

This entire updating tag call can be moved inside existing `try`, just after `await noobaaS3.createBucket({ Bucket: bucketName });`.

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
```diff
@@ -0,0 +1,173 @@
...
+        // @TODO: add a toast warning that is visible after redirection to bucket details.
+        // eslint-disable-next-line no-console
+        console.error(`Error while updating bucket tags: ${name}: ${message}`);
+      }
+    }
+
+    if (errorMessage) {
+      setInProgress(false);
+      return;
+    }
```
SanjalKatiyar, line 89: This can be moved inside existing (and only one) `catch` block.

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
SanjalKatiyar: Also, check like this `if (!errorMessage) {` (immediately after catch) on `errorMessage` will not work.

Condition `if (!errorMessage) {` will get executed before even changes of `setErrorMessage('Error while creating bucket: ${name}: ${message}')` are actually reflected.

packages/shared/src/s3/commands.ts
```diff
@@ -22,9 +29,15 @@ export class S3Commands {
   }
 
   // Bucket command members
+  createBucket: CreateBucket = (input) =>
+    this.s3Client.send(new CreateBucketCommand(input));
+
   listBuckets: ListBuckets = (input) =>
     this.s3Client.send(new ListBucketsCommand(input));
 
+  updateBucketTags: UpdateBucketTags = (input) =>
```
SanjalKatiyar, line 38: it will be easier to find, if we align the name of the member function with the actual S3 command:)
```suggestion
  putBucketTags: PutBucketTags = (input) =>
```

SanjalKatiyar: or is there any specific reason for naming it differently ??

packages/shared/src/utils/NameValueEditor.tsx
```diff
@@ -305,19 +309,25 @@ export const NameValueEditor: React.FC<NameValueEditorProps> =
...
-            <div className={classNames('col-xs-5', className)}>
-              {nameStringUpdated}
-            </div>
-            <div className={classNames('col-xs-5', className)}>
-              {valueStringUpdated}
-            </div>
-            <div className="col-xs-1 co-empty__header" />
-          </div>
-          {pairElems}
+          {hideHeaderWhenNoItems && _.isEmpty(pairElems) ? (
```
SanjalKatiyar, line 312: we are already passing `hideHeaderWhenNoItems`, why not add `_.isEmpty(pairElems)` condition is the same as well ??

`const hideHeaderWhenNoItems = _.isEmpty(pairElems) && OTHER_CONDITIONS;` and pass `hideHeaderWhenNoItems` as a prop.

SanjalKatiyar: `_.isEmpty(pairElems)` will only execute anyway when `hideHeaderWhenNoItems` is true...

SanjalKatiyar:
> Screenshot from 2024-10-03 09-26-09

Isn't the selected tile supposed to get expanded (with blue line on bottom) ?? Here looks like when "s3" option is selected, tile is shrinking instead, do we need some CSS fix ??

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
alfonsomthd: I know, but I want to keep that comment because that should be the right way to inform the user, not a `console.error`...

SanjalKatiyar: If there is any error u should see it on same "Create" page (we are using setErrorMessage, right ?), we won't be navigating in case of error.
We will only navigate in case of success.

SanjalKatiyar: that's how any existing ODF "Create" pages work...

SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/1608#discussion_r1785823360 will resolve this issue...

packages/shared/src/s3/commands.ts
alfonsomthd: I wanted to align to CRUD naming (hence `Update`). Are not we following it?

packages/shared/src/utils/NameValueEditor.tsx
alfonsomthd: The intention is not to alter the current behavior: so we need a flag for the new behavior (hiding the header) and `isEmpty` is the condition to check on the new behavior: it's been tested.

alfonsomthd:
> > Screenshot from 2024-10-03 09-26-09
> 
> Isn't the selected tile supposed to get expanded (with blue line on bottom) ?? Here looks like when "S3" option is selected, tile is shrinking instead, do we need some CSS fix ?? Maybe I am wrong, just confirming both "OBC" and "S3" tiles have similar look 'n' feel when selected. If that's the case we are good, u can ignore this comment.

The behavior is the [PF one](https://www.patternfly.org/components/tile/):

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
alfonsomthd: No, an error in teh tagging command shouldn't prevent navigation because the bucket was created successfully:
this is a secondary operation and thta is why we should infor the user but not preventing teh redirection.

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
alfonsomthd: Refactored but kept 2 separate try-catch blocks as we need to differentiate bucket creation errors (blocking) vs tag creation errors (non-blocking). That way we don't need to implement logic to check for error types.

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
alfonsomthd: Answered bove.

packages/shared/src/s3/commands.ts
SanjalKatiyar: idea was to keep name similar to it's corresponding S3 command, just camel case and removing unnecessary postfix like `Command`...
even class name is `S3Commands`...

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
```diff
@@ -0,0 +1,171 @@
...
+          isInline
+          title={t('Address form errors to proceed')}
+        />
+      )}
+      <ButtonBar errorMessage={errorMessage} inProgress={inProgress}>
+        <ActionGroup className="pf-v5-c-form">
+          <Button
+            id="create-s3-bucket-btn"
+            type="submit"
+            variant="primary"
```
SanjalKatiyar, line 153: can be addressed in followup PR...
```suggestion
            variant={ButtonVariant.primary}
```

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
```diff
@@ -0,0 +1,171 @@
...
+            data-test="obc-create"
+          >
+            {t('Create')}
+          </Button>
+          <Button
+            onClick={() => navigate(-1)}
+            type="button"
+            variant="secondary"
```
SanjalKatiyar, line 161: can be addressed in followup PR...
```suggestion
            variant={ButtonVariant.secondary}
```

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
```diff
@@ -0,0 +1,171 @@
...
+          variant="danger"
+          isInline
+          title={t('Address form errors to proceed')}
+        />
+      )}
+      <ButtonBar errorMessage={errorMessage} inProgress={inProgress}>
+        <ActionGroup className="pf-v5-c-form">
+          <Button
+            id="create-s3-bucket-btn"
+            type="submit"
```
SanjalKatiyar, line 152:
```suggestion
            type={ButtonType.submit}
```

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
```diff
@@ -0,0 +1,171 @@
...
+            variant="primary"
+            data-test="obc-create"
+          >
+            {t('Create')}
+          </Button>
+          <Button
+            onClick={() => navigate(-1)}
+            type="button"
```
SanjalKatiyar, line 160:
```suggestion
            type={ButtonType.button}
```

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
```diff
@@ -0,0 +1,171 @@
...
+    try {
+      await noobaaS3.createBucket({ Bucket: bucketName });
+    } catch ({ name, message }) {
+      bucketCreationError = `Error while creating bucket: ${name}: ${message}`;
+    }
+
+    if (bucketCreationError) {
+      setInProgress(false);
+      setErrorMessage(bucketCreationError);
+      return;
```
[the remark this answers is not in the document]
alfonsomthd: I agree.

packages/odf/components/s3-browser/create-bucket/CreateBucket.tsx
```diff
@@ -1,5 +1,7 @@
 import * as React from 'react';
 import { CreateOBC } from '@odf/core/components/mcg/CreateObjectBucketClaim';
+import CreateBucketForm from '@odf/core/components/s3-browser/create-bucket/CreateBucketForm';
```
GowthamShanmugam, line 3: no issues in this import,  it just looks very longer than './create-bucket/CreateBucketForm';
it just need an import from its own directory

packages/odf/components/s3-browser/create-bucket/CreateBucket.tsx
```diff
@@ -1,5 +1,7 @@
...
+import { NoobaaS3Provider } from '@odf/core/components/s3-browser/noobaa-context';
```
GowthamShanmugam, line 4: same here

packages/odf/components/s3-browser/create-bucket/CreateBucket.tsx
GowthamShanmugam: i can see all the files have following similar way of import, so its ok to skip this comment

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
GowthamShanmugam, line 66: This part can be a part of the catch block right? if move this inside catch we can remove bucketCreationError constant also I guess

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
```diff
@@ -0,0 +1,171 @@
...
+  });
+
+  const save = async (formData: FormData) => {
+    setInProgress(true);
+    let bucketCreationError = '';
+    setErrorMessage(bucketCreationError);
+    const { bucketName } = formData;
+    try {
+      await noobaaS3.createBucket({ Bucket: bucketName });
+    } catch ({ name, message }) {
```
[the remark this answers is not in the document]
alfonsomthd: Both are props of the [Error object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#instance_properties).

packages/odf/components/s3-browser/create-bucket/CreateBucketForm.tsx
alfonsomthd: Already done in https://github.com/red-hat-storage/odf-console/pull/1612

packages/odf/components/s3-browser/create-bucket/CreateBucket.tsx
alfonsomthd: IMHO we should favor absolute imports (vscode intellisense does it) vs relative imports due to the confusion created by the relative path hell ('../../../../../../../../../../../oh-my.ts')

PR #1617, opened 2024-10-07
Hide In-transit Encryption checkbox when MCG deployment mode is selected

malayparida2000:
https://issues.redhat.com/browse/RHSTOR-6552

The encryption in transit option is based on Ceph’s MSGR v2 protocol and is meant for traffic interacting with & within the ceph cluster (client-osd, osd-osd). MCG is not covered by this, so it doesn't make sense to show the option when deployment mode is MCG Only.


malayparida2000:
Screenshot 2024-10-07 at 4 56 58 PM
Tested the changes in cluster, Checkbox is now hidden.

bipuladh (approved):
/lgtm

PR #1626, opened 2024-10-08
Fix bundle analyzer script

GowthamShanmugam:
After the @aws-sdk/client-s3 version update, the stats.json file size for the bundle analysis became around 545M, Node has a string size limit of 500M and the current s.readFileSync(...) logic to read stats.json files fails for files >500MB. So I switched to streaming. Ref: https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/423

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,6 +128,7 @@
   },
   "devDependencies": {
     "@cypress/webpack-preprocessor": "^5.9.1",
+    "@discoveryjs/json-ext": "0.6.1",
     "@swc/core": "1.3.19",
     "@swc/jest": "^0.2.23",
     "@testing-library/jest-dom": "^6.4.5",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -124,7 +124,8 @@
     "webpack": "5.94.0",
     "webpack-bundle-analyzer": "^4.10.2",
     "webpack-cli": "4.5.x",
-    "yup": "^0.32.11"
+    "yup": "^0.32.11",
+    "@discoveryjs/json-ext": "0.6.1"
```
alfonsomthd, line 128: This should be placed in `devDependencies` as it's not required for building (right after `@cypress/webpack-preprocessor`)

alfonsomthd (changes requested):

alfonsomthd (approved):

PR #1635, opened 2024-10-10
Add support for file uploads in S3 browser

bipuladh:
[screen-capture.webm](https://github.com/user-attachments/assets/a4068639-c746-4497-8f4d-8a89ab824d6a)


packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
```diff
@@ -187,14 +187,7 @@ const BucketOverview: React.FC<{}> = () => {
...
         className="pf-v5-u-mt-md"
       />
-      {fresh ? (
-        <HorizontalNav
-          pages={navPages}
-          resource={isCreatedByOBC && noobaaObjectBucket}
-        />
-      ) : (
-        <LoadingBox />
-      )}
```
SanjalKatiyar, line 197: We have multiple tabs under this HorizontalNav (Details, Objects, YAML, more in future)... idea was to mount/un-mount entire HorizontalNav on "refresh" rather than handling it for each tab again and again.

It's fine IMO if refreshing is removing the sidebar and cancelling uploading, that's the intended affect of manual refreshing. User is intentionally clicking on that button.

packages/odf/components/s3-browser/objects-list/ObjectListWithSidebar.tsx
```diff
@@ -0,0 +1,55 @@
...
+};
+
+export const ObjectListWithSidebar: React.FC<ObjectListWithSidebarProps> = ({
+  obj: { refresh },
+}) => {
+  const [isExpanded, setExpanded] = React.useState(false);
+  const [objects, setObjects] = React.useState<UploadProgress>({});
+
+  const abortAll = React.useCallback(async () => {
+    const aborters = Object.keys(objects).map((obj) => objects?.[obj]?.abort);
```
SanjalKatiyar, line 22:
```suggestion
    const aborters = Object.keys(objects).map((obj) => object[obj]?.abort);
```

locales/en/plugin__odf-console.json
```diff
@@ -1153,6 +1153,25 @@
   "Share with presigned URL": "Share with presigned URL",
   "No objects found": "No objects found",
   "You do not have any objects in the bucket": "You do not have any objects in the bucket",
+  "Add Objects": "Add Objects",
+  "Drag and drop files/folders here.": "Drag and drop files/folders here.",
+  "Standard uploads have a size limit of up to 5GB in S3. For objects, multipart upload will upload the object in parts, which are assembled in the bucket.": "Standard uploads have a size limit of up to 5GB in S3. For objects, multipart upload will upload the object in parts, which are assembled in the bucket.",
```
SanjalKatiyar, line 1158: It's better either to remove this statement to avoid confusion or make it 5 TB... AFAIK `UploadCommand` has a limit in TBs... `PutObjectCommand` has a 5 GB limit. 
```suggestion
 "For objects, multipart upload will upload the object in parts, which are assembled in the bucket.": "For objects, multipart upload will upload the object in parts, which are assembled in the bucket.",
```

packages/shared/src/s3/commands.ts
```diff
@@ -18,11 +18,9 @@ import {
   DeleteObjects,
 } from './types';
 
-export class S3Commands {
-  private s3Client: S3Client;
-
+export class S3Commands extends S3Client {
   constructor(endpoint: string, accessKeyId: string, secretAccessKey: string) {
-    this.s3Client = new S3Client({
+    super({
```
SanjalKatiyar, line 23: nice !

locales/en/plugin__odf-console.json
```diff
@@ -1153,6 +1153,25 @@
...
+  "Upload": "Upload",
+  "Uploads": "Uploads",
+  "{{uploadedFiles}} of {{totalFiles}} files uploaded&nbsp;": "{{uploadedFiles}} of {{totalFiles}} files uploaded&nbsp;",
+  "Succeeded: {{uploadedFiles}}": "Succeeded: {{uploadedFiles}}",
+  "Failed files: {{failedFiles}}": "Failed files: {{failedFiles}}",
+  "Completion time: {{totalTimeElasped}}": "Completion time: {{totalTimeElasped}}",
+  "Total Remaining: {{totalReamining}}": "Total Remaining: {{totalReamining}}",
+  "Estimated time remaining: {{timeRemaining}}": "Estimated time remaining: {{timeRemaining}}",
+  "Transfer rate: {{uploadSpeed}}": "Transfer rate: {{uploadSpeed}}",
+  "Standard uploads have a size limit of up to 5 GB in S3. For large objects, something or the other will happen. We need to improve this alert. Or not show this. Doesnt make sense to repeat your messages in UI.": "Standard uploads have a size limit of up to 5 GB in S3. For large objects, something or the other will happen. We need to improve this alert. Or not show this. Doesnt make sense to repeat your messages in UI.",
```
SanjalKatiyar, line 1168: same as https://github.com/red-hat-storage/odf-console/pull/1635/files#r1795747673
```suggestion
  "For large objects, something or the other will happen. We need to improve this alert. Or not show this. Doesnt make sense to repeat your messages in UI.": "For large objects, something or the other will happen. We need to improve this alert. Or not show this. Doesnt make sense to repeat your messages in UI.",
```

packages/odf/components/s3-browser/upload-objects/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+  }, 0);
+  return [totalUploaded, totalObjects];
+};
+
+export const getUploadSpeed = (objects: UploadProgress): string => {
+  if (_.isEmpty(objects)) return '';
+  const files = Object.values(objects);
+  const uploadingFile = files.filter(
+    (item) => item.uploadState === UploadStatus.UPLOAD_START
+  )?.[0];
```
SanjalKatiyar, line 27: Just a doubt: is it worthwhile to take average of speeds of all files with `item.uploadState === UploadStatus.UPLOAD_START`, rather than always selecting the first (0th index) one ??

packages/odf/components/s3-browser/upload-objects/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+  const totalTime = Date.now() - (uploadingFile?.startTime || Date.now());
+  const totalUploaded = uploadingFile?.loaded ?? 0;
+  const speed = humanizeDecimalBytesPerSec(totalUploaded / (totalTime / 1000));
+  return speed.string;
+};
+
+export const getTotalRemainingFilesAndSize = (
```
SanjalKatiyar, line 34: "TotalRemaining" could include failed and the ones which have not started yet as well... hence name is bit confusing IMO...
```suggestion
export const getTotalUploadingFilesAndSize = (
```

packages/odf/components/s3-browser/upload-objects/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+  );
+  const filesCount = uploadingFiles.length;
+  const filesSize = uploadingFiles.reduce(
+    (acc, curr) => (acc += curr.total - curr.loaded),
+    0
+  );
+  return `${filesCount} files (${humanizeBinaryBytes(filesSize).string})`;
+};
+
+export const getTotalTimeReamining = (objects: UploadProgress): string => {
```
SanjalKatiyar, line 50: same as https://github.com/red-hat-storage/odf-console/pull/1635/files#r1795811268
```suggestion
export const getTotalUploadTimeRemaining = (objects: UploadProgress): string => {
```

packages/odf/components/s3-browser/upload-objects/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+  if (_.isEmpty(objects)) return '';
+  let files = Object.values(objects);
+  const uploadingFile = files.filter(
+    (item) => item.uploadState === UploadStatus.UPLOAD_START
+  )?.[0];
+  const totalTime = Date.now() - (uploadingFile?.startTime || Date.now());
+  const totalUploaded = uploadingFile?.loaded ?? 0;
+
+  if (_.isEmpty(objects)) return '';
+  files = Object.values(objects);
```
SanjalKatiyar, line 60: repeated lines, already handled before...
```suggestion
```

packages/odf/components/s3-browser/upload-objects/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+};
+
+export const getTotalTimeReamining = (objects: UploadProgress): string => {
+  if (_.isEmpty(objects)) return '';
+  let files = Object.values(objects);
+  const uploadingFile = files.filter(
+    (item) => item.uploadState === UploadStatus.UPLOAD_START
+  )?.[0];
+  const totalTime = Date.now() - (uploadingFile?.startTime || Date.now());
+  const totalUploaded = uploadingFile?.loaded ?? 0;
```
SanjalKatiyar, line 57: don't we have `getCompletedAndTotalUploadObjects` for calculating `totalUploaded` value ??

packages/odf/components/s3-browser/upload-objects/upload.scss
```diff
@@ -0,0 +1,40 @@
...
+    width: 95%;
+    margin: 10px;
+    padding: 5px;
+}
+
+.odf-upload__title {
+    border: 1px dotted var(--pf-v5-global--palette--black-500);
+    padding: 15px;
+    margin-top: 5px;
+}
```
SanjalKatiyar, line 11: can we plz use `rem` instead of `px` ?

SanjalKatiyar:
`packages/odf/components/s3-browser/upload-objects/index.ts` is an empty file...

packages/odf/components/s3-browser/upload-objects/utils.ts
SanjalKatiyar: or if we wanna call it `getTotalRemainingFilesAndSize `, then include `INIT_STATE` ones as well ??

packages/odf/components/s3-browser/upload-objects/utils.ts
SanjalKatiyar: or if we wanna call it `getTotalTimeReamining `, then include `INIT_STATE` as well ??

packages/odf/components/s3-browser/objects-list/ObjectListWithSidebar.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <UploadSidebar
+      isExpanded={isExpanded}
+      closeSidebar={() => setExpanded(false)}
+      progress={objects}
+      mainContent={
+        <>
+          <FileUploadComponent
+            client={noobaaS3}
```
SanjalKatiyar, line 41: nits (optional): given later we will add `rgwClient` as well...
```suggestion
            noobaaClient={noobaaS3}
```

packages/odf/components/s3-browser/upload-objects/uploads.ts
```diff
@@ -0,0 +1,82 @@
...
+  ): ReturnType<typeof Upload.prototype.done> => {
+    const uploader = new Upload({
+      client: client,
+      params: {
+        Bucket: bucketName,
+        Key: file.name,
+        Body: file,
+      },
+      queueSize: 4,
+    });
```
SanjalKatiyar, line 24: let's pass `contentType` param as well, it's a MIME type of the uploaded file (https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#upload-property)... should be reported by `file: File` object...

SanjalKatiyar:
```
If you do not set the ContentType, S3 will treat the file as binary/octet-stream, which might lead to improper handling of the file on download or access. It is always a best practice to specify the appropriate ContentType when uploading files to S3.
```

SanjalKatiyar: `If you do not set the ContentType, S3 will treat the file as binary/octet-stream, which might lead to improper handling of the file on download or access. It is always a best practice to specify the appropriate ContentType when uploading files to S3.`

SanjalKatiyar: check: https://github.com/aws/aws-sdk-js/issues/3085

packages/odf/components/s3-browser/objects-list/ObjectListWithSidebar.tsx
```diff
@@ -0,0 +1,55 @@
...
+            bucketName={bucketName}
+            uploadProgress={objects}
+            setUploadProgress={setObjects}
+            showSidebar={() => setExpanded(true)}
```
SanjalKatiyar, line 45: `showSidebar` is added to deps list of `useCallback` later, so better to conserve its ref here before passing it as a prop...

packages/odf/components/s3-browser/upload-objects/uploads.ts
```diff
@@ -0,0 +1,82 @@
...
+  setUploadProgress: React.Dispatch<React.SetStateAction<UploadProgress>>
+): Generator<unknown, unknown, ReturnType<typeof Upload.prototype.done>[]> {
+  const performUploadPromise = (
+    file: File
+  ): ReturnType<typeof Upload.prototype.done> => {
+    const uploader = new Upload({
+      client: client,
+      params: {
+        Bucket: bucketName,
+        Key: file.name,
```
SanjalKatiyar, line 20: u need to get current `foldersPath` (which signifies ur current folder inside the bucket) and append that to the file name... this will "logically" or "virtually" make sure that these files are uploaded inside a selected folder...
```suggestion
        Key: getPrefix(file.name, foldersPath),
```

packages/odf/components/s3-browser/upload-objects/FileUploadComponent.tsx
```diff
@@ -0,0 +1,144 @@
...
+  abortAll: () => void;
+  totalObjects?: number;
+  totalCompletedObjects?: number;
+};
+
+export const FileUploadComponent: React.FC<FileUploadComponentProps> = ({
+  client,
+  bucketName,
+  uploadProgress: totalUploadProgress,
+  setUploadProgress: setTotalUploadProgress,
```
SanjalKatiyar, line 29: we are passing `objects & setObjects` from `ObjectListWithSidebar` FC, we are receiving it in `uploadProgress & setUploadProgress` variables, then renaming it to `totalUploadProgress & setTotalUploadProgress`... bit confusing:)

packages/odf/components/s3-browser/upload-objects/uploads.ts
SanjalKatiyar: we need `file.webkitRelativePath` to make sure a folder is uploaded when selected by user...

packages/odf/components/s3-browser/upload-objects/FileUploadComponent.tsx
SanjalKatiyar: then again later it's passed onto `processFiles` function as `setProgress`... which then receives it as `setUploadProgress`...

so many names for same variables !!

packages/odf/components/s3-browser/upload-objects/FileUploadComponent.tsx
```diff
@@ -0,0 +1,144 @@
...
+          <UploadStatusBasedAlert
+            status={uploadStatus}
+            totalCompletedObjects={totalCompletedObjects}
+            showSidebar={showSidebar}
+            totalObjects={totalObjects}
+            closeAlert={closeAlert}
+            abortAll={abortAll}
+          />
+        )}
+        <Title headingLevel="h4">{t('Add Objects')}</Title>
```
SanjalKatiyar, line 105:
```suggestion
        <Title headingLevel="h4">{t('Add objects')}</Title>
```

SanjalKatiyar: there is fieldLevelHelp popup here as well (in the UX mocks)...

packages/odf/components/s3-browser/upload-objects/FileUploadComponent.tsx
```diff
@@ -0,0 +1,144 @@
...
+                <FlexItem>
+                  <Flex>
+                    <FlexItem>
+                      <Icon color="blue">
+                        <UploadIcon />
+                      </Icon>
+                    </FlexItem>
+                    <FlexItem>
+                      <Title headingLevel="h5">
+                        {t('Drag and drop files/folders here.')}
```
SanjalKatiyar, line 120:
```suggestion
                        {t('Drag and drop files/folders here or upload.')}
```

packages/odf/components/s3-browser/upload-objects/types.ts
```diff
@@ -0,0 +1,19 @@
+export enum UploadStatus {
+  INIT_STATE,
+  UPLOAD_COMPLETE,
+  UPLOAD_START,
+  UPLOAD_FAILED_PARTIALLY,
```
SanjalKatiyar, line 5: we are not using `UPLOAD_FAILED_PARTIALLY` anywhere, right ??

packages/odf/components/s3-browser/upload-objects/FileUploadComponent.tsx
```diff
@@ -0,0 +1,144 @@
...
+  });
+
+  const closeAlert = () => {
+    setUploadStatus(UploadStatus.INIT_STATE);
+  };
+  return (
+    <>
+      <div className="odf-upload-component">
+        {(uploadStatus === UploadStatus.UPLOAD_START ||
+          uploadStatus === UploadStatus.UPLOAD_COMPLETE) && (
```
SanjalKatiyar, line 95: What if all objects failed, we don't want to show any error/alert ??

packages/odf/components/s3-browser/upload-objects/UploadSidebar.tsx
```diff
@@ -0,0 +1,274 @@
+import * as React from 'react';
+import { Status, useCustomTranslation } from '@odf/shared';
+import { humanizeBinaryBytes } from '@odf/shared/utils';
+import { ResourceStatus } from '@openshift-console/dynamic-plugin-sdk';
+import { Trans } from 'react-i18next';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import AutoSizer from 'react-virtualized-auto-sizer';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { FixedSizeList as List } from 'react-window';
```
SanjalKatiyar, line 9: I am assuming we haven't added these to dependencies yet, instead suppressing linter errors because this is a WIP PR... we will clean this up when we are open for merging...

packages/odf/components/s3-browser/upload-objects/UploadSidebar.tsx
```diff
@@ -0,0 +1,274 @@
...
+};
+
+const getProgressVariant = (
+  state: UploadStatus
+): Progress['props']['variant'] => {
+  switch (state) {
+    case UploadStatus.UPLOAD_FAILED:
+      return 'danger';
+    case UploadStatus.UPLOAD_COMPLETE:
+      return 'success';
```
SanjalKatiyar, line 145:
```suggestion
      return ProgressVariant.danger;
    case UploadStatus.UPLOAD_COMPLETE:
      return ProgressVariant.success;
```

packages/odf/components/s3-browser/upload-objects/UploadSidebar.tsx
```diff
@@ -0,0 +1,274 @@
...
+              </>
+            )}
+          </Flex>
+          <DrawerActions>
+            <DrawerCloseButton onClick={onClose} />
+          </DrawerActions>
+        </DrawerHead>
+      </div>
+      <DrawerContentBody>
+        <Alert title="How uploading works?" isInline variant="info">
```
SanjalKatiyar, line 243:
```suggestion
        <Alert title="How uploading works?" isInline variant={AlertVariant.info}>
```

packages/odf/components/s3-browser/upload-objects/UploadSidebar.tsx
```diff
@@ -0,0 +1,274 @@
...
+          <Trans t={t}>
+            Standard uploads have a size limit of up to 5 GB in S3. For large
+            objects, something or the other will happen. We need to improve this
+            alert. Or not show this. Doesnt make sense to repeat your messages
+            in UI.
+          </Trans>
+        </Alert>
+        <UploadStatusList progress={Object.values(progress)} />
```
SanjalKatiyar, line 251: nits: some `pf-v5-u-mt-sm` between `UploadStatusList` and `Alert` components...

packages/odf/components/s3-browser/upload-objects/UploadStatusBasedAlert.tsx
```diff
@@ -0,0 +1,62 @@
...
+  showSidebar,
+  abortAll,
+  totalCompletedObjects,
+  totalObjects,
+  closeAlert,
+}) => {
+  const { t } = useCustomTranslation();
+  return (
+    <Alert
+      variant={UploadStatus.UPLOAD_START === status ? undefined : 'success'}
```
SanjalKatiyar, line 28:
```suggestion
      variant={UploadStatus.UPLOAD_START === status ? undefined : AlertVariant.success}
```

packages/odf/components/s3-browser/upload-objects/uploads.ts
```diff
@@ -0,0 +1,82 @@
...
+    while (true) {
+      const completed = yield Promise.all(allUploadPromise);
+      const totallyCompleted = completed.every(async (p) => {
+        try {
+          const done = await p;
+          return done;
+        } catch {
+          return true;
+        }
+      });
```
SanjalKatiyar, line 61: better to not block other uploads, when anyone of them fails... plz use `Promise.allSettled` instead of `Promise.all` :)
```suggestion
      const completed = yield Promise.allSettled(allUploadPromise);
      const totallyCompleted = completed.every((result) => result.status === 'fulfilled');
```

packages/odf/components/s3-browser/upload-objects/uploads.ts
```diff
@@ -0,0 +1,82 @@
...
+            : UploadStatus.UPLOAD_START,
+        },
+      }));
+    });
+    return uploader.done();
+  };
+  try {
+    const allUploadPromise = files.map((file) => performUploadPromise(file));
+    while (true) {
+      const completed = yield Promise.all(allUploadPromise);
```
SanjalKatiyar, line 53: shouldn't this `yield` outside `while (true) {` loop @bipuladh ??

because when `new Promise((resolve) => setTimeout(resolve, 1000)` will execute, it will start the while loop all again, this would mean your `Promise.all` (or `Promise.allSettled`) will be executed again as well ??

SanjalKatiyar, line 53: I don't think yield outputs anything which can be stored in a variable:
```
function* generator() {
  const t = yield 1;
  console.log("test: " + t);
}

const gen = generator();

console.log(gen.next().value);
console.log(gen.next().value);

// OUTPUT:
> 1
> "test: undefined"
> undefined
```

SanjalKatiyar, line 53: value of `completed` will be `undefined`... u can't store output of yield in a variable like this, example:
```
function* generator() {
  const t = yield 1;
  console.log("test: " + t);
}

const gen = generator();

console.log(gen.next().value);
console.log(gen.next().value);


// OUTPUT:
Reset
> 1
> "test: undefined"
> undefined
```

SanjalKatiyar: if u want to pass promises to `completed`, u need to do it from the caller of the generator, and pass it via `next`...
example: `uploadGenerator.next(promises)`

SanjalKatiyar:
Few things:
1. We should refresh tokens (used for pagination) after upload is complete, it's a good practice to follow to avoid undesirable results.
2. Need to add "Upload" button here as well: https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/s3-browser/objects-list/table-components.tsx#L202 (as per the UX mocks).

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
bipuladh: We are refreshing because the list of object is not being watched. However the details page and YAML wrapper should be using watchResource so there is no need to refresh the whole page. The user might want to refresh the table only to see the files being shown when the upload is going on especially in cases of large files being uploaded.

locales/en/plugin__odf-console.json
bipuladh: This is currently as it stands in the mockups. Let's update the text when the microcopy is ready.

locales/en/plugin__odf-console.json
bipuladh: This is currently as it stands in the mockups. Let's update the text when the microcopy is ready.

SanjalKatiyar: 5GB is more of a technical detail, microcopy team wouldn't know about it unless we tell them specifically... it's safer to remove than displaying incorrect info :(

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
SanjalKatiyar: Details tab and other tabs (expect YAML) **won't be** refreshing as well. We are not using web-sockets on these pages. "Refresh" button refreshes everything (even details and future tabs).

SanjalKatiyar: We are only using `watchResource` at a single place, i.e., for reading OB CRs... other than that all are REST HTTPs calls...
`useSWR` might revalidate data from time to time, but currently we were not using continous polling (because designs had "Refresh" option)... so clicking "Refresh" should mount/un-mount entire HorizontalNav (which includes Details tab as well).

SanjalKatiyar: we might consider using continuous polling for Details tab, but as per designs this "Refresh" is on top of the page which refreshes everything, not just objects list.

packages/odf/components/s3-browser/upload-objects/upload.scss
bipuladh: using standard pf variables

packages/odf/components/s3-browser/upload-objects/FileUploadComponent.tsx
bipuladh: Even if all fails uploadStatus should go to UPLOAD_COMPLETE in the end

packages/odf/components/s3-browser/objects-list/ObjectListWithSidebar.tsx
bipuladh: Let's introduce it when we actually introduce RGW. This would be a minor refactor. we could look at other ways to use the same object as well.

packages/odf/components/s3-browser/upload-objects/FileUploadComponent.tsx
bipuladh: added

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
```diff
@@ -187,14 +187,10 @@ const BucketOverview: React.FC<{}> = () => {
...
+      <HorizontalNav
+        pages={navPages}
+        resource={{ refresh: fresh, triggerRefresh } as any}
```
SanjalKatiyar, line 192:
```suggestion
        resource={{ refresh: fresh, triggerRefresh, noobaaObjectBucket } as any}
```

packages/shared/src/s3/commands.ts
```diff
@@ -36,24 +34,23 @@ export class S3Commands {
...
-    this.s3Client.send(new PutBucketTaggingCommand(input));
-
-  // Object command members (alphabetical order)
-  deleteObjects: DeleteObjects = (input) =>
-    this.s3Client.send(new DeleteObjectsCommand(input));
+    this.send(new PutBucketTaggingCommand(input));
 
-  getObject: GetObject = (input) =>
-    this.s3Client.send(new GetObjectCommand(input));
+  // Object command members
```
SanjalKatiyar, line 45:
```suggestion
// Object command members (alphabetical order)
```

packages/shared/src/s3/commands.ts
```diff
@@ -36,24 +34,23 @@ export class S3Commands {
...
   getSignedUrl: GetSignedUrl = (input, expiresIn) =>
-    getSignedUrl(this.s3Client, new GetObjectCommand(input), { expiresIn });
+    getSignedUrl(this, new GetObjectCommand(input), { expiresIn });
 
-  listObjects: ListObjectsV2 = (input) =>
-    this.s3Client.send(new ListObjectsV2Command(input));
+  getObject: GetObject = (input) => this.send(new GetObjectCommand(input));
+
+  deleteObjects: DeleteObjects = (input) =>
+    this.send(new DeleteObjectsCommand(input));
```
SanjalKatiyar, line 55: we are keeping all this in alphabetical order...

packages/odf/components/s3-browser/upload-objects/upload-component/uploads.ts
```diff
@@ -0,0 +1,75 @@
...
+      client: client,
+      params: {
+        Bucket: bucketName,
+        Key: key,
+        Body: file,
+        ...(file.type ? { ContentType: file.type } : {}),
+      },
+      partSize: 5 * 1024 * 1024,
+      queueSize: 4,
+    });
```
SanjalKatiyar, line 29: plz create a member function under `commands` file (and corresponding type) and call it here... similar to how we are doing it for `getSignedUrl`...
we are keeping everything under that class, so would not make sense to make any exception here:)

packages/odf/components/s3-browser/upload-objects/upload-component/FileUploadComponent.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+  const closeAlert = () => {
+    setUploadStatus(UploadStatus.INIT_STATE);
+  };
+
+  return (
+    <>
+      <div className="odf-upload">
+        {(uploadStatus === UploadStatus.UPLOAD_START ||
+          uploadStatus === UploadStatus.UPLOAD_COMPLETE) && (
```
SanjalKatiyar, line 102: what will we do if all objects failed ?? won't we show any alert ??

packages/odf/components/s3-browser/upload-objects/upload-component/uploads.ts
SanjalKatiyar: client will be automatically passed by the member function of the class, like we are doing for all other commands...

packages/odf/components/s3-browser/upload-objects/upload-component/uploads.ts
```diff
@@ -0,0 +1,75 @@
...
+    return uploader.done();
+  };
+  const allUploadPromise = files.map((file) => performUploadPromise(file));
+  try {
+    await Promise.allSettled(allUploadPromise);
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.error(e);
+  }
```
SanjalKatiyar, line 61: shouldn't we set status as `UploadStatus.UPLOAD_FAILED` for the files which were `rejected` or when entire `try` block encounters any exception ??

packages/odf/components/s3-browser/upload-objects/upload-component/FileUploadComponent.tsx
```diff
@@ -0,0 +1,165 @@
...
+          client,
+          bucketName,
+          setProgress,
+          foldersPath
+        );
+        setCompletionTime(completionTime);
+      } catch (e) {
+        // eslint-disable-next-line no-console
+        console.error('Error event', e);
+      }
```
SanjalKatiyar, line 62: shouldn't we set status as `UploadStatus.UPLOAD_FAILED` when we encounter this catch block ??

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
SanjalKatiyar: needed by Details tab and YAML tab...

packages/odf/components/s3-browser/upload-objects/upload-component/uploads.ts
SanjalKatiyar: also, just to point out that u r passing wrong `client` here... what u r actually passing is `noobaaS3Client` (that's why it was good to rename the variable correctly), whereas what u need to pass is `s3Client` ("this" object under `S3Commands` class)...

SanjalKatiyar: above way is probably working because of inheritance "extends" part, but logically it's wrong...

packages/odf/components/s3-browser/upload-objects/UploadSidebar.tsx
```diff
@@ -0,0 +1,174 @@
...
+  completionTime,
+}) => {
+  const { t } = useCustomTranslation();
+  const [uploadedFiles, totalFiles] =
+    getCompletedAndTotalUploadCount(uploadProgress);
+  const failedFiles = getFailedFiles(uploadProgress);
+  const uploadSpeed = getUploadSpeed(uploadProgress);
+  const totalReamining = getTotalRemainingFilesAndSize(uploadProgress);
+  const timeRemaining = getTotalTimeRemaining(uploadProgress);
+  const totalTimeElasped = getTotalTimeElasped(uploadProgress, completionTime);
```
alfonsomthd, line 49:
```suggestion
  const totalTimeElapsed = getTotalTimeElapsed(uploadProgress, completionTime);
```

packages/odf/components/s3-browser/upload-objects/UploadSidebar.tsx
```diff
@@ -0,0 +1,174 @@
...
+                  {t('Succeeded: {{uploadedFiles}}', { uploadedFiles })}
+                </FlexItem>
+                <FlexItem>
+                  {t('Failed files: {{failedFiles}}', {
+                    failedFiles: totalFiles - uploadedFiles,
+                  })}
+                </FlexItem>
+                <FlexItem>
+                  {t('Completion time: {{totalTimeElasped}}', {
+                    totalTimeElasped,
```
alfonsomthd, line 98:
```suggestion
                  {t('Completion time: {{totalTimeElapsed}}', {
                    totalTimeElapsed,
```

packages/odf/components/s3-browser/upload-objects/UploadSidebar.tsx
```diff
@@ -0,0 +1,174 @@
...
+                  })}
+                </FlexItem>
+              </>
+            ) : (
+              <>
+                <FlexItem>
+                  {t('Total Remaining: {{totalReamining}}', { totalReamining })}
```
alfonsomthd, line 105:
```suggestion
                  {t('Total Remaining: {{totalRemaining}}', { totalRemaining })}
```

packages/odf/components/s3-browser/upload-objects/UploadSidebar.tsx
```diff
@@ -0,0 +1,174 @@
...
+          title="How uploading works?"
+          isInline
+          variant={AlertVariant.info}
+          className="pf-v5-u-mb-sm"
+        >
+          <Trans t={t}>
+            Standard uploads have a size limit of up to 5 TB in S3. For large
+            objects, something or the other will happen. We need to improve this
+            alert. Or not show this. Doesnt make sense to repeat your messages
+            in UI.
```
alfonsomthd, line 134: :smile:

packages/odf/components/s3-browser/upload-objects/upload-component/FileUploadComponent.tsx
alfonsomthd: I would be a little bit more specific by saying `Error while uploading file` instead of `Error event`.

packages/odf/components/s3-browser/upload-objects/upload-status/UploadStatusBasedAlert.tsx
```diff
@@ -0,0 +1,56 @@
...
+export const UploadStatusBasedAlert: React.FC<UploadStatusBasedAlertProps> = ({
+  showSidebar,
+  abortAll,
+  completedUploads,
+  failedUploads,
+  totalUploads,
+  closeAlert,
+}) => {
+  const { t } = useCustomTranslation();
+  const showSuccess = totalUploads === completedUploads + failedUploads;
```
alfonsomthd, line 25: `showSuccess` can be misleading at first sight as in reality we're taking into account both successful + failed uploads.
Can we rename it to `isAllFinished` or similar?

packages/odf/components/s3-browser/upload-objects/utils.ts
```diff
@@ -0,0 +1,115 @@
...
+    0
+  );
+  const timeRemaining = totalRemainingBytes / speed;
+  if (timeRemaining < 1000) {
+    return humanizeSeconds(timeRemaining, 'ms').string;
+  }
+  return humanizeMinutes(timeRemaining / 1000).string;
+};
+
+export const getTotalTimeElasped = (
```
alfonsomthd, line 101:
```suggestion
export const getTotalTimeElapsed = (
```

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
bipuladh: Please introduce it when required later.

packages/shared/src/s3/commands.ts
bipuladh: any reason for it? Unless we have a linter doing this. This is sooner or later going to be messed up.

alfonsomthd:
@bipuladh rebase needed as Bucket Details Tab PR has been recently merged.

packages/odf/components/s3-browser/upload-objects/upload-component/uploads.ts
bipuladh: logically it is not incorrect because a class that extends a base class can be used in its place too.

packages/odf/components/s3-browser/upload-objects/upload-component/FileUploadComponent.tsx
bipuladh: We would show alerts but we would still reach UPLOAD_COMPELTE state.

packages/odf/components/s3-browser/upload-objects/upload-component/FileUploadComponent.tsx
```diff
@@ -0,0 +1,165 @@
...
+        </Title>
+        <div {...getRootProps()}>
+          <input {...getInputProps({ webkitdirectory: 'true' } as any)} />
+          <Flex className="odf-upload__title">
+            <FlexItem className="odf-upload-title__button">
+              <Flex>
+                <FlexItem>
+                  <Flex>
+                    <FlexItem>
+                      <Icon color="blue">
```
SanjalKatiyar, line 135: icon is still rendered as black... I don't think `color="blue"` works like this...

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
SanjalKatiyar: @alfonsomthd details tab PR is merged:)

SanjalKatiyar: and YAML tab is already existing...

packages/shared/src/s3/commands.ts
SanjalKatiyar: @alfonsomthd suggestion, which I liked/agreed-to... not sure if it's possible to configure linter for a member functions of a particular class...

packages/odf/components/s3-browser/upload-objects/upload-component/uploads.ts
SanjalKatiyar: make sense... but my point was, our custom class `S3Commands` contains much more than what is natively by "this" (base class' object), like lot of member functions which are not required to be passed to `Upload` command, hence "logically" wrong (but implementationally correct).

packages/odf/components/s3-browser/upload-objects/upload-component/FileUploadComponent.tsx
SanjalKatiyar: ack !

packages/shared/src/s3/commands.ts
bipuladh: https://typescript-eslint.io/rules/member-ordering/ 
Let's introduce this later if required for now I am against using manual effort to sort code.

packages/odf/components/s3-browser/upload-objects/upload-component/uploads.ts
bipuladh: it looks for an object not member functions. So S3Commands extends S3Clients hence I don't think it's logically wrong. It is common practice to pass derived classes where base classes are accepted.

packages/odf/components/s3-browser/upload-objects/upload-status/UploadStatusBasedAlert.tsx
bipuladh: We show success variant so I think it makes sense to keep it as is.

packages/shared/src/s3/commands.ts
SanjalKatiyar: ack, thanks !

packages/odf/components/s3-browser/upload-objects/upload-component/uploads.ts
SanjalKatiyar: ack !
but still let's move this under the class to have consistency with other commands...

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
bipuladh: I think we need to keep usage constant. Some places we are taking it from the param some places we are resolving objects by passing them under the same HorizontalNav components. I will update the code to use params in YAML as details tab is currently using params.

packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
bipuladh: I think the default behaviour of details page should be to poll. We shouldn't have to refresh details page to get more info. If we need to introduce polling let's do that but details page being updated after refresh is an-anti pattern in our UI

SanjalKatiyar: IMHO that's not a good idea... because with that reasoning we should then poll in object list page too, we don't need "Refresh" at all...
let's discuss this offline once and try to sort this...

bipuladh: Updated the code with the following updates:
1) Details page get refreshed
2) YAML page is a huge component so not refreshing it

packages/odf/components/s3-browser/upload-objects/FileUploadComponent.tsx
bipuladh: we show it. This is just about upload starting and ending regardless of failures.

packages/odf/components/s3-browser/bucket-details/BucketDetails.tsx
```diff
@@ -166,16 +169,23 @@ const BucketDetailsOverview: React.FC<{}> = ({}) => {
...
 };
 
-export const BucketDetails: React.FC<BucketDetailsProps> = ({ obj }) => {
+export const BucketDetails: React.FC<BucketDetailsProps> = ({
+  obj: { resource, refresh },
+}) => {
   const { t } = useCustomTranslation();
 
-  return (
+  return refresh && resource ? (
```
SanjalKatiyar, line 183:
```suggestion
  return refresh ? (
```

packages/odf/components/s3-browser/bucket-details/BucketDetails.tsx
```diff
@@ -166,16 +169,23 @@ const BucketDetailsOverview: React.FC<{}> = ({}) => {
...
     <>
       <BucketDetailsOverview />
-      {obj && <OBDetails obj={obj} ownerLabel={t('Owner References')} />}
+      {<OBDetails obj={resource} ownerLabel={t('Owner References')} />}
```
SanjalKatiyar, line 186: `BucketDetailsOverview` does not depends on CR (resource)...
```suggestion
      {resource && <OBDetails obj={resource} ownerLabel={t('Owner References')} />}
```

packages/odf/components/s3-browser/upload-objects/UploadSidebar.tsx
SanjalKatiyar: I know microcopy is pending, still good not to show wrong text on UI !
```suggestion
            objects,
```

plugins/odf/console-extensions.json
```diff
@@ -747,7 +747,7 @@
       }
     },
     "flags": {
-      "required": ["MCG_STANDALONE"]
+      "required": ["MCG"]
```
SanjalKatiyar, line 750: it's for MCG standalone deployments only !
```suggestion
      "required": ["MCG_STANDALONE"]
```

SanjalKatiyar (commented):
LGTM

PR #1636, opened 2024-10-14
Bucket Details Page: add 'Bucket overview' + OB Details

alfonsomthd:
Fixes: https://issues.redhat.com/browse/RHSTOR-6541
- S3 Bucket Details:
Screenshot from 2024-10-14 13-19-48
- OBC Bucket Details:
Screenshot from 2024-10-15 15-51-05


packages/odf/components/s3-browser/bucket-overview/BucketDetails.tsx
```diff
@@ -0,0 +1,144 @@
+import * as React from 'react';
```
SanjalKatiyar, line 1: `bucket-overview` is the folder which glues all components together...

we have separate folders for "upload-objects", "objects-list" etc...

better to make separate folder for "bucket-details" too and import the details tab in BucketOverview.tsx from that folder.

packages/odf/components/s3-browser/bucket-overview/BucketDetails.tsx
```diff
@@ -0,0 +1,144 @@
...
+type BucketDetailsProps = {
+  obj: K8sResourceKind;
+};
+
+export const BucketDetails: React.FC<BucketDetailsProps> = ({ obj }) => {
+  const { t } = useCustomTranslation();
+
+  return (
+    <>
+      <BucketOverview />
```
SanjalKatiyar, line 140: nits (optional): can we rename it to just `Overview` or something else ?? There is a different (already existing) file/folder with exact same name, which would cause confusion.
```suggestion
      <Overview />
```

SanjalKatiyar: Other option is to rename existing file/folder to a diff name.

alfonsomthd: It has the title "Bucket overview" in the UX design, so I'll rename it to `BucketDetailsOverview`

packages/odf/components/s3-browser/bucket-overview/BucketDetails.tsx
SanjalKatiyar: it's "R" (read) of CRUD :P 
and a different use case as well... so should qualify for its own separate folder.

but if we don't want it to be separate, then probably we need to move "objects-list" inside "bucket-overview" as well ??

packages/odf/components/s3-browser/bucket-overview/BucketDetails.tsx
SanjalKatiyar: thanks

packages/odf/components/s3-browser/bucket-overview/BucketDetails.tsx
alfonsomthd: As I see it, `bucket-overview` folder is the R (read/retrieve the bucket info); the details tab is just a part of that operation.
Anyway I'll move it to a separate folder for the sake of moving forward.

SanjalKatiyar: that make sense too... I am okay with either moving bucket-details out or moving object-details inside.

^ if latter, then in a separate PR, of course.

packages/odf/constants/s3-browser.ts
```diff
@@ -16,6 +16,11 @@ export const BUCKETS_BASE_ROUTE = '/odf/object-storage/buckets';
 export const BUCKET_CREATE_PAGE_PATH = '/odf/object-storage/create-bucket';
 
 // key to be used by SWR for caching particular API call
+export const BUCKET_ACL_CACHE_KEY_SUFFIX = 'BUCKET_ACL_CACHE_KEY';
+export const BUCKET_ENCRYPTION_CACHE_KEY_SUFFIX = 'BUCKET_ENCRYPTION_CACHE_KEY';
+export const BUCKET_POLICY_CACHE_KEY_SUFFIX = 'BUCKET_POLICY_CACHE_KEY_SUFFIX';
+export const BUCKET_TAGGING_CACHE_KEY_SUFFIX = 'BUCKET_TAGGING_CACHE_KEY';
+export const BUCKET_VERSIONING_CACHE_KEY_SUFFIX = 'BUCKET_VERSIONING_CACHE_KEY';
```
SanjalKatiyar, line 23: in a separate PR can we plz make them a function ? something like this:
 ```
 export const bucketACLCacheKeySuffix = (bucketName) => bucketName + '-' + 'BUCKET_ACL_CACHE_KEY';
 ```

SanjalKatiyar: I will do the same for `LIST_OBJECTS` key as well...

SanjalKatiyar: we will do the same for `LIST_OBJECTS` key as well...

SanjalKatiyar: this will prevent someone from using their own format of exact keys in the future by mistake...

PR #1637, opened 2024-10-14
Cleanup + Bug fixes

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-6568

Fixes for:
- Try to navigate to buckets list page Route (by specifically typing in the URL) when deployed in "Full deployment" mode - Route works for MCG standalone, but page breaks for other modes, page should not break, even if other modes are not supported.
- S3 commands and their types under shared should be defined in alphabetical order.
- Add object/bucket count in the list page pagination, eg: "1 - 300 of many".
- Add file/folder icons to objects list.
- Pagination not working correctly.

Screenshot 2024-10-14 at 5 38 34 PM


packages/odf/components/s3-browser/pagination-helper.tsx
```diff
@@ -97,14 +114,55 @@ export const continuationTokensRefresher = async <T extends ListCommandOutput>(
...
+  return count <= 0 ? 0 : count;
+};
+
+export const getPaginationCount = (
+  continuationTokens: ContinuationTokens,
+  currentPageCount: number,
+  maxResponses: number
+) => {
+  const fromCount = getFromCount(
+    continuationTokens,
```
GowthamShanmugam, line 138: optional comment:
```suggestion
    continuationTokens.previous?.length,
```

SanjalKatiyar: It's alright IMO, `getFromCount` will handle this...

PR #1639, opened 2024-10-15
[WIP] Rebase with target/base branch (GH workflows)

SanjalKatiyar:
Currently `frontend-build.yml` workflow only runs either **on the PR** or **on master after the merge/push**.

As per this update now PR will get rebased with the target/base (master or release*), then build and other tests will run on that.


alfonsomthd:
As per the [pull request event](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request), `actions/checkout@v3` is checking out at the merge commit, so this change is not needed.

PR #1648, opened 2024-10-21
Bug 2317212: Fix Clients' list page filtering

SanjalKatiyar:
**Before:**
Screenshot 2024-10-21 at 12 46 35 PM

**After:**
Screenshot 2024-10-21 at 12 51 00 PM

### DIFF
--- a/packages/odf/components/storage-consumers/client-list.tsx
+++ b/packages/odf/components/storage-consumers/client-list.tsx
@@ -46,6 +46,7 @@ import {
   clientHeartBeatFilter,
   getMajorMinorVersion,
   versionMismatchFilter,
+  storageConsumerNameFilter,
 } from './list-filter';
 import { ClientOnBoardingModal } from './onboarding-modal';
 import { RotateKeysModal } from './rotate-keys-modal';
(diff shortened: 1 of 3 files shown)


packages/odf/components/storage-consumers/list-filter.ts
```diff
@@ -75,3 +76,15 @@ export const versionMismatchFilter = (
     );
   },
 });
+
+export const nameFilterOverride = (): RowFilter<StorageConsumerKind> => ({
```
bipuladh, line 80:
```suggestion
export const storageConsumerNameFilter = (): RowFilter<StorageConsumerKind> => ({
```

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -424,14 +425,23 @@ export const ClientListPage: React.FC<ClientListPageProps> = () => {
...
-  const rowFilters = React.useMemo(
-    () => [clientHeartBeatFilter(t), versionMismatchFilter(t, serviceVersion)],
-    [t, serviceVersion]
-  );
+  // "rowFilters":
+  // - custom filters (new) that we are introducing to the list page
+  // - passing to both "useListPageFilter" hook & "ListPageFilter" component
+  // "rowFiltersWithNameOverride":
+  // - needed for overriding the filtering (default) on the CR's "name" (need to read the name from the CR's status instead)
+  // - only passing to "useListPageFilter" hook & not "ListPageFilter" component
```
bipuladh, line 433: Why not list page filter component?

SanjalKatiyar: We have a separate dropdown for "Name" filtering on the UI, already managed by `ListPageFilter` component by default. All we need to do is to override filter's definition in `useListPageFilter` hook.

ListPageFilter + rowFilters adds a different/new dropdown, passing `rowFiltersWithNameOverride` to `ListPageFilter` is adding extra empty item to that dropdown.

bipuladh (approved):
/lgtm

PR #1649, opened 2024-10-21
Bug 2312118: Fix 'Not Found' error with encryption form

SanjalKatiyar:
**Before:**
Screenshot 2024-10-21 at 4 02 26 PM

**After:**
Screenshot 2024-10-21 at 4 03 41 PM
Screenshot 2024-10-21 at 4 03 59 PM


alfonsomthd (approved):
LGTM (although IMHO `isNotFoundError` would have been more appropriate than just `isNotFound` as the shared utils have a lot of functions so better to be explicit).

PR #1661, opened 2024-11-07
Upgrade ts-node dependency

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-6587


package.json
```diff
@@ -27,7 +27,7 @@
     "build-mco": "NODE_ENV=production PLUGIN=mco I8N_NS=plugin__odf-multicluster-console yarn build:plugin",
     "build-client": "NODE_ENV=production PLUGIN=client I8N_NS=plugin__odf-client-console yarn build:plugin",
     "build-dev": "NODE_ENV=development PLUGIN=odf I8N_NS=plugin__odf-console yarn build:plugin",
-    "server:plugin": "yarn build:generate && I8N_NS=${I8N_NS} yarn ts-node --stack-size=65500 ./node_modules/.bin/webpack serve -c ./webpack.config.ts --progress",
```
SanjalKatiyar, line 30: `--stack-size` not recognised by newer version of ts-node...
also I didn't face any issue after removing it, so didn't add any alternative...

package.json
```diff
@@ -38,7 +38,7 @@
     "dev-client": "PLUGIN=client I8N_NS=plugin__odf-client-console yarn server:plugin",
     "serve-local-build": "./http-server.sh ./plugins/odf/dist",
     "serve-local-build-mco": "./http-server.sh ./plugins/mco/dist",
-    "ts-node": "ts-node -O '{\"module\":\"commonjs\"}' -I '/node_modules/(?!(@odf)/)/'",
+    "ts-node": "ts-node --transpile-only -O '{\"module\":\"commonjs\"}' -I '/node_modules/(?!(@odf)/)/'",
```
SanjalKatiyar, line 41: Looks like newer version of ts-node is more strict, even though we are using `-I` flag it still checks `./node_modules/.bin/webpack` and reports error:
```
TSError: ⨯ Unable to compile TypeScript:
node_modules/webpack/node_modules/graceful-fs/graceful-fs.js:34:3 - error TS2322: Type '(msg: string, ...param: any[]) => void' is not assignable to type '() => void'.
  Target signature provides too few arguments. Expected 1 or more, but got 0.

34   debug = util.debuglog('gfs4')
     ~~~~~
node_modules/webpack/node_modules/graceful-fs/graceful-fs.js:52:3 - error TS2741: Property '__promisify__' is missing in type '(fd: any, cb: any) => any' but required in type 'typeof close'.

52   fs.close = (function (fs$close) {
     ~~~~~~~~

  node_modules/@types/node/fs.d.ts:1251:18
    1251         function __promisify__(fd: number): Promise<void>;
                          ~~~~~~~~~~~~~
    '__promisify__' is declared here.
node_modules/webpack/node_modules/graceful-fs/graceful-fs.js:72:25 - error TS6133: 'fd' is declared but its value is never read.

72     function closeSync (fd) {
                           ~~
node_modules/webpack/node_modules/graceful-fs/graceful-fs.js:86:13 - error TS2554: Expected 0 arguments, but got 1.

86       debug(fs[gracefulQueue])
```

SanjalKatiyar: only way which worked is to use `--transpile-only`... which I think is relatively okay as editor (IDE) is still configured with typescript support and this flag only affects selective files only...

PR #1666, opened 2024-11-08
Enhance block pool mirroring status card to cover rados namespace

GowthamShanmugam:
Issue: https://issues.redhat.com/browse/RHSTOR-6578
ToDo: The overall_image_health column has to be calculated using rados image health metrics. Fetching it from the CR has some performance impact. 

Issue to track metrics change: https://issues.redhat.com/browse/RHSTOR-6579


packages/ocs/dashboards/block-pool/mirroring-card.tsx
```diff
@@ -166,26 +255,31 @@ export const MirroringCard: React.FC = () => {
...
-                  icon={healthStateMapping[mirroringImageHealth]?.icon}
+                  title={mirroringHealth?.health}
+                  icon={mirroringHealth?.icon}
                   className="pf-v5-u-ml-xs"
                 />
               </MirroringCardItem>
-              {!_.isEmpty(obj.status?.mirroringStatus?.summary?.states) && (
+              {!_.isEmpty(states) && isLoaded && !isLoadError && (
                 <>
                   <MirroringCardItem>
```
bipuladh, line 280: Should loading and error should be handled by `MirroringCardItrem` rather than a conditional render?

bipuladh (commented):
looks good. minor comment.

packages/ocs/dashboards/block-pool/mirroring-card.tsx
GowthamShanmugam: ack, i  will change this

GowthamShanmugam:
Screenshot 2024-11-12 at 4 27 14 PM
Screenshot 2024-11-12 at 4 26 52 PM

packages/ocs/types.ts
```diff
@@ -49,3 +54,19 @@ export type CephFilesystemKind = K8sResourceCommon & {
 };
 
 export type ODFSystemParams = { namespace: string; systemName: string };
+
+export type CephBlockPoolRadosNamespaceKind = K8sResourceCommon & {
+  spec: DataPool & {
+    blockPoolName: string;
+    mirroring: {
+      mode: string;
```
alfonsomthd, line 62: If mirroring mode is a field that both RBD & CephFS have, you can add `mode` to `DataPool` type.

alfonsomthd:
LGTM. @SanjalKatiyar & @bipuladh for tagging.

packages/ocs/types.ts
GowthamShanmugam: Ah! right i dont even need DataPool for CephBlockPoolRadosNamespaceKind. i am removing .

packages/ocs/dashboards/block-pool/mirroring-card-item.tsx
```diff
@@ -27,25 +28,27 @@ export const MirroringCardItem: React.FC<MirroringCardItemProps> = React.memo(
...
-            data-test="mirroring-card-item-title"
-          >
-            {title}
-          </dt>
-        )}
-        <dd
-          className={classNames(
-            'odf-block-pool__mirroring-card-item-dd',
-            valueClassName
+      !isHidden && (
```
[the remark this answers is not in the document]
GowthamShanmugam: My initial plan to hide the entire MirroringCardItem render but as per this comment, I moved the hiding part inside this component: https://github.com/red-hat-storage/odf-console/pull/1666#discussion_r1837549495

I am ok with both the approach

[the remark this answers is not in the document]
SanjalKatiyar: what I understood from the comment is that for: `!_.isEmpty(states) && isLoaded && !isLoadError`, why to use "isLoaded" and "isLoadError" outside when we can pass it as a prop to `MirroringCardItem`, which internally handles it.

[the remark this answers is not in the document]
SanjalKatiyar: but anyway, I am okay with current changes as well...

packages/ocs/dashboards/block-pool/mirroring-card.tsx
```diff
@@ -18,16 +26,99 @@ import {
...
 
+const aggregateMirroringHealth = (
+  poolObj: StoragePoolKind,
+  radosNamespaces: CephBlockPoolRadosNamespaceKind[],
+  t: TFunction
+): HealthMappingValues[] => {
+  const healthStateMap = healthStateMapping(t);
+  let mirroringHealth: HealthMappingValues =
+    healthStateMap[poolObj.status?.mirroringStatus?.summary?.health];
+  let imageHealth: HealthMappingValues =
```
SanjalKatiyar, line 51: I saw a comment below `@ToDo: Fetch aggregated image health from metrics: https://issues.redhat.com/browse/RHSTOR-6579`, but here you are using CR. Which one we need to use CR or metrics ??

packages/ocs/dashboards/block-pool/mirroring-card.tsx
```diff
@@ -18,16 +26,99 @@ import {
...
 import { healthStateMapping, ImageStateLegendMap } from './states';
 import './mirroring-card.scss';
 
+const aggregateMirroringHealth = (
+  poolObj: StoragePoolKind,
+  radosNamespaces: CephBlockPoolRadosNamespaceKind[],
+  t: TFunction
+): HealthMappingValues[] => {
+  const healthStateMap = healthStateMapping(t);
+  let mirroringHealth: HealthMappingValues =
```
SanjalKatiyar, line 49: same here... we need to rely on CR or metrics ??

packages/ocs/dashboards/block-pool/mirroring-card.tsx
```diff
@@ -150,12 +240,29 @@ const MirroringImageHealthChart: React.FC<MirroringImageHealthChartProps> = ({
...
+    },
+  });
 
   const mirroringStatus: boolean = obj.spec?.mirroring?.enabled;
-  const mirroringImageHealth: string =
-    obj.status?.mirroringStatus?.summary?.image_health;
-  const lastChecked: string = obj.status?.mirroringStatus?.lastChecked;
-  const formatedDateTime = lastChecked
+  const [mirroringHealth, imageHealth]: HealthMappingValues[] =
+    aggregateMirroringHealth(obj, radosNamespaces, t);
```
SanjalKatiyar, line 259: Function is called `aggregateMirroringHealth`, but u r aggregating image health too... either create 2 different functions or rename current one IMO.

packages/ocs/dashboards/block-pool/mirroring-card-item.tsx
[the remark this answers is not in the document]
GowthamShanmugam: oh ok! my bad

packages/ocs/dashboards/block-pool/mirroring-card.tsx
GowthamShanmugam: mirroring_image health is a subset of mirroring health

packages/ocs/dashboards/block-pool/mirroring-card.tsx
GowthamShanmugam: For the last page, I want mirror image health for all block pools. If I read from CR, I will end up listening to all radio namespaces. To improve my performance I want to replace the CR read with metrics on list page.  

Mirroring card under dashboard is for a specific pool, I am adding the watchers agianst the rados namespace using  pool name label. Also, here I want the image state information, so I have to use CR.

packages/ocs/dashboards/block-pool/mirroring-card.tsx
GowthamShanmugam: https://github.com/red-hat-storage/odf-console/pull/1666#discussion_r1838471183

alfonsomthd:
LGTM. @SanjalKatiyar for tagging.

SanjalKatiyar:
> LGTM. @SanjalKatiyar for tagging.

We are waiting for the confirmation from PM before merging this PR. IIUC that's why it is on hold.

GowthamShanmugam:
I removed hold, i am not removing the overall image health column, and I modified the logic with metrics instead of CR.

packages/ocs/dashboards/block-pool/mirroring-card.tsx
```diff
@@ -18,16 +27,99 @@ import {
...
+        states[state] = oldCount + count;
+      });
+    }
+  });
+  return states;
+};
+
+const getPieChartHeight = (numOfStates: number) => {
+  if (numOfStates <= 3) {
+    return 150;
```
bipuladh, line 116: What are these magic numbers?

packages/ocs/dashboards/block-pool/states.ts
```diff
@@ -26,36 +20,29 @@ export const ImageStateLegendMap = (
...
+    case HealthState.OK:
+      return t('Ok');
+    case HealthState.UNKNOWN:
+      return t('Unknown');
+    case HealthState.WARNING:
+      return t('Warning');
+    case HealthState.ERROR:
+      return t('Error');
+    default:
+      return t('-');
```
SanjalKatiyar, line 44: no need to translate this...
```suggestion
      return '-';
```

packages/ocs/storage-pool/StoragePoolListPage.tsx
```diff
@@ -258,21 +276,18 @@ const RowRenderer: React.FC<RowProps<StoragePool, CustomData>> = ({
...
   const mirroringStatus: boolean = obj.spec?.mirroring?.enabled;
-  const mirroringImageHealth: string = mirroringStatus
-    ? obj.status?.mirroringStatus?.summary?.image_health
-    : '-';
-  const lastChecked: string = obj.status?.mirroringStatus?.lastChecked;
-  const formatedDateTime = lastChecked
-    ? twelveHoursdateTimeNoYear.format(new Date(lastChecked))
-    : '-';
+  const imageHealth =
+    !(!!mirroringImageHealthError && mirroringImageHealthLoading) &&
```
SanjalKatiyar, line 287: `!(!!mirroringImageHealthError && mirroringImageHealthLoading)` will translate to `!mirroringImageHealthError || !mirroringImageHealthLoading`...
is that what was expected ??
```suggestion
    (!mirroringImageHealthError && !mirroringImageHealthLoading) &&
```

packages/ocs/storage-pool/StoragePoolListPage.tsx
```diff
@@ -339,12 +354,10 @@ const RowRenderer: React.FC<RowProps<StoragePool, CustomData>> = ({
         {mirroringStatus ? t('Enabled') : t('Disabled')}
       </TableData>
       <TableData {...tableColumnInfo[7]} activeColumnIDs={activeColumnIDs}>
-        <Tooltip content={`${t('Last synced')} ${formatedDateTime}`}>
```
SanjalKatiyar, line 342: removing the tooltip ?? isn't this UX change ??

packages/ocs/storage-pool/StoragePoolListPage.tsx
```diff
@@ -246,10 +250,24 @@ const RowRenderer: React.FC<RowProps<StoragePool, CustomData>> = ({
...
+    mirroringImageHealthLoading,
+  ] = useCustomPrometheusPoll({
+    endpoint: 'api/v1/query' as any,
+    query: getPoolQuery(
+      [name],
+      StorageDashboardQuery.POOL_MIRRORING_IMAGE_HEALTH,
+      managedByOCS
+    ),
+    basePath: usePrometheusBasePath(),
+  });
```
SanjalKatiyar, line 270: instead of reading/polling metrics for each row, can't we do it once on table level (for all pools) and just filter later ??

packages/ocs/dashboards/block-pool/mirroring-card.tsx
```diff
@@ -18,16 +27,99 @@ import {
...
+      radosNamespace.status?.mirroringStatus?.summary?.states ?? {};
+    if (!!radosNamespaceStates) {
+      Object.entries(radosNamespaceStates).forEach(([state, count]) => {
+        const oldCount = states?.[state] || 0;
+        states[state] = oldCount + count;
+      });
+    }
+  });
+  return states;
+};
```
SanjalKatiyar, line 112: general question: doesn't/shouldn't any health/state status on BlockPool level incorporate status of Rados Namespace level too ?? what's the difference/significance of the status reported by the upper level BlockPool ??

GowthamShanmugam: The problem is rook has to check each rados namespace status + default namespace status on some sync interval. When the number of rados namespace counts increases, the rook's performance will be impacted. But UI checks all the rados namespace only when the page is rendered. so we decided to go with the UI approach.

packages/ocs/storage-pool/StoragePoolListPage.tsx
GowthamShanmugam: i will check this

packages/ocs/storage-pool/StoragePoolListPage.tsx
GowthamShanmugam: It's a minor change, CR has a file called last check, so we used it. But for the metrics approach, I skipped it. It is not like we are completely ignoring this field. The mirroring card still has this info. Basically, This list page has too much info, which is the same as a mirroring card.

GowthamShanmugam: cc: @Chandan141996

packages/ocs/dashboards/block-pool/mirroring-card.tsx
GowthamShanmugam: the image pie chart has 6 states, 
Starting replay, Replaying, Stopping replay, Stopped, Error, Unknown.

Each image from Blackpool can have different states so that the user may see all of these states at the same time or a few states in the UI. 

Based on the state count, I am adjusting the height of the pie chart, if we use a maximum height but only one state occurs, then we see a lot of empty space on top of the pie chart.

GowthamShanmugam: with height 240(maximum height to cover all legends) :
Screenshot 2024-11-18 at 9 01 15 PM

With height, condition numOfStates <= 3 with 150 height:
Screenshot 2024-11-18 at 9 00 49 PM

PR #1673, opened 2024-11-12
Add CA issue prerequisites

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-6222

Screenshot 2024-11-12 at 5 20 13 PM


packages/odf/components/s3-browser/ca-error/CAErrorMessage.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const errorType = error?.name;
+  if (
+    ['TypeError', 'NetworkingError', 'NetworkError'].includes(errorType) &&
+    (errorMessage.includes('certificate') ||
+      errorMessage.includes('authority') ||
+      errorMessage.includes('ssl') ||
+      errorMessage.includes('self-signed') ||
+      errorMessage.includes('failed to fetch') ||
+      errorMessage.includes('load failed') ||
+      errorMessage.includes('network'))
```
SanjalKatiyar, line 32: @alfonsomthd @bipuladh @GowthamShanmugam I couldn't find a better way than this. There is only limited info present in the "error" object response from the SDK (no status code, nothing concrete).

alfonsomthd: Tested in Fedora 40 with GChrome & Firefox: it works, LGTM.

SanjalKatiyar: I tested on Mac as well (Chrome & Firefox).

packages/odf/components/s3-browser/ca-error/CAErrorMessage.tsx
```diff
@@ -0,0 +1,89 @@
...
+          titleText={t('Could not load information')}
+          icon={<EmptyStateIcon icon={RedTimesCircleIcon} />}
+          headingLevel="h4"
+        />
+        <EmptyStateBody>
+          {t(
+            'The browser cannot connect securely to this endpoint because it does not recognize the SSL certificate. This occurs when the certificate of the endpoint is not issued by a trusted Certificate Authority (CA).'
+          )}
+          <br />
+          <br />
```
alfonsomthd, line 55: Instead of using several `<br>`, what about wrapping those phrases with `<p>` elements? Or using CSS margin?

packages/shared/src/constants/doc.ts
```diff
@@ -1,5 +1,5 @@
-export const ODF_DEFAULT_DOC_VERSION = '4.16';
```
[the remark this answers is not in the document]
SanjalKatiyar: It won't be used in prod, just there for safety... but this should be 4.18 as we in that release...

packages/odf/components/s3-browser/ca-error/CAErrorMessage.tsx
SanjalKatiyar: either is fine by me, any particular reason for not using `<br>` ?? I can update the PR if so...

packages/shared/src/constants/doc.ts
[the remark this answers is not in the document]
SanjalKatiyar: we are fine by not updating in 4.17 as well...

packages/odf/components/s3-browser/ca-error/CAErrorMessage.tsx
alfonsomthd: It's not a best practice to use <br> for adding space between paragraphs: it's meant to add a line break and nothing else.

PR #1675, opened 2024-11-13
Cypress: enable Test Isolation

alfonsomthd:
https://docs.cypress.io/app/core-concepts/test-isolation

Now after each test the session is cleared, hence the following refactoring:
- Login occurs in global `beforeEach` hook (and in `before` hooks where is required).
- Removed Logout as no longer necessary (and it's not part of any plugin).

### DIFF
--- a/cypress/support.ts
+++ b/cypress/support.ts
@@ -4,18 +4,26 @@ import {
   STORAGE_SYSTEM_NAME,
   OCS_SC_STATE,
   ODF_OPERATOR_NAME,
+  SECOND,
+  MINUTE,
 } from './consts';
 import './support/selectors';
 import './support/login';
 
-before(() => {
-  // Disable Cypress's default behavior of logging all XMLHttpRequests and fetches.
+const disableIrrelevantLogging = () => {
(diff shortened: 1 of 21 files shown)


cypress/support.ts
```diff
@@ -4,18 +4,26 @@ import {
...
+  disableIrrelevantLogging(); // Required as 'before' actions precede 'beforeEach' actions.
 });
 
 beforeEach(() => {
-  // Disable Cypress's default behavior of logging all XMLHttpRequests and fetches.
-  cy.intercept({ resourceType: /xhr|fetch/ }, { log: false });
+  disableIrrelevantLogging();
+  cy.login();
+  cy.visit('/');
+  cy.install();
```
GowthamShanmugam, line 26: if there is not logout, do we really need these three in beforeEach? is this can be moved to before?

alfonsomthd: No, after each test all is cleared, so we have to perform these actions always. bear in mind that session is cached so it's really performed 1 time.

alfonsomthd: The test isolation ensures that any test could be run isolated.

PR #1676, opened 2024-11-13
UI check to avoid peering a same managed cluster with more than one cluster

GowthamShanmugam:
Issue: https://issues.redhat.com/browse/RHSTOR-6634


GowthamShanmugam:
Screenshot 2024-11-13 at 8 12 22 PM

packages/mco/components/create-dr-policy/selected-cluster-validator.tsx
```diff
@@ -57,9 +57,26 @@ const checkClientToODFPeering = (clusters: ManagedClusterInfoType[]): boolean =>
   !!clusters[0]?.odfInfo?.storageClusterInfo?.clientInfo !==
   !!clusters[1]?.odfInfo?.storageClusterInfo?.clientInfo;
 
+const verifyMirrorPeerExistance = (
```
alfonsomthd, line 60:
```suggestion
const verifyMirrorPeerExistence = (
```

PR #1677, opened 2024-11-14
Upgrade Jest & React Testing Library

alfonsomthd:
IMPORTANT: We have to stick with `@testing-library/react@12.x` as 13+ versions require `react@18+` but `react` is a shared module with `@openshift-console/dynamic-plugin-sdk` that in turn requires 17.x version :disappointed: 
https://github.com/testing-library/react-testing-library?tab=readme-ov-file#installation


package.json
```diff
@@ -131,13 +131,12 @@
...
-    "@swc/jest": "^0.2.23",
-    "@testing-library/jest-dom": "^6.4.5",
-    "@testing-library/react": "^12",
+    "@swc/core": "^1.9.2",
+    "@swc/jest": "^0.2.37",
+    "@testing-library/jest-dom": "^6.6.3",
+    "@testing-library/react": "^12.1.5",
     "@testing-library/react-hooks": "^8.0.1",
-    "@testing-library/user-event": "^14.5.1",
-    "@types/jest": "29.2.2",
```
alfonsomthd, line 140: Jest already has this package as dependency so let's allow jest to manage it.

GowthamShanmugam (approved):

PR #1679, opened 2024-11-14
Upgrade Webpack & Sass

alfonsomthd:
UPGRADED:
- `webpack`
- webpack plugins.
  - Due to `ts-node` crashing with unexpected errors, it has been replaced with simpler `tsx` (works out-of-the-box).
    - Also `tsm` has been replaced by `tsx` (it was redundant).
    - See [TypeScript runtime comparison tables](https://github.com/privatenumber/ts-runtime-comparison?tab=readme-ov-file#transformation).
  - `style-loader` is kept in 3.x due to v4 throwing unexpected errors.
- `Sass`
  - `sass` dependency has been replaced with `sass-embedded` (recommended).
  - Removed [deprecated `~` in module paths](https://webpack.js.org/loaders/sass-loader/#resolving-import-and-use-at-rules). 
  - TODO (future PR): fix Sass deprecation warnings.


package.json
```diff
@@ -111,19 +111,18 @@
     "react-window": "^1.8.10",
     "redux": "4.0.1",
     "resolve-url-loader": "^5.0.0",
-    "sass": "^1.55.0",
-    "sass-loader": "^13.0.2",
+    "sass-embedded": "^1.80.7",
```
SanjalKatiyar, line 114: shouldn't we update `/shared/package.json` as well ??

alfonsomthd: Done.

PR #1692, opened 2024-11-18
Hide NFS builtin pool

alfonsomthd:
Fixes: https://issues.redhat.com/browse/DFBUGS-871

BEFORE:
cephnfs-builtin-pool-sc-form
cephnfs-builtin-pool

AFTER:
cephnfs-builtin-pool-sc-form-after
cephnfs-builtin-pool-after

### DIFF
--- a/cypress/support.ts
+++ b/cypress/support.ts
@@ -60,6 +60,10 @@ Cypress.Commands.add('install', () => {
       cy.contains('Create StorageSystem', { timeout: 15 * SECOND }).should(
         'be.visible'
       );
+
+      // Uncomment next line only if the cluster has enough resources.
+      // cy.get('label[for="enable-nfs"]').click();
+
       cy.get('button').contains('Next').click();
       // @TODO: Do we still want to uncheck the already unchecked 'Taint nodes' checkbox?
       // If yes, we should scroll down (needed after adding the performance profile selection)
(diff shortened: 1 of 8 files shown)


cypress/support.ts
```diff
@@ -60,6 +60,7 @@ Cypress.Commands.add('install', () => {
       cy.contains('Create StorageSystem', { timeout: 15 * SECOND }).should(
         'be.visible'
       );
+      cy.get('label[for="enable-nfs"]').click();
```
SanjalKatiyar, line 63: Is just clicking checkbox deploying NFS without any errors ??
AFAIR there were some manual steps as well when we added this feature which were needed to be enabled from CLI to use NFS.

SanjalKatiyar: It needed more resources than our current nodes in the CI, that's why I never added e2e for NFS feature.

alfonsomthd: The NFS pool is created and we're testing that is not exposed to the user.

alfonsomthd: > It needed more resources than our current nodes in the CI, that's why I never added e2e for NFS feature.

1. We need to increase our test coverage.
2. Those things (needing more resources) we need to document it, otherwise nobody knows why we're not covering features. When you say resources you mean at the CPU/RAM level or do you mean cluster nodes?

PR #1744, opened 2024-11-28
Debug: add instructions & vscode 'launch.json' template

alfonsomthd:
image


GowthamShanmugam:
Nice!! I tested for ODF and MCO, it works for both. It is really helpful for debugging.

PR #1783, opened 2024-12-17
Fix 'nanoid' CVE

SanjalKatiyar:
https://issues.redhat.com/browse/DFBUGS-1193
https://issues.redhat.com/browse/DFBUGS-1194
https://issues.redhat.com/browse/DFBUGS-1195

### DIFF
--- a/package.json
+++ b/package.json
@@ -189,7 +189,8 @@
     "webpack-dev-server/express": "^4.21.0",
     "tough-cookie": "^4.1.3",
     "cross-spawn": "^7.0.6",
-    "postcss": "^8.4.49"
+    "postcss": "^8.4.49",
+    "nanoid": "^3.3.8"
   },
   "engines": {
     "node": ">=20.x"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -189,7 +189,8 @@
     "webpack-dev-server/express": "^4.21.0",
     "tough-cookie": "^4.1.3",
     "cross-spawn": "^7.0.6",
-    "postcss": "^8.4.49"
+    "postcss": "^8.4.49",
+    "nanoid": "^3.3.8"
```
alfonsomthd, line 193: We should sort dependencies alphabetically. We can use a plugin to enforce this.

SanjalKatiyar: plugin sounds good...

alfonsomthd: Can we do it in this PR anyway?

SanjalKatiyar: a master only PR would be better IMO, this one needs to be backported till 4.14.z release...

PR #1791, opened 2024-12-18
Sort package.json alphabetically

SanjalKatiyar:
Add https://www.npmjs.com/package/sort-package-json to `husky`.


.husky/pre-commit
```diff
@@ -2,3 +2,5 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npx lint-staged
+yarn sort:packagejson
+git add .
```
alfonsomthd, line 6: This is dangeorus as I can have some files that I don't want it to be added to the commit that'll be added silently.
```suggestion
git add package.json
```

alfonsomthd: @SanjalKatiyar  Instead of addign any file, we can make pre-commit fail so the contributor can run `yarn sort:packagejson` manually and commit again. Thoughts?

SanjalKatiyar: let's move forward with https://github.com/red-hat-storage/odf-console/pull/1792 instead of any manual efforts... I will close this PR.

PR #1833, opened 2025-01-23
Object Browser: access S3 endpoint via console proxy

alfonsomthd:
IMPORTANT: merge after https://github.com/red-hat-storage/odf-operator/pull/519
Fixes: https://issues.redhat.com/browse/RHSTOR-6844

### DIFF
--- a/packages/odf/components/s3-browser/noobaa-context.tsx
+++ b/packages/odf/components/s3-browser/noobaa-context.tsx
@@ -2,16 +2,31 @@ import * as React from 'react';
 import { useSafeK8sGet } from '@odf/core/hooks';
 import { useODFNamespaceSelector } from '@odf/core/redux';
 import { StatusBox } from '@odf/shared/generic/status-box';
-import { SecretModel, RouteModel } from '@odf/shared/models';
-import { S3Commands } from '@odf/shared/s3';
-import { SecretKind, K8sResourceKind } from '@odf/shared/types';
+import { SecretModel } from '@odf/shared/models';
+import {
+  ODF_S3_PROXY_PATH,
+  S3_INTERNAL_ENDPOINT_PORT,
+  S3_INTERNAL_ENDPOINT_PREFIX,
+  S3_INTERNAL_ENDPOINT_SUFFIX,
+  S3_LOCAL_ENDPOINT,
+  S3Commands,
+} from '@odf/shared/s3';
+import { SecretKind } from '@odf/shared/types';
+import type { HttpRequest } from '@smithy/types';
 import * as _ from 'lodash-es';
 import {
   NOOBAA_ADMIN_SECRET,
-  NOOBAA_S3_ROUTE,
   NOOBAA_ACCESS_KEY_ID,
   NOOBAA_SECRET_ACCESS_KEY,
 } from '../../constants';
+
+const getS3Url = (odfNamespace: string) => {
+  return new URL(
+    window.location.hostname.includes('localhost')
+      ? S3_LOCAL_ENDPOINT
+      : `${S3_INTERNAL_ENDPOINT_PREFIX}${odfNamespace}${S3_INTERNAL_ENDPOINT_SUFFIX}`
+  );
+};
 
 type NoobaaS3ContextType = {
   noobaaS3: S3Commands;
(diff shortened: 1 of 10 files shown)


packages/shared/src/s3/commands.ts
```diff
@@ -35,6 +42,37 @@ import {
...
+    sign: (request: HttpRequest) => {
+      const originalPath = request.path;
+      request.path = originalPath.replace(
+        '/api/proxy/plugin/odf-console/s3-proxy',
+        ''
+      );
+      request.headers['host'] = request.headers['host'].includes('localhost')
+        ? 'localhost:6001'
+        : (request.headers['host'] =
+            's3.openshift-storage.svc.cluster.local:443');
```
alfonsomthd, line 66: @SanjalKatiyar ToDo: to move it to proper `constants` file.

packages/odf/components/s3-browser/noobaa-context.tsx
```diff
@@ -40,17 +39,12 @@ export const NoobaaS3Provider: React.FC<NoobaaS3ProviderType> = ({
...
-  );
-
   const s3Route = React.useRef<string>();
 
   const [noobaaS3, noobaaS3Error]: [S3Commands, unknown] = React.useMemo(() => {
-    if (!_.isEmpty(secretData) && !_.isEmpty(routeData)) {
+    if (!_.isEmpty(secretData)) {
       try {
-        s3Route.current = `https://${routeData.spec.host}`;
+        s3Route.current = `${window.location.protocol}//${window.location.host}/api/proxy/plugin/odf-console/s3-proxy`;
```
GowthamShanmugam, line 47: We started using console proxy in a lot of places so that we can keep the common part of the endpoint constant, for example:
CONSOLE_PROXY_ENDPOINT="/api/proxy/plugin"

packages/shared/src/s3/commands.ts
```diff
@@ -35,6 +42,37 @@ import {
...
+    sha256: Sha256,
+  });
+
+  return {
+    sign: (request: HttpRequest) => {
+      const originalPath = request.path;
+      request.path = originalPath.replace(
+        '/api/proxy/plugin/odf-console/s3-proxy',
+        ''
+      );
```
GowthamShanmugam, line 62: Do we need to update the request path? Is path can be stored in a separate variable and passed to signedRequest.path? or both are same?

packages/shared/src/s3/commands.ts
```diff
@@ -35,6 +42,37 @@ import {
...
+      request.headers['host'] = request.headers['host'].includes('localhost')
+        ? 'localhost:6001'
```
GowthamShanmugam, line 64: localhost:6001 can be a constant?

packages/shared/src/s3/commands.ts
SanjalKatiyar, line 66: we shouldn't hardcode `openshift-storage` part... reaad the NS from `useODFNamespaceSelector` hook...

packages/shared/src/s3/commands.ts
```diff
@@ -35,6 +42,37 @@ import {
...
+        : (request.headers['host'] =
+            's3.openshift-storage.svc.cluster.local:443');
+
+      return signer.sign(request).then((signedRequest) => {
+        signedRequest.path = originalPath;
+        return signedRequest;
```
SanjalKatiyar, line 70: what's the need to this part ? why are we replacing back with the original path ?

packages/shared/src/s3/commands.ts
SanjalKatiyar, line 66: aren't both syntax same ?
```suggestion
        : s3.openshift-storage.svc.cluster.local:443';
```

alfonsomthd: It's a draft PR for early review of the logic: I plan to clean it up.

packages/shared/src/s3/commands.ts
alfonsomthd: https://github.com/red-hat-storage/odf-console/pull/1833/files#r1926923997

packages/shared/src/s3/commands.ts
alfonsomthd: @SanjalKatiyar @GowthamShanmugam As I said in our previous conversations, we're behind a proxy so this is the required workflow:
- The signer reads the host header and the **request path** from the original request: we have to set the host header & the path as if we were reaching out to the S3 point directly (thus the removal of the proxy path).
- The signer returns the signed request: we have to reset the original path AFTER the signing so the request can reach out the proxy.

packages/shared/src/s3/commands.ts
alfonsomthd: Yes, that was accidental (due to previous experiments).

packages/shared/src/s3/commands.ts
GowthamShanmugam: ack

packages/odf/components/s3-browser/noobaa-context.tsx
alfonsomthd: Done.

packages/shared/src/s3/commands.ts
alfonsomthd: Done.

GowthamShanmugam (commented):
LGTM

packages/shared/src/s3/commands.ts
```diff
@@ -85,7 +86,18 @@ export class S3Commands extends S3Client {
     this.send(new ListObjectVersionsCommand(input));
 
   getSignedUrl: GetSignedUrl = (input, expiresIn) =>
-    getSignedUrl(this, new GetObjectCommand(input), { expiresIn });
+    getSignedUrl(this, new GetObjectCommand(input), { expiresIn }).then(
+      (url) => {
+        // We must set the proxy URL because the S3 client
+        // doesn't execute 'finalizeRequest' step for this action.
```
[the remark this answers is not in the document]
alfonsomthd: 1. The S3 client sends a `GetObject` request (thus executing the `build` & `finalizeRequest`) to check that the object exists and is retrievable.
2. It executes only the `build` step of  `GetObject` to construct the presigned URL based on the request (that points to the S3 endpoint). Therefore we need to adapt the URL to the user so it points to the proxy.

package.json
```diff
@@ -88,6 +88,7 @@
     "@patternfly/react-table": "5.4.14",
     "@patternfly/react-tokens": "5.4.1",
     "@patternfly/react-topology": "5.4.1",
+    "@smithy/types": "4.1.0",
```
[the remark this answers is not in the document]
alfonsomthd: No, otherwise the production build breaks:
>58.44 ERROR in ../../packages/odf/components/s3-browser/noobaa-context.tsx:15:29
58.44 TS2307: Cannot find module '@smithy/types' or its corresponding type declarations.
58.44     13 | } from '@odf/shared/s3';
58.44     14 | import { SecretKind } from '@odf/shared/types';
58.44   > 15 | import { HttpRequest } from '@smithy/types';
58.44        |                             ^^^^^^^^^^^^^^^

packages/shared/src/s3/commands.ts
```diff
@@ -85,7 +86,18 @@ export class S3Commands extends S3Client {
...
+        const proxyUrl = new URL(url);
+        proxyUrl.protocol = window.location.protocol;
+        proxyUrl.hostname = window.location.hostname;
+        proxyUrl.port = window.location.port;
+        proxyUrl.pathname = `${ODF_S3_PROXY_PATH}${proxyUrl.pathname}`;
```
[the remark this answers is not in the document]
alfonsomthd: We can, but although the behavior is similar, in 1 place we modify an `HttpRequest` object type and the other is an `URL` object type so we'll have to do type check as the `path` property is not identical on both types: we'll end up over-complicating things, let's keep it simple.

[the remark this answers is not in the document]
SanjalKatiyar: ack

package.json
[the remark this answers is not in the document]
SanjalKatiyar: okay ! name of the package is confusing, why call it `types` when it also contains runtime code.
Anyway, can you maybe try `import type { HttpRequest } from '@smithy/types';` (notice `import type`) after adding it to devDependency ??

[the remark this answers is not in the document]
alfonsomthd: Same error:
>60.09 ERROR in ../../packages/odf/components/s3-browser/noobaa-context.tsx:15:34
60.09 TS2307: Cannot find module '@smithy/types' or its corresponding type declarations.
60.09     13 | } from '@odf/shared/s3';
60.09     14 | import { SecretKind } from '@odf/shared/types';
60.09   > 15 | import type { HttpRequest } from '@smithy/types';
60.09        |                                  ^^^^^^^^^^^^^^^

packages/odf/components/s3-browser/noobaa-context.tsx
```diff
@@ -33,48 +48,73 @@ export const NoobaaS3Provider: React.FC<NoobaaS3ProviderType> = ({
...
+        // We must include the port in the host header as the proxy does (it's omitted
+        // if the port is the protocol default port e.g. 443 for 'https').
+        // It must be done BEFORE signature calculation.
+        client.middlewareStack.add(
+          (next) => (args) => {
+            const request: Partial<HttpRequest> = args.request;
+            if (s3Url.protocol === 'https:') {
+              request.headers[
+                'host'
+              ] = `${s3Url.hostname}:${S3_INTERNAL_ENDPOINT_PORT}`;
```
GowthamShanmugam, line 87: is appending S3_INTERNAL_ENDPOINT_PORT with s3Url on line 66 won't work?

alfonsomthd: In the comments I already explained the reason (the port is removed if it matches the default port for the protocol used). You can also try:
```
const url = new URL('https://localhost:443');
console.log(url.port);
```

GowthamShanmugam: oh! got it

PR #1851, opened 2025-02-05
Add support for Bucket Policy (S3)

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-6859

Screenshot 2025-02-05 at 4 09 04 PM
Screenshot 2025-02-05 at 4 09 35 PM
Screenshot 2025-02-05 at 4 09 53 PM
Screenshot 2025-02-05 at 4 10 07 PM
Screenshot 2025-02-05 at 4 10 24 PM


locales/en/plugin__odf-console.json
```diff
@@ -1176,6 +1176,27 @@
...
+  "Drag a file here, upload files, or start from scratch.": "Drag a file here, upload files, or start from scratch.",
+  "Start from scratch or use predefined policy configuration": "Start from scratch or use predefined policy configuration",
+  "Apply policy": "Apply policy",
+  "Save changes": "Save changes",
+  "Grant Public Read Access to All Objects": "Grant Public Read Access to All Objects",
+  "Allows anyone to read all objects in the bucket": "Allows anyone to read all objects in the bucket",
+  "Allow Access to a Specific S3 Account": "Allow Access to a Specific S3 Account",
+  "Grants full access to the bucket to another S3 account": "Grants full access to the bucket to another S3 account",
+  "Enforce Secure Transport (HTTPS)": "Enforce Secure Transport (HTTPS)",
+  "Denies access to the bucket if the reguest is not made over HTTPS": "Denies access to the bucket if the reguest is not made over HTTPS",
```
alfonsomthd, line 1194:
```suggestion
  "Denies access to the bucket if the request is not made over HTTPS": "Denies access to the bucket if the request is not made over HTTPS",
```

packages/odf/components/s3-browser/bucket-policy/PreConfiguredPolicies.tsx
```diff
@@ -0,0 +1,112 @@
...
+      </HelperText>
+    </>
+  ),
+  [Policies.EnforceSecureTransportHTTPS]: (
+    <>
+      {t('Enforce Secure Transport (HTTPS)')}
+      <HelperText>
+        <HelperTextItem variant="indeterminate">
+          {t(
+            'Denies access to the bucket if the reguest is not made over HTTPS'
```
alfonsomthd, line 51:
```suggestion
            'Denies access to the bucket if the request is not made over HTTPS'
```

packages/odf/components/s3-browser/bucket-policy/BucketPolicy.tsx
```diff
@@ -0,0 +1,373 @@
...
+  const noPolicyExists =
+    error?.name === 'NoSuchBucketPolicy' && !policyData?.Policy;
+
+  // initial fetch on first mount or remounts only
+  React.useEffect(() => {
+    const fetchPolicies = async () => {
+      try {
+        await trigger();
+      } catch (err) {
+        // no need to handle any error here, use "error" object directly from the "useSWRMutation" hook
```
alfonsomthd, line 305: If there is no error handling required here, can we drop the try-catch statement?

packages/odf/components/s3-browser/bucket-policy/BucketPolicy.tsx
```diff
@@ -0,0 +1,373 @@
...
+        );
+        setCode(formattedPolicy);
+      } catch {
+        setCode(unformattedPolicy);
+      }
+    }
+  }, [isLoading, error, policyData]);
+
+  if (isLoading || (error && !noPolicyExists)) {
+    return <StatusBox loaded={!isLoading} loadError={isLoading ? '' : error} />;
```
alfonsomthd, line 331: Is the loadError ternary operator really needed?
```suggestion
    return <StatusBox loaded={!isLoading} loadError={error} />;
```

packages/odf/modals/s3-browser/bucket-policy/DeleteBucketPolicy.tsx
```diff
@@ -0,0 +1,120 @@
+import * as React from 'react';
+import { ButtonBar } from '@odf/shared/generic/ButtonBar';
+import { CommonModalProps } from '@odf/shared/modals';
+import { S3Commands } from '@odf/shared/s3';
+import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
+import { TFunction } from 'react-i18next';
+import { Trans } from 'react-i18next';
```
alfonsomthd, line 7:
```suggestion
import { TFunction, Trans } from 'react-i18next';
```

packages/odf/components/s3-browser/bucket-policy/BucketPolicy.tsx
SanjalKatiyar: yeah it was needed earlier when I was using `useSWR` hook... but later I moved to `useSWRMutation` so not needed anymore. I will update, thanks !

packages/odf/components/s3-browser/bucket-policy/BucketPolicy.tsx
SanjalKatiyar: `trigger` still returns a promise which should be handled IMO, else browser will catch it and execute some default behaviour. In dev mode this default behaviour is to show error message on the screen itself (on top of your application window) (haven't checked in prod though).

alfonsomthd: @SanjalKatiyar what should be handled if there is no error handling implemented? I'm not sure what you mean.

alfonsomthd: What I mean is that the purpose of try-catch is defeated here: either we do error handling or is meaningless, right?

SanjalKatiyar: I can use`.catch` as well, but idea is to catch the failed promised. Else if `trigger` fails, it will be an uncaught promise and you will see this default behaviour from the browser everytime you navigate to this page (in dev mode):
Screenshot 2025-02-07 at 4 36 32 PM

alfonsomthd: So the intention is to silent the error: if it's handled at a later stage, then it's OK.

Issue #1874, opened 2025-02-14
Cancel icon not working ("Create new BackingStore" modal)


SanjalKatiyar:
Go to: "Storage > Object Storage > Bucket Class > Create Bucket Class > Select "BucketClass type" as "Standard", type any "BucketClass name" and select "Next" > On the third "Resources" step, select "Create BackingStore" > try closing the modal via cross/cancel icon on top-right corner of the modal"

Modal should get closed on clicking the cross/cancel icon.

<img width="1719" alt="Image" src="https://github.com/user-attachments/assets/882a3895-82ba-4d95-912c-ef00df1581df" />

PR #1876, opened 2025-02-20
Cancel icon not working ("Create new BackingStore" modal) #1874

NIKHITHAVADDEMPUDI:
Fixes: https://github.com/red-hat-storage/odf-console/issues/1874
Added "onClose={closeModal}" to the Modal in the file "odf-console/packages/odf/components/create-bs/create-bs-modal.tsx" to make the close icon work, after this change the close icon is working to close create new backing store.
 go to:
"Storage > Object Storage > Bucket Class > Create Bucket Class > Select "BucketClass type" as "Standard", type any "BucketClass name" and select "Next" > On the third "Resources" step, select "Create BackingStore" > try closing the modal via close icon on top-right corner of the modal"
Modal is now getting closed on clicking the cross/cancel icon.

Screenshot 2025-02-20 at 11 11 15 AM

### DIFF
--- a/packages/odf/components/create-bs/create-bs-modal.tsx
+++ b/packages/odf/components/create-bs/create-bs-modal.tsx
@@ -16,7 +16,12 @@ const CreateBackingStoreFormModal: React.FC<CreateBackingStoreFormModalProps> =
 
     const Header = <ModalHeader>{t('Create new BackingStore')}</ModalHeader>;
     return (
-      <Modal isOpen={isOpen} variant={ModalVariant.small} header={Header}>
+      <Modal
+        isOpen={isOpen}
+        variant={ModalVariant.small}
+        header={Header}
+        onClose={closeModal}
+      >
         <div className="nb-endpoints__modal">
           <ModalBody>
             <p>


weirdwiz:
thanks for the PR! 

can you rename the PR into something descriptive? it will help other people understand what is the issue about
also in the description for the PR, just add `Fixes: #[issue number]` in the end, so that github can link your PR to that issue, and it will automatically close it when your PR merges.

.DS_Store
weirdwiz, line 1: no need to add these files to the commit, as these are OS specific and not needed for the change :D

packages/.DS_Store
weirdwiz, line 1: same here :)

weirdwiz (changes requested):
Small nits, but rest looks okay to me :D 
cc @SanjalKatiyar

weirdwiz:
Can you also keep the title of the commit message subject line to 50 charecters? 
this is a good source for how to shape your commit message, https://cbea.ms/git-commit/

weirdwiz (changes requested):
Tiny nit, the commit message contains a stray 4 in the beginning. It would be nice to have the Subject line be a little imperative, and have some description of what this commit contains. 
Although for this commit the change seems straightforward, so you can also leave it at the subject line.

Rest of it LGTM

PR #1877, opened 2025-02-20
Add lifecycle rules list, edit and delete pages

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-6949
https://issues.redhat.com/browse/RHSTOR-6950

Screenshot 2025-02-25 at 12 30 30 PM
Screenshot 2025-02-25 at 12 30 40 PM
Screenshot 2025-02-25 at 12 35 02 PM


packages/odf/components/s3-browser/create-or-edit-lifecycle-rules/useEditLifecycleRule.ts
```diff
@@ -0,0 +1,164 @@
...
+    } else {
+      prefix = rule.Filter?.Prefix || '';
+      sizeGreater = rule.Filter?.ObjectSizeGreaterThan;
+      sizeLess = rule.Filter?.ObjectSizeLessThan;
+    }
+  }
+  ruleState.conditionalFilters.prefix = prefix;
+  ruleState.conditionalFilters.objectTags = tags;
+
+  if (!!sizeGreater && typeof sizeGreater === 'number') {
```
alfonsomthd, line 66: Optional:
```suggestion
  if (!!sizeGreater && !Number.isNaN(sizeGreater)) {
```

packages/odf/components/s3-browser/create-or-edit-lifecycle-rules/useEditLifecycleRule.ts
```diff
@@ -0,0 +1,164 @@
...
+    ruleState.conditionalFilters.minObjectSize.isChecked = true;
+    ruleState.conditionalFilters.minObjectSize.sizeInB = sizeGreater;
+    ruleState.conditionalFilters.minObjectSize.unit = SizeUnit.KiB;
+    ruleState.conditionalFilters.minObjectSize.size = humanizeBinaryBytes(
+      sizeGreater,
+      SizeUnit.B,
+      SizeUnit.KiB
+    ).value;
+  }
+  if (!!sizeLess && typeof sizeLess === 'number') {
```
alfonsomthd, line 76: Optional:
```suggestion
  if (!!sizeLess && !Number.isNaN(sizeLess)) {
```

packages/odf/components/s3-browser/create-or-edit-lifecycle-rules/useEditLifecycleRule.ts
alfonsomthd: Alternatively, there is the `_.isNumber()` lodash method.

alfonsomthd:
LGTM. @bipuladh @GowthamShanmugam to tag.

packages/odf/components/s3-browser/create-or-edit-lifecycle-rules/useEditLifecycleRule.ts
SanjalKatiyar: Using `isNumber` and `typeof sizeGreater === 'number'` is exactly same, right ?? Is there any instance/example where these will output different results ??

Even though API won't allow `sizeGreater` or `sizeLess` to be any value other than a number, still using `!Number.isNaN` seems conceptually incorrect to me. Ex: `!Number.isNaN('some string')` === `true` as well, which I something I do not want.

SanjalKatiyar: Am I missing anything ??

packages/odf/modals/s3-browser/delete-lifecyclerules/DeleteLifecycleRuleModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+              onClick={closeModal}
+              className="pf-v5-u-ml-xs"
+            >
+              {t('Cancel')}
+            </Button>
+          </span>
+        </ButtonBar>,
+      ]}
+    >
+      {' '}
```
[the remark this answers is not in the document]
SanjalKatiyar: yeah, `Modal` needs `children` as a required prop.

packages/odf/components/s3-browser/create-or-edit-lifecycle-rules/useEditLifecycleRule.ts
alfonsomthd: > Using `isNumber` and `typeof sizeGreater === 'number'` is exactly same, right ?? Is there any instance/example where these will output different results ??

AFAIK They're equal, just that `_.isNumber` is self-explanatory, shorter IMHO. But as I said above: it was an optional suggestion, not a requested change.
> Even though S3 won't allow `sizeGreater` or `sizeLess` to be any value other than a number, still using `!Number.isNaN` seems conceptually incorrect to me. Examples: `!Number.isNaN('some string')` === `true` and `!Number.isNaN(undefined)` === `true` as well, which is something I do not want.

:see_no_evil: Honestly I didn't expect this:
```
>  Number.isNaN('some string')
false
```

alfonsomthd: @SanjalKatiyar Bear in mind that `NaN` qualifies as a number, so maybe you want to add that additional check or even better using `Number.isFinite`:
```
> typeof NaN
'number'
> Number.isFinite(NaN)
false
> Number.isFinite(0)
true
> Number.isFinite('some string')
false
> Number.isFinite(Infinity)
false
```

SanjalKatiyar: ack !
just adding for ref: if by any chance say `sizeGreater === NaN`, then `!!sizeGreater &&` part of the check will handle that case !

packages/odf/components/s3-browser/create-or-edit-lifecycle-rules/useEditLifecycleRule.ts
alfonsomthd: Closed thread:
https://github.com/red-hat-storage/odf-console/pull/1877#discussion_r1969904894

packages/odf/components/s3-browser/create-or-edit-lifecycle-rules/useEditLifecycleRule.ts
alfonsomthd: > ack ! just adding for ref: if by any chance say `sizeGreater === NaN`, then `!!sizeGreater &&` part of the check will handle that case !

Right! But It  will not handle `Infinity` (:stuck_out_tongue_closed_eyes: I know very edgy case haha):
```
> !!Infinity
true
```

packages/odf/utils/s3-browser.ts
```diff
@@ -137,3 +139,38 @@ export const getNavigationURL = (
...
+      !ObjectSizeLessThan
+    ) {
+      return true;
+    }
+
+    return false;
+  }
+
+  if (
+    !!filter.Prefix ||
```
bipuladh, line 167: Isn't this covered in line 148?

packages/odf/modals/s3-browser/delete-lifecyclerules/DeleteLifecycleRuleModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+              variant={ButtonVariant.link}
+              onClick={closeModal}
+              className="pf-v5-u-ml-xs"
+            >
+              {t('Cancel')}
+            </Button>
+          </span>
+        </ButtonBar>,
+      ]}
+    >
```
bipuladh, line 126:
```suggestion
    children={''}/>
```
nit: Perhaps do this and use a single tag

packages/odf/modals/s3-browser/delete-lifecyclerules/DeleteLifecycleRuleModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+  isOpen,
+  extraProps: { noobaaS3, bucketName, mutate, ruleName, ruleHash },
+}) => {
+  const { t } = useCustomTranslation();
+
+  const [inProgress, setInProgress] = React.useState<boolean>(false);
+  const [error, setError] = React.useState<Error>();
+
+  const onDelete = async (event) => {
+    event.preventDefault();
```
bipuladh, line 63: What does it do by default?

packages/odf/modals/s3-browser/delete-lifecyclerules/DeleteLifecycleRuleModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+          {t(
+            'Deleting this lifecycle rule may prevent the removal of existing objects, potentially increasing storage costs.'
+          )}
+        </div>
+      }
+      variant={ModalVariant.medium}
+      actions={[
+        <ButtonBar
+          inProgress={inProgress}
+          errorMessage={error?.message || error}
```
bipuladh, line 105:
```suggestion
          errorMessage={error?.message || JSON.stringify(error)}
```
Should we do this just to be safe? I think usually errors are objects.

packages/odf/components/s3-browser/lifecycle-rules-list/LifecycleRulesList.tsx
```diff
@@ -0,0 +1,391 @@
...
+              loaded
+              hideLabelFilter={true}
+              onFilterChange={onFilterChange}
+              hideColumnManagement={true}
+            />
+            <RulesTable
+              data={filteredRules}
+              unfilteredData={rules}
+              loaded
+              loadError={null}
```
bipuladh, line 383:
```suggestion
              loadError={error}
```
Shouldn't this be error?

packages/odf/components/s3-browser/lifecycle-rules-list/LifecycleRulesList.tsx
```diff
@@ -0,0 +1,391 @@
...
+              rowData={{ navigate, launcher, mutate, noobaaS3, bucketName }}
```
bipuladh, line 384: Do we need to pass `navigate` here. I think we can just use it inside `RuleRow`.

packages/odf/components/s3-browser/lifecycle-rules-list/LifecycleRulesList.tsx
SanjalKatiyar: not needed... for such case we already have `StatusBox` above... `loadError` is a required prop, hence sending `null` instead...

packages/odf/components/s3-browser/lifecycle-rules-list/LifecycleRulesList.tsx
SanjalKatiyar: just saving some calls... using directly inside `RuleRow` works as well, but we will unnecessarily be initialising new instance of `useNavigate` hook per each row...
similar is true for `launcher` and `noobaaS3`...

packages/odf/modals/s3-browser/delete-lifecyclerules/DeleteLifecycleRuleModal.tsx
SanjalKatiyar: I did not try without it (just a copy/paste from other files), but isn't this something we are already doing in every modal/form that we have in our repo ??

packages/odf/utils/s3-browser.ts
SanjalKatiyar: this is different, we can have an entry like this as well: `{ Filter: { Prefix: '' }}`, which would still mean global scope. But, any other values like `undefined` etc would mean it's not global.

packages/odf/modals/s3-browser/delete-lifecyclerules/DeleteLifecycleRuleModal.tsx
SanjalKatiyar: ack... will update in next PR, I need to do it at few more places (other modals) as well...

packages/odf/modals/s3-browser/delete-lifecyclerules/DeleteLifecycleRuleModal.tsx
bipuladh: I think with forms we require this because of the `onSubmit` I am not sure if this is required.

packages/odf/components/s3-browser/lifecycle-rules-list/LifecycleRulesList.tsx
bipuladh: It should have shared scope. So I am not sure if a new instance will be generated. You will save some type imports. Leave the rest up to you.

SanjalKatiyar: `console.log(useNavigate() === useNavigate());`, this results `false`.

packages/odf/modals/s3-browser/delete-lifecyclerules/DeleteLifecycleRuleModal.tsx
SanjalKatiyar: make sense, we probably need to do this cleanup in multiple files (modals), even the "shared" modal has this line: https://github.com/red-hat-storage/odf-console/blob/master/packages/shared/src/modals/DeleteModal.tsx#L60

alfonsomthd: For `onSubmit` event makes sense, so in case of doing a cleanup let's analyze case by case.

PR #1878, opened 2025-02-24
[RDR] PVC label selector causes infinite blinking when applying DRPolicy

GowthamShanmugam:
**Fix: Prevent Unnecessary Re-renders in AppManagePoliciesModal Using useDeepCompareMemoize**

Screenshot 2025-02-17 at 6 10 29 PM

**Problem Context:**
The AppManagePoliciesModal component is invoked by multiple parsers, specifically application-set-parser.tsx and subscription-parser.tsx. These parsers watch different types of Application Custom Resources (CRs) along with Ramen CRs. As a result:

* **Multiple CRs trigger repeated renders:** Each parser watches separate resources, leading to frequent updates.
* **New** `applicationInfo` **objects are created on each render:** Even though the content of the object remains the same, React sees it as a new reference.
* **The modal re-renders unnecessarily,** causing UI issues, including the **PVC label selector blinking infinitely**, which was reported by the QE team.

**Solution:**
To prevent unnecessary re-renders, the `useDeepCompareMemoize` **hook is applied to** `applicationInfo`.

**Blocker:**
Yes

**Backport Required Upto:**
4.18

**Issue**
https://issues.redhat.com/browse/DFBUGS-1663

### DIFF
--- a/packages/mco/components/modals/app-manage-policies/app-manage-policies-modal.tsx
+++ b/packages/mco/components/modals/app-manage-policies/app-manage-policies-modal.tsx
@@ -18,7 +18,7 @@ import {
 import { ApplicationType, DRInfoType, DRPolicyType } from './utils/types';
 
 export const ModalContextViewer: React.FC<ModalContextViewerProps> = ({
-  applicaitonInfo,
+  applicationInfo,
   state,
   matchingPolicies,
   dispatch,
@@ -47,13 +47,13 @@ export const ModalContextViewer: React.FC<ModalContextViewerProps> = ({
     <>
       {state.modalViewContext === ModalViewContext.MANAGE_POLICY_VIEW && (
         <ManagePolicyView
-          drInfo={applicaitonInfo?.drInfo as DRInfoType}
-          workloadNamespace={applicaitonInfo?.workloadNamespace}
+          drInfo={applicationInfo?.drInfo as DRInfoType}
+          workloadNamespace={applicationInfo?.workloadNamespace}
           eligiblePolicies={matchingPolicies}
           isSubscriptionAppType={
-            applicaitonInfo?.type === DRApplication.SUBSCRIPTION
+            applicationInfo?.type === DRApplication.SUBSCRIPTION
           }
-          unProtectedPlacementCount={applicaitonInfo?.placements?.length}
+          unProtectedPlacementCount={applicationInfo?.placements?.length}
           dispatch={dispatch}
           setModalContext={setModalContext}
           setModalActionContext={setModalActionContext}
@@ -64,7 +64,7 @@ export const ModalContextViewer: React.FC<ModalContextViewerProps> = ({
       )}
       {state.modalViewContext === ModalViewContext.ASSIGN_POLICY_VIEW && (
         <AssignPolicyView
-          applicaitonInfo={applicaitonInfo}
+          applicationInfo={applicationInfo}
           matchingPolicies={matchingPolicies}
           state={state.assignPolicyView}
           modalActionContext={state.modalActionContext}
@@ -78,7 +78,7 @@ export const ModalContextViewer: React.FC<ModalContextViewerProps> = ({
 };
 
 export const AppManagePoliciesModal: React.FC<AppManagePoliciesModalProps> = ({
-  applicaitonInfo,
+  applicationInfo,
   matchingPolicies,
   loaded,
   loadError,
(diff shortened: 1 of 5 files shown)


packages/mco/components/modals/app-manage-policies/app-manage-policies-modal.tsx
```diff
@@ -78,13 +79,15 @@ export const ModalContextViewer: React.FC<ModalContextViewerProps> = ({
...
-  applicaitonInfo,
+  applicationInfo: appInfo,
   matchingPolicies,
   loaded,
   loadError,
   isOpen,
   close,
 }) => {
+  // Perform deep comparison to prevent unnecessary re-renders
+  const applicationInfo = useDeepCompareMemoize(appInfo);
```
SanjalKatiyar, line 90:
```suggestion
  const applicationInfo = useDeepCompareMemoize(appInfo, true);
```
`appInfo` is a very deep object (contains multiple CRs)... can you try with `true` argument and check if that works ??

packages/mco/components/modals/app-manage-policies/app-manage-policies-modal.tsx
```diff
@@ -114,7 +117,7 @@ export const AppManagePoliciesModal: React.FC<AppManagePoliciesModalProps> = ({
       onClose={close}
     >
       <ModalContextViewer
-        applicaitonInfo={applicaitonInfo}
+        applicaitonInfo={applicationInfo}
```
alfonsomthd, line 120:
```suggestion
        applicationInfo={applicationInfo}
```

packages/mco/components/modals/app-manage-policies/app-manage-policies-modal.tsx
```diff
@@ -78,13 +79,15 @@ export const ModalContextViewer: React.FC<ModalContextViewerProps> = ({
 };
 
 export const AppManagePoliciesModal: React.FC<AppManagePoliciesModalProps> = ({
-  applicaitonInfo,
+  applicationInfo: appInfo,
```
bipuladh, line 82: Perhaps this could be passed after memoization at a higher level.

bipuladh (approved):

packages/mco/components/modals/app-manage-policies/app-manage-policies-modal.tsx
alfonsomthd: @SanjalKatiyar does this provide better performance?

bipuladh:
Minor comment.
/lgtm cancel

packages/mco/components/modals/app-manage-policies/app-manage-policies-modal.tsx
GowthamShanmugam: it's a small object, it wont impact the performance

SanjalKatiyar: looks like I was wrong: https://www.samanthaming.com/tidbits/33-how-to-compare-2-objects/#json-stringify-vs-lodash-s-isequal-performance

SanjalKatiyar: for small objects stringfy is better and for large nested objects lodash is better !

packages/mco/components/modals/app-manage-policies/app-manage-policies-modal.tsx
GowthamShanmugam: Right, for now, the problem is only for ApplicaitonSet parser, I dont have to keep this for all.

packages/mco/components/modals/app-manage-policies/app-manage-policies-modal.tsx
GowthamShanmugam: i will check with true option

GowthamShanmugam: Based on the comment: [[GitHub link](https://github.com/red-hat-storage/odf-console/pull/1878#discussion_r1969562795)](https://github.com/red-hat-storage/odf-console/pull/1878#discussion_r1969562795), I have moved the memoization one step back. The issue specifically occurs for `appset`, so I have relocated the memoization logic inside `application-set-parser.tsx`.  

In this section ([[GitHub link](https://github.com/red-hat-storage/odf-console/pull/1878/files#diff-ecc0b5ba3f2a4cf0a22eec5c4aed84f598b5114bfb31fcb8acee4ecee28ddf93R105)](https://github.com/red-hat-storage/odf-console/pull/1878/files#diff-ecc0b5ba3f2a4cf0a22eec5c4aed84f598b5114bfb31fcb8acee4ecee28ddf93R105)), the object size is slightly larger, so I have used Lodash for optimization.

Issue #1873, opened 2025-02-14
Add margin/padding ("Create storage pool" page)


SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-7022

Go to: "Storage > Data Foundation > Overview > Click on StorageSystem name > Storage pools > Create storage pool".

Add top margin or padding between the header and "Volume type" label.

<img width="1668" alt="Image" src="https://github.com/user-attachments/assets/ae256b23-69fa-40e7-849d-f58350af7117" />

PR #1879, opened 2025-02-25
Add spacing class in body.tsx to insert space between the header and …

NIKHITHAVADDEMPUDI:
Add padding ("Create storage pool" page) #1873

https://issues.redhat.com/browse/RHSTOR-7022
Go to: "Storage > Data Foundation > Overview > Click on StorageSystem name > Storage pools > Create storage pool".

Add spacing class in odf-console/packages/ocs/storage-pool/body.tsx to insert space between the header and volume type label
Screenshot 2025-02-25 at 12 35 52 PM

### DIFF
--- a/packages/ocs/storage-pool/body.tsx
+++ b/packages/ocs/storage-pool/body.tsx
@@ -261,7 +261,7 @@ export const StoragePoolBody: React.FC<StoragePoolBodyProps> = ({
 
   return isClusterReady || !showPoolStatus ? (
     <>
-      <FormGroup label={t('Volume type')} isRequired>
+      <FormGroup label={t('Volume type')} className="pf-v6-u-pt-xl" isRequired>
         <div className="pf-v5-u-display-flex pf-v5-u-flex-direction-row ceph-pool__radio-flex">
           <Radio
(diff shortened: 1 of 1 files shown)


packages/ocs/storage-pool/body.tsx
```diff
@@ -261,6 +261,7 @@ export const StoragePoolBody: React.FC<StoragePoolBodyProps> = ({
 
   return isClusterReady || !showPoolStatus ? (
     <>
+      <div className="ws-example-flex-item pf-v6-u-pt-xl"></div>
       <FormGroup label={t('Volume type')} isRequired>
```
SanjalKatiyar, line 265:
```suggestion
      <FormGroup label={t('Volume type')} className='pf-v5-u-pt-xl' isRequired>
```

SanjalKatiyar: Few things:
1. `FormGroup` supports `className` prop, so extra `div` is not needed.
2. We don't need `ws-example-flex-item`.
3. We are currently using PatternFly `v5` (not migrated to `v6` yet).

PR #1885, opened 2025-02-27
Add versioning support for S3

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-6861

Screenshot 2025-03-10 at 6 41 32 PM
Screenshot 2025-03-10 at 6 41 43 PM
Screenshot 2025-03-10 at 6 41 55 PM
Screenshot 2025-03-10 at 6 42 35 PM
Screenshot 2025-03-10 at 6 42 47 PM
Screenshot 2025-03-10 at 6 42 59 PM
Screenshot 2025-03-10 at 6 43 23 PM


packages/shared/src/s3/utils.ts
```diff
@@ -1,2 +1,26 @@
...
+  versioningData: GetBucketVersioningCommandOutput,
+  t: TFunction
+): string => {
+  if (!versioningData?.Status) return t('Disabled');
+  return versioningData.Status === BucketVersioningStatus.Enabled
+    ? t('Enabled')
+    : t('Suspended');
+};
+
+export const getIsVersioningEnabled = (
```
alfonsomthd, line 20:
```suggestion
export const isVersioningEnabled = (
```

packages/shared/src/s3/utils.ts
```diff
@@ -1,2 +1,26 @@
...
+  versioningData: GetBucketVersioningCommandOutput
+) => versioningData?.Status === BucketVersioningStatus.Enabled;
+
+export const getIsVersioningSuspended = (
```
alfonsomthd, line 24:
```suggestion
export const isVersioningSuspended = (
```

packages/odf/modals/s3-browser/delete-objects/DeleteObjectsModal.tsx
```diff
@@ -57,19 +63,70 @@ const getTextInputLabel = (t: TFunction) => (
...
+  if (showVersioning)
+    return isMultiDelete ? t('Delete versions?') : t('Delete version?');
+
+  return isMultiDelete ? t('Delete objects?') : t('Delete object?');
+};
+
+const getDescription = (showVersioning: boolean, t: TFunction) => {
+  if (showVersioning) {
+    return (
+      <Trans t={t} ns="plugin__odf-console">
```
alfonsomthd, line 81: @SanjalKatiyar provided this is going to be used by another plugin (client), can we skip setting `ns` ?

packages/odf/modals/s3-browser/delete-objects/DeleteObjectsModal.tsx
```diff
@@ -219,6 +346,32 @@ const DeleteObjectsModal: React.FC<
           )}
         </ListFilter>
       </div>
+      {allowPermanentDeletion && (
+        <Checkbox
+          label={
+            <Trans t={t} ns="plugin__odf-console">
```
alfonsomthd, line 352: https://github.com/red-hat-storage/odf-console/pull/1885/files#r1989438104

packages/shared/src/s3/utils.ts
SanjalKatiyar: this is intentional, I am using `isVersioningEnabled` as a variable name to store the value returned by the function. Hence, function itself is called `get...`.

packages/shared/src/s3/utils.ts
SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/1885#discussion_r1991116033

alfonsomthd:
LGTM. @bipuladh @GowthamShanmugam to tag.

packages/odf/components/s3-browser/bucket-details/BucketDetails.tsx
```diff
@@ -85,7 +100,7 @@ const getEncryptionDescription = (
   }
 };
 
-const BucketDetailsOverview: React.FC<{}> = ({}) => {
+const S3BucketDetailsOverview: React.FC<{}> = ({}) => {
```
bipuladh, line 103:
```suggestion
const S3BucketDetailsOverview: React.FC = () => {
```
Just curious why we need to use those curly braces?

packages/odf/components/s3-browser/object-details/ObjectDetailsSidebar.tsx
```diff
@@ -38,53 +47,252 @@ import {
...
+    }
+
+    allObjects.sort((a, b) => {
+      let lastModifiedB: string | number = b?.apiResponse?.lastModified;
+      let lastModifiedA: string | number = a?.apiResponse?.lastModified;
+      if (lastModifiedB === DASH || !lastModifiedB) lastModifiedB = 0;
+      if (lastModifiedA === DASH || !lastModifiedA) lastModifiedA = 0;
+      return (
+        new Date(lastModifiedB).getTime() - new Date(lastModifiedA).getTime()
+      );
```
bipuladh, line 154: Can we move this to a sort function outside? Maybe creation a function named `sortByLastModified`?

packages/odf/components/s3-browser/objects-list/useObjectsList.ts
```diff
@@ -0,0 +1,284 @@
...
+          );
+        }
+      );
+      dataLatestVersions?.Contents?.forEach((content: Content) => {
+        objects.push(convertObjectDataToCrFormat(content, t, false, false));
+      });
+    }
+
+    return objects;
+  }, [dataLatestVersions, dataAllVersions, loadedWOError, listAllVersions, t]);
```
bipuladh, line 171: Should you be doing a string comparison for `dataLatestVersions` it might not be referentially equal? Does useSWR poll or only performs one request? If it performs one request do we really need useMemo?

packages/odf/components/s3-browser/bucket-details/BucketDetails.tsx
SanjalKatiyar: no particular reason, copy-paste from other files I guess (we are using that format other places too)... anything is fine (using and not using it)...

packages/odf/components/s3-browser/objects-list/useObjectsList.ts
SanjalKatiyar: `useSWRMutation` will perform single request, every time `trigger` is called with pagination params.

`useObjectsList` hook itself can re-render due to other reasons (not just because of useSWR), `useMemo` is used to keep ref of `structuredObjects` constant and to make sure `convertObjectDataToCrFormat` is not called again and again unnecessarily.

packages/odf/components/s3-browser/object-details/ObjectDetailsSidebar.tsx
SanjalKatiyar: ack !

PR #1888, opened 2025-03-04
Add support for SC distribution modal for storage clients

bipuladh:
Screenshot 2025-03-05 at 3 08 50 PM


packages/shared/src/skeletal-loader/tableSkeleton.scss
```diff
@@ -0,0 +1,9 @@
+.odf-skeleton-table__skeletal-loader {
+    display: flex;
+    justify-content: space-between;
+}
+
+.odf-skeleton-table-skeletal-loader__item {
+    margin-left: 2px;
+    margin-right: 2px;
+}
```
SanjalKatiyar, line 9: nits: u can directly use this instead of creating a new scss file
`pf-v5-u-display-flex pf-v5-u-justify-content-space-between`
`pf-v5-u-mx-sm`

packages/shared/src/storage-classes/constants.ts
```diff
@@ -0,0 +1,14 @@
...
+  'noobaa.io/obc',
+  'ceph.rook.io/bucket',
+];
+
+// Block and file storage provisioners
+export const CEPH_PROVISIONERS = [
+  'ceph.rook.io/block',
+  'cephfs.csi.ceph.com',
+  'rbd.csi.ceph.com',
+];
```
SanjalKatiyar, line 14: IMHO this should not be under its own folder, will be very hard to find/locate this way... highly likely we will add here and forget about it...
 should be part of `shared/constants` directory directly...

packages/shared/src/storage-classes/selectors.ts
```diff
@@ -0,0 +1,4 @@
+import { StorageClassResourceKind } from '../types';
+
+export const getProvisioner = (storageClass: StorageClassResourceKind) =>
+  storageClass?.provisioner;
```
SanjalKatiyar, line 4: same here... we already have `shared/selectors` directory... u can create a separate `storage-class.tsx` file there, instead of a new folder here...

packages/shared/src/storage-classes/utils.ts
```diff
@@ -0,0 +1,8 @@
+import * as _ from 'lodash-es';
+import { CEPH_PROVISIONERS } from './constants';
+
+export const isCephProvisioner = (scProvisioner: string): boolean => {
+  return CEPH_PROVISIONERS.some((provisioner: string) =>
+    _.endsWith(scProvisioner, provisioner)
+  );
+};
```
SanjalKatiyar, line 8: same here... should be part of `shared/utils` directory...

packages/odf/components/ResourceDistribution/ResourceDistributionModal.tsx
```diff
@@ -0,0 +1,251 @@
...
+              {t('Save changes')}
+            </Button>
+          </FlexItem>
+        </Flex>
+      </ModalFooter>
+    </Modal>
+  );
+};
+
+export default DistributeResourceModal;
```
SanjalKatiyar, line 251: as per our current structure we have `odf/modals` directory, so ideally this directory should be part of that...

packages/odf/components/ResourceDistribution/ResourceDistributionModal.tsx
```diff
@@ -0,0 +1,251 @@
...
+    <Tr key={getUID(storageClass)}>
+      <Td
+        select={{
+          rowIndex,
+          onSelect: (_event, isSelecting) => onSelect(isSelecting),
+          isSelected,
+        }}
+      />
+      <Td>{getName(storageClass)}</Td>
+      <Td>{storageClass?.provisioner}</Td>
```
SanjalKatiyar, line 75:
```suggestion
      <Td>{getProvisioner(storageClass)}</Td>
```

packages/odf/components/ResourceDistribution/ResourceDistributionModal.tsx
```diff
@@ -0,0 +1,251 @@
...
+      })
+      .catch((err) => {
+        setProgress(false);
+        setError(err.message);
+      });
+  };
+  return (
+    <Modal
+      className="odf-modal"
+      title={'Manage distribution of resources'}
```
SanjalKatiyar, line 198:
```suggestion
      title={t('Manage distribution of resources')}
```

packages/odf/components/ResourceDistribution/ResourceDistributionModal.tsx
```diff
@@ -0,0 +1,251 @@
...
+            loaded={isVolumeSnaphostClassDataLoaded}
+            columns={[t('Name'), t('Driver'), t('Deletion policy')]}
+            resourceType="volumeSnapshotClass"
+          />
+        </Tab>
+      </Tabs>
+      <ModalFooter inProgress={inProgress} errorMessage={error}>
+        <Flex direction={{ default: 'row' }}>
+          <FlexItem>
+            <Button key="Cancel" variant="secondary" onClick={closeModal}>
```
SanjalKatiyar, line 236:
```suggestion
            <Button key="Cancel" variant={ButtonVariant.secondary} onClick={closeModal}>
```

packages/odf/components/ResourceDistribution/ResourceDistributionModal.tsx
```diff
@@ -0,0 +1,251 @@
...
+              {t('Cancel')}
+            </Button>
+          </FlexItem>
+          <FlexItem>
+            <Button key="Confirm" variant="primary" onClick={onConfirm}>
```
SanjalKatiyar, line 241:
```suggestion
            <Button key="Confirm" variant={ButtonVariant.primary} onClick={onConfirm}>
```

packages/odf/components/ResourceDistribution/resourceDistributionTable.scss
```diff
@@ -0,0 +1,9 @@
+.odf-skeleton-table__skeletal-loader {
+    display: flex;
+    justify-content: space-between;
+}
+
+.odf-skeleton-table-skeletal-loader__item {
+    margin-left: 2px;
+    margin-right: 2px;
+}
```
SanjalKatiyar, line 9: nits: u can directly use this instead of creating a new scss file
`pf-v5-u-display-flex pf-v5-u-justify-content-space-between`
`pf-v5-u-mx-sm`

packages/odf/components/ResourceDistribution/ResourceDistributionTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+  return !loaded ? (
+    <TableSkeletonLoader columns={4} rows={8} />
+  ) : (
+    <>
+      <ListPageFilter
+        data={unfileteredData}
+        loaded={loaded}
+        onFilterChange={onFilterChange}
+        hideColumnManagement={true}
+      />
```
SanjalKatiyar, line 104: IIRC `ListPageFilter` directly mutates your Route (adds query params) for filtering, plz test once that filtering in one Tab (say StorageClasses) won't affect the other Tab (VolumeSnapshotClasses in this case)...

packages/odf/components/ResourceDistribution/ResourceDistributionModal.tsx
```diff
@@ -0,0 +1,251 @@
...
+        closeModal();
+      })
+      .catch((err) => {
+        setProgress(false);
+        setError(err.message);
+      });
+  };
+  return (
+    <Modal
+      className="odf-modal"
```
SanjalKatiyar, line 197: is this file importing this class `odf-modal` ??

packages/odf/components/ResourceDistribution/ResourceDistributionTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+      <Table variant="compact" data-test-id="resource-distribution-table">
```
SanjalKatiyar, line 105:
```suggestion
      <Table variant={TableVariant.compact} data-test-id="resource-distribution-table">
```

packages/shared/src/storage-classes/constants.ts
SanjalKatiyar: under `shared/constants` u could create a separate `storage-classes.ts` file...

packages/odf/components/ResourceDistribution/ResourceDistributionModal.tsx
```diff
@@ -0,0 +1,251 @@
...
+        <Tab
+          eventKey={1}
+          title={<TabTitleText>{t('VolumeSnapshot classes')}</TabTitleText>}
+        >
+          <ResourceDistributionTable
+            resources={filteredVolumeSnapshotClasses}
+            selectedResources={selectedResources}
+            setSelectedResources={setSelectedResources}
+            RowGenerator={VolumeSnapshotClassRowGenerator}
+            loaded={isVolumeSnaphostClassDataLoaded}
```
alfonsomthd, line 227:
```suggestion
            loaded={isVolumeSnapshotClassDataLoaded}
```

packages/odf/components/ResourceDistribution/ResourceDistributionTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+    [resourceType, selectedResources, setSelectedResources]
+  );
+
+  const isRowSelected = React.useCallback(
+    (resource: K8sResourceCommon) =>
+      selectedResources?.[getName(resource)]?.selected,
+    [selectedResources]
+  );
+
+  const [unfileteredData, filteredData, onFilterChange] =
```
alfonsomthd, line 81:
```suggestion
  const [unfilteredData, filteredData, onFilterChange] =
```

packages/shared/src/skeletal-loader/TableSkeleton.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <>
+      {Array.from({ length: rows }).map((_item1, index1) => (
+        <div>
+          <div className="odf-skeleton-table__skeletal-loader">
+            {Array.from({ length: columns }).map((_item2, index2) => (
+              <Skeleton
+                key={index1 + index2}
+                width="80%"
+                className="odf-skeleton-table-skeletal-loader__item"
```
alfonsomthd, line 24: The loader item is still an element, right?
```suggestion
                className="odf-skeleton-table__skeletal-loader-item"
```

packages/odf/components/ResourceDistribution/ResourceDistributionTable.tsx
bipuladh: tested it they are mutually exclusive. the two instances

packages/odf/components/ResourceDistribution/ResourceDistributionModal.tsx
bipuladh: `  variant?: 'primary' | 'secondary' | 'tertiary' | 'danger' | 'warning' | 'link' | 'plain' | 'control';` 
This is the type definition so we can use string and still retain typescript checks.

packages/odf/components/ResourceDistribution/ResourceDistributionModal.tsx
bipuladh: `  variant?: 'primary' | 'secondary' | 'tertiary' | 'danger' | 'warning' | 'link' | 'plain' | 'control';` 
This is the type definition so we can use string and still retain typescript checks.

packages/odf/components/ResourceDistribution/ResourceDistributionTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+            <RowGenerator
+              key={resource.metadata.uid}
+              resource={resource}
+              rowIndex={index}
+              onSelect={onSelectRow(resource)}
+              isSelected={isRowSelected(resource)}
+            />
+          ))}
+        </Tbody>
+      </Table>
```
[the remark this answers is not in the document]
bipuladh: I looked into it. In my use case it slightly doesn't fit. I will have to refactor that component a bit which will cascade to all the components that use `selectable-table`

packages/odf/components/ResourceDistribution/ResourceDistributionTable.tsx
bipuladh: Same here as button: 
`  variant?: TableVariant | 'compact';`

packages/shared/src/skeletal-loader/TableSkeleton.tsx
bipuladh: I think this is correct. `odf-skeleton-table-skeletal-loader` would be the block and `item` would be element here.

packages/shared/src/skeletal-loader/tableSkeleton.scss
bipuladh: I want to avoid these classes. IIRC the upgrade tool last time dint pick up class names used inside of tsx files.

packages/odf/components/ResourceDistribution/ResourceDistributionModal.tsx
bipuladh: removed

packages/shared/src/storage-classes/constants.ts
bipuladh: moved

packages/shared/src/skeletal-loader/tableSkeleton.scss
alfonsomthd: @bipuladh @SanjalKatiyar me must  achieve an agreement about encouraging/discouraging those classes for code style consistency, as 1 reviewer can request to use it and not the other...

packages/shared/src/skeletal-loader/TableSkeleton.tsx
alfonsomthd: Then you're not following the same pattern as per the above class `odf-skeleton-table__skeletal-loader` it's clear that the common block is `odf-skeleton-table`. It's better to follow a consistent pattern.

packages/odf/components/ResourceDistribution/ResourceDistributionModal.tsx
alfonsomthd: @bipuladh as per my understanding @SanjalKatiyar is requesting this to maintain the code style and consistency in all files. We must reach out an agreement here either to continue to enforce it or not because usually all contributors are requested to use the proper enums , so we should all follow it (or drop the enforcement).

packages/shared/src/skeletal-loader/tableSkeleton.scss
SanjalKatiyar: IMO we should take advantage of utility classes provided by the library where ever we can. Basic CSS like margin/padding/flex/width/height etc can easily be handled without our custom code (saves us from maintaining our own files).

packages/odf/components/ResourceDistribution/ResourceDistributionTable.tsx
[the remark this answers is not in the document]
SanjalKatiyar: ack !

packages/shared/src/skeletal-loader/TableSkeleton.tsx
bipuladh: my bad updating

packages/odf/components/ResourceDistribution/ResourceDistributionModal.tsx
bipuladh: @alfonsomthd I understand that. If the goal is to maintain a consistent coding style, this should be formally documented in a style guide, and the enforcement should be clear to all contributors. It should not be a subjective decision in individual PRs. 
Let's discuss this with everyone and work on a style guide. I see that we are already using `primary` in our codebase as well.

packages/shared/src/skeletal-loader/tableSkeleton.scss
bipuladh: Removing scss file and using utility but we should use a style guide.

packages/odf/components/ResourceDistribution/ResourceDistributionModal.tsx
alfonsomthd: @bipuladh I agree on having a style guide document (I proposed it when I started contributing) but I have seen this request about using the proper enums in all contributor PRs without exception. @SanjalKatiyar we can move forward as it's a minor issue but let's set up a meeting to enforce aminimum set of styling rules.

PR #1856, opened 2025-02-12
RHSTOR-6968: Disaster recovery UI for Kubevirt VM - managed application

GowthamShanmugam:
ACM custom extension: https://github.com/stolostron/console/pull/4252

**Existing Flow:**
The parser always triggers the modal, even while the watcher is fetching data, displaying it with a loading: false argument.

Screenshot 2025-02-17 at 6 10 29 PM

**Problem in existing flow for VM:** (Delay in modal loading)
The application parser is determined only after the VM parser identifies the appropriate parser to call. This causes a delay in modal loading, making the modal appear a few seconds after the user clicks the button.

Screenshot 2025-02-17 at 6 10 43 PM

**New Flow:**
The modal will always load, while the modal body will be rendered based on the parsers.

Screenshot 2025-02-17 at 6 24 27 PM

PR #1889, opened 2025-03-05
RHSTOR-7003: Disaster recovery UI for Kubevirt VM - Discovered(Standalone)

GowthamShanmugam:
On top of: https://github.com/red-hat-storage/odf-console/pull/1856


GowthamShanmugam:
Screenshot 2025-03-12 at 2 36 51 PM
Screenshot 2025-03-12 at 2 36 54 PM
Screenshot 2025-03-12 at 2 36 58 PM

packages/mco/components/modals/app-manage-policies/app-manage-policies-modal.tsx
```diff
@@ -16,19 +17,35 @@ export const AppManagePoliciesModal: React.FC<AppManagePoliciesModalProps> = ({
...
 
-  const application =
-    'apigroup' in resource ? convertSearchResult(resource) : resource;
+  // Problem: Without useCallback, the function reference changes on each render,
+  // causing child components to re-render unnecessarily.
+  // Fix: useCallback ensures the function has a stable reference.
+  const setCurrentModalContext = React.useCallback(
+    (context: ModalViewContext) => setModalContext(context),
+    []
+  );
```
SanjalKatiyar, line 30: why do you need `setCurrentModalContext` when you already have `setModalContext` ??

Ref of `setModalContext` will always be stable and you can use it similarly by passing `setModalContext(context)`, there is no need for  `setCurrentModalContext`.

packages/mco/components/modals/app-manage-policies/parsers/discovered-vm-parser.tsx
```diff
@@ -0,0 +1,123 @@
...
+  PVCQueryFilter,
+} from '../utils/types';
+
+export const DiscoveredVMParser: React.FC<DiscoveredVMParserProps> = ({
+  virtualMachine,
+  cluster,
+  pvcQueryFilter,
+  setCurrentModalContext,
+}) => {
+  // Fetch resources
```
SanjalKatiyar, line 44:
```suggestion
```

SanjalKatiyar: it's understood...

packages/shared/src/input-with-requirements/nameInputWithValidation.tsx
```diff
@@ -0,0 +1,107 @@
...
+  name: string;
+  existingNames?: string[];
+  label?: string;
+  helperText?: string;
+  fieldId?: string;
+  dataTest?: string;
+  onChange: (text: string) => void;
+};
+
+export default NameInput;
```
SanjalKatiyar, line 107:
```suggestion
export default NameInputWithValidation;
```

SanjalKatiyar: file name should be same as well, i.e., `NameInputWithValidation.tsx`...

packages/shared/src/input-with-requirements/nameInputWithValidation.tsx
```diff
@@ -0,0 +1,107 @@
...
+    control,
+    watch,
+    formState: { isValid },
+  } = useForm({ ...formSettings, resolver });
+
+  const newName = watch(fieldId);
+
+  React.useEffect(() => {
+    onChange(isValid ? newName : '');
+  }, [isValid, name, onChange, newName]);
```
SanjalKatiyar, line 68: `textInputProps` has a `onChange` prop, do we need `useEffect` explicitly ??

SanjalKatiyar: what's the use of this `useEffect` ??

SanjalKatiyar, line 68: this change doesn't seems like something which should be part of `shared` component...

SanjalKatiyar, line 68:
```suggestion
  React.useEffect(() => {
    onChange(isValid ? newName : '');
  }, [isValid, onChange, newName]);
```

SanjalKatiyar, line 68: make sure ref of `onChange` is stable if we are adding it to the dep list of the hook...

packages/mco/components/modals/app-manage-policies/app-manage-policies-modal.tsx
GowthamShanmugam: The app-manage-policies-modal.tsx file contains a parent component. Any state changes in this parent trigger re-renders of all child components, including parsers — which handle most of the DR-related logic.

I need to update the header text based on whether the modal is in the "Assign Policy" view or the "DR Info" view. To manage this, I’ve introduced setCurrentModalContext and currentModalContext in the parent component.

However, directly updating modalContext causes the parent and all its children to re-render, which is a performance-heavy operation. To avoid this, I wrapped the state update logic inside setCurrentModalContext, ensuring only the header text updates without triggering unnecessary re-renders across the entire component tree.

i also `AppManagePoliciesModalBody` inside `usememo` so it wont trigger re-render until unless any argument is changed `application, cluster, setCurrentModalContext`

SanjalKatiyar: not sure if I follow, but to me `setCurrentModalContext` and `setModalContext` are exactly same and can be used interchangeably, refs of both are stable too...

GowthamShanmugam: You are right, I misunderstood. Both are the same. The fix is wrapping AppManagePoliciesModalBody useMemo

packages/shared/src/input-with-requirements/nameInputWithValidation.tsx
GowthamShanmugam: The design of `TextInputWithFieldRequirements` relies on receiving the `TextInput` value from the `watch` function, which is why `useEffect` is used throughout. It’s reset to an empty value intentionally to prevent the user from progressing to the next wizard step, as the validation condition for this step is `name !== ''`.

GowthamShanmugam: it will work for all, https://github.com/red-hat-storage/odf-console/pull/1889#discussion_r2002786865

SanjalKatiyar: I got your point that other places are using `useEffect` as well, but they are just misusing the hook... a generic component should not make same mistake again... instead of `useEffect` u can utilise the `onChange` prop...
check: https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/attach-storage-storagesystem/storageclass-form.tsx#L252-L257

GowthamShanmugam: i added useRef for onChange, please check

GowthamShanmugam: I reviewed this logic. The val is compared with isValid before the new val is fully validated, meaning isValid may still hold an outdated result. Only after the state updates does isValid refresh, but by then, the state value is already set incorrectly.

SanjalKatiyar: ack !!

packages/shared/src/input-with-requirements/nameInputWithValidation.tsx
```diff
@@ -0,0 +1,112 @@
...
+    watch,
+    formState: { isValid },
+  } = useForm({ ...formSettings, resolver });
+
+  const newName = watch(fieldId);
+
+  const stableOnChange = React.useRef(onChange);
+  React.useEffect(() => {
+    stableOnChange.current = onChange;
+  }, [onChange]);
```
SanjalKatiyar, line 69: do we need `useEffect` here ?? can't we do `stableOnChange.current = onChange` directly during render...

locales/en/recipeNamespace.json
```diff
@@ -0,0 +1,3 @@
+{
+  "Recipe namespace: {{namespace}}": "Recipe namespace: {{namespace}}"
+}
```
SanjalKatiyar, line 3: why do we have this file ??

packages/shared/src/input-with-requirements/nameInputWithValidation.tsx
```diff
@@ -0,0 +1,112 @@
...
+  name: string;
+  existingNames?: string[];
+  label?: string;
+  helperText?: string;
+  fieldId?: string;
+  dataTest?: string;
+  onChange: (text: string) => void;
+};
+
+export default NameInputWithValidation;
```
SanjalKatiyar, line 112: format for this file name is wrong... u missed this comment: https://github.com/red-hat-storage/odf-console/pull/1889#discussion_r2000403353

locales/en/recipeNamespace.json
GowthamShanmugam: oh! something wrong

packages/shared/src/input-with-requirements/nameInputWithValidation.tsx
GowthamShanmugam: We can assign it directly too — both approaches ensure stableOnChange.current(isValid ? newName : '') doesn’t run repeatedly. I added useEffect to prevent re-assignment of stableOnChange.current on each re-render caused by other state or prop changes.

Both are correct, this useeffect adds some extra safety

locales/en/recipeNamespace.json
GowthamShanmugam: the argument name `namespace ` is the problem

packages/mco/components/modals/app-manage-policies/utils/reducer.ts
```diff
@@ -1,4 +1,4 @@
-import { DRPolicyType } from './types';
+import { DRPolicyType, VMProtectioType } from './types';
```
alfonsomthd, line 1:
```suggestion
import { DRPolicyType, VMProtectionType } from './types';
```

packages/shared/src/input-with-requirements/nameInputWithValidation.tsx
```diff
@@ -0,0 +1,110 @@
+import * as React from 'react';
```
GowthamShanmugam, line 1: strange in my local it is `NameInputWithValidation.tsx` only

GowthamShanmugam:
> /override ci/prow/odf-console-e2e-aws

MCO dont have e2e

PR #1896, opened 2025-03-10
DFBUGS-1627: Add contrast between text and background

NIKHITHAVADDEMPUDI:
[DFBUGS-1627](https://issues.redhat.com//browse/DFBUGS-1627): Add contrast between text and background

https://issues.redhat.com/browse/DFBUGS-1627  
  
  Add a title component with fill: `var(--pf-v5-global--Color--100)` in file packages/ocs/dashboards/common/capacity-card/capacity-card.tsx  to add contrast between text and background

pie graph text has minimal contrast  before- is dark against a dark background

After the above change 
Screenshot 2025-03-10 at 12 58 58 PM
Screenshot 2025-03-10 at 12 55 14 PM


SanjalKatiyar:
/retitle DFBUGS-1627: Add contrast between text and background

PR #1899, opened 2025-03-11
RHSTOR-7003: PVC and VM label update for DR protection

GowthamShanmugam:
On top of: https://github.com/red-hat-storage/odf-console/pull/1889
Corresponding ramen backend PR: https://github.com/RamenDR/ramen/pull/190

---

- The UI **labels** the VM and PVCs with `ramendr.openshift.io/k8s-resource-selector: unique-name`.  
- It sends **`unique-name`** as `K8S_RESOURCE_SELECTOR` and `PVC_RESOURCE_SELECTOR` to the backend.  
- The backend **replaces** those placeholders with the label key at runtime, generating an **in-memory recipe** named `vm-recipe` in `openshift-dr-ops` for the DRPC.

---

### **1. DRPC YAML Configuration:**
```yaml
apiVersion: ramendr.openshift.io/v1alpha1
kind: DRPlacementControl
metadata:
  name: unique-name-drpc
  namespace: openshift-dr-ops
spec:
  drPolicyRef:
    apiVersion: ramendr.openshift.io/v1alpha1
    kind: DRPolicy
    name: odr-policy-5m
  kubeObjectProtection:
    captureInterval: 5m0s
    recipeParameters:
      K8S_RESOURCE_SELECTOR:
        - unique-name
      PROTECTED_VMS:
        - vm-cirros
      PVC_RESOURCE_SELECTOR:
        - unique-name
    recipeRef:
      name: vm-recipe
      namespace: openshift-dr-ops
  placementRef:
    apiVersion: cluster.open-cluster-management.io/v1beta1
    kind: Placement
    name: unique-name-placement
    namespace: openshift-dr-ops
  preferredCluster: dr2
  protectedNamespaces:
    - kubevirt-test
  pvcSelector: {}
```

---

### **2. UI Applies Labels to VM & PVCs**

The UI forcefully adds the label `ramendr.openshift.io/k8s-resource-selector: unique-name` to both the VM and its PVCs:

####  VM:
```yaml
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  name: vm-cirros
  namespace: kubevirt-test
  labels:
    ramendr.openshift.io/k8s-resource-selector: unique-name
spec:
  running: true
  template:
    spec:
      domain:
        cpu:
          cores: 2
        devices:
          disks:
            - disk: {}
              name: rootdisk
      volumes:
        - name: rootdisk
          persistentVolumeClaim:
            claimName: vm-cirros-pvc
```

#### PVC:
[...]

PR #1909, opened 2025-03-18
 RHSTOR-7112: Disaster recovery UI for Kubevirt VM - Discovered(Shared)

GowthamShanmugam:
On top of: https://github.com/red-hat-storage/odf-console/pull/1899


GowthamShanmugam:
Screenshot 2025-03-19 at 12 47 29 AM
Screenshot 2025-03-19 at 12 47 40 AM
Screenshot 2025-03-19 at 12 47 56 AM
Screenshot 2025-03-19 at 12 48 00 AM

locales/en/plugin__odf-console.json
```diff
@@ -349,6 +350,18 @@
...
+  "Protect this VM independently without associating it with an existing DR placement control.": "Protect this VM independently without associating it with an existing DR placement control.",
+  "Standalone": "Standalone",
+  "Add this VM to an existing DR placement control for consistent failover and recovery. This method is only available for discovered VMs.": "Add this VM to an existing DR placement control for consistent failover and recovery. This method is only available for discovered VMs.",
+  "Shared": "Shared",
+  "Protection type": "Protection type",
+  "Choose how you would like to protect this VM:": "Choose how you would like to protect this VM:",
+  "Shared protection is not available for managed VMs.": "Shared protection is not available for managed VMs.",
+  "Virtual machines ({{count}})_one": "Virtual machines ({{count}})",
+  "Virtual machines ({{count}})_other": "Virtual machines ({{count}})",
+  "<0>Showing all VMs grouped under <2></2>{getName(viewSharedVMGroup)}</0>": "<0>Showing all VMs grouped under <2></2>{getName(viewSharedVMGroup)}</0>",
```
alfonsomthd, line 364: `getName(viewSharedVMGroup)` shouldn't be part of the translation, right?

packages/mco/components/modals/app-manage-policies/utils/k8s-utils.ts
```diff
@@ -195,3 +220,190 @@ export const assignPromises = (
...
+                namespace: DISCOVERED_APP_NS,
+              },
+            },
+            data: patch,
+          })
+        );
+      }
+    })
+    /* eslint-disable @typescript-eslint/no-empty-function */
+    .catch((_error) => {}); // Promise error will be handled by on onSubmit using promises
```
alfonsomthd, line 388: Better than a silent death, you can log with `console.error`.

packages/mco/utils/managed-cluster-view.ts
```diff
@@ -0,0 +1,144 @@
...
+    name: viewName,
+    ns: clusterName,
+  });
+
+export const pollManagedClusterView = (
+  viewName: string,
+  clusterName: string,
+  t: TFunction
+): Promise<PollManagedClusterViewResult> => {
+  let retries = process.env.NODE_ENV === 'development' ? 0 : 20;
```
alfonsomthd, line 40: @GowthamShanmugam this won't work at runtime (and in my `Object Browser on consumer clusters` PR that I'll raise ASAP `process.env` references in our browser code will be deleted @bipuladh @SanjalKatiyar ). For a quick workaround use this instead (like in `noobaa-context.tsx`):
```suggestion
  let retries = window.location.hostname.includes('localhost') ? 0 : 20;
```

SanjalKatiyar: btw do we really need this condition ?? what's wrong with `retries = 20` in dev mode ??

packages/mco/components/modals/app-manage-policies/utils/k8s-utils.ts
SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/1899#discussion_r2002637312

packages/mco/utils/managed-cluster-action.ts
```diff
@@ -0,0 +1,152 @@
...
+                new Error(
+                  t(
+                    'There was an error while performing the managed cluster resource action. Ensure the cluster is online and the work manager pod in namespace open-cluster-management-agent-addon is healthy.'
+                  )
+                )
+              );
+            }
+
+            deleteManagedClusterAction(actionName, clusterName);
+          } else if (retries-- > 0) {
```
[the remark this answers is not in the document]
GowthamShanmugam: sorry retries-- > 0 is the correct one
for example: 
```
let retries = 3;
console.log(--retries > 0); // true (retries becomes 2, then checks the condition)
console.log(retries); // 2
```
for (> 0) retires-- is correct, for (>=0) --retires is correct.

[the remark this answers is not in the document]
alfonsomthd: OK, I just wanted to confirm that comparing BEFORE decreasing if what you really wanted.
```
> r=1; --r > 0
false
> r=1; r-- > 0
true

```

locales/en/plugin__odf-console.json
GowthamShanmugam: ack

packages/shared/src/utils/table.ts
```diff
@@ -11,11 +11,11 @@ export const sortRows = (
...
   b: any,
   c: SortByDirection,
-  sortField: string,
+  sortField?: string,
   favoriteNames?: string[]
 ) => {
-  let aValue = _.get(a, sortField, '').toString();
-  let bValue = _.get(b, sortField, '').toString();
+  let aValue = sortField ? _.get(a, sortField, '').toString() : a;
+  let bValue = sortField ? _.get(b, sortField, '').toString() : b;
```
SanjalKatiyar, line 18: why do we need this ternary condition ? 3rd argument to `_.get` is '' (empty string), which is the default value for any edge case.

if `sortField` is `undefined`, u will automatically get '' (empty string) as output.

GowthamShanmugam: ack

locales/en/plugin__odf-console.json
GowthamShanmugam: updating the code

packages/shared/src/utils/table.ts
```diff
@@ -11,11 +11,11 @@ export const sortRows = (
...
   b: any,
   c: SortByDirection,
-  sortField: string,
+  sortField?: string,
   favoriteNames?: string[]
 ) => {
-  let aValue = _.get(a, sortField, '').toString();
-  let bValue = _.get(b, sortField, '').toString();
+  let aValue = _.get(a, sortField, a).toString();
+  let bValue = _.get(b, sortField, b).toString();
```
SanjalKatiyar, line 18: why do we want to default to `a` and `b` ? is there any issue with existing behaviour ?
```suggestion
  let aValue = _.get(a, sortField, '').toString();
  let bValue = _.get(b, sortField, '').toString();
```

SanjalKatiyar: I would prefer not to make changes in shared, unless really needed...

SanjalKatiyar: it could cause a bug at other places...

GowthamShanmugam: if the `a&b` is an object, to fetch the particular field from the object, we are using sortField. In my case, it is a string. so sortField is not required. 

ok! the problem is, it might break for the object which does not have sortField, I will correct this.

GowthamShanmugam:
```
export const sortRows = (
  a: any,
  b: any,
  c: SortByDirection,
  sortField?: string,
  favoriteNames?: string[]
) => {
  let aValue,
bValue;

  if (typeof a === 'string' && typeof b === 'string') {
    // Directly compare if both are strings
    aValue = a;
    bValue = b;
  } else {
    // Default behavior with sortField
    aValue = _.get(a, sortField, '').toString();
    bValue = _.get(b, sortField, '').toString();
  }

  if (!!favoriteNames) {
    aValue = favoriteNames.includes(aValue).toString();
    bValue = favoriteNames.includes(bValue).toString();
  }
  return sort(aValue, bValue, c);
};
```

PR #1911, opened 2025-03-19
Fix styling issues due to OCP's PF6 migration (StatusBox and Heading tags)

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-7183

### DIFF
--- a/packages/shared/src/styles/_status-box.scss
+++ b/packages/shared/src/styles/_status-box.scss
@@ -1,0 +1,98 @@
+.odf-status-box {
+    padding: 40px 20px;
+  
+    &__title {
+      color: var(--pf-v5-global--Color--100);
+      font-size: 18px;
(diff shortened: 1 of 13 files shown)


packages/shared/src/styles/_status-box.scss
```diff
@@ -0,0 +1,98 @@
+.odf-status-box {
```
alfonsomthd, line 1: @SanjalKatiyar are the values of this file taken from original console to keep the same style?
My concern is that we're using `px` (which is a fixed unit) for CSS properties like `font-size` or `max-width` where a more responsive unit like `rem` would be a better choice. WDYT?

SanjalKatiyar: yeah these classes are copied from console directly... I haven't updated anything...

PR #1918, opened 2025-03-21
Fix S3 bucket overview routing

SanjalKatiyar:
Without this fix routing breaks for `basePath/href` tabs' routes (when we have nested tabs).

### DIFF
--- a/packages/shared/src/utils/Tabs.tsx
+++ b/packages/shared/src/utils/Tabs.tsx
@@ -105,7 +105,8 @@ const Tabs: React.FC<TabsProps> = ({
       (href) =>
         location.hash === href ||
         currentLocation.endsWith(href) ||
-        currentLocation.endsWith(`${href}/`)
+        currentLocation.endsWith(`${href}/`) ||
+        currentLocation.includes(href)
     );
     const trueActiveTab = hrefToTabMap[firstMatchKey];
     if (trueActiveTab && trueActiveTab !== activeTab) {


packages/shared/src/utils/Tabs.tsx
```diff
@@ -105,7 +105,8 @@ const Tabs: React.FC<TabsProps> = ({
       (href) =>
         location.hash === href ||
         currentLocation.endsWith(href) ||
-        currentLocation.endsWith(`${href}/`)
+        currentLocation.endsWith(`${href}/`) ||
+        currentLocation.includes(href)
```
alfonsomthd, line 109: @SanjalKatiyar just a suggestion: `startsWith` could avoid in the future weird bugs related to nested tabs having same name as parents...
```suggestion
        currentLocation.startsWith(href)
```

SanjalKatiyar: I don't think `startsWith` will work.
Example: `location = /odf/cluster/buckets/test/permissions/cors`

I have tested my change (using `includes`) for all the places where we use `Tabs` and it works, so for now lets keep it that way IMO.
In future if we face any issue we can replace and test all places with `startsWith` as well.

SanjalKatiyar: It might not work because of the example I mentioned above:
`location.pathname = /odf/cluster/buckets/test/permissions/cors` here tab's `basePath = permissions` and `href = cors` but URL startswith `/odf/cluster/buckets/...` not `basePath` or `href`.

PR #1923, opened 2025-03-26
RHSTOR-7113: Disable DR for Standalone and Shared VMs

GowthamShanmugam:
### **Disable DR Behavior for VMs**  

We have disabled Disaster Recovery (DR) for **both** managed and discovered VMs.  

#### **Managed VMs**  
- The behavior is the same as managed applications.  
- DR is fully disabled when required.  

#### **Discovered VMs**  
The behavior differs slightly from discovered applications:  

1. **Discovered Applications**  
   - DR is disabled by deleting the **DRPC (DRPlacementControl)** and the **dummy placement**.  

2. **Discovered VMs** (Classified into Shared and Standalone types)  
   - **Shared VMs** (VMs belonging to a group):  
     - **Do not delete DRPC** unless it's the last VM in the group.  
     - Remove the **ramen unique label**.  from VM and its PVCs.
     - Remove the VM from the **DRPC VM list**.  
   - **Standalone VMs** (Independent VMs):  
     - Remove all relevant **labels** from VM and its PVCs.
     - **Delete the DRPC**.  
     - **Delete the dummy placement**.


packages/mco/components/modals/app-manage-policies/manage-policy-view.tsx
```diff
@@ -486,32 +485,28 @@ export const ManagePolicyView: React.FC<ManagePolicyViewProps> = ({
...
+      );
+    } catch (error) {
+      setLocalModalActionContext(
+        ModalActionContext.DISABLE_DR_PROTECTION_FAILED
+      );
+      setErrorMessage(getErrorMessage(error) || error);
+    } finally {
+      setInProgress(false);
+    }
     setInProgress(true);
```
SanjalKatiyar, line 509: this is a bug :)
```suggestion
```

packages/mco/components/modals/app-manage-policies/utils/k8s-utils.ts
```diff
@@ -101,43 +102,121 @@ export const doNotDeletePVCAnnotationPromises = (
...
+    ];
+    await updateDRPC(drpcName, DISCOVERED_APP_NS, patch);
+  } else {
+    await Promise.all([
+      deleteDRPC(drpcName, DISCOVERED_APP_NS),
+      deleteDummyPlacement(placementName),
+    ]);
+  }
+
   return promises;
```
SanjalKatiyar, line 191: u r returning empty array...
```suggestion
```

SanjalKatiyar:
Do we support disable DR for managed/discovered apps too ??

GowthamShanmugam:
> Do we support disable DR for managed/discovered apps too ??

Yes, we have disabled DR for both—Disable DR for the managed VM behaviour, which is the same as the managed application.  https://github.com/red-hat-storage/odf-console/pull/1923#issue-2949799757

packages/mco/components/modals/app-manage-policies/manage-policy-view.tsx
GowthamShanmugam: ack

packages/mco/components/modals/app-manage-policies/utils/k8s-utils.ts
GowthamShanmugam: ah! my bad

PR #1928, opened 2025-03-28
Capacity Autoscaling modal: show 'Failed' state timestamp & message.

alfonsomthd:
- Delete preexisting autoscaler CR on storage cluster creation to avoid misconfiguration.
- Minor fixes, improvements & refactoring.
day2-3-modal-enabled-failed

### DIFF
--- a/packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-autoscaling.tsx
+++ b/packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-autoscaling.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import { OSD_CAPACITY_SIZES } from '@odf/core/constants';
 import {
   useCustomTranslation,
   TypeaheadDropdown,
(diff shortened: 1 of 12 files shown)


packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-autoscaling.tsx
```diff
@@ -21,22 +23,29 @@ import {
...
-const getItem = (value: number) => ({
-  value: `${value}${StorageSizeUnit.Ti}`,
-  children: getFormattedCapacity(`${value}${StorageSizeUnit.Ti}`),
-});
+const getItem = (capacityLimit: number) => {
+  const limit = Number.isInteger(capacityLimit)
+    ? capacityLimit
+    : capacityLimit.toFixed(2);
+  return {
+    'data-limit': limit,
```
SanjalKatiyar, line 31: shouldn't this be this as well convention ??
```suggestion
    dataLimit: limit,
```

alfonsomthd: As per our conversation, these items later become HTML elements so I prefer to store this data in a standard `data-*` attribute.

PR #1929, opened 2025-03-31
Openshift-storage namespace cleanup script

GowthamShanmugam:


scripts/ns-cleanup.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+NAMESPACE="openshift-storage"
```
alfonsomthd, line 3:
```suggestion
NAMESPACE=${NAMESPACE:-openshift-storage}
```

GowthamShanmugam: ack

bipuladh:

E2E is not required for such additions.

PR #1931, opened 2025-04-01
Typeahead dropdown: allow selected value to be a number + unit tests

alfonsomthd:
Unit tests ensure the expected behavior remains after refactoring & minor improvements.

### DIFF
--- a/packages/shared/src/dropdown/TypeaheadDropdown.tsx
+++ b/packages/shared/src/dropdown/TypeaheadDropdown.tsx
@@ -16,13 +16,13 @@ import {
 import { TimesIcon } from '@patternfly/react-icons';
 
 type TypeaheadDropdownProps = {
-  ariaLabel: string;
-  className: string;
-  id: string;
-  onSelect: (selected: string) => void;
+  ariaLabel?: string;
+  className?: string;
+  id?: string;
+  onSelect?: (selected: string) => void;
   items: SelectOptionProps[];
-  placeholder: string;
-  selectedValue: string;
+  placeholder?: string;
+  selectedValue?: string | number;
 };
 
 export const TypeaheadDropdown: React.FC<TypeaheadDropdownProps> = ({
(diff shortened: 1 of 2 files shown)


packages/shared/src/dropdown/TypeaheadDropdown.tsx
```diff
@@ -16,34 +16,34 @@ import {
 import { TimesIcon } from '@patternfly/react-icons';
 
 type TypeaheadDropdownProps = {
-  ariaLabel: string;
-  className: string;
-  id: string;
-  onSelect: (selected: string) => void;
+  ariaLabel?: string;
```
GowthamShanmugam, line 19: are all the props optional?

packages/shared/src/dropdown/TypeaheadDropdown.tsx
```diff
@@ -52,38 +52,33 @@ export const TypeaheadDropdown: React.FC<TypeaheadDropdownProps> = ({
...
+          value: NO_RESULTS,
+        },
+      ];
     }
 
-    setSelectOptions(newSelectOptions);
-  }, [filterValue, isOpen, items, NO_RESULTS, t]);
+    // Open the menu when the input value changes and the new value is not empty
+    if (!isOpen) {
+      setIsOpen(true);
```
GowthamShanmugam, line 79: is setIsOpen becomes true only when filterValue is not empty?

packages/shared/src/dropdown/TypeaheadDropdown.tsx
```diff
@@ -112,28 +107,33 @@ export const TypeaheadDropdown: React.FC<TypeaheadDropdownProps> = ({
...
     value: string | number | undefined
   ) => {
-    if (value && value !== NO_RESULTS) {
-      const optionText = selectOptions.find(
-        (option) => option.value === value
-      )?.children;
-      selectOption(value, optionText);
-    }
+    selectOption(value);
+    onSelect(value as string);
```
GowthamShanmugam, line 128: if we converting value as string always for  onSelect, is this conversion is required? just a doubt.
https://github.com/red-hat-storage/odf-console/pull/1931/files#diff-18c0d896528823859556d4b20064ad9c04335f662b2a727072b9ea5e1b379a32R39

packages/shared/src/dropdown/TypeaheadDropdown.tsx
alfonsomthd: No, it's also `true` on click & keyboard events (otherwise the dropdown would not be usable).

PR #1933, opened 2025-04-03
Minor fixes related to StorageClient refactors

bipuladh:


packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -377,6 +376,7 @@ const StorageClientRow: React.FC<
                   {
                     key: 'GENERATE_ONBOARDING_TOKEN',
                     value: t('Generate client onboarding token'),
+                    isDisabled: isLocalClient,
```
SanjalKatiyar, line 379: there is one more case where we might need to disable this: https://ibm-systems-storage.slack.com/archives/C06DWD30QTZ/p1744366090266389?thread_ts=1744171121.198289&cid=C06DWD30QTZ

do you want to incorporate that in this PR ??

packages/odf/modals/ResourceDistributionModal/ResourceDistributionModal.tsx
```diff
@@ -237,7 +237,29 @@ export const DistributeResourceModal: React.FC<
 
   const onConfirm = () => {
     // Todo(bipuladh): Implement patching
```
SanjalKatiyar, line 239: can we remove this Todo now ??

SanjalKatiyar (commented):
there is another bug: https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/odf-dashboard/status-card/status-card.tsx#L152
this route need to be updated.

Do you wanna do it in this PR ?? or maybe let QE open a new bug ?? (ur call)

packages/odf/components/storage-consumers/client-list.tsx
bipuladh: let's not do that

bipuladh: onboarded client detection is a headache

packages/odf/modals/ResourceDistributionModal/ResourceDistributionModal.tsx
bipuladh: yes thank you

PR #1935, opened 2025-04-04
Fix for ManagedClusterAction mca-64x44 violates policy 299

GowthamShanmugam:
The ManagedClusterAction CRD does not include these spec fields, which leads to a policy violation warning being triggered in the console.


GowthamShanmugam:
Screenshot 2025-04-04 at 3 17 35 PM

PR #1985, opened 2025-04-10
CVE fix for minimist and showdown 

GowthamShanmugam:
https://issues.redhat.com/browse/DFBUGS-2005
https://issues.redhat.com/browse/DFBUGS-2009

The package @openshift-console/plugin-shared@^0.0.3 strictly depends on showdown@1.8.6, which contains known CVEs. Since it locks the version without a range (^ or ~), it prevents Yarn resolutions from upgrading showdown to a safer version (e.g., 1.9.1 or above).

This PR removes the @openshift-console/plugin-shared dependency from the console.

### DIFF
--- a/packages/shared/src/overview-page/OverviewDetailItem.tsx
+++ b/packages/shared/src/overview-page/OverviewDetailItem.tsx
@@ -1,0 +1,55 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+import './OverviewDetailItem.scss';
+
+export type OverviewDetailItemProps = {
+  /** Details card title */
+  title: string;
+  children: React.ReactNode;
+  /** Trigger skeleton loading component during the loading phase. */
+  isLoading?: boolean;
+  /** Value for a className */
+  valueClassName?: string;
+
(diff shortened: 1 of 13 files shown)


alfonsomthd:
LGTM. @SanjalKatiyar to tag.

packages/shared/src/overview-page/OverviewDetailItem.tsx
```diff
@@ -0,0 +1,53 @@
...
+  children,
+  error,
+  valueClassName,
+}) => {
+  let status: React.ReactNode;
+
+  if (error) {
+    status = <span className="text-secondary">{error}</span>;
+  } else if (isLoading) {
+    status = <div className="skeleton-text" />;
```
SanjalKatiyar, line 30: where are `text-secondary` and `skeleton-text` defined ? this could break if console does not provide these classes...

alfonsomthd: @SanjalKatiyar those classes are already used in our code.

PR #2029, opened 2025-04-11
Fix for DRPolicy creation issue for MDR

GowthamShanmugam:
When both selected clusters have deploymentType set to external, storage class validation is not required. An early return in the useStorageClassValidation hook was added to skip the validation logic entirely if both clusters are external.

For the external cluster, `ramendr.openshift.io/storageid` was added by the MCO operator after the DRPolicy & MirrorPeer creation. so it is not possible to do the storage class validation. 

Issue: https://issues.redhat.com/browse/DFBUGS-2229


alfonsomthd:
LGTM. @SanjalKatiyar to tag.

PR #2053, opened 2025-04-22
Introduces new flag for detection of internal mode

bipuladh:
A flag is introduced by client plugin to disable itself 
Flag is based on the present of successfully loaded odf-console plugin

Tested locally by running two plugins simultaneously

Fixes Issue:  https://issues.redhat.com/browse/DFBUGS-2248

### DIFF
--- a/plugins/client/console-extensions.json
+++ b/plugins/client/console-extensions.json
@@ -9,7 +9,7 @@
       "href": "/odf/cluster"
     },
     "flags": {
-      "disallowed": ["PROVIDER_MODE"]
+      "disallowed": ["ODF_MODEL"]
     }
   },
   {
@@ -20,7 +20,7 @@
       "component": { "$codeRef": "dashboard.default" }
     },
     "flags": {
-      "disallowed": ["PROVIDER_MODE"]
+      "disallowed": ["ODF_MODEL"]
     }
   },
   {
@@ -34,7 +34,7 @@
       "component": { "$codeRef": "install.default" }
     },
     "flags": {
-      "disallowed": ["PROVIDER_MODE"]
+      "disallowed": ["ODF_MODEL"]
     }
   },
   {
@@ -45,7 +45,7 @@
       }
     },
     "flags": {
-      "disallowed": ["PROVIDER_MODE"]
+      "disallowed": ["ODF_MODEL"]
     }
   },
   {
@@ -55,7 +55,7 @@
       "reducer": { "$codeRef": "features" }
     },
     "flags": {
-      "disallowed": ["PROVIDER_MODE"]
+      "disallowed": ["ODF_MODEL"]
     }
   },
   {
@@ -66,7 +66,7 @@
       }
     },
     "flags": {
-      "disallowed": ["PROVIDER_MODE"]
+      "disallowed": ["ODF_MODEL"]
     }
   }
 ]
--- a/packages/shared/src/global.d.ts
+++ b/packages/shared/src/global.d.ts
@@ -44,7 +44,9 @@ declare global {
     windowError?: string;
     __REDUX_DEVTOOLS_EXTENSION_COMPOSE__?: Function;
     store?: {}; // Redux store
-    pluginStore?: {}; // Console plugin store
+    pluginStore?: {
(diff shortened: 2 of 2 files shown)


packages/client/src/features.ts
```diff
@@ -1,3 +1,16 @@
...
+
 export { default } from '@odf/core/redux';
 export * from '@odf/core/redux';
 export { detectRGW } from '@odf/core/features';
+
+const INTERNAL_MODE_PRESENT = 'ODF_INTERNAL_MODE_PRESENT';
+
+export const detectInternalMode = async (setFlag: SetFeatureFlag) => {
+  const isInternalModePresent = !_.isEmpty(
+    window.pluginStore?.loadedDynamicPlugins.get('odf-console@0.0.0')
```
SanjalKatiyar, line 12: in local it's `0.0.0`, but in production won't it be exact version that we set during build: https://github.com/red-hat-storage/odf-console/blob/master/scripts/generatePluginPackage.mts#L31 ??

packages/client/src/features.ts
```diff
@@ -1,3 +1,16 @@
+import { SetFeatureFlag } from '@openshift-console/dynamic-plugin-sdk';
+import * as _ from 'lodash-es';
+
 export { default } from '@odf/core/redux';
 export * from '@odf/core/redux';
 export { detectRGW } from '@odf/core/features';
+
+const INTERNAL_MODE_PRESENT = 'ODF_INTERNAL_MODE_PRESENT';
+
+export const detectInternalMode = async (setFlag: SetFeatureFlag) => {
```
SanjalKatiyar, line 10: does return type need to be `Promise<void>` ? what's the need of async ??
```suggestion
export const detectInternalMode = (setFlag: SetFeatureFlag) => {
```

packages/client/src/features.ts
SanjalKatiyar, line 12: If odf-console is loaded first and then client-console, everything will work fine. But, can there be a case where client-console is loaded first and then odf-console is loaded (with few secs delay, as we don't have control over it) ?? (this logic will not work in that case)

SanjalKatiyar: On second thought, maybe it will work, as we have refresh popover when each plugin is loaded.

packages/client/src/features.ts
```diff
@@ -1,3 +1,16 @@
+import { SetFeatureFlag } from '@openshift-console/dynamic-plugin-sdk';
+import * as _ from 'lodash-es';
+
 export { default } from '@odf/core/redux';
 export * from '@odf/core/redux';
 export { detectRGW } from '@odf/core/features';
+
+const INTERNAL_MODE_PRESENT = 'ODF_INTERNAL_MODE_PRESENT';
```
SanjalKatiyar, line 8: Can we name it to `ODF_PLUGIN_PRESENT` instead ?

SanjalKatiyar: because that's what u are checking ideally...

SanjalKatiyar:
can't we use this flag https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/features.ts#L74-L75 instead of creating a new flag ??

plugins/client/console-extensions.json
```diff
@@ -9,7 +9,7 @@
       "href": "/odf/cluster"
     },
     "flags": {
-      "disallowed": ["PROVIDER_MODE"]
+      "disallowed": ["ODF_INTERNAL_MODE_PRESENT"]
```
GowthamShanmugam, line 12: Just a doubt, 
   Which plugin gets enabled first? Are plugins initialized sequentially or in parallel? if a client plugin loads before the ODF-Console plugin, users might need to refresh twice, and some UI elements may briefly appear and disappear, which can look buggy.

bipuladh:
> can't we use this flag https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/features.ts#L74-L75 instead of creating a new flag ??

This would be a better way to do it.

plugins/client/console-extensions.json
bipuladh: after each plugin is installed a refresh is required. A refresh is required for plugin to load. So we should not have to refresh it twice.

PR #2054, opened 2025-04-23
Convert dl, dd, dt to PF DescriptionList + Remove DetailsBody component

SanjalKatiyar:
https://issues.redhat.com/browse/DFBUGS-2327

Console removed `DetailsBody` component from the internal SDK.
Console removed the styling for `dl, dd, dt` HTML elements.
Ref: https://github.com/openshift/console/pull/14947 for more details.

**Before:**
Screenshot 2025-04-23 at 1 50 48 PM

**After:**
Screenshot 2025-04-23 at 2 01 51 PM

### DIFF
--- a/packages/ocs/dashboards/block-pool/mirroring-card-body.tsx
+++ b/packages/ocs/dashboards/block-pool/mirroring-card-body.tsx
@@ -1,9 +1,10 @@
 import * as React from 'react';
+import { DescriptionList } from '@patternfly/react-core';
 import './mirroring-card.scss';
 
 export const MirroringCardBody: React.FC<MirroringCardBodyProps> = ({
   children,
-}) => <dl className="odf-block-pool__mirroring-card-body">{children}</dl>;
+}) => <DescriptionList>{children}</DescriptionList>;
 
 type MirroringCardBodyProps = {
   children: React.ReactNode;
--- a/packages/shared/src/details-page/DetailsPage.tsx
+++ b/packages/shared/src/details-page/DetailsPage.tsx
@@ -18,6 +18,10 @@ import {
   Popover,
   Split,
   SplitItem,
+  DescriptionList,
+  DescriptionListDescription,
+  DescriptionListGroup,
+  DescriptionListTerm,
 } from '@patternfly/react-core';
 import { PencilAltIcon } from '@patternfly/react-icons';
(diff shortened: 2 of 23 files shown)


packages/ocs/dashboards/block-pool/mirroring-card-body.tsx
```diff
@@ -1,9 +1,10 @@
 import * as React from 'react';
+import { DescriptionList } from '@patternfly/react-core';
 import './mirroring-card.scss';
 
 export const MirroringCardBody: React.FC<MirroringCardBodyProps> = ({
   children,
-}) => <dl className="odf-block-pool__mirroring-card-body">{children}</dl>;
```
alfonsomthd, line 6: This css class should be removed also, right?

packages/shared/src/details-page/DetailsPage.tsx
```diff
@@ -310,7 +310,7 @@ export const ResourceSummary: React.FC<ResourceSummaryProps> = ({
   const launchModal = useModal();
 
   return (
-    <dl data-test-id="resource-summary" className="odf-m-pane__details">
```
alfonsomthd, line 313: The css class should also be removed, right?

packages/ocs/dashboards/block-pool/mirroring-card-body.tsx
SanjalKatiyar: yes, not needed.

packages/shared/src/details-page/DetailsPage.tsx
SanjalKatiyar: both having and not having won't cause any issue (as per the testing)... I just removed it instead.

alfonsomthd: @SanjalKatiyar `odf-m-pane__details` class will become dead code if not removed.

packages/ocs/dashboards/block-pool/mirroring-card-body.tsx
alfonsomthd: I mean from the `.scss` file to avoid having dead code.

SanjalKatiyar: updated...

PR #2067, opened 2025-05-02
Add target size ratio value 0.49 to custom cephBlockPool during creation

malayparida2000:
Custom cephBlockPool are not reconciled by ocs-operator, so if the user does not set the target size ratio value, it will be not set. This would affect the scaling of the pool.

So we need to add the target size ratio default value to the custom cephBlockPool during creation.


malayparida2000:
Testing
Target Size ratio present in newly created custom cephBlockPool
Screenshot 2025-05-03 at 12 54 41 PM

SanjalKatiyar:
@malayparida2000 couple of things:
1. Don't we want to pass targetSizeRatio for CephFS data pools created from UI (via `storageCluster.spec.managedResources.cephFilesystems.additionalDataPools`) ??
2. There is another flow where we create CBP: https://github.com/red-hat-storage/ocs-operator/blob/main/services/ux-backend/handlers/expandstorage/handler.go#L142-L181, we need changes here in OCS as well (cc @parth-gr @vbnrh FYI).

malayparida2000:
Created the ocs-operator PR here https://github.com/red-hat-storage/ocs-operator/pull/3203

PR #2069, opened 2025-05-05
CI catalogsource: use latest-stable tag

alfonsomthd:
We should use this tag according to the build team. cc @b-ranto


b-ranto (approved):
lgtm

bipuladh (approved):

PR #2071, opened 2025-05-05
Fixes storage resource distribution

bipuladh:
Removes VolumeGroupSnapshot from UI as well

Recording
[bug-fix.webm](https://github.com/user-attachments/assets/6671627a-e199-4ee8-8c6c-8c9aadbfa9cc)

Fixes https://issues.redhat.com/browse/DFBUGS-2336


packages/odf/components/ResourceDistribution/ResourceDistributionTable.tsx
```diff
@@ -10,34 +10,22 @@ import * as _ from 'lodash-es';
 import { Thead, Tr, Th, Tbody, Table } from '@patternfly/react-table';
 
 export type SelectedResources = {
-  [name: string]: {
-    selected: boolean;
-    resourceType:
-      | 'storageClass'
-      | 'volumeSnapshotClass'
-      | 'volumeGroupSnapshotClass';
```
alfonsomthd, line 18: @bipuladh the removal of `volumeGroupSnapshotClass` is intentional, right? Not to be ditributed?

alfonsomthd: Now I see it's mentioned in the PR description.

packages/odf/components/ResourceDistribution/utils.ts
```diff
@@ -2,91 +2,93 @@ import { StorageConsumerKind } from '@odf/shared';
...
+      path,
+      value: resourceNames.map((name) => ({ name })),
     });
   }
-
-  if (
-    !isVolumeSnapshotClassesDistributed &&
-    !_.isEmpty(volumeSnapshotClassNames)
-  ) {
+  if (addedResources.length > 0) {
```
SanjalKatiyar, line 47: shouldn't this `else if` instead of `if` ?

alfonsomthd: AFAIK this is adding patches to an array. The array is later used in a single call.

SanjalKatiyar: my bad, got it !

PR #2123, opened 2025-06-17
RHSTOR-6175 Automate key rotation for encrypted RBD storage classes

black-dragon74:
This patch sets the key rotation schedule annotation by default on RBD storage classes created from the UI; enabling automatic key rotation.

### DIFF
--- a/packages/ocs/storage-class/mutators.ts
+++ b/packages/ocs/storage-class/mutators.ts
@@ -22,11 +22,15 @@ export const addClusterParams = (sc: StorageClass): StorageClass => {
   return sc;
 };
 
-export const addKubevirtAnnotations = (sc: StorageClass): StorageClass => {
-  if (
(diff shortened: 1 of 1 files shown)


packages/ocs/storage-class/mutators.ts
```diff
@@ -36,6 +36,19 @@ export const addKubevirtAnnotations = (sc: StorageClass): StorageClass => {
   return addClusterParams(sc);
 };
 
+export const addKeyRotationAnnotation = (sc: StorageClass): StorageClass => {
+  if (
+    sc?.parameters?.hasOwnProperty('encrypted') &&
+    sc?.parameters?.['encrypted'] === 'true'
+  ) {
```
bipuladh, line 43: Please create a function something like `isStorageClassEncrypted` and use it in L27 & L28 as well.

bipuladh (changes requested):

PR #2126, opened 2025-06-18
Add checkbox for enabling volume consistency groups

vbnrh:
Screenshot 2025-06-18 at 3 37 35 PM
Screenshot 2025-06-18 at 4 41 06 PM


packages/mco/components/discovered-application-wizard/utils/k8s-utils.ts
```diff
@@ -135,6 +143,11 @@ export const createPromise = (
     })
   );
 
+  let labels: ObjectMetadata['labels'];
+  const annotations = isVolumeConsistencyEnabled && {
```
bipuladh, line 147:
```suggestion
  const annotations = isVolumeConsistencyEnabled ? {
```

packages/mco/components/discovered-application-wizard/utils/k8s-utils.ts
```diff
@@ -135,6 +143,11 @@ export const createPromise = (
...
+    [IS_CG_ENABLED_ANNOTATION]: 'true',
+  };
```
bipuladh, line 149:
```suggestion
  } : {};
```
I would prefer ternary operator to make things clear. You are applying an annotation only when it is enabled else annotations is empty. Also are we sure there will be no other annotations?

packages/mco/components/discovered-application-wizard/utils/k8s-utils.ts
```diff
@@ -135,6 +143,11 @@ export const createPromise = (
     })
   );
 
+  let labels: ObjectMetadata['labels'];
+  const annotations = isVolumeConsistencyEnabled && {
+    [IS_CG_ENABLED_ANNOTATION]: 'true',
```
bipuladh, line 148: Is `true` value required I guess they will only look for the annotation key in which case we can keep it empty here as well to signify the key is only what matters.

packages/mco/components/discovered-application-wizard/utils/reducer.ts
```diff
@@ -34,6 +35,7 @@ export type EnrollDiscoveredApplicationState = {
     name: string;
   };
   configuration: {
+    isVolumeConsistencyEnabled?: boolean;
```
bipuladh, line 38:
```suggestion
    isVolumeConsistencyEnabled: boolean;
```
Is it an optional field in the CRD specification?

packages/mco/components/enable-volume-consistency-group/enable-volume-consistency-group.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import { Checkbox } from '@patternfly/react-core';
+
+type VolumeConsistencyCheckboxProps = {
+  isChecked: boolean;
+  onChange: (checked: boolean) => void;
+  t: (key: string) => string;
```
bipuladh, line 7: This is wrong type definition t is a react-i18n translate function. It might have a similar signature but better to use it correctly. See https://github.com/red-hat-storage/odf-console/blob/3baa1610ab7189400736438a4d72728e4a8bfcbd/packages/odf/constants/tooltips.tsx#L2 for example.

packages/mco/components/modals/app-manage-policies/utils/k8s-utils.ts
```diff
@@ -282,6 +286,9 @@ export const assignPromisesForManaged = async (
         : placementRuleAssignPromise(placement)
     );
 
+    const annotations = enableVolumeConsistencyGroup && {
+      [IS_CG_ENABLED_ANNOTATION]: 'true',
+    };
```
bipuladh, line 291:
```suggestion
    const annotations = enableVolumeConsistencyGroup ? {
      [IS_CG_ENABLED_ANNOTATION]: 'true',
    } : {};
```

packages/mco/components/discovered-application-wizard/utils/k8s-utils.ts
vbnrh: They are going to use the value as well. Confirmed from Ramen team

packages/mco/components/discovered-application-wizard/utils/reducer.ts
vbnrh: Its not in the CRD specification, its optional to enable it

PR #2142, opened 2025-07-08
RHSTOR-7577 Remove file-loader dependency: use asset module

Soundarya189r:
https://issues.redhat.com/browse/RHSTOR-7577

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,6 @@
     "classnames": "^2.5.1",
     "copy-webpack-plugin": "^13.0.0",
     "css-loader": "^7.1.2",
-    "file-loader": "^6.2.0",
     "fork-ts-checker-webpack-plugin": "^9.1.0",
     "fuzzysearch": "^1.0.3",
     "i18next": "^25.2.1",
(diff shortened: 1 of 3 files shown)


SanjalKatiyar:
this is what I get when I `exec` into the console pod:
```
$ find . -type d
.
./.pki
./.pki/nssdb
./compatibility
./compatibility/locales
./compatibility/locales/en
./compatibility/locales/es
./compatibility/locales/fr
./compatibility/locales/ja
./compatibility/locales/ko
./compatibility/locales/zh
./locales
./locales/en
./locales/es
./locales/fr
./locales/ja
./locales/ko
./locales/zh
./nginx-start
```

alfonsomthd:
> @alfonsomthd I tried to replicate this issue, I don't see asset folder getting generated prior to these changes.

@Soundarya189r probably because some recent change/PR in master had made that svg no longer necessary. Have you tested including an image in a component (no need to commit that, just test) and then confirming that the image was added to dist/asset/ folder in the build?

PR #2145, opened 2025-07-08
RHSTOR-7589 Remove style-loader (use mini-css-extract-plugin)

Soundarya189r:
https://issues.redhat.com/browse/RHSTOR-7589


webpack.config.ts
```diff
@@ -85,6 +86,13 @@ const config: webpack.Configuration & DevServerConfiguration = {
         exclude: /(build|dist)/, // Ignore shared build folder.
         use: [
           { loader: 'cache-loader' },
+          { loader: MiniCssExtractPlugin.loader },
+          {
+            loader: 'css-loader',
+            options: {
+              sourceMap: true,
+            },
+          },
```
alfonsomthd, line 95: @Soundarya189r Please place MiniCssExtractPlugin & css-loader back to their oiriginal order (under thread-loader).

alfonsomthd (changes requested):

webpack.config.ts
alfonsomthd: @Soundarya189r Then put only MiniCssExtractPlugin before `thread-loader` as per [this example](https://github.com/webpack-contrib/thread-loader/blob/master/test/sass-loader-example/webpack.config.js). There are more examples [here](https://github.com/webpack-contrib/thread-loader/tree/master/test).

webpack.config.ts
```diff
@@ -21,6 +22,7 @@ const NODE_ENV = (process.env.NODE_ENV ||
 const PLUGIN = process.env.PLUGIN;
 const OPENSHIFT_CI = process.env.OPENSHIFT_CI;
 const IS_PRODUCTION = NODE_ENV === 'production';
+const useThreads = Number(process.env.threads) > 0;
```
alfonsomthd, line 25: @Soundarya189r is this always 0? Why do we need this?

webpack.config.ts
```diff
@@ -85,7 +87,8 @@ const config: webpack.Configuration & DevServerConfiguration = {
         exclude: /(build|dist)/, // Ignore shared build folder.
         use: [
           { loader: 'cache-loader' },
-          {
+          { loader: MiniCssExtractPlugin.loader },
+          useThreads && {
```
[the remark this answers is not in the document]
Soundarya189r: Removed

webpack.config.ts
alfonsomthd: @SanjalKatiyar we can temporarily go ahead with putting `css-loader` before thread-loader while we investigate the issue.

SanjalKatiyar: @Soundarya189r what error are u getting when putting css-loader after thread-loader ?? (as Alfonso suggested above).

Adding MiniCssExtractPlugin loader before thread-loader make sense, but what's the issue with below:
[
    .
    .
    MiniCssExtractPlugin.loader,
    thread-loader,
    css-loader,
    .
    .
]

Soundarya189r: > @SanjalKatiyar Please find the error below
> 
> return JSON.stringify(loaderContext.utils.contextify(loaderContext.context || loaderContext.rootContext, request)); ^
> 
> TypeError: loaderContext.utils.contextify is not a function

@SanjalKatiyar @alfonsomthd 

I am running in the above error whenever I try to add MiniCssExtractPlugin.loader or css-loader after the thread-loader.

As the loaders are evaluated/executed from right to left (or from bottom to top), and considering the limitations of thread-loader, the threads don't have access to Webpack internal API's ,it throws runtime errors like 
TypeError: loaderContext.utils.contextify is not a function.

As the thread-loaders are for expensive operations should we just wrap the sass-loader with thread-loader?

[
.
.
MiniCssExtractPlugin.loader,
css-loader,
resolve-url-loader,
thread-loader,
sass-loader
.
.
]

Please let me know your thoughts on this.

SanjalKatiyar: I don't find any evidence online on whether resolve-url-loader is "thread-safe" or not. So, I agree to keep thread-loader as top as possible unless we are getting any error (like we got now for css-loader and Mini loader).

PR #2146, opened 2025-07-09
Adds support for loading screen while OCS is getting installed

bipuladh:
Screenshot: 
Screenshot 2025-07-09 at 2 48 41 PM

Recording: 

https://github.com/user-attachments/assets/63949e31-9300-44ea-bb57-77d43d056608

Bug Link: https://issues.redhat.com/browse/DFBUGS-2963

### DIFF
--- a/packages/odf/components/create-storage-system/create-storage-system.tsx
+++ b/packages/odf/components/create-storage-system/create-storage-system.tsx
@@ -4,15 +4,33 @@ import {
   StorageClassWizardStepExtensionProps as ExternalStorage,
   isStorageClassWizardStep,
 } from '@odf/odf-plugin-sdk/extensions';
+import { useK8sList } from '@odf/shared';
 import { DEFAULT_INFRASTRUCTURE } from '@odf/shared/constants';
 import { useK8sGet } from '@odf/shared/hooks/k8s-get-hook';
-import { useK8sList } from '@odf/shared/hooks/useK8sList';
-import { InfrastructureModel, NamespaceModel } from '@odf/shared/models';
-import { InfrastructureKind } from '@odf/shared/types';
+import {
+  CustomResourceDefinitionModel,
+  InfrastructureModel,
+  NamespaceModel,
+  StorageClusterModel,
+} from '@odf/shared/models';
+import {
+  CustomResourceDefinitionKind,
+  InfrastructureKind,
+} from '@odf/shared/types';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
 import { getInfrastructurePlatform } from '@odf/shared/utils';
-import { useResolvedExtensions } from '@openshift-console/dynamic-plugin-sdk';
+import {
+  useK8sWatchResource,
+  useResolvedExtensions,
+} from '@openshift-console/dynamic-plugin-sdk';
 import { Wizard, WizardStep } from '@patternfly/react-core/deprecated';
+import * as _ from 'lodash-es';
+import {
+  EmptyState,
+  EmptyStateHeader,
+  EmptyStateIcon,
+  Spinner,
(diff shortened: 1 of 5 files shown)


packages/odf/components/create-storage-system/create-storage-system.tsx
```diff
@@ -131,4 +171,45 @@ const CreateStorageSystem: React.FC<{}> = () => {
...
+
+  React.useEffect(() => {
+    let timer: NodeJS.Timeout;
+
+    if (isCrdPresent) {
+      timer = setTimeout(() => setDelayedShow(true), 5000);
+    }
+
+    return () => clearTimeout(timer);
+  }, [isCrdPresent]);
```
SanjalKatiyar, line 196: instead of using multiple useState and useEffect, can't we just use `useK8sWatch` on the CRD ??
If concern is that connection will remain open unnecessarily, then that's true, but it will only happen till deployment flow is mounted, once user completes deployment, connection will get removed.

packages/odf/components/create-storage-system/create-storage-system.tsx
```diff
@@ -23,6 +41,28 @@ import { CreateStorageSystemHeader } from './header';
...
+    },
+    isList: true,
+    optional: true,
+  });
+  const storageClusterCRD = crds?.filter(
+    (crd) => crd.spec.names.kind === StorageClusterModel.kind
+  );
+
+  const isStorageClusterCRDPresent =
+    crdsLoaded && _.isEmpty(crdsError) && storageClusterCRD.length > 0;
```
SanjalKatiyar, line 61:
```suggestion
  const isStorageClusterCRDPresent = storageClusterCRD.length > 0;
```

packages/odf/components/create-storage-system/create-storage-system.tsx
```diff
@@ -131,4 +171,45 @@ const CreateStorageSystem: React.FC<{}> = () => {
...
+    let timer: NodeJS.Timeout;
+
+    if (isCrdPresent) {
+      timer = setTimeout(() => setDelayedShow(true), 5000);
+    }
+
+    return () => clearTimeout(timer);
+  }, [isCrdPresent]);
+
+  return !isCrdPresent || !delayedShow ? (
```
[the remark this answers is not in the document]
bipuladh: We need delayedShow to prevent the user from seeing error messages.

packages/odf/components/create-storage-system/create-storage-system.tsx
bipuladh: We have `isCrdPresent` to make sure that in case for some reason we get to loading state or error state in the poller after we make sure StorageCluster CRD is present that does not cause flicker in our screen. 

We have `delayedShow` to prevent the user from getting an error message. I see that during testing if we directly render the wizard flow we see a red error for a few seconds briefly.

packages/odf/components/create-storage-system/create-storage-system.tsx
bipuladh: short circuits in case it's loading or has an error.

packages/odf/components/create-storage-system/create-storage-system.tsx
SanjalKatiyar: discussed offline, will merge for now and create a Github issue to later debug and remove this 5 sec delay...

packages/odf/components/create-storage-system/create-storage-system.tsx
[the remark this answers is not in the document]
SanjalKatiyar: discussed offline, will merge for now and create a Github issue to later debug and remove this 5 sec delay...

PR #2160, opened 2025-07-22
Fixes cypress tests

bipuladh:
Fixes: https://issues.redhat.com/browse/RHSTOR-7230


packages/shared/src/kebab/kebab.tsx
```diff
@@ -294,7 +296,7 @@ export const Kebab: React.FC<KebabProps> & KebabStaticProperties = ({
       }
     >
       <Dropdown
-        data-test="kebab-button"
+        data-test={dataTestId ?? 'kebab-button'}
```
alfonsomthd, line 299: Nullish coalescing operator only checks `null` or `undefined`, but `dataTestId` could also be an empty string or similar falsy value:
```suggestion
        data-test={dataTestId || 'kebab-button'}
```

bipuladh: is someone wants to purposefully disable it by passing an empty string we should allow it.

alfonsomthd: What is the purpose of passing a `data-test` empty value? It doesn't make any sense to me.

packages/odf/components/Sections/InitialEmptyStatePage.tsx
```diff
@@ -116,6 +116,7 @@ const InitialEmptyStatePage: React.FC = () => {
         <EmptyStateFooter>
           <EmptyStateActions>
             <Button
+              data-test="configure-data-foundation"
```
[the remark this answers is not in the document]
bipuladh: Created an issue
https://github.com/red-hat-storage/odf-console/issues/2169

PR #2175, opened 2025-08-01
Fix storage pool Cypress test failures

weirdwiz:
Fix failing Cypress tests for storage pool creation and updates by adding missing data-test attributes.

### DIFF
--- a/cypress/views/pvc.ts
+++ b/cypress/views/pvc.ts
@@ -37,4 +37,9 @@ export const deletePVCFromCLI = (pvcName: string, ns = 'default') => {
       throw new Error(`PVC deletion failed: ${result.stderr}`);
     }
   });
+
+  cy.log('Waiting for PVC deletion to complete');
+  cy.exec(`oc wait --for=delete pvc/${pvcName} -n ${ns} --timeout=60s`, {
+    failOnNonZeroExit: false,
+  });
 };
--- a/packages/ocs/storage-pool/StoragePoolListPage.tsx
+++ b/packages/ocs/storage-pool/StoragePoolListPage.tsx
@@ -43,7 +43,6 @@ import { Link, useLocation } from 'react-router-dom-v5-compat';
 import { Tooltip, Label } from '@patternfly/react-core';
 import { sortable, wrappable } from '@patternfly/react-table';
 import {
-  OCS_DEVICE_REPLICA,
   PoolType,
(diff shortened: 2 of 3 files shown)


packages/ocs/storage-pool/StoragePoolListPage.tsx
```diff
@@ -361,9 +360,7 @@ const RowRenderer: React.FC<RowProps<StoragePool, CustomData>> = ({
         {poolType}
       </TableData>
       <TableData {...tableColumnInfo[4]} activeColumnIDs={activeColumnIDs}>
-        {poolType === PoolType.BLOCK && replica
-          ? OCS_DEVICE_REPLICA[replica]
-          : '-'}
+        <span data-test={`${name}-replicas`}>{replica}</span>
```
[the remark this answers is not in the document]
SanjalKatiyar: aah okay !
this change is fine then, let's keep it what it originally was (i.e., `<span data-test={`${name}-replicas`}>{replica}</span>`).

packages/ocs/storage-pool/StoragePoolListPage.tsx
```diff
@@ -701,7 +693,7 @@ const StoragePoolList: React.FC<StoragePoolListProps> = ({
           data={filteredData}
           unfilteredData={data}
           loaded={loaded}
-          loadError={error}
```
SanjalKatiyar, line 704: why this change ??
this was error before as well: https://github.com/red-hat-storage/odf-console/blob/release-4.19/packages/ocs/storage-pool/StoragePoolListPage.tsx#L626

cypress/views/pvc.ts
```diff
@@ -37,4 +37,9 @@ export const deletePVCFromCLI = (pvcName: string, ns = 'default') => {
       throw new Error(`PVC deletion failed: ${result.stderr}`);
     }
   });
+
+  cy.log('Waiting for PVC deletion to complete');
+  cy.exec(`oc wait --for=delete pvc/${pvcName} -n ${ns} --timeout=60s`, {
+    failOnNonZeroExit: false,
+  });
```
SanjalKatiyar, line 44: are tests failing if PVC is not deleted:
- If so, then we shouldn't use `failOnNonZeroExit: false`.
- If not so, then is this "oc wait" block really needed ??

PR #2185, opened 2025-08-08
Disable resource distribution for internal clients

SanjalKatiyar:
https://issues.redhat.com/browse/DFBUGS-2567

Screenshot 2025-08-08 at 6 55 53 PM
Screenshot 2025-08-08 at 6 56 15 PM
Screenshot 2025-08-08 at 6 56 43 PM

- Disable resource distribution for internal clients.
- Fix filtering on Storage Consumers list page.
- Disable token generation kebab item for already onboarded clients.

### DIFF
--- a/packages/odf/components/storage-consumers/client-list.tsx
+++ b/packages/odf/components/storage-consumers/client-list.tsx
@@ -43,6 +43,11 @@ import { Button, Popover, PopoverPosition } from '@patternfly/react-core';
 import { sortable } from '@patternfly/react-table';
 import { useODFNamespaceSelector } from '../../redux';
 import {
+  getClientClusterId,
+  isLocalClientCluster,
+  isClientClusterOnboarded,
+} from '../../utils';
+import {
   clientHeartBeatFilter,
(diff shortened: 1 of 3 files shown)


packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -352,7 +357,7 @@ const StorageClientRow: React.FC<
                   {
                     key: 'GENERATE_ONBOARDING_TOKEN',
                     value: t('Generate client onboarding token'),
-                    isDisabled: isLocalClient,
+                    isDisabled: isLocalClient || isClientOnboarded,
```
bipuladh, line 360: I am not sure if we want to disable this. Just because the secret is created doesn't guarantee that the client is onboarded. I think it's the heartbeat through which we know about onboarding status. Also pressing this again basically refreshes the onboarding token essentially a new secret gets created.

bipuladh (commented):
Looks good but this is making code a bit complicated please just filter out the local client from the list rather than disabling it.

PR #2208, opened 2025-09-01
User can not forcefully delete StorageConsumer that stuck in deletion

Soundarya189r:
JIRA: https://issues.redhat.com/browse/DFBUGS-2819

Please find below the screen recording

https://github.com/user-attachments/assets/42ca5451-ddf9-486c-96db-5c64a500a1ec

https://github.com/user-attachments/assets/37efb745-3964-4d19-b07f-492841b1592c

### DIFF
--- a/packages/odf/components/storage-consumers/client-list.tsx
+++ b/packages/odf/components/storage-consumers/client-list.tsx
@@ -301,6 +301,7 @@ const StorageClientRow: React.FC<
     : t('Unlimited');
   const isLocalClient = isLocalClientCluster(obj, localClusterId);
   const isClientOnboarded = isClientClusterOnboarded(obj);
+  const hasDeletionTimestamp = !!obj?.metadata?.deletionTimestamp;
   return (
     <>
       {tableColumns.map((tableColumn) => {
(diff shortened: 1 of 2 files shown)


packages/shared/src/kebab/kebab.tsx
```diff
@@ -74,6 +74,7 @@ type KebabProps = {
     resource: K8sResourceCommon;
     resourceModel: K8sModel;
     [key: string]: any;
+    allowMultipleDeletion?: boolean;
```
SanjalKatiyar, line 77: nits: `Kebab` is a generic component, it doesn't need to know whether it's single/multi deletion scenario. Probably should be renamed to something more generic and use case specific.

Soundarya189r: I have renamed to forceDeletion.

packages/odf/components/storage-consumers/client-list.tsx
```diff
@@ -353,11 +354,15 @@ const StorageClientRow: React.FC<
                     component: React.lazy(
                       () => import('./update-storage-quota-modal')
                     ),
+                    isDisabled: !!obj?.metadata?.deletionTimestamp,
```
SanjalKatiyar, line 357: nits:
```
const hasDeletionTimestamp = !!obj?.metadata?.deletionTimestamp;
```

use `hasDeletionTimestamp` everywhere...

Soundarya189r: Updated

PR #2249, opened 2025-09-11
Fixing the issue of donut chart text not appearing clearly in dark mode

Harshvardhan-handa123:
The text in the donut chart was not appearing clearly in dark mode. I fixed this by adding the PatternFly CSS property fill: var(--pf-v5-global--Color--200), which dynamically adjusts the text color based on the UI’s dark or light mode.


Harshvardhan-handa123:
After fix:
After-dark
After-Light

before fix: 
Before

PR #2255, opened 2025-09-15
Block storage cluster capacity addition if pods are down

bipuladh:
Fixes https://issues.redhat.com/browse/DFBUGS-1755

Screenshot: 
Screenshot 2025-09-15 at 2 33 04 PM


packages/odf/modals/add-capacity/add-capacity-modal.tsx
```diff
@@ -100,6 +101,36 @@ export const StorageClassDropdown: React.FC<StorageClassDropdownProps> = ({
...
+    )?.status.replicas > 0;
+  const isOCSOperatorPresent =
+    isLoaded &&
+    !loadError &&
+    deployments?.find(
+      (deployment) => deployment.metadata.name === 'ocs-operator'
+    )?.status.replicas > 0;
+  if (!isLoaded || loadError) {
+    return true;
+  }
```
SanjalKatiyar, line 130: I can understand why we return `true` for `loadError`.

But, we shouldn't return `true` for `!isLoaded`. It might cause unnecessary flickering where initially "Add" was enabled and then got disabled later (in edge case, most cases I know it would be fast enough that users won't realise it).

packages/odf/modals/add-capacity/add-capacity-modal.tsx
```diff
@@ -100,6 +101,36 @@ export const StorageClassDropdown: React.FC<StorageClassDropdownProps> = ({
...
+    isList: true,
+    namespace: clusterNamespace,
+  };
+  const [deployments, isLoaded, loadError] =
+    useK8sWatchResource<DeploymentKind[]>(deploymentResource);
+  const isRookCephOperatorPresent =
+    isLoaded &&
+    !loadError &&
+    deployments?.find(
+      (deployment) => deployment.metadata.name === 'rook-ceph-operator'
```
SanjalKatiyar, line 120:
```suggestion
      (deployment) => getName(deployment) === 'rook-ceph-operator'
```

packages/odf/modals/add-capacity/add-capacity-modal.tsx
```diff
@@ -100,6 +101,36 @@ export const StorageClassDropdown: React.FC<StorageClassDropdownProps> = ({
...
+    )?.status.replicas > 0;
+  const isOCSOperatorPresent =
+    isLoaded &&
+    !loadError &&
+    deployments?.find(
+      (deployment) => deployment.metadata.name === 'ocs-operator'
```
SanjalKatiyar, line 126:
```suggestion
      (deployment) => getName(deployment) === 'ocs-operator'
```

SanjalKatiyar, line 126: we already have a const variable for this under `shared/src/constants/common.ts`...

packages/odf/modals/add-capacity/add-capacity-modal.tsx
SanjalKatiyar, line 120: probably add it under `shared/src/constants/common.ts` as well...

SanjalKatiyar (changes requested):

packages/odf/modals/add-capacity/add-capacity-modal.tsx
bipuladh: It should be the oppposite. It should be disabled at first and then only enable. If we don't disable it while loading chances are users will click it quickly.

SanjalKatiyar: I agree, but here in this case while API call is in-progress `!isLoaded` will be `true` and it will be enabled first and then got disabled later.

PR #2264, opened 2025-09-15
Fix consistency group volume type filter

SanjalKatiyar:
https://issues.redhat.com/browse/DFBUGS-4065

UI retrieves all protected PVCs from the VRG's status. It then looks for PVCs that have the `ramendr.openshift.io/consistency-group` label and treats the value of that label as the name of the consistency group (CG). Then, for determining whether volume type is Block/Filesystem UI used-to-check/now-checks below respectively:

 **Previously -** If the CG name starts with `cephfs-`, the UI identifies it as a Filesystem-type CG else Block-type.

**Update -** Using PVC's `volumeMode` (https://github.com/red-hat-storage/ramen/blob/main/api/v1alpha1/volumereplicationgroup_types.go#L332) reported by VRG's status instead.

### DIFF
--- a/packages/mco/components/modals/app-manage-policies/helper/consistency-groups.tsx
+++ b/packages/mco/components/modals/app-manage-policies/helper/consistency-groups.tsx
@@ -21,7 +21,7 @@ import {
   StatusBox,
   useCustomTranslation,
 } from '@odf/shared';
-import { BLOCK, FILESYSTEM } from '@odf/shared/constants/common';
+import { BLOCK, FILESYSTEM, PVCVolumeMode } from '@odf/shared/constants';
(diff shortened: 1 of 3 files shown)


SanjalKatiyar:
@ShyamsundarR can you please check this PR/description and confirm if the updated way of determining whether the type is Block/Filesystem is correct or not ??

SanjalKatiyar:
/hold cancel
we need to merge it by today or mark this bug as blocker

ShyamsundarR:
Joining the review late, but for RBD backed PVCs it could still report a volume mode of Filesystem, as it is how the volume is [accessed](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#volume-modes). So this method would flag those as filesystem volumes?

PR #2291, opened 2025-09-23
Fix MCG-only deployment + MultiCluster deployment

SanjalKatiyar:
https://issues.redhat.com/browse/DFBUGS-4141
https://issues.redhat.com/browse/DFBUGS-4139

Introduced changes:
- Adds `NamespaceSafetyBox` wrapper and `useODFNamespaceSelector` loading/error states (unrelated to this bug fix, but good/imp to have, we missed it at many places).
- Enable `Object storage` nav item for non-admin (unrelated to this bug fix, but imp to have, we disabled it by mistake).
- Fixes MCG-only deployment flow and shifted to `StartingPoint` enum every where to reduce such bugs.
- Fixes multi-cluster (external) deployment flow -
   - Namespace was not getting created.
   - Namespace was not getting labelled with cluster monitoring label.

Screenshot 2025-09-19 at 4 47 27 PM
Screenshot 2025-09-23 at 2 29 39 PM

### DIFF
--- a/packages/ocs/storage-pool/StoragePoolListPage.tsx
+++ b/packages/ocs/storage-pool/StoragePoolListPage.tsx
@@ -465,7 +465,11 @@ const getResources = (
 };
 
 const StoragePoolListPage: React.FC = () => {
-  const { odfNamespace: clusterNs } = useODFNamespaceSelector();
+  const {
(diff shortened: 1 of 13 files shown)


bipuladh (approved):
/lgtm

PR #2320, opened 2025-10-07
Adjust Minimum CPU Requirements for s390x based environment

malayparida2000:
In s390x based environments(i.e IBM Z & IBM LinuxOne) individual CPUs are much more powerful than other platforms like x86. Hence the CPU requirements are reduced in s390x based environments. Also adjust the resource calculation to be based on 3 OSDs instead of 6


packages/odf/utils/ocs.ts
```diff
@@ -114,14 +115,23 @@ export const isFlexibleScaling = (
...
-  const extraOsds = osdAmount - 6;
+  let { minCpu, minMem, osd } = RESOURCE_PROFILE_REQUIREMENTS_MAP[profile];
+
+  if (isS390x) {
+    const s390xAdjustments = S390X_CPU_ADJUSTMENTS[profile];
+    minCpu = s390xAdjustments.minCpu;
+    osd.cpu = s390xAdjustments.osdCpu;
+  }
+
+  const extraOsds = osdAmount - 3;
```
alfonsomthd, line 134: @malayparida2000 Why are we altering this amount for every case instead for S390X only? I remember you suggested that amount in the first place for a correct calculation...

packages/odf/utils/ocs.ts
```diff
@@ -136,16 +146,22 @@ export const getResourceProfileRequirements = (
...
+ * @param isS390x Whether the platform is s390x.
  * @returns boolean
  */
 export const isResourceProfileAllowed = (
   profile: ResourceProfile,
   cpu: number,
   memory: number,
-  osdAmount: number
+  osdAmount: number,
+  isS390x: boolean = false
```
alfonsomthd, line 158: Suggestion: instead of adding a flag param for a certain architecture (which eventually will lead to adding more & more flags for different architectures), let's pass the architecture and invoke `isS390xArchitecture` inside:
```suggestion
  architecture: string
```

packages/odf/constants/common.ts
```diff
@@ -60,6 +60,26 @@ export const RESOURCE_PROFILE_REQUIREMENTS_MAP: ResourceProfileRequirementsMap =
     },
   };
 
+/**
+ * s390x specific CPU adjustments
+ * in s390x env Individual CPUs are much more powerful than other platforms like x86
+ * Hence the CPU requirements are reduced in s390x env
+ */
+export const S390X_CPU_ADJUSTMENTS = {
```
alfonsomthd, line 68: @malayparida2000 this is a problem that I pointed out in the first implementation: as the backend is not exposing this info as a source of truth for the frontend, eventually the UI will be broken over and over again as not being updated at the same time as the backend.
Better to expose the profile requirements on the backend so the UI can always retrieve the correct values from the backend.
If we keep doing this, we'll keeping opening bugs against the UI component and it's not sustainable.

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/configure-performance.spec.tsx
```diff
@@ -118,7 +118,7 @@ describe('Configure Performance', () => {
       name: /options menu/i,
     });
     expect(dropdown).toHaveTextContent('Balanced mode');
-    expect(screen.getByText(/36 CPUs/i)).toBeInTheDocument();
-    expect(screen.getByText(/87 GiB/i)).toBeInTheDocument();
+    expect(screen.getByText(/42 CPUs/i)).toBeInTheDocument();
```
alfonsomthd, line 121: I see you're just altering the regular case test. Why don't you add an additional test that covers the S390x architecture instead?

packages/odf/constants/common.ts
```diff
@@ -60,6 +60,26 @@ export const RESOURCE_PROFILE_REQUIREMENTS_MAP: ResourceProfileRequirementsMap =
     },
   };
 
+/**
+ * s390x specific CPU adjustments
+ * in s390x env Individual CPUs are much more powerful than other platforms like x86
```
SanjalKatiyar, line 65: nits:
```suggestion
 * in s390x env individual CPUs are much more powerful than other platforms like x86
```

packages/odf/utils/ocs.ts
```diff
@@ -136,16 +146,22 @@ export const getResourceProfileRequirements = (
  * @param profile A resource profile.
  * @param cpu The amount CPUs.
  * @param memory The amount of selected nodes' memory in GiB.
- * @param memory The amount of OSD pods.
+ * @param osdAmount The amount of OSD pods.
+ * @param isS390x Whether the platform is s390x.
```
SanjalKatiyar, line 150: nits:
```suggestion
 * @param isS390x whether the platform is s390x.
```

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/configure-performance.spec.tsx
malayparida2000: I had to alter these cases as the calculation base for regular case changed to be based upon 3 OSD instead of 6.

packages/odf/constants/common.ts
SanjalKatiyar: (this discussion is out of scope of this bug fix)

@malayparida2000 Aren't there 3 source of truths (https://github.com/red-hat-storage/ocs-operator/blob/5a5a31c92630c97160112214638ab17889414025/controllers/defaults/resources.go#L75-L232) ??
What Alfonso meant is to have single source for both OCS and UI.

Documentation update is a different topic.

malayparida2000: @SanjalKatiyar The Resouce Requirements for some of the Individual Daemons(Rook & Noobaa are stored on ocs as you shared), but apart from that the minimum requirement need to take into account also all the CSI related pods, All the operator pods themselves too. 
UI & doc both need the total minimum including everything(operators, rook, noobaa, csi), but ocs operator does not and can not store it.

malayparida2000:
Added test screenshots that I could grab from someone in the IBM Z team

PR #2322, opened 2025-10-08
DFBUGS-2803: [release-4.19] Remove Automatic capacity scaling option

NIKHITHAVADDEMPUDI:
https://issues.redhat.com/browse/DFBUGS-2803

Removed "Automatic capacity scaling option" from the kebab menu in storage system tab of data foundation.

**UI BEFORE CHANGES:**
external-mode-storage-system-action-menu

**UI AFTER CHANGES:**
Screenshot 2025-10-08 at 12 01 31 PM


packages/odf/components/system-list/odf-system-list.tsx
```diff
@@ -295,23 +291,6 @@ const StorageSystemRow: React.FC<RowProps<StorageSystemKind, CustomData>> = ({
     },
   ];
 
-  if (
-    isCapacityAutoScalingAllowed(
```
bipuladh, line 299: You should add a check for the storageCluster resource to be an internal mode cluster. If it is then only add this option.

bipuladh (approved):
/lgtm

PR #2360, opened 2025-10-15
Add RGW + Dev user support to the object browser

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-7747
https://issues.redhat.com/browse/RHSTOR-7748
https://issues.redhat.com/browse/RHSTOR-7749
https://issues.redhat.com/browse/RHSTOR-7743
https://issues.redhat.com/browse/RHSTOR-7744
https://issues.redhat.com/browse/RHSTOR-7746
https://issues.redhat.com/browse/RHSTOR-8054

Endpoints used are introduced here: https://github.com/red-hat-storage/ocs-operator/pull/3536

Screenshot 2025-10-27 at 2 07 54 PM
Screenshot 2025-10-27 at 2 52 15 PM
Screenshot 2025-10-27 at 2 53 30 PM
Screenshot 2025-10-27 at 2 59 28 PM


packages/odf/components/s3-browser/bucket-overview/BucketOverview.tsx
```diff
@@ -344,13 +348,14 @@ const BucketOverviewContent: React.FC<BucketOverviewContentProps> = ({
...
             bucketName={bucketName}
+            providerType={providerType}
             foldersPath={foldersPath}
             currentFolder={currentFolder}
             isCreatedByOBC={isCreatedByOBC}
             noobaaObjectBucket={noobaaObjectBucket}
           />
         }
-        actions={actions(noobaaS3)}
+        actions={actions(s3Client, providerType)}
```
bipuladh: All these extra params can be avoided if we make providerType an internal property of s3Client

packages/odf/redux/provider-hooks/useODFSystemFlags.ts
```diff
@@ -88,6 +89,10 @@ const useODFSystemFlagsPayload = ({
               isCephAvailable: !!ceph,
               // Based on the existence of CephObjectStore CR
               isRGWAvailable: !!cephObjStore,
+              // Based on the existence of CephObjectStore CR
+              ...(!!cephObjStore && {
+                rgwSecureEndpoint: getRGWSecureEndpoint(cephObjStore),
+              }),
```
bipuladh: secureEndpoint is not a flag. Can we just use it where required based on `isRGWAvailable` flag?

packages/odf/components/mcg/ObjectBucketClaim.tsx
```diff
@@ -280,11 +275,9 @@ export const OBCListPage: React.FC<ObjectBucketClaimsPageProps> = (props) => {
   return (
     <>
       <ListPageHeader title="">
-        {!hasNone ? (
```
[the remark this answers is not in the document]
SanjalKatiyar: The entire "Object Storage" nav item is currently hidden for the client. This flow is not supported there yet.

When/If we decide to support it, the existing MCG/RGW flag logic in odf's feature.ts won't work because these calls go through ux-backend-server, which only runs on the provider side (due to the non-admin ODF use case).

We'll need to implement new logic in the client's corresponding feature.ts file to handle this (when/if needed).

packages/odf/redux/provider-hooks/useODFSystemFlags.ts
SanjalKatiyar: Yeah right, we are calling this hook "useODFSystemFlags" !!

When u say `we just use it where required based on isRGWAvailable flag`, do u mean if `isRGWAvailable === true` then send a new API call to fetch COS (that would be unnecessary when we are already fetching COS on global level) ?? Maybe u meant something else ??

We could also just rename `useODFSystemFlags` to `useODFSystem` or `useODFStorages` (this is name of the API endpoint as well, for non-admin flow). Maybe we can do renaming in future too, for now I can add a comment saying: `// Rename the hook if we need to store information beyond just flags (e.g., additional details)`

PR #2378, opened 2025-10-31
Fallback to node CR when Prometheus unavailable

weirdwiz:
Fixes: https://issues.redhat.com/browse/DFBUGS-4424

On ROSA HCP clusters, Prometheus may be unreachable causing
node selection table to fail during storage system creation.

Modified useNodesData to check for prometheus errors and return
nodes with undefined metrics, allowing fallback to node.status.capacity.

### DIFF
--- a/packages/odf/hooks/useNodesData.ts
+++ b/packages/odf/hooks/useNodesData.ts
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { nodeResource } from '@odf/core/resources';
+import { getName } from '@odf/shared';
 import {
   useCustomPrometheusPoll,
   usePrometheusBasePath,
@@ -25,18 +26,19 @@ export const useNodesData = (
 ): [NodeData[], boolean, any] => {
   const [nodes, nodesLoaded, nodesLoadError] =
     useK8sWatchResource<NodeKind[]>(nodeResource);
-  const [utilization, promError, promLoading] = useCustomPrometheusPoll({
+  const [utilization, , promLoading] = useCustomPrometheusPoll({
     query: allNodesUtilizationQueries[NodeQueries.ALL_NODES_MEMORY_TOTAL],
     endpoint: 'api/v1/query' as any,
     basePath: usePrometheusBasePath(),
   });
 
   const loaded = nodesLoaded && !promLoading;
-  const error = nodesLoadError || promError;
+  const error = nodesLoadError;
+
   const nodesData = React.useMemo(() => {
     let nodesData = [];
-    // For data consistency, we must return nodes with their metrics.
-    if (nodes && utilization && loaded && !error) {
+    // Fallback to node.status.capacity when prometheus is unavailable
+    if (nodes && loaded && !error) {
       nodesData = nodes
         .filter(
           (node) =>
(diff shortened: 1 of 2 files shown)


weirdwiz:
waiting for a cluster to verify this PR

packages/odf/hooks/useNodesData.ts
```diff
@@ -46,10 +48,12 @@ export const useNodesData = (
             )
         )
         .map((node: Partial<NodeData>): NodeData => {
-          const metric = _.find(utilization.data.result, [
-            'metric.instance',
-            node.metadata.name,
-          ]);
+          const metric = utilization?.data?.result
+            ? _.find(utilization.data.result, [
```
bipuladh:
```suggestion
            ? _.find(utilization?.data?.result, [
```

packages/odf/hooks/useNodesData.ts
```diff
@@ -46,10 +48,12 @@ export const useNodesData = (
...
+                'metric.instance',
+                node.metadata.name,
```
bipuladh:
```suggestion
                getName(node),
```

packages/odf/hooks/useNodesData.ts
```diff
@@ -31,12 +32,14 @@ export const useNodesData = (
     basePath: usePrometheusBasePath(),
   });
 
-  const loaded = nodesLoaded && !promLoading;
-  const error = nodesLoadError || promError;
+  const promReady = !promLoading && (utilization || !promError);
+  const loaded = nodesLoaded && promReady;
```
SanjalKatiyar: how is this `loaded` ever getting `true` in case of ROSA ??

U have `promError` with bad-gateway error, hence `!promError` and in-turn `promReady` and `loaded` are set to always `false`.

packages/odf/hooks/useNodesData.ts
```diff
@@ -46,16 +49,18 @@ export const useNodesData = (
...
-          const metric = _.find(utilization.data.result, [
-            'metric.instance',
-            node.metadata.name,
-          ]);
+          const metric = utilization?.data?.result
+            ? _.find(utilization?.data?.result, [
+                'metric.instance',
+                getName(node),
+              ])
+            : undefined;
```
SanjalKatiyar: nits: `metric` will automatically be undefined, if `utilization?.data?.result` is empty.
```suggestion
          const metric = _.find(utilization?.data?.result || [], [
                'metric.instance',
                getName(node),
              ]);
```

packages/odf/hooks/useNodesData.ts
```diff
@@ -25,18 +26,19 @@ export const useNodesData = (
...
   const loaded = nodesLoaded && !promLoading;
-  const error = nodesLoadError || promError;
+  const error = nodesLoadError;
+
   const nodesData = React.useMemo(() => {
     let nodesData = [];
-    // For data consistency, we must return nodes with their metrics.
-    if (nodes && utilization && loaded && !error) {
+    // Fallback to node.status.capacity when prometheus is unavailable
+    if (nodes && loaded) {
```
SanjalKatiyar:
```suggestion
    if (nodes && loaded && !error) {
```

PR #2380, opened 2025-11-04
Remove deprecated Dropdown components in PF5

weirdwiz:
Screenshot 2025-11-04 at 3 15 16 PM

Screenshot 2025-11-04 at 3 14 19 PM
Screenshot 2025-11-04 at 1 58 07 PM
Screenshot 2025-11-04 at 2 00 20 PM
Screenshot 2025-11-04 at 2 00 35 PM
Screenshot 2025-11-04 at 3 13 44 PM


packages/shared/src/dropdown/ResourceDropdown.tsx
```diff
@@ -229,63 +230,67 @@ const ResourceDropdown: ResourceDropdown = <T extends unknown>({
     setSearchText('');
   };
 
+  const toggle = (toggleRef: React.Ref<MenuToggleElement>) => (
+    <MenuToggle
+      ref={toggleRef}
+      onClick={loaded && !loadError ? onToggle : () => null}
```
bipuladh: Why null? Isn't disabled handling that case?

packages/mco/components/modals/app-failover-relocate/subscriptions/dr-policy-selector.tsx
```diff
@@ -70,23 +72,22 @@ export const DRPolicySelector: React.FC<DRPolicySelectorProps> = ({
     () =>
       drPolicyList.map((drPolicy, index) => (
         <DropdownItem
-          id={index.toString()}
           data-test={`dr-policy-dropdown-item-${index}`}
           key={getUID(drPolicy)}
-          value={getName(drPolicy)}
-          component="button"
+          value={index}
```
bipuladh: Why are we passing `index` as value we should pass the name instead.

packages/mco/components/modals/app-failover-relocate/subscriptions/target-cluster-selector.tsx
```diff
@@ -229,44 +229,50 @@ export const TargetClusterSelector: React.FC<TargetClusterSelectorProps> = ({
...
+            {state.selectedTargetCluster?.clusterInfo?.clusterName}
+          </FlexItem>
+          <FlexItem>
+            <TargetClusterStatus
+              isClusterAvailable={
+                state.selectedTargetCluster?.isClusterAvailable
+              }
+            />
+          </FlexItem>
+        </Flex>
```
bipuladh: This can be refactored into a component. But I guess this was the original code so we can let this be for now. We can keep it in our issues for later.

PR #2400, opened 2025-11-25
RHSTOR-7956: Minimal IAM Support - View User and DeleteUser

Soundarya189r:
Figma: https://www.figma.com/design/Oi1myOBISCru4T7nzmar7B/IAM?node-id=1-6&p=f&t=bQPU9m4fhuOgaciH-0
Epic:https://issues.redhat.com/browse/RHSTOR-7353

image

image

PR #2403, opened 2025-11-26
RHSTOR-7951: IAM CreateUser flow

NIKHITHAVADDEMPUDI:
https://issues.redhat.com/browse/RHSTOR-7951

**CreateUserForm.tsx** - React form component for creating IAM users with username validation, dynamic tag management, and a modal that displays generated access keys and secret keys with copy/download functionality.

**EmptyIamUserPage.tsx** - Main IAM dashboard component that checks for existing users and displays either an empty state with a "Create user" action or a user list view, with loading states during data fetch. An alert when there is an error

**AddKeyValuePair.tsx** - component for managing key-value pair tags using LazyNameValueEditor supporting add/edit/remove operations with validation and a 50-pair limit.

**useIAMUserFormValidation.tsx** - contain validation logic for user name input.

**packages/odf/utils/s3-iam.ts**- contain validation functions for key and value.

**during fetch error:**
Screenshot 2025-11-28 at 3 47 41 PM

**CreateUserFlow**

https://github.com/user-attachments/assets/3debd88e-8a40-4618-91d5-33921e64a072

used the file mock.ts and replaced const iamClient = createMockIamClient(); to test 

[mocks.ts](https://github.com/user-attachments/files/23825792/mocks.ts)

Screenshot 2025-11-28 at 6 41 51 PM


packages/odf/components/IAM/IAM.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { NoobaaIAMContext } from './iam-context';
+
+const IAM = () => {
+  const { t } = useCustomTranslation();
+  const navigate = useNavigate();
+  const { noobaaIAM } = React.useContext(NoobaaIAMContext);
+  const [hasNoUsers, setHasNoUsers] = React.useState<boolean>(null);
+  const [hasError, setHasError] = React.useState<Error>(null);
+
+  (async () => {
```
TimothyAsirJeyasing: An async IIFE is executed directly in the component body, which will run on every render and cause infinite re-renders. Can we use useEffect

Something like is possible?
React.useEffect(() => {
  let isMounted = true;
  const fetchUsers = async () => {
    try {
      const response = await noobaaIAM.listIAMUsers({});
      if (isMounted) {
        setHasNoUsers(response.Users.length === 0);
      }
    } catch (error) {
      if (isMounted) {
        setHasError(error instanceof Error ? error : new Error(String(error)));
      }
    }
  };
  fetchUsers();
  return () => {
    isMounted = false;
  };
}, [noobaaIAM]);

packages/odf/components/IAM/validationUtils.tsx
```diff
@@ -0,0 +1,30 @@
+export const validateKey = (key: string): string => {
+  const allowedCharsRegex = /^[\p{L}\p{Z}\p{N}_.:/=+\-@]+$/u;
+
+  // Check all conditions
+  const hasMinLengthError = key.length < 1;
+  const hasMaxLengthError = key.length > 128;
```
TimothyAsirJeyasing: Numbers and validation constants are used instead of being centralized

packages/odf/components/IAM/IAM.tsx
```diff
@@ -0,0 +1,90 @@
...
+            </Title>
+            <p className="pf-v5-u-pt-md pf-v5-u-disabled-color-100">
+              {t(
+                'User is an identity who has specific permissions and policies depending on the assigned role'
+              )}
+            </p>
+          </div>
+          <EmptyState variant={EmptyStateVariant.lg}>
+            <EmptyStateHeader
+              titleText="Create user"
```
TimothyAsirJeyasing: "Create user" is hardcoded instead of using translation ?

packages/odf/components/IAM/IAM.tsx
```diff
@@ -0,0 +1,90 @@
...
+  EmptyStateVariant,
+  EmptyStateBody,
+  EmptyStateActions,
+  EmptyStateHeader,
+  EmptyStateFooter,
+  EmptyStateIcon,
+  Alert,
+} from '@patternfly/react-core';
+import { PlusCircleIcon } from '@patternfly/react-icons';
+import { NoobaaIAMContext } from './iam-context';
```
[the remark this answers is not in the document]
NIKHITHAVADDEMPUDI: This file is in soundarya's Prs which is yet to be merged i followed that PR to include this path. https://github.com/red-hat-storage/odf-console/pull/2400/

packages/odf/components/IAM/AddKeyValuePair.tsx
```diff
@@ -0,0 +1,191 @@
...
+  Value: string;
+};
+
+type AddKeyValuePairsProps = {
+  pairs: KeyValuePair[];
+  setPairs: React.Dispatch<React.SetStateAction<KeyValuePair[]>>;
+  disableAddTags: boolean;
+};
+
+const CustomTagInput: React.FC<CustomInputProps> = ({
```
SanjalKatiyar: use default behaviour of NameValueEditor, we shouldn't keep on create diff components for same use case... there should be consistency between diff pages and that's why we have shared components...
```suggestion
```

packages/odf/components/IAM/AddKeyValuePair.tsx
```diff
@@ -0,0 +1,191 @@
...
+          <HelpIcon color="var(--pf-v5-global--disabled-color--100)" />
+        </FlexItem>
+
+        <FlexItem>
+          <Button
+            variant="link"
+            icon={<TagIcon />}
+            iconPosition="left"
+            style={{
+              color: disableAddTags
```
SanjalKatiyar: why do u ever want to disable tags ??
user can add tags first then the username, they are not restricted to add name first and then tag...

plugins/odf/console-extensions.json
```diff
@@ -606,6 +606,29 @@
...
     }
   },
+  {
+    "type": "odf.horizontalNav/tab",
+    "properties": {
+      "id": "iam",
+      "after": "objectbucketclaims",
+      "name": "%plugin__odf-console~IAM%",
+      "contextId": "odf-object-service",
+      "href": "IAM",
```
SanjalKatiyar: plz check and follow convention what's used at other places...
```suggestion
      "href": "iam",
```

packages/odf/components/IAM/CreateUserForm.tsx
```diff
@@ -0,0 +1,444 @@
...
+import { NoobaaIAMContext } from './iam-context';
+import useIAMUserFormValidation from './useIAMUserFormValidation';
+import { validateKey, validateValue } from './validationUtils';
+
+type FormData = {
+  userName: string;
+};
+
+export const CreateUserForm = () => {
+  const descriptiontagRef = React.useRef<HTMLInputElement>(null);
```
SanjalKatiyar: why we need ref and why can't we use event ??

packages/odf/components/IAM/CreateUserForm.tsx
```diff
@@ -0,0 +1,444 @@
...
+    URL.revokeObjectURL(url);
+  };
+
+  return (
+    <Modal
+      variant={ModalVariant.medium}
+      isOpen={isOpen}
+      onClose={() => {
+        setIsOpen(false);
+        navigate?.('IAM');
```
SanjalKatiyar: plz fix link...

packages/odf/components/IAM/CreateUserForm.tsx
```diff
@@ -0,0 +1,444 @@
...
+          </Title>
+          <Button variant="plain" icon={<HelpIcon />} />
+        </div>
+      }
+    >
+      <Form>
+        <Alert
+          variant="warning"
+          title={t(
+            'Important : Write down or download secret key as it will be displayed once.'
```
SanjalKatiyar: fix text and match with mockups...

packages/odf/components/IAM/CreateUserForm.tsx
```diff
@@ -0,0 +1,444 @@
...
+                </Label>
+                <Label className="pf-v5-u-ml-xs" color="cyan">
+                  {t('Primary')}
+                </Label>
+              </>
+            }
+          >
+            <TextInput readOnly readOnlyVariant="plain" value={AccessKeyId} />
+          </FormGroup>
+          <FormGroup label={t('Secretkey')}>
```
SanjalKatiyar:
```suggestion
          <FormGroup label={t('Secret key')}>
```

packages/odf/components/IAM/validationUtils.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const validateValue = (value: string): string => {
+  const allowedCharsRegex = /^[\p{L}\p{Z}\p{N}_.:/=+\-@]*$/u;
+
+  // Check all conditions
+  const hasMaxLengthError = value.length > 256;
+  const hasInvalidChars = value.length > 0 && !allowedCharsRegex.test(value);
+
+  // Return single comprehensive message listing ALL requirements if any error exists
+  if (hasMaxLengthError || hasInvalidChars) {
+    return 'Value cannot exceed 256 characters and can only contain letters, numbers, spaces, and the following special characters: _ . : / = + - @';
```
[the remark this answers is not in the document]
NIKHITHAVADDEMPUDI: I referred the IAM docs for validations https://docs.aws.amazon.com/IAM/latest/APIReference/API_Tag.html

packages/odf/modals/s3-iam/create-user/AccessKeySecretKeyDispModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+          }
+        />
+        <FormSection title={t('Access key information')} titleElement="div">
+          <FormGroup
+            label={
+              <>
+                {t('Access key ')}
+                <Label
+                  className="pf-v5-u-ml-xs"
+                  icon={<CheckCircleIcon />}
```
Soundarya189r: You can use `<GreenCheckCircleIcon />` instead

packages/odf/components/s3-iam/create-user/AddKeyValuePair.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+export const AddKeyValuePairs: React.FC<AddKeyValuePairsProps> = ({
+  pairs,
+  setPairs,
+}) => {
+  const { t } = useCustomTranslation();
+
+  // Convert KeyValuePair[] to nameValuePairs format: [name, value, index][]
+  const nameValuePairs = React.useMemo(() => {
+    return pairs.map((pair, index) => [pair.Key, pair.Value, index]);
```
bipuladh: Why not a map instead of an array for key value pairs maps are better IMO.

packages/odf/components/s3-iam/create-user/CreateUserForm.tsx
```diff
@@ -0,0 +1,490 @@
...
+  const noOp = () => {
+    // Errors are handled by useSWRMutation and surfaced via useEffect
+  };
+
+  const {
+    trigger: updateAccessKey,
+    isMutating: isUpdatingAccessKey,
+    error: updateAccessKeyError,
+  } = useSWRMutation(
+    'update-access-key-cleanup',
```
bipuladh: Make a constant out of it.

packages/odf/components/s3-iam/create-user/AddKeyValuePair.tsx
NIKHITHAVADDEMPUDI: We're constrained by the LazyNameValueEditor API, which requires the array format [name, value, index][]. 
We already use a Map in the reverse conversion (handleUpdateParentData), where it adds value. For the forward conversion (nameValuePairs), 
We're mapping KeyValuePair[] → [name, value, index][] 
We need to preserve the index for the editor

packages/odf/components/s3-iam/create-user/CreateUserForm.tsx
```diff
@@ -0,0 +1,489 @@
...
+          </TextContent>
+          <FormSection>
+            <Alert
+              variant="info"
+              title={t('Creating user will generate access key and secret key')}
+              ouiaId="InfoAlert"
+            />
+
+            <TextInputWithFieldRequirements
+              control={control as any}
```
bipuladh:
```suggestion
              control={control}
```

bipuladh (approved):

PR #2421, opened 2025-12-05
minimal IAM Support - Improvements and Fixes

Soundarya189r:
1. UsersListTable - Tags are not part of the listUsers response - Updated the table to show the Creation Date and Last Activity
FYI -  https://docs.aws.amazon.com/cli/latest/reference/iam/list-users.html
image

2. We need to Delete the Accesskeys before trying to DeleteUser 
FYI - https://docs.aws.amazon.com/IAM/latest/APIReference/API_DeleteUser.html

https://github.com/user-attachments/assets/c143f6dd-308d-4c78-90a5-28356641f2a2

3. Added an empty state when No Tags

https://github.com/user-attachments/assets/cc88304a-00e9-48db-9d49-af78540dfe05

4. Update the EditDescription Modal to call refreshTokens so the Description Tags are updated on the AccesskeyCard

https://github.com/user-attachments/assets/6b8cbce5-d153-4947-8912-5e2f8873179a


packages/odf/modals/s3-iam/DeleteIamUserModal.tsx
```diff
@@ -140,17 +151,26 @@ export const DeleteIamUserModal: React.FC<
...
 
-      setInProgress(false);
-      setDeactivateSuccess(true);
-      refreshTokens();
+      const deletePromises = accessKeys.map((key) =>
+        iamClient.deleteAccessKey({
+          UserName: key.UserName,
+          AccessKeyId: key.AccessKeyId,
+        })
+      );
```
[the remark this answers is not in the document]
SanjalKatiyar: okay, so https://docs.aws.amazon.com/IAM/latest/UserGuide/id_users_remove.html#id_users_deleting_cli clarifies mu doubt.
I understand that cleanup is required. What I was trying to confirm is who is responsible for cleaning up the resources (keys, policies, etc), especially since users can also use the CLI. Looks like CLI users must delete everything manually, which means the cleanup is handled by the console, not the endpoint.

packages/odf/modals/s3-iam/DeleteIamUserModal.tsx
```diff
@@ -107,6 +113,11 @@ export const DeleteIamUserModal: React.FC<
     setInProgress(true);
 
     try {
+      await iamClient.deleteUserPolicy({
```
Soundarya189r: @SanjalKatiyar I have added the deleteUserPolicy here 
This change can be merged after : https://github.com/red-hat-storage/odf-console/pull/2423

Soundarya189r: yeah sure , we can take it as a  bug fix

PR #2502, opened 2026-01-23
DFBUGS-5212: Fix HealthOverviewCard messages in no ceph case

aruniiird:
A waiting message was shown in 'HealthOverviewCard' in case of no OSDs (or no Ceph). This change is to fix it by providing appropriate messages in those cases.

For more details see: https://issues.redhat.com/browse/DFBUGS-5212

### DIFF
--- a/packages/odf/components/overview/Overview.spec.tsx
+++ b/packages/odf/components/overview/Overview.spec.tsx
@@ -10,6 +10,7 @@ import { BrowserRouter } from 'react-router-dom-v5-compat';
 import Overview from './Overview';
 
 const odfNamespace = 'test-ns';
+
 jest.mock('@odf/core/redux/selectors', () => ({
   useODFNamespaceSelector: () => ({
     odfNamespace,
@@ -17,6 +18,16 @@ jest.mock('@odf/core/redux/selectors', () => ({
     odfNsLoadError: null,
     isNsSafe: true,
     isFallbackSafe: true,
+  }),
+  useODFSystemFlagsSelector: () => ({
+    systemFlags: {
+      [odfNamespace]: {
+        isInternalMode: true,
+        isExternalMode: false,
+        isNoobaaStandalone: false,
+      },
+    },
+    areFlagsSafe: true,
   }),
 }));
 
--- a/packages/odf/components/overview/Overview.tsx
+++ b/packages/odf/components/overview/Overview.tsx
@@ -4,11 +4,13 @@ import { ExternalSystemsCard } from '@odf/core/components/overview/external-syst
 import { ObjectStorageCard } from '@odf/core/components/overview/object-storage-card/ObjectStorageCard';
 import { StorageClusterCard } from '@odf/core/components/overview/storage-cluster-card/StorageClusterCard';
 import { StorageClusterCreateModal } from '@odf/core/modals/ConfigureDF/StorageClusterCreateModal';
+import { useODFSystemFlagsSelector } from '@odf/core/redux/selectors';
 import { PageHeading, useCustomTranslation } from '@odf/shared';
 import { useModalWrapper } from '@odf/shared';
 import { Helmet } from 'react-helmet';
 import { useLocation } from 'react-router-dom-v5-compat';
 import { Grid, GridItem } from '@patternfly/react-core';
+import { hasAnyInternalCeph } from '../../utils';
 import { HealthOverviewCard } from './health-overview-card/HealthOverviewCard';
 import './Overview.scss';
 
@@ -20,6 +22,12 @@ const Overview: React.FC = () => {
   const searchParams = new URLSearchParams(location.search);
   const showWelcomeModal = searchParams.get('show-welcome-modal');
   const launchModal = useModalWrapper();
+
+  // Show health card only for internal Ceph clusters.
+  // Can't use hasAnyInternalOCS because MCG standalone is also internal mode,
+  // but it has no Ceph, so ocs_health_score metric won't exist.
+  const { systemFlags, areFlagsSafe } = useODFSystemFlagsSelector();
+  const showHealthCard = areFlagsSafe && hasAnyInternalCeph(systemFlags);
 
   React.useEffect(() => {
     if (showWelcomeModal === 'true') {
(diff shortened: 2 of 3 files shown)


aruniiird:
Attaching the screenshot

Screenshot 2026-05-07 at 4 52 28 PM

@bipuladh , @SanjalKatiyar , @weirdwiz , please let me know if the wordings has to be corrected/changed.

packages/odf/components/overview/health-overview-card/HealthOverviewCard.tsx
```diff
@@ -157,6 +171,31 @@ export const HealthOverviewCard: React.FC = () => {
...
+            <EmptyStateBody>
+              {t(
+                'Infrastructure health monitoring is unavailable because no OSDs are present in the cluster.'
+              )}
+            </EmptyStateBody>
+          </EmptyState>
+        </CardBody>
+      </Card>
+    );
+  }
```
SanjalKatiyar: we already have `showEmptyState` block below, u should re-use that instead of adding another.

SanjalKatiyar:
If issue is external and mcg standalone mode, we can simply remove/hide the card in such scenarios instead of exposing terms like "OSDs" on the UI which user might/might-not know about (text added in this PR is just a suggestion from QE side, not an absolute requirement). Plz discuss such changes with UX team.

And, in this PR we are checking presence of Ceph Cluster CR and adding text suggested by QE. Ceph cluster CR is present in external mode as well, so even current fix is not always accurate.

aruniiird:
@SanjalKatiyar , added the changes to hide the card entirely if it is an external or MCG only cluster. Please take a look

aruniiird:
Here is the screen shot on a non-ceph cluster,

Screenshot 2026-05-18 at 5 32 48 PM

packages/odf/components/overview/Overview.tsx
```diff
@@ -21,6 +23,10 @@ const Overview: React.FC = () => {
   const showWelcomeModal = searchParams.get('show-welcome-modal');
   const launchModal = useModalWrapper();
 
+  // Check if any internal Ceph cluster exists (not external mode or MCG standalone)
+  const { systemFlags, areFlagsSafe } = useODFSystemFlagsSelector();
+  const showHealthCard = areFlagsSafe && hasAnyInternalCeph(systemFlags);
```
SanjalKatiyar, line 30: why not simply use `hasAnyInternalOCS` ?? when we only need this card for internal mode ??

packages/odf/components/overview/Overview.spec.tsx
```diff
@@ -10,14 +10,32 @@ import { BrowserRouter } from 'react-router-dom-v5-compat';
...
-    odfNsLoadError: null,
-    isNsSafe: true,
-    isFallbackSafe: true,
-  }),
+  useODFNamespaceSelector: () => mockODFNamespaceSelector,
+}));
+
+jest.mock('@odf/core/redux', () => ({
+  useODFNamespaceSelector: () => mockODFNamespaceSelector,
+  useODFSystemFlagsSelector: () => mockSystemFlagsSelector,
```
SanjalKatiyar: we have 2 mocks:
`@odf/core/redux/selectors`
`@odf/core/redux`

one of them should be redundant.

aruniiird: Tried to remove the duplication, but due to following reasons unable to do so,

```
Both mocks ARE needed and cannot be consolidated due to technical constraints:

  1. @odf/core/redux re-exports from ./provider-hooks which has initialization side effects (calls
  referenceForModel at module load time)
  2. @odf/core/redux/selectors is a "lighter" import that only includes the selector hooks without the
  provider-hooks side effects
  3. StorageClusterCard.tsx intentionally imports from @odf/core/redux/selectors to avoid triggering these
  side effects during testing
  4. Overview.tsx imports from @odf/core/redux because it needs useODFSystemFlagsSelector which is defined
  in provider-hooks
  5. Jest treats these as separate modules for mocking purposes, so both paths need to be mocked
```

Added a comment to make the reason explicit for others
@SanjalKatiyar , please let me know if there is a way to simplify this.

packages/odf/components/overview/Overview.tsx
aruniiird: Noobaa-only cluster (MCG standalone) is also considered as an internal-mode cluster. But we require to hide the card in both cases, MCG standalone and for external mode clusters.
Added some extra comments to clarify this

packages/odf/components/overview/health-overview-card/HealthOverviewCard.tsx
aruniiird: not needed, as we are hiding the card

packages/odf/components/overview/Overview.tsx
```diff
@@ -4,11 +4,13 @@ import { ExternalSystemsCard } from '@odf/core/components/overview/external-syst
 import { ObjectStorageCard } from '@odf/core/components/overview/object-storage-card/ObjectStorageCard';
 import { StorageClusterCard } from '@odf/core/components/overview/storage-cluster-card/StorageClusterCard';
 import { StorageClusterCreateModal } from '@odf/core/modals/ConfigureDF/StorageClusterCreateModal';
+import { useODFSystemFlagsSelector } from '@odf/core/redux';
```
SanjalKatiyar:
```suggestion
import { useODFSystemFlagsSelector } from '@odf/core/redux/selectors';
```

packages/odf/components/overview/Overview.spec.tsx
```diff
@@ -10,14 +10,36 @@ import { BrowserRouter } from 'react-router-dom-v5-compat';
...
```
SanjalKatiyar: try this: https://github.com/red-hat-storage/odf-console/pull/2502/changes#r3280047662

packages/odf/components/overview/Overview.tsx
SanjalKatiyar: ack.

packages/odf/components/overview/Overview.spec.tsx
aruniiird: Done

SanjalKatiyar:
/retitle [DFBUGS-5212](https://redhat.atlassian.net/browse/DFBUGS-5212): Fix HealthOverviewCard messages in no ceph case

PR #2519, opened 2026-01-27
Allow only firing alerts to be silenced

aruniiird:
Only alerts that are still firing, that is alerts with no 'endTime' set, will be allowed to be silenced.
'Silence' button won't be enabled for already ended alerts.

BZ: https://redhat.atlassian.net/browse/DFBUGS-5374


aruniiird:
Screen recording of the fix

https://github.com/user-attachments/assets/f9dc2725-09fa-4d05-a449-192bdfd04466

bipuladh (approved):

PR #2617, opened 2026-03-13
Fix stalling of 'ocs_health_score' value  in HealthOverview card

aruniiird:
The overall health percentage on HealthOverview card was showing an old value, while current 'ocs_health_score' metric value was different. Previously we were using the query_range endpoint to fetch the data for the time series graph and (from the same range) use the last value as the HealthScore percentage value. There was a time-gap/step-size of ~24mins from one fetch to the next.

In the fix, we are trying a hybrid approach, where for the chart, we use a moderate sampled range-query and for the health score display we use instant-query. If we hit a staled chart, we will automatically trigger refetch.

### DIFF
--- a/packages/odf/components/overview/health-overview-card/HealthOverviewCard.tsx
+++ b/packages/odf/components/overview/health-overview-card/HealthOverviewCard.tsx
@@ -56,13 +56,21 @@ export const HealthOverviewCard: React.FC = () => {
   const [chartRef, chartWidth] = useRefWidth();
   const basePath = usePrometheusBasePath();
 
-  // Fetch health score metric over time (range query)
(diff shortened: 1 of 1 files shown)


yati1998 (commented):
LGTM

aruniiird:
Screenshots

Actual score from prometheus metrics

Screenshot 2026-05-14 at 11 16 49 AM

.
Before the fix (showing the old health-score value)

Screenshot 2026-05-14 at 11 33 43 AM

.
After the fix (showing the current health-score value)
Screenshot 2026-05-14 at 11 28 11 AM

bipuladh (approved):

PR #2640, opened 2026-03-23
[release-4.20-compatibility] DFBUGS-4747: Unable to select NAD from dropdown list when creating storage cluster when using Multus

gourhari1:
https://redhat.atlassian.net/browse/DFBUGS-4747


bipuladh (changes requested):
Please fix it with current component rather than reverting to deprecated one's

bipuladh (approved):

PR #2641, opened 2026-03-24
Upgrade packages in both package.json files

NIKHITHAVADDEMPUDI:
https://redhat.atlassian.net/browse/RHSTOR-8593

Upgrade packages in both package.json files

details regarding upgradation:
https://docs.google.com/document/d/1b4GJ2zxx8a-UXmLxae84kzMDJ-1NCXd-xqTa6GGaPvI/edit?usp=sharing


packages/shared/src/global.d.ts
```diff
@@ -1,11 +1,11 @@
...
-  declare const PLUGIN_BUILD_NAME: string;
-  declare const PLUGIN_BUILD_VERSION: string;
-  declare const OPENSHIFT_CI: string;
+  const PLUGIN_BUILD_I8N_NS: string;
+  const PLUGIN_BUILD_NAME: string;
+  const PLUGIN_BUILD_VERSION: string;
+  const OPENSHIFT_CI: string;
 
-  declare interface Window {
+  interface Window {
```
SanjalKatiyar, line 8: why removing `declare` ??
plz add back if removed by mistake...

NIKHITHAVADDEMPUDI: i intentionally removed declare because of this error https://github.com/red-hat-storage/odf-console/pull/2641#issuecomment-4160582093

bipuladh (approved):

PR #2661, opened 2026-04-01
added support for TNF in UI

gourhari1:
issues: https://redhat.atlassian.net/browse/RHSTOR-8630


packages/odf/components/create-storage-system/create-storage-system.tsx
```diff
@@ -67,6 +90,154 @@ const useIsStorageClusterCRDPresent = (): boolean => {
...
+
+  const [drbdAutostartDaemonSet, drbdAutostartLoaded, drbdAutostartError] =
+    useK8sGet<K8sResourceCommon>(
+      DaemonSetModel,
+      'drbd-autostart',
+      'openshift-kmm'
+    );
+
+  const [drbdConfigureConfigMap, drbdConfigureLoaded, drbdConfigureError] =
+    useK8sGet<ConfigMapKind>(
```
bipuladh: Do you need to watch for this resource instead? Could it be created after the user reaches this step?

packages/odf/components/create-storage-system/create-storage-system.tsx
```diff
@@ -67,6 +90,154 @@ const useIsStorageClusterCRDPresent = (): boolean => {
...
+
+    if (drbdAutostartDaemonSet) {
+      return true;
+    }
+
+    return false;
+  }, [
+    isTNFEnabled,
+    drbdAutostartLoaded,
+    drbdAutostartDaemonSet,
```
bipuladh: Use https://github.com/red-hat-storage/odf-console/blob/3f3d9d139969dc8bd439db0fd0c1462008477b71/packages/shared/src/hooks/deep-compare-memoize.ts#L4 here with JSON resources if you are watching them. Else it will compare with reference.

packages/odf/components/create-storage-system/create-storage-system.tsx
```diff
@@ -278,6 +463,36 @@ const CreateStorageSystemWtihLoader: React.FC = () => {
         </EmptyStateBody>
       </EmptyState>
     </>
+  ) : isTNFEnabled && !isTNFValidated ? (
+    <>
+      <CreateStorageSystemHeader state={{} as any} />
```
bipuladh: This can be shared with https://github.com/red-hat-storage/odf-console/pull/2661/changes#diff-6ccff839065ed7fca7aa4a0aa8622fba73bc23ec06286a38c1ac54cc96d12e2cR451

packages/odf/components/create-storage-system/create-storage-system.tsx
gourhari1: we are not watching the resource here rather we are using usek8Get , which is a one-shot get. it will only refetch when kind, name, namespace, or cluster changes which i think will remain constant while validation is loading.

packages/odf/components/create-storage-system/create-storage-system.tsx
gourhari1: I have confirmed with parth , these resources should be present before we perform any validations, hence we use usek8Get and not watching the resource.

packages/odf/components/create-storage-system/useTNFValidation.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  const [infra, infraLoaded, infraLoadError] = useK8sGet<InfrastructureKind>(
+    InfrastructureModel,
+    DEFAULT_INFRASTRUCTURE
+  );
+
+  const [drbdAutostartDaemonSet, drbdAutostartLoaded, drbdAutostartError] =
+    useK8sGet<K8sResourceCommon>(
+      DaemonSetModel,
+      'drbd-autostart',
```
SanjalKatiyar: const variable, outside FC...

packages/odf/components/create-storage-system/useTNFValidation.ts
```diff
@@ -0,0 +1,118 @@
...
+  ]);
+
+  const isDrbdConfigMapPresent = React.useMemo(() => {
+    if (!isTNFEnabled || drbdConfigureError || !drbdConfigureLoaded) {
+      return false;
+    }
+
+    if (drbdConfigureConfigMap) {
+      return true;
+    }
```
SanjalKatiyar: no need for `useMemo`...

packages/odf/components/create-storage-system/footer.tsx
```diff
@@ -265,8 +266,9 @@ const canJumpToNextStep = (
       );
     case StepsName(t)[Steps.CapacityAndNodes]: {
       const architecture = getNodeArchitectureFromState(nodes);
+      const minNodes = isTNFEnabled ? 2 : MINIMUM_NODES;
       return (
-        nodes.length >= MINIMUM_NODES &&
+        nodes.length >= minNodes &&
```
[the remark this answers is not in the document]
SanjalKatiyar: that should be sufficient...

what I meant above is to have 2 paths, if it's TNF then validations specific to it, else what we currently...

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -275,9 +290,14 @@ const SelectedCapacityAndNodes: React.FC<SelectedCapacityAndNodesProps> = ({
...
       pvsBySc.length
     ) {
-      const pvNodes = getAssociatedNodes(pvsBySc);
+      const pvNodes = getAssociatedNodes(pvsForNodeDiscovery);
       const filteredNodes = allNodes.filter((node) =>
-        pvNodes.includes(node.metadata.name)
+        isTNFEnabled
+          ? pvNodes.includes(node.metadata.name) ||
+            pvNodes.includes(
+              node.metadata?.labels?.['kubernetes.io/hostname'] as string
```
bipuladh:
```suggestion
              getLabel(node, 'kubernetes.io/hostname', '')
```

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -251,10 +252,24 @@ const SelectedCapacityAndNodes: React.FC<SelectedCapacityAndNodesProps> = ({
   const [zones, setZones] = React.useState([]);
 
   const pvsBySc = React.useMemo(
-    () => getSCAvailablePVs(pvs, storageClassName),
+    () => getSCAvailablePVs(pvs ?? [], storageClassName),
     [pvs, storageClassName]
   );
 
+  const pvsForNodeDiscovery = React.useMemo(() => {
```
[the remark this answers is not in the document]
gourhari1: if we only use Available PVs to derive nodes, we can get no nodes even when the cluster contains local storage on workers so we are filtering the PV which is already Bound.

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -275,9 +290,14 @@ const SelectedCapacityAndNodes: React.FC<SelectedCapacityAndNodesProps> = ({
...
-      const pvNodes = getAssociatedNodes(pvsBySc);
+      const pvNodes = getAssociatedNodes(pvsForNodeDiscovery);
       const filteredNodes = allNodes.filter((node) =>
-        pvNodes.includes(node.metadata.name)
+        isTNFEnabled
+          ? pvNodes.includes(node.metadata.name) ||
+            pvNodes.includes(
+              node.metadata?.labels?.['kubernetes.io/hostname'] as string
+            )
+          : pvNodes.includes(node.metadata.name)
```
SanjalKatiyar: here as well u don't need separation based on `isTNFEnabled`... 
```suggestion
pvNodes.includes(node.metadata.name) || pvNodes.includes(
              node.metadata?.labels?.['kubernetes.io/hostname'] as string
            )
```

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -500,7 +536,7 @@ export const CapacityAndNodes: React.FC<CapacityAndNodesProps> = ({
           systemNamespace={systemNamespace}
         />
       )}
-      {(!isNoProvisioner || nodes.length > 0) && (
+      {!isTNFEnabled && (!isNoProvisioner || nodes.length > 0) && (
```
SanjalKatiyar: wren't we supposed to disable unsupported options as per UX ?? or UX wanted them hidden ??

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -513,17 +549,18 @@ export const CapacityAndNodes: React.FC<CapacityAndNodesProps> = ({
...
-          capacityLimit={capacityLimit}
-          className="pf-v6-u-w-75"
-          enable={enableAutoScaling}
-          onChange={onCapacityAutoscalingChange}
-          onLimitSelect={onCapacityAutoscalingSelect}
-          osdAmount={osdAmount}
-          osdSize={String(capacity)}
-        />
-      )}
+      {!isTNFEnabled &&
```
SanjalKatiyar: wren't we supposed to disable unsupported options as per UX ?? or UX wanted them hidden ??

packages/odf/components/create-storage-system/create-storage-system.tsx
```diff
@@ -261,25 +283,59 @@ const CreateStorageSystemWtihLoader: React.FC = () => {
...
+          <EmptyStateFooter>
+            <EmptyStateActions>
+              <Button
+                variant={ButtonVariant.link}
+                component={'a'}
+                href={tnfHomePage(DOC_VERSION)}
+                target="_blank"
+                rel="noopener noreferrer"
+              >
+                {t('Setting up TNF homepage')}
```
SanjalKatiyar: we already have button component wrapper for doc link, plz use that...

SanjalKatiyar (changes requested):
StorageClass dropdown in backingstore wizard step would still show provisioner based classes... we need to filter them out for TNF cases...

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -365,12 +399,12 @@ const SelectedCapacityAndNodes: React.FC<SelectedCapacityAndNodesProps> = ({
                   </Trans>
                 </Content>
               </Content>
-              <Content />
+              {!isTNFEnabled && <Content />}
```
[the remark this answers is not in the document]
gourhari1: removed the guard

PR #2683, opened 2026-04-10
Refine CephFS subvolume card per review feedback

TimothyAsirJeyasing:
Fix the PF6 dropdown wrapper class for the header layout. 
Use the shared CephFS subvolume docs helper instead of a local URL helper. 
Rename the pods fetch helper to avoid shadowing.

## Summary by CodeRabbit

- **Documentation**
  - Updated the CephFS subvolume metrics “Learn more” link to direct users to the relevant version-specific Red Hat documentation.

- **Style**
  - Adjusted the metric selector layout for improved presentation within the CephFS subvolume card.

### DIFF
--- a/packages/ocs/dashboards/persistent-internal/cephfs-subvolume-card/cephfs-subvolume-card.tsx
+++ b/packages/ocs/dashboards/persistent-internal/cephfs-subvolume-card/cephfs-subvolume-card.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { getName, getNamespace } from '@odf/shared';
-import { odfDocBasePath } from '@odf/shared/constants/doc';
+import { cephFSSubvolumeMetricsDoc } from '@odf/shared/constants/doc';
 import { FieldLevelHelp } from '@odf/shared/generic/FieldLevelHelp';
 import { LoadingInline } from '@odf/shared/generic/status-box';
 import { DOC_VERSION } from '@odf/shared/hooks';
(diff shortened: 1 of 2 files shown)


packages/ocs/dashboards/persistent-internal/cephfs-subvolume-card/cephfs-subvolume-card.tsx
```diff
@@ -135,7 +130,7 @@ const CephFSSubvolumeRow: React.FC<RowComponentType<SubvolumeRow>> = ({
   extraProps,
 }) => {
   const { t } = useCustomTranslation();
-  const { metricLabel, fetchPodsForPVC: fetchPods }: CephFSSubvolumeExtraProps =
+  const { metricLabel, fetchPods: fetchPodsForRow }: CephFSSubvolumeExtraProps =
```
SanjalKatiyar, line 133: 1. Why do we need to add `fetchPods` as part of `ExtraProps` ?? What is the point of that ?
2. What is the point of first calling the function `fetchPods` on the top and then locally renaming it to something else ??

TimothyAsirJeyasing: @SanjalKatiyar Sorry for the very late reply — I missed this comment somehow.

You're right, this is over-engineered. I was trying to avoid variable shadowing and ended up making it more confusing.

`fetchPods` does not need to be in `extraProps`. It is already defined in the same file, so the row can call it directly. I'll keep only `metricLabel` in `extraProps` (that value comes from the parent). I'll also drop the extra rename and keep a single descriptive name (`fetchPodsForPVC`).

I'll rebase this PR and update it with that change.

TimothyAsirJeyasing: Sorry for the very late reply. I missed this comment somehow.
You are right,  I was trying to avoid variable shadowing and ended up making it more confusing.
fetchPods does not need to be in extraProps as It is already defined in the same file, so the row can call it directly. I'll keep only metricLabel in extraProps (that value comes from the parent). also I will drop the extra rename and keep a single descriptive name and will rebase this PR and update it with that change.

PR #2685, opened 2026-04-13
DFBUGS-6348: [release-4.21-compatibility] pf6 upgrade

Harshvardhan-handa123:
Bug Link: https://redhat.atlassian.net/browse/DFBUGS-6348


packages/mco/style.scss
```diff
@@ -1 +1,2 @@
-@import '@odf/shared/style.scss';
+@forward '@odf/shared/style.scss';
+@import '@patternfly/patternfly/patternfly.css';
```
SanjalKatiyar, line 2: why is this added here ??

packages/odf/modals/lun-group/DeleteLUNModal.tsx
```diff
@@ -82,20 +81,17 @@ const DeleteLUNModal: React.FC<DeleteLUNModalProps> = ({
...
 
   const location = useLocation();
   const navigate = useNavigate();
 
   const isConfirmNameValid = fsName === confirmName;
 
-  const filteredStorageClasses = storageClasses?.filter(
-    (sc) => sc?.parameters?.volBackendFs === fsName
-  );
-
```
SanjalKatiyar, line 98: why these changes are here ??

SanjalKatiyar: removing this is regression...

.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
SanjalKatiyar, line 1: why this file is here when we use yarn v1 ??

packages/mco/style.scss
Harshvardhan-handa123: During pf-6 upgrade the pf-6 components like buttons , modals etc. were not getting rendered in UI. 
@import adds PF6 base styles so buttons and links render correctly.
This change was introduced in the pr https://github.com/red-hat-storage/odf-console/pull/2552

.yarnrc.yml
Harshvardhan-handa123: My fault, I will revert it.

PR #2690, opened 2026-04-14
Upgrade @openshift-console/dynamic-plugin-sdk

NIKHITHAVADDEMPUDI:
https://redhat.atlassian.net/browse/DFBUGS-6223

Upgrade @openshift-console/dynamic-plugin-sdk to 4.22.0-prerelease.2  and all the dependencies 

**SDK:** Move @openshift-console/dynamic-plugin-sdk, …-internal, and …-webpack to 4.22.0-prerelease.2 so the plugin matches the Console version this branch targets.

**Shared dependencies:** Align with Console-provided modules: React 18, Redux 5, react-redux 9, and react-i18next 16.5.8 (exact + Yarn resolution) so remote-plugin validation matches Console’s expected react-i18next ~16.5.8 line.

**Webpack**: Point react/jsx-runtime(.js) (and dev runtime) at the real React files so react-dnd (and similar) don’t fail on React 18’s package.json exports rules.

**TypeScript / i18n:** Add types/console-upgrade.d.ts (wired in tsconfig) so:
allowObjectInHTMLChildren is enabled for i18next typings (needed for classic <Trans> {{ var }} patterns).
Trans typing is relaxed slightly so older interpolation usage still type-checks.

**SDK API updates:** Adjust code for SDK changes (e.g. Extension typing, RecentEventsBody prop shape) and React 18 typing (children, FC, small HealthBody / HealthItem / StatusPopupSection fixes).

**Tests (Jest):** Add a tiny PaneBody mock + moduleNameMapper so ListPageBody doesn’t accidentally use identity-obj-proxy for @console/.... Add react-router-dom so react-router-dom-v5-compat can load in Jest. Update label-expression tests for PF6 text layout.


packages/client/src/components/dashboards/cards/status-card.tsx
```diff
@@ -23,6 +23,9 @@ import {
 } from '@patternfly/react-core';
 import '../../../../style.scss';
 
+const HealthBodyCompat =
+  HealthBody as React.ComponentType<React.PropsWithChildren>;
+
```
bipuladh, line 28: Why do we need this?

packages/odf/components/create-storage-system/create-storage-system.tsx
```diff
@@ -160,7 +160,7 @@ const CreateStorageSystem: React.FC<CreateStorageSystemProps> = ({
         ...EXTERNAL_CEPH_STORAGE,
         ...(extensions?.map(
           (vendor) => vendor.properties
-        ) as ExternalStorage[]),
+        ) as unknown as ExternalStorage[]),
```
bipuladh, line 163: Why?

packages/shared/src/dropdown/ResourceDropdown.tsx
```diff
@@ -117,10 +117,15 @@ type ResourceDropdownProps<T> = {
   'data-test'?: string;
   id?: string;
   selectedResource?: T;
+  children?: (
+    loaded: boolean,
+    loadError: unknown,
```
bipuladh, line 122: why?

packages/client/src/components/dashboards/cards/status-card.tsx
NIKHITHAVADDEMPUDI: The SDK’s HealthBody type does not declare children, while our markup still nests content inside HealthBody as before the upgrade; this wrapper applies a narrow type assertion so TypeScript accepts that usage without altering runtime behavior.

packages/odf/components/create-storage-system/create-storage-system.tsx
NIKHITHAVADDEMPUDI: as unknown as ExternalStorage[]: After extension resolution, vendor.properties is the data our wizard expects, but its static type from useResolvedExtensions does not line up with ExternalStorage[]; the unknown intermediate cast is the standard TypeScript pattern to bridge that mismatch safely.

packages/shared/src/dropdown/ResourceDropdown.tsx
NIKHITHAVADDEMPUDI: We type children as an optional function because this dropdown calls it as children(loaded, loadError, resources) to let callers render extra UI from the load result; it is not normal nested markup, so calling it a ReactNode would be wrong and would break the type checker where a function is passed in.

NIKHITHAVADDEMPUDI:
will reopen and work on this later when the plugin issue is resolved

PR #2700, opened 2026-04-15
DFBUGS-6375: [release-4.21] fix: replace deprecated useSafetyFirst

TimothyAsirJeyasing:
bz: https://redhat.atlassian.net/browse/DFBUGS-6375

Manuel cherry pick from https://github.com/red-hat-storage/odf-console/pull/2615

Remove useSafetyFirst import from @openshift-console/dynamic-plugin-sdk
Replace it with React.useState for 'loading' and 'isAllowed' state
Add 'isMounted' handling in 'useEffect' to avoid state updates after unmount

We’re not using that SDK hook anymore. We use plain React.useState and implement the same behavior in useEffect with an isMounted flag: we set isMounted = true at the start, only call setAllowed/setLoading when isMounted is true inside the .then()/.catch(), and in the effect cleanup we set isMounted = false. So we still avoid updating state after unmount, but without depending on the deprecated SDK API.

### DIFF
--- a/packages/shared/src/hooks/rbac-hook.ts
+++ b/packages/shared/src/hooks/rbac-hook.ts
@@ -4,7 +4,6 @@ import {
   K8sVerb,
   SelfSubjectAccessReviewKind,
   AccessReviewResourceAttributes,
-  useSafetyFirst,
 } from '@openshift-console/dynamic-plugin-sdk';
 import * as _ from 'lodash-es';
 import { ProjectModel, SelfSubjectAccessReviewModel } from '../models';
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -193,6 +193,7 @@
     "stylelint-order": "^7.0.0",
     "webpack-dev-server": "^5.2.1"
   },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e",
```
SanjalKatiyar: plz remove unnecessary changes...

SanjalKatiyar (changes requested):
1. Is this the only file which uses this hook ?
2. Why we need manual backport, is automated having conflicts ?
3. Plz raise a PR for compatibility branch as well.

package.json
TimothyAsirJeyasing: Sure,

yes this is the only file which uses this hook in our odf-consle repo.
the earlier PR that merged to master included yarn.lock (and related) changes. I used a manual backport so we only carry the hook change and avoid unnecessary lockfile

SanjalKatiyar:
Open a PR for compatibility branch...

SanjalKatiyar:
re-opening, let's merge both in release-4.21 and release-4.21-compatibility, plz do proper testing though...

SanjalKatiyar:
@TimothyAsirJeyasing plz test the changes and add screenshots...

TimothyAsirJeyasing:
image
image

PR #2701, opened 2026-04-16
Skip mon-ip validation for cluster network CIDR

TimothyAsirJeyasing:
https://redhat.atlassian.net/browse/DFBUGS-5756

The initial requirement was to validate both the public network and the cluster network. However, based on later discussions, it was clarified that the public network uses network.rook.io/mon-ip for MON placement, while the cluster network CIDR does not require MON IP checks, as per guidance from the storage team.

Therefore, this PR removes validation for the cluster network MON IP and validates only the cluster network CIDR format.


TimothyAsirJeyasing:
validating only public network, it will not validate for cluster network
Screenshot 2026-04-15 at 11 37 34 PM

even though we give some random ip for the client network it does not do mon ip validation it does only the format check
Screenshot 2026-04-16 at 12 05 07 AM

bipuladh (approved):

PR #2726, opened 2026-04-24
fix bucket-list table layout

Harshvardhan-handa123:
Jira bug link: https://redhat.atlassian.net/browse/DFBUGS-5962
Before SS: 
image

After SS: 
image

Fav icon alignment fix -> 
image (2)

### DIFF
--- a/packages/odf/components/s3-browser/buckets-list-page/bucketListTable.tsx
+++ b/packages/odf/components/s3-browser/buckets-list-page/bucketListTable.tsx
@@ -34,6 +34,7 @@ import {
 } from '@patternfly/react-table';
 import { getProviderLabel } from '../../../utils';
 import { S3Context } from '../s3-context';
+import './bucketListTable.scss';
 
 const getRowActions = (
   t: TFunction<string>,
@@ -96,10 +97,17 @@ const getHeaderColumns = (t: TFunction<string>, favorites: string[]) => {
     {
       columnName: columnNames[0],
       sortFunction: (a, b, c) => sortRows(a, b, c, 'metadata.name', favorites),
+      thProps: {
+        modifier: 'fitContent' as const,
+        className: 'odf-bucket-list__favorites-th',
+      },
     },
     {
       columnName: columnNames[1],
       sortFunction: (a, b, c) => sortRows(a, b, c, 'metadata.name'),
+      thProps: {
+        className: 'pf-v6-u-w-16-on-lg',
+      },
     },
     {
       columnName: columnNames[2],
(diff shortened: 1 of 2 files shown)


Harshvardhan-handa123:
I created a new scss file to add a CSS property so that the star icon in the table header aligns with the other star icons in the column.

packages/odf/components/s3-browser/buckets-list-page/bucketListTable.tsx
```diff
@@ -96,10 +97,17 @@ const getHeaderColumns = (t: TFunction<string>, favorites: string[]) => {
     {
       columnName: columnNames[0],
       sortFunction: (a, b, c) => sortRows(a, b, c, 'metadata.name', favorites),
+      thProps: {
+        modifier: 'fitContent' as const,
+        className: 'odf-bucket-list__favorites-th',
```
SanjalKatiyar, line 102: seems more like a hack-ish fix... why other table headers (name, storage endpoint etc) are aligned with its row content and only favourites column needs this fix ??

packages/odf/components/s3-browser/buckets-list-page/bucketListTable.tsx
```diff
@@ -96,10 +97,17 @@ const getHeaderColumns = (t: TFunction<string>, favorites: string[]) => {
...
+      },
     },
     {
       columnName: columnNames[1],
       sortFunction: (a, b, c) => sortRows(a, b, c, 'metadata.name'),
+      thProps: {
+        className: 'pf-v6-u-w-16-on-lg',
+      },
```
SanjalKatiyar, line 110: create a bucket with maximum name length allowed and check once how the table looks... plz add the screenshot too...

Harshvardhan-handa123: image

packages/odf/components/s3-browser/buckets-list-page/bucketListTable.tsx
Harshvardhan-handa123: I tried a lot of changes to make the header star icon be aligned with rest of the column. Only this soln seemed to be working.

Harshvardhan-handa123: Reason-> 
1- modifier: 'fitContent' as const: column width to only as wide as its content needs. Earlier there was an issue that the star column was as broad as the name col, for which you told the star column width must be small.
2- className: 'odf-bucket-list__favorites-th' : This is done so that the header star column is aligned with the rest of the column. Before this due to some reason it was not aligned.

PR #2729, opened 2026-04-28
fix-aws-regions

Harshvardhan-handa123:
Jira bug : https://redhat.atlassian.net/browse/DFBUGS-5330
After fix screen recording: 

https://github.com/user-attachments/assets/d0a670a1-364c-4055-835f-b07138831143


packages/odf/constants/mcg.ts
```diff
@@ -52,27 +52,46 @@ export const BUCKET_LABEL_NOOBAA_MAP = {
...
+  'ap-south-2',
   'me-south-1',
+  'me-central-1',
   'sa-east-1',
   'us-gov-west-1',
   'us-gov-east-1',
+  'af-south-1',
+  'il-central-1',
+  'mx-central-1',
+  'cn-north-1',
```
Soundarya189r, line 93: @Harshvardhan-handa123 
I don't see `cn-north-1` and `cn-northwest-1 `in the list  https://github.com/noobaa/noobaa-operator/blob/582ec04ea09c4382036cec898c4b8d425b720112/pkg/util/util.go#L1214-L1253

Harshvardhan-handa123: removed

SanjalKatiyar:
This is not a complete fix:
1. We can't have a dropdown whose height is growing uncontrollably with the number of added items. Please use limited height and proper overflow CSS properties.
2. Please check the comments on the bug, along with updating the list we also need to provide a dropdown with a text input.

Harshvardhan-handa123:
> This is not a complete fix:
> 
> 1. We can't have a dropdown whose height is growing uncontrollably with the number of added items. Please use limited height and proper overflow CSS properties.
> 2. Please check the comments on the bug, along with updating the list we also need to provide a dropdown with a text input.

Point 1 -> Done
Point 2-> This uses StaticDropown which internally uses pf dropdown, it does not support user text input. What should I do in this case ?? https://www.patternfly.org/components/menus/dropdown

SanjalKatiyar:
>Point 2-> This uses StaticDropown which internally uses pf dropdown, it does not support user text input. What should I do in this case ?? https://www.patternfly.org/components/menus/dropdown

Plz check if any other component in PF supports that, else talk to UX and we can create our own new component. It's straightforward to combine TextInput + Dropdown to create the expected behaviour.

Harshvardhan-handa123:
> > Point 2-> This uses StaticDropown which internally uses pf dropdown, it does not support user text input. What should I do in this case ?? https://www.patternfly.org/components/menus/dropdown
> 
> Plz check if any other component in PF supports that, else talk to UX and we can create our own new component. It's straightforward to combine TextInput + Dropdown to create the expected behaviour.

Yes, pf selectMenu supports that. I can use that here.

packages/shared/src/dropdown/TypeaheadDropdown.tsx
```diff
@@ -50,30 +54,36 @@ export const TypeaheadDropdown: React.FC<TypeaheadDropdownProps> = ({
...
-          children: t('No results found for {{ filterValue }}', {
-            filterValue,
-          }),
-          value: NO_RESULTS,
-        },
-      ];
+      selectOptions = isCreatable
+        ? [
+            {
+              children: t('Create new option "{{filterValue}}"', {
```
bipuladh: This might be causing issues

Harshvardhan-handa123: Sure, thanks !

packages/odf/constants/mcg.ts
```diff
@@ -52,27 +52,44 @@ export const BUCKET_LABEL_NOOBAA_MAP = {
...
   'ap-south-1',
+  'ap-south-2',
   'me-south-1',
+  'me-central-1',
   'sa-east-1',
   'us-gov-west-1',
   'us-gov-east-1',
+  'af-south-1',
+  'il-central-1',
+  'mx-central-1',
```
SanjalKatiyar, line 92: I see 2 more: https://github.com/noobaa/noobaa-operator/blob/5.22/pkg/util/util.go#L1270-L1271

packages/odf/components/mcg-endpoints/s3-endpoint-type.tsx
```diff
@@ -15,7 +15,7 @@ import {
 } from '@patternfly/react-core';
 import { AWS_REGIONS, StoreProviders, StoreType } from '../../constants';
 import './noobaa-provider-endpoints.scss';
-import { awsRegionItems, endpointsSupported } from '../../utils';
```
SanjalKatiyar, line 18: plz cleanup un-used exports...
if `awsRegionItems` not used anywhere else, plz remove it entirely from the corresponding file...

packages/shared/src/dropdown/TypeaheadDropdown.tsx
```diff
@@ -126,10 +144,25 @@ export const TypeaheadDropdown: React.FC<TypeaheadDropdownProps> = ({
...
   ) => {
-    const optionValue = value || '';
-    selectOption(optionValue);
-    onSelect(optionValue);
-    setIsUserSelection(true);
+    if (value) {
+      if (isCreatable && value === CREATE_NEW) {
+        const newOption: SelectOptionProps = {
+          value: filterValue,
+          children: filterValue,
```
SanjalKatiyar, line 154: why are u using `filterValue` instead of directly `value` ??

packages/shared/src/dropdown/TypeaheadDropdown.tsx
```diff
@@ -126,10 +144,25 @@ export const TypeaheadDropdown: React.FC<TypeaheadDropdownProps> = ({
...
+        };
+        setCreatedOptions((prev) => [...prev, newOption]);
+        setSelected(filterValue);
+        setInputValue(filterValue);
+        setFilterValue('');
```
SanjalKatiyar: lot of these are already handled by `selectOption(optionValue);`, why not use that ??

packages/shared/src/dropdown/TypeaheadDropdown.tsx
Harshvardhan-handa123: Value is CREATE_NEW (A constant) for that.
When I enter the filterValue and no option matches that -> A new option is displayed whose
1- children (displayed value)  -> "Create option`filterValue`".
2- value -> CREATE_NEW (see line 148)
When I click on this option two things happen:
1- filterValue is selected as the selectedValue (line 154)
2- A new option in the dropdown is added whose value and children is filterValue.

packages/shared/src/dropdown/TypeaheadDropdown.tsx
Harshvardhan-handa123: Yes, I'll use that.

packages/odf/components/mcg-endpoints/s3-endpoint-type.tsx
Harshvardhan-handa123: Sure, I'll remove this export

packages/odf/constants/mcg.ts
Harshvardhan-handa123: Done

PR #2731, opened 2026-04-29
Count the only the unique active alerts by name

aruniiird:
Since from the backend, we calculate the 'HealthScore' using only the active alerts, unique by name, let's do the same in the UI calculation (for minor/major/critical) as well.

BZ: https://redhat.atlassian.net/browse/DFBUGS-5919


aruniiird:
Before

Screenshot 2026-04-29 at 11 32 15 AM

After

Screenshot 2026-04-29 at 11 20 40 AM

yati1998:
this looks fine, thank you for the update

bipuladh (approved):

PR #2736, opened 2026-04-29
Update the Add Capacity modal

NIKHITHAVADDEMPUDI:
https://redhat.atlassian.net/browse/DFBUGS-5050

Update the Add Capacity modal to only list StorageClasses already in use by existing deviceSet

created a new storage class set1 not in use by any device sets.

**before fix:**

Screenshot 2026-04-28 at 2 13 12 PM

**after fix:**

https://github.com/user-attachments/assets/30043968-1dd8-4128-8508-c42926d1e8a7

## Summary by CodeRabbit

- **New Features**
  - The Add Capacity dialog now lists only storage classes already used by the cluster.
  - Capacity additions now target an existing matching device set and update its capacity.

- **Bug Fixes**
  - Prevented capacity additions from creating new device sets when no matching device set exists.
  - The operation now stops safely when no suitable device set is available.

### DIFF
--- a/cypress/tests/multiple-storageclass-selection.spec.ts
+++ b/cypress/tests/multiple-storageclass-selection.spec.ts
@@ -1,17 +1,13 @@
 import { K8sResourceKind, getCurrentDeviceSetIndex } from '../mocks/ksm';
-import {
-  testEbsSC,
-  testNoProvisionerSC,
-  getPVJSON,
-} from '../mocks/storageclass';
+import { testEbsSC, testNoProvisionerSC } from '../mocks/storageclass';
 import {
   withJSONResult,
   fetchStorageClusterJson,
-  fetchWorkerNodesJson,
   addCapacity,
-  newStorageClassTests,
+  assertStorageClassNamesAbsentInAddCapacityDropdown,
   existingStorageClassTests,
   IndexAndDeviceSet,
+  openAddCapacityModal,
   UidAndDeviceSet,
 } from '../views/multiple-storageclass';
 
@@ -28,18 +24,6 @@ describe('Add capacity using multiple storage classes', () => {
     cy.exec(
       `echo '${JSON.stringify(testNoProvisionerSC)}' | kubectl apply -f -`
     );
-    fetchWorkerNodesJson().then((res) => {
-      const nodes = JSON.parse(res.stdout);
-      const { name: scName } = testNoProvisionerSC.metadata;
-      nodes.items.forEach((node, id) => {
-        const nodeName = node.metadata.name;
-        cy.exec(
-          `echo '${JSON.stringify(
-            getPVJSON(id, nodeName, scName)
-          )}' | kubectl apply -f -`
-        );
-      });
-    });
   });
 
   beforeEach(() => {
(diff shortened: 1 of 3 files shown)


cypress/tests/multiple-storageclass-selection.spec.ts
```diff
@@ -70,16 +69,6 @@ describe('Add capacity using multiple storage classes', () => {
...
 
-  it('Add capacity with a new storage class having EBS as provisioner', () => {
-    const { name: scName } = testEbsSC.metadata;
-    const iAndD: IndexAndDeviceSet = { index: 0, deviceSets: [] };
-    addCapacity(scName);
-    fetchStorageClusterJson().then((res) => {
-      withJSONResult(res, scName, iAndD);
-      newStorageClassTests(beforeCapacityAddition, iAndD, true);
-    });
-  });
```
bipuladh, line 81: Why are we removing this test?

cypress/tests/multiple-storageclass-selection.spec.ts
```diff
@@ -94,14 +83,4 @@ describe('Add capacity using multiple storage classes', () => {
...
-
-  it(`Add capacity with a new storage class having NO-PROVISIONER as provisioner`, () => {
-    const { name: scName } = testNoProvisionerSC.metadata;
-    const iAndD: IndexAndDeviceSet = { index: 0, deviceSets: [] };
-    addCapacity(scName);
-    fetchStorageClusterJson().then((res) => {
-      withJSONResult(res, scName, iAndD);
-      newStorageClassTests(beforeCapacityAddition, iAndD, false);
-    });
-  });
```
bipuladh, line 106: We should modify this test to not show the storage class rather than to delete it?

cypress/tests/multiple-storageclass-selection.spec.ts
```diff
@@ -94,14 +83,4 @@ describe('Add capacity using multiple storage classes', () => {
...
 });
```
SanjalKatiyar, line 107: instead of just removing the existing tests, plz add a new test for the updated use case, test which ensures a "new" StorageClass is not rendered/displayed by the modal...

cypress/views/multiple-storageclass.ts
```diff
@@ -33,24 +33,6 @@ export const addCapacity = (scName: string) => {
...
-  const portabilityStatus = portability ? 'enabled' : 'disabled';
-  cy.log('New device set is created');
-  expect(iAndD.deviceSets.length).to.equal(
-    beforeCapacityAddition.deviceSets.length + 1
-  );
-
-  cy.log('Device count is 1 in the new device set');
-  expect(iAndD.deviceSets[iAndD.index].count).to.equal(1);
-
-  cy.log(`Osd portability is ${portabilityStatus} in the new device set`);
```
SanjalKatiyar, line 50: instead of just removing the existing tests, plz add a new test for the updated use case, test which ensures a "new" StorageClass is not rendered/displayed by the modal...

packages/odf/modals/add-capacity/add-capacity-modal.tsx
SanjalKatiyar: this flow shouldn't get triggered now... `deviceSetIndex` should never be "-1" if we are not allowing new StorageClasses.

plz remove this and add `console.error` log instead here... and test it accordingly...

cypress/tests/multiple-storageclass-selection.spec.ts
NIKHITHAVADDEMPUDI: the no-provisioner case is covered in https://github.com/red-hat-storage/odf-console/pull/2736/changes#diff-065f7cec8bbb5d4254705a33c61a9a72ddb03b37c7605cdc2b8f99a21d2aeb21R45-R50:~:text=it(%27Add%20Capacity%20modal,%5D)%3B by applying test-no-prov-sc and asserting it never appears in the Add Capacity StorageClass dropdown via assertStorageClassNamesAbsentInAddCapacityDropdown, which matches the modal logic in add-capacity-modal.tsx that only shows StorageClasses already referenced by storageDeviceSets

cypress/tests/multiple-storageclass-selection.spec.ts
```diff
@@ -28,18 +24,6 @@ describe('Add capacity using multiple storage classes', () => {
...
-      nodes.items.forEach((node, id) => {
-        const nodeName = node.metadata.name;
-        cy.exec(
-          `echo '${JSON.stringify(
-            getPVJSON(id, nodeName, scName)
-          )}' | kubectl apply -f -`
-        );
-      });
-    });
   });
```
SanjalKatiyar, line 43: why did we remove this ??

cypress/tests/multiple-storageclass-selection.spec.ts
```diff
@@ -56,35 +40,31 @@ describe('Add capacity using multiple storage classes', () => {
...
+    openAddCapacityModal();
+    assertStorageClassNamesAbsentInAddCapacityDropdown([
+      testEbsSC.metadata.name,
+      testNoProvisionerSC.metadata.name,
+    ]);
   });
 
   it('Add capacity with an existing storage class having EBS as provisioner', () => {
-    const { name: scName } = testEbsSC.metadata;
+    const { name: preferredScName } = testEbsSC.metadata;
```
SanjalKatiyar: why the variable name change ??

cypress/tests/multiple-storageclass-selection.spec.ts
```diff
@@ -56,35 +40,31 @@ describe('Add capacity using multiple storage classes', () => {
...
     const iAndD: IndexAndDeviceSet = { index: 0, deviceSets: [] };
     const { deviceSets } = beforeCapacityAddition;
-    const index = getCurrentDeviceSetIndex(deviceSets, scName);
+    let index = getCurrentDeviceSetIndex(deviceSets, preferredScName);
+    let scName = preferredScName;
```
SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/2736/changes#r3265952334

name change and assigning back to same name ??

NIKHITHAVADDEMPUDI:
Test 1: Creates StorageClasses that exist but are not on any storageDeviceSet, opens Add Capacity, and explicitly checks test-ebs-sc and test-no-prov-sc are not in the dropdown

https://github.com/red-hat-storage/odf-console/pull/2736/changes#diff-065f7cec8bbb5d4254705a33c61a9a72ddb03b37c7605cdc2b8f99a21d2aeb21R45-R50:~:text=it(%27Add%20Capacity%20modal,%5D)%3B

Test 2: Finds a StorageClass that is on a device set (getCurrentDeviceSetIndex), runs Add Capacity for that name, then explicitly checks via kubectl that the same device set’s count increased by 1 and no new device set was created. Dropdown presence is implicit via a successful click, not a separate should('exist') assertion.

https://github.com/red-hat-storage/odf-console/pull/2736/changes#:~:text=53-,it(%27Add%20capacity%20with%20an%20existing%20storage%20class%20having%20EBS,%2D,-it(%60Add

_____________________________________________________________
https://github.com/red-hat-storage/odf-console/pull/2736/changes#diff-065f7cec8bbb5d4254705a33c61a9a72ddb03b37c7605cdc2b8f99a21d2aeb21L31:~:text=%2D-,fetchWorkerNodesJson().then((res)%20%3D%3E%20%7B,%7D)%3B,-43 

we removed the above lines because 

It was only needed for the old test that added capacity with a new no-provisioner StorageClass. That test is removed because the modal no longer lets you pick unused StorageClasses.

https://github.com/red-hat-storage/odf-console/pull/2736/changes#diff-065f7cec8bbb5d4254705a33c61a9a72ddb03b37c7605cdc2b8f99a21d2aeb21R45-R50:~:text=it(%27Add%20Capacity%20modal,%5D)%3B

The new test opens Add Capacity and checks that test-ebs-sc and test-no-prov-sc (created on the cluster but not used by any device set) do not show up in the dropdown.

The new test only checks the dropdown. It doesn’t add capacity with those StorageClasses, so we don’t need PVs on every worker.

bipuladh (approved):

SanjalKatiyar:
/hold
https://github.com/red-hat-storage/odf-console/pull/2947 needs to be merged first.

PR #2737, opened 2026-04-29
Add Lazy loading to IAM

NIKHITHAVADDEMPUDI:
**_An internal dev change without a bug_**

Add lazy-loaded chunks for the IAM user overview Details and Access Keys tabs using the same React.lazy pattern as S3 bucket and vector bucket overviews.

Switch those tab modules to a default export so dynamic imports resolve cleanly. 

No functional change intended beyond loading those panels on demand when the tab is opened.

everything seems working as usual

https://github.com/user-attachments/assets/f3fd2637-9306-4bc4-aa33-477d30b3dc11


SanjalKatiyar:
/hold
https://github.com/red-hat-storage/odf-console/pull/2947 needs to be merged first.

PR #2739, opened 2026-04-29
fix: TNF day-1 device set count/replica

TimothyAsirJeyasing:
https://redhat.atlassian.net/browse/DFBUGS-6474

TNF day‑1 was hardcoding count: 1 and replica: 2 in getTNFSpecs, which is wrong for host failure domain.  Now it sets replica: 1 and count from available PVs via getDeviceSetCount,
and getOCSRequestData passes availablePvsCount into getTNFSpecs.

### DIFF
--- a/packages/odf/components/utils/common.ts
+++ b/packages/odf/components/utils/common.ts
@@ -516,7 +516,7 @@ export const getOCSRequestData = ({
       },
     };
   } else if (isTNFEnabled) {
-    requestData.spec = getTNFSpecs(scName, storage);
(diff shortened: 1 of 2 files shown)


TimothyAsirJeyasing:
Screenshot 2026-05-09 at 1 46 13 AM


image

spec:
  flexibleScaling: true
  storageDeviceSets: [
    0: { name: "ocs-deviceset", config: {}, count: 1, replica: 1, portable: false, ... }
  ]

PR #2742, opened 2026-04-30
Fix null reference crashes in DR components

vbnrh:
[DFBUGS-6377](https://redhat.atlassian.net/browse/DFBUGS-6377): Add null guards to prevent TypeError when resource is undefined in DRStatus and AppManagePoliciesModal components (the `in` operator requires an object operand), and initialize the AbortController ref to avoid accessing an uninitialized signal in useSafeFetch.

### DIFF
--- a/packages/shared/src/hooks/custom-prometheus-poll/safe-fetch-hook.ts
+++ b/packages/shared/src/hooks/custom-prometheus-poll/safe-fetch-hook.ts
@@ -10,10 +10,10 @@ export type SafeFetchProps = {
 };
 
 export const useSafeFetch = () => {
-  const controller = useRef<AbortController>();
+  const controller = useRef<AbortController>(new AbortController());
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


packages/shared/src/hooks/custom-prometheus-poll/safe-fetch-hook.ts
```diff
@@ -10,7 +10,7 @@ export type SafeFetchProps = {
 };
 
 export const useSafeFetch = () => {
-  const controller = useRef<AbortController>();
+  const controller = useRef<AbortController>(new AbortController());
   useEffect(() => {
     controller.current = new AbortController();
     return () => controller.current.abort();
```
SanjalKatiyar: it's still buggy if u are creating new object instance in `useEffect`...
```suggestion
  const controller = useRef<AbortController>(new AbortController());
useEffect(() => {
return () => controller.current.abort();
```

PR #2756, opened 2026-05-07
DFBUGS-6793: cluster name overlapping with dotted line boundary fix

Harshvardhan-handa123:
Jira link: https://redhat.atlassian.net/browse/DFBUGS-6793

Before SS: 
image

After SS: 
image


SanjalKatiyar:
/retitle [DFBUGS-6793](https://redhat.atlassian.net/browse/DFBUGS-6793): cluster name overlapping with dotted line boundary fix

PR #2759, opened 2026-05-07
Fix sidebar DR status not auto-refreshing

vbnrh:
## Summary
- Fixes [DFBUGS-6718](https://redhat.atlassian.net/browse/DFBUGS-6718): Applications sidebar DR Status not auto-refreshing in topology view
- Sidebar was reading data from `selectedElement.getData()` which is updated by `controller.fromModel()` in a post-render effect — always one render behind
- Derives sidebar data directly from the reactive `model` (computed from context maps via `useMemo`) using the selected element's ID as a lookup key
- Closes the sidebar when the selected element disappears from the model (e.g. operation completes and the failover node is removed)

### DIFF
--- a/packages/mco/components/topology/Topology.tsx
+++ b/packages/mco/components/topology/Topology.tsx
@@ -126,9 +126,27 @@ const TopologyViewComponent: React.FC = () => {
 
   const controlButtons = useTopologyControls({ controller });
 
-  // Get the selected element's data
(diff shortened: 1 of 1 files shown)


vbnrh:
https://github.com/user-attachments/assets/d4b2d08e-08a9-4767-9b4b-701eef20ed27

bipuladh (approved):

PR #2760, opened 2026-05-07
Allow distribution of NFS StorageClass

SanjalKatiyar:
https://redhat.atlassian.net/browse/DFBUGS-5068

Screenshot 2026-05-07 at 2 42 42 PM
Screenshot 2026-05-07 at 2 43 11 PM

### DIFF
--- a/packages/shared/src/constants/storageClasses.ts
+++ b/packages/shared/src/constants/storageClasses.ts
@@ -19,3 +19,5 @@ export const CEPH_PROVISIONERS = [
   'cephfs.csi.ceph.com',
   'rbd.csi.ceph.com',
 ];
+
+export const CEPH_NFS_PROVISIONER = 'nfs.csi.ceph.com';
--- a/packages/shared/src/utils/storage.ts
+++ b/packages/shared/src/utils/storage.ts
@@ -6,6 +6,7 @@ import {
   TIB_CONVERSION_DIVISOR,
   STORAGE_SIZE_UNIT_NAME_MAP,
   NOOBAA_PROVISIONER,
+  CEPH_NFS_PROVISIONER,
 } from '@odf/shared/constants';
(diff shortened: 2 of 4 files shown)


packages/shared/src/constants/storageClasses.ts
```diff
@@ -19,3 +19,5 @@ export const CEPH_PROVISIONERS = [
   'cephfs.csi.ceph.com',
   'rbd.csi.ceph.com',
 ];
+
+export const CEPH_NFS_PROVISIONER = 'nfs.csi.ceph.com';
```
bipuladh, line 23: This same array can be expanded with this just update the comment.

packages/shared/src/utils/storage.ts
```diff
@@ -94,5 +95,16 @@ export const isCephProvisioner = (scProvisioner: string): boolean => {
 export const isCephDriver = (driver: string): boolean =>
   _.endsWith(driver, 'ceph.com');
 
-export const isCephOrNooBaaProvisioner = (provisioner: string): boolean =>
```
bipuladh, line 97: Keep this as it is. And just update the old existing array.

bipuladh (changes requested):

packages/shared/src/constants/storageClasses.ts
SanjalKatiyar: `CEPH_PROVISIONERS` is used at multiple places, which might/might-not need NFS.
NFS is a special case, this is optionally available, hence it's better not to add it as part of global array.

packages/shared/src/utils/storage.ts
SanjalKatiyar: It's intentional that way.

https://github.com/red-hat-storage/odf-console/pull/2760#discussion_r3242277642

PR #2761, opened 2026-05-07
Fix sidebar showing DRPC name instead of application name

vbnrh:
## Summary
- Fixes [DFBUGS-6707](https://redhat.atlassian.net/browse/DFBUGS-6707): Applications sidebar shows DRPC name instead of application name
- Uses existing `getApplicationName(pav)` utility which reads `applicationRef.name` from PAV status, falling back to PAV metadata name

### DIFF
--- a/packages/mco/components/topology/sidebar/AppSidebar.tsx
+++ b/packages/mco/components/topology/sidebar/AppSidebar.tsx
@@ -142,7 +142,7 @@ export const AppSidebar: React.FC<AppSidebarProps> = ({ edgeData }) => {
     );
     const volumeLastGroupSyncTime = op.drpc?.status?.lastGroupSyncTime;
     return {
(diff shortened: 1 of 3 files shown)


vbnrh:
image

SanjalKatiyar:
@vbnrh few questions:
1. What do we show in protected applications list page, PAV name or DRPC name ? We should the same here in Topology, right ?
2. In which case PAV and DRPC names are diff ? In case of Subscription apps ?

vbnrh:
> @vbnrh few questions:
> 
>     1. What do we show in protected applications list page, PAV name or DRPC name ? We should the same here in Topology, right ?
> 
>     2. In which case PAV and DRPC names are diff ? In case of Subscription apps ?

1. Its the application's name that resides in PAV status
2. They are the same in all cases

PR #2762, opened 2026-05-07
Fix policy column showing cluster names instead of policy name

vbnrh:
## Summary
- Fixes [DFBUGS-6795](https://redhat.atlassian.net/browse/DFBUGS-6795): Policy column in operations sidebar shows failover direction (e.g., `Failover: local-cluster → dr1-may-7-26`) instead of the DR policy name (e.g., `odr-policy-5m`)
- Uses existing `getPAVDRPolicyName()` utility to read the actual policy name from PAV status


vbnrh:
image

PR #2763, opened 2026-05-07
Update deleting phase of cluster in status card.

NIKHITHAVADDEMPUDI:
bug link: https://redhat.atlassian.net/browse/DFBUGS-4273

**description :** When storage cluster is stuck in Deleting state, on Storage cluster page it's still green, as if everything is OK and its misleading.

**cause:** When this dashboard was observed, CephCluster was not it deletion phase yet. 

**fix:** Updated the ui to show a spinner and show deleting status in the status card as suggested and approved by UX team

**before fix:**

https://github.com/user-attachments/assets/7bdd6caa-7124-46c8-bcfe-c5df748a8ce8

**after fix:**

https://github.com/user-attachments/assets/cc972ec9-ecd0-4c39-9b6d-590fb953e5a8

## Summary by CodeRabbit

- **New Features**
  - Status cards now include StorageCluster health when determining overall storage health.
  - Added clearer progress reporting while a storage cluster is being deleted.

- **Bug Fixes**
  - Health checks are no longer shown while storage cluster information is loading or during deletion.
  - Deleting storage clusters now display a “Deleting” progress state instead of unrelated health results.
  - Ceph health remains available as a fallback when StorageCluster health cannot be determined.

### DIFF
--- a/packages/ocs/dashboards/persistent-internal/status-card/status-card.tsx
+++ b/packages/ocs/dashboards/persistent-internal/status-card/status-card.tsx
@@ -1,11 +1,16 @@
 import * as React from 'react';
 import { useGetInternalClusterDetails } from '@odf/core/redux/utils';
-import { getResourceInNs as getCephClusterInNs } from '@odf/core/utils';
+import {
+  getResourceInNs as getCephClusterInNs,
+  getStorageClusterInNs,
+  isClusterDeleting,
+} from '@odf/core/utils';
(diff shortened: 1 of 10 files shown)


SanjalKatiyar:
We should keep Overview dashboard status in sync with Block and File dashboard, else we will get mismatch bug later, @bipuladh wdyt ?

NIKHITHAVADDEMPUDI:
image
image

PR #2800, opened 2026-05-12
fix show success decorator for FailedOver on DR topology

TimothyAsirJeyasing:
https://redhat.atlassian.net/browse/DFBUGS-6797

getDecoratorForStatus mapped DRStatus.FailedOver to the warning triangle while sidebar-utils and disaster-recovery already treat FailedOver as a successful completed failover. Topology app nodes use getDecoratorForStatus for quadrant decorators, which caused the BZ mismatch on the target cluster.

Align FailedOver with Relocated/Healthy: CheckCircle and NodeStatus.success.


TimothyAsirJeyasing:
before:
image


after:
image

PR #2807, opened 2026-05-13
Include isNoProvisioner check to validateAdvancedSettingsStep

Soundarya189r:
https://redhat.atlassian.net/browse/DFBUGS-6863

### DIFF
--- a/packages/odf/components/create-storage-system/footer.tsx
+++ b/packages/odf/components/create-storage-system/footer.tsx
@@ -149,7 +149,8 @@ const validateAdvancedSettingsStep = (
   advancedSettings: WizardState['advancedSettings'],
   type: BackingStorageType,
   nodeCount: number,
-  flexibleScaling: boolean
+  flexibleScaling: boolean,
+  isNoProvisioner: boolean
 ) => {
   const {
     useErasureCoding,
@@ -165,7 +166,7 @@ const validateAdvancedSettingsStep = (
 
   // The Next button is disabled only when no VolumeSnapshotClass is selected for MCG Standalone when automatic backup option enabled.
   const hasForcefulDeploymentButNoConfirmation =
-    type === BackingStorageType.LOCAL_DEVICES &&
+    (type === BackingStorageType.LOCAL_DEVICES || isNoProvisioner) &&
     enableForcefulDeployment &&
     !(forcefulDeploymentConfirmation?.trim() === 'CONFIRM');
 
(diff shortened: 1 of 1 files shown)


SanjalKatiyar:
@Soundarya189r I didn't get this fix ??
```
When enabling forceful deployment in the "Create Storage System" wizard (Advanced Settings step), the UI displays a warning and asks the user to type CONFIRM before proceeding. However, the input field validation is broken: the Next button becomes active regardless of what text is entered in the confirmation field. Any arbitrary string (e.g. abc) allows the user to proceed.
```
^ is the ask in the Jira, fix in this PR is diff.

Soundarya189r:
> @Soundarya189r I didn't get this fix ??
> 
> ```
> When enabling forceful deployment in the "Create Storage System" wizard (Advanced Settings step), the UI displays a warning and asks the user to type CONFIRM before proceeding. However, the input field validation is broken: the Next button becomes active regardless of what text is entered in the confirmation field. Any arbitrary string (e.g. abc) allows the user to proceed.
> ```
> 
> ^ is the ask in the Jira, fix in this PR is diff.


**When Enable forceful deployment is shown**

It appears on the Advanced Settings step of the Create storage system wizard.

That step (and the checkbox) are shown only when both of these are true:

- Deployment type is not MCG-only (not NooBaa standalone).
And either:
- Backing storage is Local devices, or
- The selected storage class uses the no provisioner provisioner (NO_PROVISIONER — typical for local / existing storage).

**BugFix:** 

Next stays disabled until we type CONFIRM, but only for Local devices backing storage, so I added the isNoProvisioner to the footer.

Link : https://github.com/red-hat-storage/odf-console/pull/2807/changes#:~:text=isNoProvisioner%20&&

packages/odf/components/create-storage-system/footer.tsx
```diff
@@ -166,6 +167,7 @@ const validateAdvancedSettingsStep = (
   // The Next button is disabled only when no VolumeSnapshotClass is selected for MCG Standalone when automatic backup option enabled.
   const hasForcefulDeploymentButNoConfirmation =
     type === BackingStorageType.LOCAL_DEVICES &&
+    isNoProvisioner &&
```
SanjalKatiyar: plz check and let me know if you feel it still should be `&&` condition (not `||`):
```suggestion
(type === BackingStorageType.LOCAL_DEVICES ||
    isNoProvisioner) &&
```

Soundarya189r: image

image

PR #2812, opened 2026-05-18
Fix S3 format validation not gating DRPolicy creation

vbnrh:
## Summary

The Create button for third-party storage DRPolicies was enabled as long as S3 fields were non-empty, ignoring format validation errors (bucket name regex, URL format, profile name pattern) displayed in the UI. This allowed submitting invalid S3 details that corrupt the shared `ramen-hub-config` ConfigMap.

This fix lifts the S3 form's format validation state into the reducer so `validateDRPolicyInputs()` blocks submission when format errors exist.

Fixes: https://redhat.atlassian.net/browse/DFBUGS-4003


packages/mco/components/create-dr-policy/add-s3-bucket-details/s3-bucket-details-form.tsx
```diff
@@ -75,32 +88,64 @@ export const ClusterS3BucketDetailsForm: React.FC<
...
+  ]);
+
+  React.useEffect(() => {
+    return () => {
+      dispatch({
+        type: DRPolicyActionType.SET_S3_FORMAT_VALID,
+        payload: false,
+      });
+    };
+  }, [dispatch]);
```
SanjalKatiyar: why is this needed ??

packages/mco/components/create-dr-policy/utils/reducer.ts
```diff
@@ -210,6 +217,12 @@ export const drPolicyReducer = (
         selectedClustersHaveODF: action.payload,
       };
     }
+    case DRPolicyActionType.SET_S3_FORMAT_VALID: {
+      return {
+        ...state,
+        isS3FormatValid: action.payload,
+      };
+    }
```
SanjalKatiyar: do we really need new error state ??

entire `state` is accessible to the FC, we can on the go (dynamically, during render) just validate it, right ??

if concern is that regex is expensive (which is ideally not for this case), we can use `useMemo` too ??

SanjalKatiyar (changes requested):

packages/mco/components/create-dr-policy/utils/reducer.ts
vbnrh: Removed the new state to dynamically check the validation on each step

PR #2813, opened 2026-05-18
Make S3 form read-only when DRClusters already exist

vbnrh:
## Summary
When creating a new DRPolicy for third-party storage where both DRClusters already exist (e.g. offload policy), the S3 bucket detail fields should be read-only since those details are already configured.

- Disable all S3 bucket detail fields and hide "use same connection" checkbox when `areDRClustersAlreadyCreated` is true
- Skip S3 format validation for the Create button when both DRClusters exist
- Skip `prepareThirdPartyPeering()` on submit to avoid overwriting existing secrets/ConfigMap with empty credentials from pre-fill

Fixes: https://issues.redhat.com/browse/DFBUGS-4146

### DIFF
--- a/packages/mco/components/create-dr-policy/utils/k8s-utils.ts
+++ b/packages/mco/components/create-dr-policy/utils/k8s-utils.ts
@@ -1,5 +1,6 @@
 import {
   BackendType,
+  MAX_ALLOWED_CLUSTERS,
   ODFMCO_OPERATOR_NAMESPACE,
   RBD_IMAGE_FLATTEN_LABEL,
   ReplicationType,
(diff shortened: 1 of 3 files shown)


packages/mco/components/create-dr-policy/utils/k8s-utils.ts
```diff
@@ -145,24 +146,36 @@ export const createPolicyPromises = async (
...
-      secrets: [],
-      profiles: [],
-      drClusters: [],
-    };
-    try {
-      await prepareThirdPartyPeering(state, selectedDRClusters, created);
+    const allDRClustersExist =
+      selectedDRClusters?.length === MAX_ALLOWED_CLUSTERS;
+
+    if (allDRClustersExist) {
```
SanjalKatiyar, line 170: this block also needs try/catch ??

packages/mco/components/create-dr-policy/utils/k8s-utils.ts
SanjalKatiyar, line 139: same here ??

https://github.com/red-hat-storage/odf-console/pull/2813/changes#r3332237467

vbnrh: Moved inside try and catch.

vbnrh:
Added more granular checks for each cluster

PR #2814, opened 2026-05-19
Remove silenced alerts from HealthScore active alerts list

aruniiird:
BZ: https://redhat.atlassian.net/browse/DFBUGS-5419

### DIFF
--- a/packages/odf/components/HealthOverview/hooks.ts
+++ b/packages/odf/components/HealthOverview/hooks.ts
@@ -492,7 +492,37 @@ const mapIndefiniteSilencesToRows = (
 };
 
 /**
- * Checks if an alert matches a silence's matchers
+ * Checks if a value matches a silence matcher's criteria.
+ * Handles regex patterns, equality, and negation.
+ */
+const valueMatchesMatcher = (
+  value: string,
+  matcher: { value: string; isRegex: boolean; isEqual: boolean }
+): boolean => {
+  if (matcher.isRegex) {
+    try {
+      const regex = new RegExp(matcher.value);
+      const matches = regex.test(value);
+      return matcher.isEqual ? matches : !matches;
+    } catch {
+      return false;
+    }
+  }
+
+  const matches = value === matcher.value;
+  return matcher.isEqual ? matches : !matches;
+};
+
+/**
+ * Checks if an alert matches a silence's matchers.
+ *
+ * Key considerations:
+ * 1. 'alertstate' matchers are skipped since alertstate is a runtime state
+ *    indicator (firing/pending) that can vary over time.
+ * 2. Infrastructure labels (prometheus, job, etc.) added by the monitoring
+ *    stack may not be present in historical Prometheus alert data. We only
+ *    match on labels that the alert actually has.
+ * 3. The 'alertname' matcher must always be present and match.
  */
 const alertMatchesSilence = (
   alert: AlertRowData,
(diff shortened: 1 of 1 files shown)


aruniiird:
Here is the description:
The issue here is, when user try to 'Silence' a HealthScore alert through ODF console's Observe -> Alerting page, the silenced alert still shows up as active in the HealthScore's main page (in the alerts list). [Please see the screen [recording of the issue](https://url.corp.redhat.com/3a8bd4a) in the bug]
After the fix the issue is resolved and we can see once the alert is Silenced (either through the Console Observe -> Alerting page or through the HealthScore main page silencing mechanism), it will be correctly removed from the list.

Attaching a screen recording

https://github.com/user-attachments/assets/0da6fd75-4891-45a2-b394-8692bb520577

packages/odf/components/HealthOverview/hooks.ts
```diff
@@ -502,12 +510,54 @@ const alertMatchesSilence = (
...
+          const regex = new RegExp(alertnameMatcher.value);
+          const matches = regex.test(alert.alertname);
+          return alertnameMatcher.isEqual ? matches : !matches;
+        } catch {
+          return false;
+        }
+      })()
+    : alertnameMatcher.isEqual
+      ? alert.alertname === alertnameMatcher.value
+      : alert.alertname !== alertnameMatcher.value;
```
bipuladh: move this to a different function

packages/odf/components/HealthOverview/hooks.ts
```diff
@@ -502,12 +510,54 @@ const alertMatchesSilence = (
     return false;
   }
 
-  // Check if all matchers match the alert
-  return silence.matchers.every((matcher) => {
+  // Filter out 'alertstate' matchers - alertstate is a runtime state indicator
+  // that shouldn't be used for matching (same logic as getNormalizedAlertKey)
+  const nonAlertStateMatchers = silence.matchers.filter(
+    (matcher) => matcher.name !== 'alertstate'
+  );
```
bipuladh: if same logic as `getNormalizedAlertKey` why not use that here as well?

packages/odf/components/HealthOverview/hooks.ts
```diff
@@ -502,12 +510,54 @@ const alertMatchesSilence = (
     return false;
   }
 
-  // Check if all matchers match the alert
-  return silence.matchers.every((matcher) => {
+  // Filter out 'alertstate' matchers - alertstate is a runtime state indicator
+  // that shouldn't be used for matching (same logic as getNormalizedAlertKey)
+  const nonAlertStateMatchers = silence.matchers.filter(
```
bipuladh:
```suggestion
  const sanitizedMatchers = silence.matchers.filter(
```

packages/odf/components/HealthOverview/hooks.ts
aruniiird: Nope, they are used in different contexts,
- getNormalizedAlertKey works with Record<string, string> (alert labels) and produces a string key
- This code filters silence.matchers, which is an array of matcher objects with name, value, isRegex,
  isEqual properties

packages/odf/components/HealthOverview/hooks.ts
aruniiird: Done

packages/odf/components/HealthOverview/hooks.ts
aruniiird: Done

bipuladh (approved):

PR #2840, opened 2026-05-21
Enhance FileSystemDetails component to include navigation on cancel 

bipuladh:
- Added useNavigate hook and implemented click event handling to navigate

Fixes https://redhat.atlassian.net/browse/DFBUGS-5776

### DIFF
--- a/packages/odf/components/FileSystem/FileSystemDetails.tsx
+++ b/packages/odf/components/FileSystem/FileSystemDetails.tsx
@@ -1,14 +1,17 @@
 import * as React from 'react';
 import { SAN_STORAGE_SYSTEM_NAME } from '@odf/core/constants';
 import { FileSystemKind } from '@odf/core/types/scale';
-import { DetailsPage, FileSystemModel, getName, Kebab } from '@odf/shared';
+import {
+  DetailsPage,
+  FileSystemModel,
+  getName,
+  Kebab,
+  ResourceYAMLEditorWrapped,
+} from '@odf/shared';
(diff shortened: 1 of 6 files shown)


packages/odf/components/FileSystem/FileSystemDetails.tsx
```diff
@@ -35,6 +35,23 @@ const FileSystemDetails: React.FC<FileSystemDetailsProps> = (props) => {
...
+        event.stopPropagation();
+        navigate(-1);
+      }
+    };
+
+    document.addEventListener('click', handleClick, true);
+    return () => document.removeEventListener('click', handleClick, true);
+  }, [navigate]);
+
   return <ResourceYAMLEditor initialResource={obj} />;
```
SanjalKatiyar: Isn't this an issue which affects all our YAML pages ?? If so, we should create a temporary ResourceYAMLEditorWrapper and use that everywhere instead meanwhile.

bipuladh: In some places we are using it directly like here and in some places we are using it via the shared YAMLEditor component we have. That would cause a bit more of refactor. Let create an issue for this and use one YAMLEditor for us. Perhaps moving to PF component.

SanjalKatiyar: `YAMLEditor` also uses `ResourceYAMLEditor` ultimately (https://github.com/red-hat-storage/odf-console/blob/master/packages/shared/src/yaml-editor/YAMLEditor.tsx#L191)... so `ResourceYAMLEditorWrapper` would/should fix all, right ??

PR #2851, opened 2026-05-26
HealthOverview page should list only firing alerts

aruniiird:
In HealthOverview main page, we were listing alerts which were in pending state as well (along with other firing state alerts). This change make sure we list only firing alerts.

BZ: https://redhat.atlassian.net/browse/DFBUGS-6732 (failed QA)


aruniiird:
After the fix, the pending alert (shown below) is not coming in the HealthOverview alert list page

image

Error state alert (which is in firing state) is in the list, but the Warning state alert is not in the list

image

aruniiird:
@SanjalKatiyar , @bipuladh : please take a look

weirdwiz (approved):
looks good to me, but it will be helpful to list out the `ALERT{}` metric in the PR

PR #2889, opened 2026-06-08
Upgrade sdk-internal + Fix activity card

SanjalKatiyar:
https://redhat.atlassian.net/browse/DFBUGS-7074

#### Before:

Screenshot 2026-06-08 at 1 36 19 PM

#### After:

Screenshot 2026-06-08 at 1 35 58 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,13 +70,13 @@
     "@types/react-dom": "^17",
     "ajv-formats": "2.1.1",
     "ajv-keywords": "5.1.0",
+    "fast-xml-parser": "5.5.8",
     "fork-ts-checker-webpack-plugin/ajv": "8.18.0",
     "fork-ts-checker-webpack-plugin/schema-utils": "4.3.3",
     "immutable": "5.1.5",
     "qs": "6.14.1",
     "react-router": "6.30.3",
-    "schema-utils@npm:4.3.3/ajv": "8.18.0",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -85,7 +85,7 @@
     "@aws-sdk/lib-storage": "^3.1011.0",
     "@aws-sdk/s3-request-presigner": "^3.1011.0",
     "@openshift-console/dynamic-plugin-sdk": "4.21.0",
-    "@openshift-console/dynamic-plugin-sdk-internal": "4.21.0",
+    "@openshift-console/dynamic-plugin-sdk-internal": "4.22.0-prerelease.3",
```
bipuladh, line 88: [4.23.0-prerelease.2](https://www.npmjs.com/package/@openshift-console/dynamic-plugin-sdk/v/4.23.0-prerelease.2) should be used here for master

SanjalKatiyar: @bipuladh this is unrelated to the bug fix (we need to backport to `4.22` and `4.21-compatibility`). We are currently using `4.21.0` on master, which is even worse. We can merge this PR as-is and update master in a separate PR as part of `4.23` dev cycle.

SanjalKatiyar: As part of 4.23 we can update all the sdk versions (not just internal).

PR #2890, opened 2026-06-08
Show only cluster network for FDF host networking in create wizard

TimothyAsirJeyasing:
bz: https://redhat.atlassian.net/browse/DFBUGS-7002

Fusion does not support public network for this flow; only cluster network (CIDR) is exposed to users.


TimothyAsirJeyasing:
image
image

packages/odf/components/create-storage-system/reducer.ts
```diff
@@ -505,7 +505,12 @@ export const reducer: WizardReducer = (prevState, action) => {
       break;
     case 'securityAndNetwork/setNetworkType':
       newState.securityAndNetwork.networkType = action.payload;
-      if (action.payload !== NetworkType.NIC) {
+      if (action.payload === NetworkType.NIC) {
+        // FDF host networking: cluster network only; clear any public network state.
+        newState.securityAndNetwork.usePublicNetwork = false;
+        newState.securityAndNetwork.useClusterNetwork = true;
+        newState.securityAndNetwork.addressRanges.public = [];
```
SanjalKatiyar: plz clean up the entire reducer states, remove any keys which are not needed anymore instead of hackishly resetting the value like this...

TimothyAsirJeyasing: Sure will do it

PR #2899, opened 2026-06-11
Fix topology view showing "Unknown" DR status during relocate after failover

bipuladh:
Fixes https://redhat.atlassian.net/browse/DFBUGS-7001

## SUMMARY
 - Fixes DR status displaying "Unknown" instead of "Relocating" on the topology view when performing a relocate operation after a completed failover
 - Unifies DR status handling across both the topology view and the protected-applications list by adding an action parameter to getDRStatus(), ensuring both views show consistent status during transitional phases (Initiating/Deploying)
 - Maps transitional DRPC phases (Initiating/Deploying) to the correct action-specific status using spec.action

## RCA
When a relocate is initiated after a completed failover, the Ramen controller transitions the DRPC's status.phase through Initiating → Relocating. The useActiveDROperations hook correctly detects these as active operations (they're in ACTIVE_PHASES), but:

getEffectiveDRStatus (used by node rendering + sidebar) had no case for Phase.Initiating/Phase.Deploying in its switch — falling to default: return DRStatus.Unknown.
getDRStatus (used by protected-applications list and topology node labels) also doesn't explicitly handle these transitional phases, causing it to fall through to health-based evaluation instead of showing the in-progress operation status.
This resulted in the topology showing "Unknown" and the protected-applications list showing "Healthy" during the transitional window before the phase advances to Relocating.

## FIX
 - Added an optional action parameter to getEffectiveDRStatus and mapped Phase.Initiating/Phase.Deploying → FailingOver/Relocating based on spec.action
 - Added an optional action parameter to getDRStatus with the same transitional phase mapping, ensuring the protected-applications list also shows the correct in-progress status
 - Simplified computeOperationDRStatus in node-generator.ts — removed the manual effectivePhase mapping and instead passes action directly to getDRStatus
 - Updated useProtectedAppsByCluster.ts to pass drpc?.spec?.action to getDRStatus
 - Updated dr-status-popover.tsx to pass action through getDRStatusDetails to getDRStatus
 - Updated all topology callers to pass the action field
 - Added test cases for Initiating/Deploying + action=Relocate/Failover scenarios
 - When action is undefined (e.g., initial enrollment), existing behavior is preserved — no change to Protecting/health logic paths

### DIFF
--- a/packages/mco/components/dr-status-popover/dr-status-popover.spec.tsx
+++ b/packages/mco/components/dr-status-popover/dr-status-popover.spec.tsx
@@ -1,7 +1,7 @@
 /* eslint-disable import/no-extraneous-dependencies */
 import React from 'react';
 import { describe, expect, it } from '@jest/globals';
-import { VolumeReplicationHealth } from '@odf/mco/constants';
+import { DRActionType, VolumeReplicationHealth } from '@odf/mco/constants';
 import { DRPlacementControlConditionReason, Phase } from '@odf/mco/types';
 import { K8sResourceConditionStatus } from '@odf/shared/types';
 import { render, screen } from '@testing-library/react';
@@ -856,6 +856,42 @@ describe('DRStatusPopover Component', () => {
             status: K8sResourceConditionStatus.False,
             reason: DRPlacementControlConditionReason.Error,
           },
+        },
+      },
+      {
+        label: 'Initiating with action=Relocate -> Relocating',
+        expectedStatus: 'Relocating',
+        overrides: {
+          phase: Phase.Initiating,
+          progression: 'Initial',
+          action: DRActionType.RELOCATE,
+        },
+      },
+      {
+        label: 'Initiating with action=Failover -> FailingOver',
+        expectedStatus: 'FailingOver',
+        overrides: {
+          phase: Phase.Initiating,
+          progression: 'Initial',
+          action: DRActionType.FAILOVER,
+        },
+      },
+      {
+        label: 'Deploying with action=Relocate -> Relocating',
+        expectedStatus: 'Relocating',
+        overrides: {
+          phase: Phase.Deploying,
+          progression: 'Deploying',
+          action: DRActionType.RELOCATE,
+        },
+      },
+      {
+        label: 'Deploying with action=Failover -> FailingOver',
+        expectedStatus: 'FailingOver',
+        overrides: {
+          phase: Phase.Deploying,
+          progression: 'Deploying',
+          action: DRActionType.FAILOVER,
         },
       },
       {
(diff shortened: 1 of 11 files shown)


packages/mco/components/topology/factory/MCOStyleAppNode.tsx
```diff
@@ -74,7 +74,14 @@ const MCOStyleAppNodeComponent: React.FC<MCOStyleAppNodeProps> = ({
...
     data.isStatic && data.appStatus
       ? data.appStatus
-      : getEffectiveDRStatus(data?.phase, data?.progression);
+      : getEffectiveDRStatus(
+          data?.phase,
+          data?.progression,
+          undefined,
+          undefined,
+          undefined,
+          data?.action
```
vbnrh: Can we use optional object instead of using undefined positional args

packages/mco/components/topology/sidebar/OperationSidebar.tsx
```diff
@@ -23,7 +23,14 @@ import { DRPCFilterToolbar } from './DRPCFilterToolbar';
...
 
 const getOperationStatus = (op: DROperationInfo) =>
-  getEffectiveDRStatus(op.phase, op.progression, op.hasProtectionError);
+  getEffectiveDRStatus(
+    op.phase,
+    op.progression,
+    op.hasProtectionError,
+    undefined,
+    undefined,
+    op.action
```
vbnrh: Same use an object and pass relevant fields

packages/mco/utils/dr-status.ts
```diff
@@ -290,7 +300,8 @@ export const getEffectiveDRStatus = (
   progression?: string,
   hasProtectionError?: boolean,
   protectedCondition?: K8sResourceCondition,
-  volumeLastGroupSyncTime?: string
+  volumeLastGroupSyncTime?: string,
+  action?: DRActionType | string
```
vbnrh:
```suggestion
  action?: DRActionType
```

Can we use a singular type ? string encompasses everything and weakens the typing

packages/mco/components/dr-status-popover/dr-status-popover.spec.tsx
```diff
@@ -774,7 +774,7 @@ describe('DRStatusPopover Component', () => {
           volumeReplicationHealth: VolumeReplicationHealth.CRITICAL,
           protectedCondition: {
             status: K8sResourceConditionStatus.True,
-            reason: DRPlacementControlConditionReason.Success,
+            reason: 'Success',
```
vbnrh: Why are we hardcoding if its available in the enum

vbnrh (commented):
I don't like that we are maintaining two separate functions to maintain parity between protected applications page and topology view. 

Can we use a single function to get one singular status ?

vbnrh:
LGTM. One small thing: Progression.FailedToFailover and Progression.FailedToRelocate don't exist in ramen's API, so they're dead code. After this PR the only remaining usage is in decorator-helpers.ts. Worth removing those enum values and the decorator-helpers cases while you're in here.

PR #2925, opened 2026-06-16
Show DRPolicy name in DR Operations sidebar Policy column

vbnrh:
## Summary
- Fixes the Policy column in the DR Operations sidebar (topology view) which was displaying cluster names and failover direction instead of the actual DRPolicy name
- Uses `getPAVDRPolicyName` to extract the policy name from PAV, consistent with how AppSidebar already handles this

Fixes: https://redhat.atlassian.net/browse/DFBUGS-6795

### DIFF
--- a/packages/mco/components/topology/sidebar/OperationSidebar.tsx
+++ b/packages/mco/components/topology/sidebar/OperationSidebar.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { getDRStatus } from '@odf/mco/utils/dr-status';
+import { DASH } from '@odf/shared/constants';
 import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
 import { useNavigate } from 'react-router-dom-v5-compat';
 import {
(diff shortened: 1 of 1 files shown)


packages/mco/components/topology/sidebar/OperationSidebar.tsx
```diff
@@ -167,20 +164,7 @@ const OperationsTableView: React.FC<{
...
-                            style={{
-                              fontSize: '0.875rem',
-                              color: 'var(--pf-v6-global--Color--200)',
-                            }}
-                          >
-                            {operation.sourceCluster} →{' '}
-                            {operation.targetCluster}
-                          </span>
-                        </div>
+                        {operation.pav ? getPAVDRPolicyName(operation.pav) : ''}
```
SanjalKatiyar: `DASH` is in /constants...
```suggestion
                        {operation.pav ? getPAVDRPolicyName(operation.pav) : DASH}
```

PR #2934, opened 2026-06-19
Fix topology sidebar showing Unknown DR status by computing full health data

bipuladh:
Fixes https://redhat.atlassian.net/browse/DFBUGS-7001`

The AppSidebar and OperationSidebar were calling getDRStatus with only phase/progression/action, missing volumeReplicationHealth and schedulingInterval. When status evaluation reached the health check with an empty replicationHealths array, it returned Unknown instead of the correct status.

Extracted a shared getOperationDRStatus helper in sidebar-utils that looks up the DR policy from context to derive the full health data, matching the computation in node-generator's computeOperationDRStatus.

### DIFF
--- a/packages/mco/components/topology/utils/sidebar-utils.tsx
+++ b/packages/mco/components/topology/utils/sidebar-utils.tsx
@@ -9,10 +9,89 @@ import { referenceForModel } from '@odf/shared/utils';
 import { InProgressIcon } from '@patternfly/react-icons';
 import { t_color_blue_40 as blueInfoColor } from '@patternfly/react-tokens';
 import { NodeStatus } from '@patternfly/react-topology';
+import { ClusterPairPoliciesMap } from '../../../hooks/useDRPoliciesByClusterPair';
+import { DRPolicyKind } from '../../../types';
+import {
+  getReplicationHealth,
+  getReplicationType,
+  getProtectedCondition,
+} from '../../../utils';
 import {
   DRStatus,
+  getDRStatus,
+  isCleanupRequired,
   isUserActionRequired as isUserActionRequiredUtil,
 } from '../../../utils/dr-status';
+import { DROperationInfo } from '../types';
+
+/**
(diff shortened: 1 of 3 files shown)


packages/mco/components/topology/utils/sidebar-utils.tsx
```diff
@@ -9,10 +9,89 @@ import { referenceForModel } from '@odf/shared/utils';
...
+  drPolicyByName: Map<string, DRPolicyKind>
+): DRStatus => {
+  const drpc = op.drpc;
+  const drPolicyName = drpc?.spec?.drPolicyRef?.name;
+  const drPolicy = drPolicyName ? drPolicyByName.get(drPolicyName) : undefined;
+  const schedulingInterval = drPolicy?.spec?.schedulingInterval;
+  const protectedCondition = getProtectedCondition(drpc);
+  const volumeLastGroupSyncTime = drpc?.status?.lastGroupSyncTime;
+
+  const volumeReplicationHealth =
```
TimothyAsirJeyasing, line 60: Should we call getReplicationHealth here even when volumeLastGroupSyncTime is missing, same as in useProtectedAppsByCluster and node-generator? Just want to make sure sidebar status matches the topology nodes.

packages/mco/components/topology/utils/sidebar-utils.tsx
```diff
@@ -9,10 +9,89 @@ import { referenceForModel } from '@odf/shared/utils';
...
+    });
+  });
+  return map;
+};
+
+/**
+ * Computes the full DR status for an operation using all available data.
+ * Mirrors the computation in node-generator's computeOperationDRStatus.
+ */
+export const getOperationDRStatus = (
```
TimothyAsirJeyasing, line 49: This looks very similar to computeOperationDRStatus in node-generator.ts. Would it be worth extracting a shared helper and reusing it in both places, so they don’t drift apart again?

TimothyAsirJeyasing (changes requested):

PR #2935, opened 2026-06-19
DFBUGS-7396: show loading spinner instead of 'Cluster not found' flash

ttrishag:
Jira : https://redhat.atlassian.net/browse/DFBUGS-7396

Screen recording after the fix:

https://github.com/user-attachments/assets/26abd879-9ec0-4ca0-8db7-6ad1a3b8e4db


SanjalKatiyar:
/retitle [DFBUGS-7396](https://redhat.atlassian.net/browse/DFBUGS-7396): show loading spinner instead of 'Cluster not found' flash

PR #2938, opened 2026-06-23
RHSTOR-9145: Adds support for validating KMM operator installation

bipuladh:
Checks KMM is Ready before enabling SAN option

### Before KMM is installed
Screenshot 2026-06-23 at 9 31 07 AM

### After KMM is installed
Screenshot 2026-06-23 at 9 30 27 AM


locales/en/plugin__odf-console.json
```diff
@@ -2253,6 +2253,7 @@
...
   "The graph displays the total number of block volumes inbound snapshots, by cluster, from all ApplicationSet and Subscription type applications. Applications that use file volumes are excluded in the total snapshot count.": "The graph displays the total number of block volumes inbound snapshots, by cluster, from all ApplicationSet and Subscription type applications. Applications that use file volumes are excluded in the total snapshot count.",
   "The IP address is already registered": "The IP address is already registered",
+  "The Kernel Module Management(KMM) operator is required to connec to a Storage Area Network. Install the operator from ": "The Kernel Module Management(KMM) operator is required to connec to a Storage Area Network. Install the operator from ",
```
SanjalKatiyar:
```suggestion
  "The Kernel Module Management (KMM) operator is required to connect to a Storage Area Network (SAN). Install the operator from ": "The Kernel Module Management (KMM) operator is required to connect to a Storage Area Network (SAN). Install the operator from ",
```

packages/odf/modals/ConfigureDF/ExternalSystemsModal.tsx
```diff
@@ -58,6 +72,21 @@ const ConfigureExternalSystems: React.FC<ConfigureDFSelectionsProps> = ({
...
+      isList: true,
+    });
+
+  const kmmOperator = kmmImageRegistry?.find((csv) =>
+    getName(csv).includes(KMM_OPERATOR_NAME)
+  );
+  const isKMMInstalled =
+    kmmOperator?.status?.phase === 'Succeeded' &&
+    kmmImageRegistryLoaded &&
+    !kmmImageRegistryError;
```
SanjalKatiyar: nits: we have useFetchCsv hook... can we use that ??

packages/odf/modals/ConfigureDF/ExternalSystemsModal.tsx
```diff
@@ -248,6 +277,21 @@ const ConfigureExternalSystems: React.FC<ConfigureDFSelectionsProps> = ({
...
+            <FlexItem>
+              <Content component="small">
+                <Icon className="pf-v6-u-mr-xs">
+                  <InfoCircleIcon color={t_color_blue_40.value} />
+                </Icon>
+                {t(
+                  'The Kernel Module Management(KMM) operator is required to connec to a Storage Area Network. Install the operator from '
+                )}
+                <Link to="/catalog" reloadDocument>
+                  Software Catalog
```
SanjalKatiyar, line 278: not sure, do we need to translate this ??

SanjalKatiyar (changes requested):

packages/odf/modals/ConfigureDF/ExternalSystemsModal.tsx
```diff
@@ -248,6 +277,21 @@ const ConfigureExternalSystems: React.FC<ConfigureDFSelectionsProps> = ({
...
+          {!isKMMInstalled && (
+            <FlexItem>
+              <Content component="small">
+                <Icon className="pf-v6-u-mr-xs">
+                  <InfoCircleIcon color={t_color_blue_40.value} />
+                </Icon>
+                {t(
+                  'The Kernel Module Management(KMM) operator is required to connec to a Storage Area Network. Install the operator from '
+                )}
+                <Link to="/catalog" reloadDocument>
```
SanjalKatiyar, line 277: is `reloadDocument` really needed ??

packages/odf/modals/ConfigureDF/ExternalSystemsModal.tsx
bipuladh: software catalog not really

packages/odf/modals/ConfigureDF/ExternalSystemsModal.tsx
bipuladh: Yes without it we need to do complex state handling for closing the modal. Life is easier this way.

packages/odf/modals/ConfigureDF/ExternalSystemsModal.tsx
bipuladh: We cannot because the name of the csv has the version appended to it hence we are use includes here. Pretty much it's the same so I don't think we need to use `useFetchCSV` hook here.

SanjalKatiyar: Name of the CSV always has a version attached to it (even for other operators), IIRC `useFetchCSV` address that by matching Subscription's spec.
It would make sense to use a common hook, since we have one already IMHO. Instead of repeating the same logic.

PR #2942, opened 2026-06-23
Refactor configure performance to Configure performance profile page

NIKHITHAVADDEMPUDI:
**Epic link:** https://redhat.atlassian.net/browse/RHSTOR-8629

**story link:** https://redhat.atlassian.net/browse/RHSTOR-9082

**Figma link:** https://www.figma.com/design/zIHZR2BEoe54aND03T7PZJ/Support-Performance-profiles-for-MCG?node-id=12168-1643&t=NssvOIyXYVjlXgHI-1

**Aim:** Previously performance profiles for Ceph is used via a modal, now move it to configure performance profile page to keep it consistent with the design of the performance profiles for mcg.

**Will add performance profiles for mcg to the page in sprint-2**

**Changes:** 

1. **Added a new Configure performance profile full page at** /odf/system/ns/:namespace/ocs.openshift.io~v1~storagecluster/:resourceName/~configureperformanceprofile, registered in console-extensions.json and console-plugin.json.

2. **Introduced configure-performance-profile/ module with:**
**configure-performance-profile.tsx** — page shell with PageHeading, shared form state (useReducer), section visibility, and Save/Cancel orchestration via Promise.all.
**core-storage-section.tsx** — Day-2 Ceph (core storage) profile UI; reuses existing ConfigurePerformance dropdown and validation; patches StorageCluster.spec.resourceProfile on save; shows aggregated requirements and cluster-available CPU/RAM.
**object-access-section.tsx** — scaffold for MCG Object Access (UI shell + empty submitObjectAccessProfile() stub; hidden until sprint-2).
**state.ts, utils.ts, configure-performance-profile-footer.tsx, and configure-performance-profile.scss** — form state, visibility/save-disabled helpers, footer, and layout styling.

3.**Updated entry points to navigate to the page instead of opening a modal:**

 Storage cluster kebab menu in StorageClusterSection.tsx
CSV actions in csv-actions.ts

4.**Removed configure-performance-modal.tsx and its SCSS (modal flow fully replaced).** (code in core-storage-section is kept as similar as possible to configure-performance-modal.tsx)

5. **Added shared visibility helpers** (isCoreStorageSectionVisible, isObjectAccessSectionVisible, isConfigurePerformanceProfileVisible) so the action appears when either core storage or MCG object access applies (not blocked solely by provider mode when NooBaa is available).

6. **Extended Day-1 configure-performance.tsx** with optional showDescription={false} so the page heading does not duplicate the intro text on the new page.

**Out of scope (sprint-2): MCG performance profile selection, patch logic, and enabling the Object Access section (showObjectAccess is currently gated off).**

**AFTER CHANGES**

https://github.com/user-attachments/assets/9312989d-5473-47e9-ae7a-5dcccf9aef69


packages/odf/components/configure-performance-profiles/core-storage-section.tsx
```diff
@@ -0,0 +1,282 @@
...
+};
+
+export const submitCoreStorageProfile = async ({
+  storageCluster,
+  resourceProfile,
+  validation,
+}: SubmitCoreStorageProfileParams): Promise<void> => {
+  if (isCoreStorageSaveDisabled(resourceProfile, validation)) {
+    return;
+  }
```
Soundarya189r, line 35: I see  `await labelNodes(selectedNodes, storageClusterNs);` is called before patching resourceProfile in the old modal.

packages/odf/components/actions/csv-actions.ts
```diff
@@ -133,18 +146,17 @@ const AddCapacityStorageSystem = (
...
     id: 'configure-performance-storage-system',
     label: 'Configure performance',
     insertAfter: 'add-capacity-storage-system',
-    cta: () => {
-      launchModal(ConfigurePerformanceModal, {
-        extraProps: { storageCluster },
-        isOpen: true,
-      });
+    cta: {
+      href: `/odf/system/ns/${getNamespace(storageCluster)}/${referenceForModel(
```
Soundarya189r: Can we move this to a constant/util function ?

packages/odf/components/Sections/StorageClusterSection.tsx
```diff
@@ -80,6 +80,22 @@ const storageClusterActions =
...
+        storageCluster,
+        hasExternalMode: isExternalMode,
+        isProviderMode,
+        isNoobaaAvailable,
+      })
+    ) {
+      customKebabItems.push({
+        key: 'CONFIGURE_PERFORMANCE',
+        value: t('Configure performance'),
+        redirect: `/odf/system/ns/${getNamespace(storageCluster)}/${referenceForModel(
```
Soundarya189r: Can we move this to a constant/util function ?

Soundarya189r:
@NIKHITHAVADDEMPUDI  Previously we had the option to pick nodes via `SelectNodesTable`. Is the explicit node selection is no longer needed?

packages/odf/components/configure-performance-profile/configure-performance-profile.scss
```diff
@@ -0,0 +1,23 @@
+.configure-performance-profile {
+  display: flex;
+  flex-direction: column;
+  height: var(--configure-performance-profile-height);
```
Soundarya189r: I don't see this variable defined anywhere `--configure-performance-profile-height`

packages/odf/components/configure-performance-profile/core-storage-section.tsx
```diff
@@ -0,0 +1,282 @@
...
+  clusterCpu?: number;
+  clusterMemoryGiB?: number;
+};
+
+const ClusterResourcesContext = React.createContext<{
+  clusterCpu: number;
+  clusterMemoryGiB: number;
+} | null>(null);
+
+const ProfileRequirementsText: React.FC<ProfileRequirementsTextProps> = ({
```
Soundarya189r: I see that `configure-performance.tsx `already exports `ProfileRequirementsText` is it possible to reuse them with optional props ?

NIKHITHAVADDEMPUDI:
> @NIKHITHAVADDEMPUDI Previously we had the option to pick nodes via `SelectNodesTable`. Is the explicit node selection is no longer needed?

Screenshot 2026-06-19 at 10 00 06 PM

Hi @Soundarya189r yes it is no longer needed, it is removed from the design based on the comments of Eran - "_**When a customer added nodes for ODF, it will be automatically be usable for any of our pods. Having the nodes in that context would overload the experience and can even confuse the customer IMHO.**_ "

SanjalKatiyar:
Haven't reviewed the PR, but from the attached recording:

1. IIRC, the existing day-2 UX automatically disables profile options in the dropdown that aren't supported based on Cpu/Mem (plz correct me if I am mistaken). If that's the current behaviour, why r we changing that ??
Why are we now allowing users to select an unsupported profile, only to disable the `Next` button later and show an error afterward ??
2. `object-access` is **not** the name which is suggested by the PM on the mockups. Plz check the discussion threads and use the appropriate name for the files and text on the UI.

NIKHITHAVADDEMPUDI:
> Haven't reviewed the PR, but from the attached recording:
> 
> 1. IIRC, the existing day-2 UX automatically disables profile options in the dropdown that aren't supported based on Cpu/Mem (plz correct me if I am mistaken). If that's the current behaviour, why r we changing that ??
>    Why are we now allowing users to select an unsupported profile, only to disable the `Next` button later and show an error afterward ??
> 2. `object-access` is **not** the name which is suggested by the PM on the mockups. Plz check the discussion threads and use the appropriate name for the files and text on the UI.

Hi @SanjalKatiyar 

1. Day-2 reuses the same ConfigurePerformance dropdown as the install wizard. Options are only disabled when forceLean applies (cluster can’t support Balanced) — not per-profile based on CPU/mem.
If a user picks an unsupported profile, the wizard already shows a dropdown error, a validation message, and blocks Next. Day-2 does the same with Save. We’re not changing that behavior — just aligning day-2 with the existing wizard pattern.

2. Sorry i missed Eran's comment to rename "_Object Access should be renamed to Multicloud Object Gateway performance profiles._", I will ask Brian to suggest the tile, whether it is Multicloud Object Gateway alone or Multicloud Object Gateway performance profiles since performance profile is already in the title. and I will change file names and strings accordingly based on the reply

packages/odf/components/configure-performance-profiles/core-storage-section.tsx
NIKHITHAVADDEMPUDI: Thank you for the review @Soundarya189r , i added labelNodes which labels all unlabelled nodes in the cluster

packages/odf/components/actions/csv-actions.ts
NIKHITHAVADDEMPUDI: Hi @Soundarya189r, could you please provide more context, like which part to move to constants, i tried to keep similar to attach storage system page https://github.com/red-hat-storage/odf-console/blob/cc265d5211c6c190cf4527fdb5bf06e0f9e1d21f/packages/odf/components/actions/csv-actions.ts#L109-L113

packages/odf/components/Sections/StorageClusterSection.tsx
NIKHITHAVADDEMPUDI: https://github.com/red-hat-storage/odf-console/pull/2942/changes#r3480640032

packages/odf/components/configure-performance-profile/configure-performance-profile.scss
NIKHITHAVADDEMPUDI: removed this variable in the latest commit

packages/odf/components/configure-performance-profile/core-storage-section.tsx
NIKHITHAVADDEMPUDI: i reused that component with some modification in the latest commit

NIKHITHAVADDEMPUDI:
> > Haven't reviewed the PR, but from the attached recording:
> > 
> > 1. IIRC, the existing day-2 UX automatically disables profile options in the dropdown that aren't supported based on Cpu/Mem (plz correct me if I am mistaken). If that's the current behaviour, why r we changing that ??
> >    Why are we now allowing users to select an unsupported profile, only to disable the `Next` button later and show an error afterward ??
> > 2. `object-access` is **not** the name which is suggested by the PM on the mockups. Plz check the discussion threads and use the appropriate name for the files and text on the UI.
> 
> Hi @SanjalKatiyar
> 
> 1. Day-2 reuses the same ConfigurePerformance dropdown as the install wizard. Options are only disabled when forceLean applies (cluster can’t support Balanced) — not per-profile based on CPU/mem.
>    If a user picks an unsupported profile, the wizard already shows a dropdown error, a validation message, and blocks Next. Day-2 does the same with Save. We’re not changing that behavior — just aligning day-2 with the existing wizard pattern.
> 2. Sorry i missed Eran's comment to rename "_Object Access should be renamed to Multicloud Object Gateway performance profiles._", I will ask Brian to suggest the tile, whether it is Multicloud Object Gateway alone or Multicloud Object Gateway performance profiles since performance profile is already in the title. and I will change file names and strings accordingly based on the reply

1. reusing component was giving the current behaviour, in UX designs it suggests to disable , @SanjalKatiyar please suggest which behaviour to be kept.
2. Now I updated the filenames and titles as per the designs https://www.figma.com/design/zIHZR2BEoe54aND03T7PZJ/Support-Performance-profiles-for-MCG?node-id=12264-2939&t=FqV2cx4Cay2gV27A-4

packages/odf/components/actions/csv-actions.ts
Soundarya189r: Hi @NIKHITHAVADDEMPUDI  okay, so we have functions like `getBucketOverviewBaseRoute`, I was referring to that.

https://github.com/red-hat-storage/odf-console/blob/cc265d5211c6c190cf4527fdb5bf06e0f9e1d21f/packages/odf/constants/s3-browser.ts#L37

packages/odf/components/Sections/StorageClusterSection.tsx
NIKHITHAVADDEMPUDI: i moved it to constants in the latest commit

packages/odf/components/actions/csv-actions.ts
NIKHITHAVADDEMPUDI: i moved it to constants in the latest commit

NIKHITHAVADDEMPUDI:
https://github.com/user-attachments/assets/07999686-4e7d-4553-93b2-2bc69ec470d3

packages/odf/constants/storage-system.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const getStorageClusterBaseRoute = (
+  namespace: string,
+  resourceName: string
+): string =>
+  `/odf/system/ns/${namespace}/${referenceForModel(
+    StorageClusterModel
+  )}/${resourceName}`;
+
+export const getAttachStorageRoute = (
```
bipuladh, line 16: Do these need to be exported?

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/configure-performance.tsx
```diff
@@ -72,18 +72,70 @@ type ProfileRequirementsTextProps = {
...
+}) => {
   const { t } = useCustomTranslation();
   const { minCpu, minMem } = getResourceProfileRequirements(
     selectedProfile,
     osdAmount,
     architecture,
     enableNFS
   );
+
+  if (variant === 'inline') {
```
bipuladh, line 113: Why do we have multiple variants?

packages/odf/components/configure-performance-profiles/core-storage-section.tsx
```diff
@@ -0,0 +1,204 @@
...
+        {t('Core storage')}
+      </Content>
+      <Content
+        component={ContentVariants.small}
+        id="core-storage-desc"
+        className="pf-v6-u-mb-xl"
+      >
+        <Trans
+          t={t}
+          ns="plugin__odf-console"
```
bipuladh: Whys is ns required?

packages/odf/components/configure-performance-profiles/core-storage-section.tsx
```diff
@@ -0,0 +1,204 @@
...
+          i18nKey="Optimize CPU and memory allocation for Block, File, and RADOS Gateway services and storage performance. Available nodes can be added to the storage cluster to add additional resources. Learn more about <0>scaling storage</0> for more cluster resources."
```
bipuladh: I am confused about what `i18nKey` is and why are we using it?

packages/odf/components/configure-performance-profiles/core-storage-section.tsx
```diff
@@ -0,0 +1,204 @@
...
+  clusterNodes,
+}) => {
+  const { t } = useCustomTranslation();
+  const { resourceProfile, validation } = state;
+
+  const osdAmount = storageCluster?.spec?.storageDeviceSets
+    ?.map((deviceSet: DeviceSet) =>
+      getOsdAmount(deviceSet.count, deviceSet.replica)
+    )
+    .reduce((accumulator: number, current: number) => accumulator + current, 0);
```
bipuladh: Do we really need map and reduce? Reduce by itself should help you get everything done right?

SanjalKatiyar:
as per recent/latest confirmation, need to add new UX for core...
that is, disabling the unsupported option in the dropdown (rather than allowing user to select and then disable "Save" and showing error message)...
plz make sure we display "aggregate resource requirement" and "cluster resources available" info and each dropdown item should have description displaying info about cpu/mem "required" for each profile...

packages/odf/components/configure-performance-profiles/core-storage-section.tsx
NIKHITHAVADDEMPUDI: i removed the redundant map in the latest commit

packages/odf/components/configure-performance-profiles/core-storage-section.tsx
NIKHITHAVADDEMPUDI: removed the redundant keys ns and i18nkey in the latest commit

packages/odf/components/configure-performance-profiles/core-storage-section.tsx
NIKHITHAVADDEMPUDI: removed the redundant keys ns and i18nkey in the latest commit

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/configure-performance.tsx
NIKHITHAVADDEMPUDI: we have multiple variants because ConfigurePerformance is shared between two UIs that need different requirements text below the dropdown.

**wizard (default)** — install wizard (day-1):

Heading: “Aggregate resource requirements for {profile} mode”
CPU and memory on separate lines with smaller labels
**inline** — configure performance profile page (day-2), set via profileRequirementsVariant="inline" in core-storage-section.tsx:

“Aggregated resource requirements: X CPUs and Y GiB RAM” on one line
“Cluster resources available: …” on a second line (Figma layout)
The dropdown and profile logic are the same; only the requirements display changes. Without the variant, day-2 would either reuse the wizard layout or duplicate the whole component.

packages/odf/constants/storage-system.ts
NIKHITHAVADDEMPUDI: Yes — both need to be exported because they’re imported by multiple files outside this file: getStorageClusterBaseRoute is used in attach-storage.tsx and configure-performance-profile.tsx (and internally by the other route helpers), and getAttachStorageRoute is used in StorageClusterSection.tsx and csv-actions.ts.

i made the changes with reference from this comment https://github.com/red-hat-storage/odf-console/pull/2942#discussion_r3490705308

NIKHITHAVADDEMPUDI:
in core section, we prefer to  consider only ocs labelled nodes
for mcg section , we prefer to consider all of the nodes irrespective of the labels

packages/odf/components/configure-performance-profiles/core-storage-section.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const clusterCpu = getTotalCpu(ocsLabeledNodes);
+  const clusterMemoryGiB = getTotalMemoryInGiB(ocsLabeledNodes);
+
+  const onProfileChange = React.useCallback(
+    (newProfile: ResourceProfile): void => {
+      dispatch({
+        type: ConfigurePerformanceProfileActionType.SET_RESOURCE_PROFILE,
+        payload: newProfile,
+      });
+    },
```
bipuladh:
```suggestion
    (newProfile: ResourceProfile): void =>
      dispatch({
        type: ConfigurePerformanceProfileActionType.SET_RESOURCE_PROFILE,
        payload: newProfile,
      });
```

packages/odf/components/configure-performance-profiles/core-storage-section.tsx
```diff
@@ -0,0 +1,131 @@
...
+  );
+
+  React.useEffect(() => {
+    if (storageCluster && !resourceProfile) {
+      dispatch({
+        type: ConfigurePerformanceProfileActionType.SET_RESOURCE_PROFILE,
+        payload: storageCluster.spec?.resourceProfile,
+      });
+    }
+  }, [storageCluster, resourceProfile, dispatch]);
```
bipuladh: Can't we do this initially rather than having to rely on useEffect?

packages/odf/components/configure-performance-profiles/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+  storageCluster,
+  hasExternalMode,
+  isProviderMode,
+}: ConfigurePerformanceProfileVisibility): boolean =>
+  !!storageCluster && !hasExternalMode && !isProviderMode;
+
+export const isMcgPerformanceSectionVisible = ({
+  isNoobaaAvailable,
+}: Pick<ConfigurePerformanceProfileVisibility, 'isNoobaaAvailable'>): boolean =>
+  isNoobaaAvailable;
```
bipuladh: Are we using this function just for typecasting? Is it really required?

packages/odf/components/configure-performance-profiles/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+  visibility: ConfigurePerformanceProfileVisibility
+): boolean =>
+  isCoreStorageSectionVisible(visibility) ||
+  isMcgPerformanceSectionVisible(visibility);
+
+export const isCoreStorageSaveDisabled = (
+  resourceProfile: ResourceProfile | null
+): boolean => !resourceProfile;
+
+export const isMcgPerformanceSaveDisabled = (): boolean => true;
```
bipuladh:
```suggestion
// Todo: Add logic later
export const isMcgPerformanceSaveDisabled = (): boolean => true;
```

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
```diff
@@ -0,0 +1,209 @@
...
+  isMcgPerformanceSaveDisabled,
+  isMcgPerformanceSectionVisible,
+} from './utils';
+import './configure-performance-profile.scss';
+
+const ConfigurePerformanceProfile: React.FC = () => {
+  const { t } = useTranslation();
+  const { resourceName, namespace } = useParams();
+  const navigate = useNavigate();
+  const [state, dispatch] = React.useReducer(
```
bipuladh, line 48: We can initialize the state by passing a function that can wrap a storage cluster resource. Instead of using `useEffect` at https://github.com/red-hat-storage/odf-console/pull/2942/changes#r3570228332

packages/odf/components/configure-performance-profiles/utils.ts
NIKHITHAVADDEMPUDI: It isn’t for typecasting — it mirrors **isCoreStorageSectionVisible** so MCG visibility stays in one shared helper used by both the page and **isConfigurePerformanceProfileVisible**, rather than inlining isNoobaaAvailable at each call site.

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
NIKHITHAVADDEMPUDI: **Included this suggestion in the latest commit** 
**changes:**
Moved profile initialization out of the useEffect in core-storage-section.tsx by adding initProfileStates(storageCluster) in state.ts to seed resourceProfile from storageCluster.spec.resourceProfile, and wired lazy useReducer init (useReducer(configurePerformanceProfileReducer, storageCluster, initProfileStates)) in a ConfigurePerformanceProfileForm child that mounts only after the StorageCluster is loaded, so form state is set from the CR on first render without a sync effect (with the same pattern ready for MCG profile fields later).

packages/odf/components/configure-performance-profiles/core-storage-section.tsx
NIKHITHAVADDEMPUDI: https://github.com/red-hat-storage/odf-console/pull/2942#discussion_r3571156767

packages/odf/components/configure-performance-profiles/core-storage-section.tsx
NIKHITHAVADDEMPUDI: included this change in the latest commit

packages/odf/components/configure-performance-profiles/utils.ts
NIKHITHAVADDEMPUDI: included this change in the latest commit

locales/en/plugin__odf-console.json
```diff
@@ -1666,6 +1670,8 @@
   "Operators health": "Operators health",
   "Opt in for automatic backup for MultiCloud Object Gateway metadata database": "Opt in for automatic backup for MultiCloud Object Gateway metadata database",
   "Opt-in to automatically add additional raw capacity equivalent to the configured deployment size whenever used capacity reaches 70%. This ensures your deployment scales seamlessly to meet demand.": "Opt-in to automatically add additional raw capacity equivalent to the configured deployment size whenever used capacity reaches 70%. This ensures your deployment scales seamlessly to meet demand.",
+  "Optimize CPU and memory allocation for Block, File, and RADOS Gateway services and storage performance. Available nodes can be added to the storage cluster to add additional resources. Learn more about <2>scaling storage</2> for more cluster resources.": "Optimize CPU and memory allocation for Block, File, and RADOS Gateway services and storage performance. Available nodes can be added to the storage cluster to add additional resources. Learn more about <2>scaling storage</2> for more cluster resources.",
```
SanjalKatiyar, line 1707: `Optimize CPU and memory allocation for Block, File, and RADOS Gateway services and storage performance.`

this should also be added to the day-1 deployment flow... as a description to the perf profile section...

packages/odf/components/actions/csv-actions.ts
```diff
@@ -69,10 +78,16 @@ export const useCsvActions = ({
...
-        items.push(
-          ConfigurePerformanceStorageSystem(launchModal, storageCluster)
-        );
+      if (
+        isConfigurePerformanceProfileVisible({
+          storageCluster,
+          hasExternalMode,
+          isProviderMode,
+          isNoobaaAvailable:
+            !!systemFlags[resource.spec.namespace]?.isNoobaaAvailable,
```
SanjalKatiyar: why casting when isNoobaaAvailable already a boolean ??
```suggestion
            systemFlags[resource.spec.namespace]?.isNoobaaAvailable,
```

packages/odf/components/actions/csv-actions.ts
```diff
@@ -69,10 +78,16 @@ export const useCsvActions = ({
       isOCSStorageSystem(resource)
     ) {
       items.push(AddCapacityStorageSystem(launchModal, storageCluster));
-      if (!isProviderMode) {
-        items.push(
-          ConfigurePerformanceStorageSystem(launchModal, storageCluster)
-        );
+      if (
+        isConfigurePerformanceProfileVisible({
```
SanjalKatiyar:
```suggestion
        isConfigurePerformanceProfileAllowed({
```
Or,
```suggestion
        shouldConfigurePerformanceProfile({
```

packages/odf/components/actions/csv-actions.ts
```diff
@@ -48,6 +54,9 @@ export const useCsvActions = ({
   );
   const launchModal = useModalWrapper();
   const isProviderMode = useFlag(PROVIDER_MODE);
+  const { systemFlags } = useODFSystemFlagsSelector();
+  const externalClusterDetails = useGetExternalClusterDetails();
+  const hasExternalMode = externalClusterDetails.clusterName !== '';
```
SanjalKatiyar: u don't have to overuse hooks where it's not required...
```suggestion
```

packages/odf/components/actions/csv-actions.ts
```diff
@@ -93,6 +108,8 @@ export const useCsvActions = ({
     launchModal,
     infrastructure,
     isProviderMode,
+    hasExternalMode,
+    systemFlags,
```
SanjalKatiyar:
```suggestion
    systemFlags,
```

packages/odf/components/actions/csv-actions.ts
```diff
@@ -69,10 +78,16 @@ export const useCsvActions = ({
...
     ) {
       items.push(AddCapacityStorageSystem(launchModal, storageCluster));
-      if (!isProviderMode) {
-        items.push(
-          ConfigurePerformanceStorageSystem(launchModal, storageCluster)
-        );
+      if (
+        isConfigurePerformanceProfileVisible({
+          storageCluster,
+          hasExternalMode,
```
SanjalKatiyar:
```suggestion
          systemFlags[resource.spec.namespace]?.isExternalMode,
```

packages/odf/components/actions/csv-actions.ts
SanjalKatiyar: also using `useGetExternalClusterDetails` here is wrong...

SanjalKatiyar:
/lgtm cancel
need some changes (will add more details in a while)

packages/odf/components/configure-performance-profiles/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export const getOcsLabeledWizardNodes = (
+  nodes: WizardNodeState[],
+  namespace: string
+): WizardNodeState[] =>
+  nodes.filter((node) => isOcsLabeledWizardNode(node, namespace));
+
+export type ConfigurePerformanceProfileVisibility = {
+  storageCluster?: StorageClusterKind;
+  hasExternalMode: boolean;
```
SanjalKatiyar: why u need to know "hasExternalMode", instead of whether a selected cluster "isExternalMode" ??

You can have internal and external together, "has" just tells entire cluster state, we need to know single system state...
```suggestion
  isExternalMode: boolean;
```

packages/odf/components/configure-performance-profiles/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+  isProviderMode: boolean;
+  isNoobaaAvailable: boolean;
+};
+
+export const isCoreStorageSectionVisible = ({
```
SanjalKatiyar:
```suggestion
export const shouldShowCoreStorageSection = ({
```

packages/odf/components/configure-performance-profiles/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+  storageCluster,
+  hasExternalMode,
+  isProviderMode,
+}: ConfigurePerformanceProfileVisibility): boolean =>
+  !!storageCluster && !hasExternalMode && !isProviderMode;
+
+export const isMcgPerformanceSectionVisible = ({
```
SanjalKatiyar:
```suggestion
export const shouldShowMcgPerformanceSection = ({
```

packages/odf/components/actions/csv-actions.ts
```diff
@@ -107,9 +124,7 @@ const AttachStorageStorageSystem = (resource: StorageSystemKind): Action => {
     label: 'Attach Storage',
     insertBefore: 'add-capacity-storage-system',
     cta: {
-      href: `/odf/system/ns/${getNamespace(resource)}/${referenceForModel(
-        StorageClusterModel
-      )}/${getName(resource)}/~attachstorage`,
+      href: getAttachStorageRoute(getNamespace(resource), getName(resource)),
```
SanjalKatiyar, line 117: plz make sure there is no regression...

packages/odf/components/actions/csv-actions.ts
```diff
@@ -133,18 +148,18 @@ const AddCapacityStorageSystem = (
...
-      launchModal(ConfigurePerformanceModal, {
-        extraProps: { storageCluster },
-        isOpen: true,
-      });
+    cta: {
+      href: getConfigurePerformanceProfileRoute(
+        getNamespace(storageCluster),
+        getName(storageCluster)
+      ),
+      external: false,
```
SanjalKatiyar: what the need for `external: false` ??

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
```diff
@@ -0,0 +1,243 @@
...
+const ConfigurePerformanceProfile: React.FC = () => {
+  const { t } = useTranslation();
+  const { resourceName, namespace } = useParams();
+  const navigate = useNavigate();
+
+  const isProviderMode = useFlag(PROVIDER_MODE);
+  const { systemFlags, areFlagsLoaded, flagsLoadError } =
+    useODFSystemFlagsSelector();
+  const externalClusterDetails = useGetExternalClusterDetails();
+  const hasExternalMode = externalClusterDetails.clusterName !== '';
```
SanjalKatiyar: same here...
```suggestion
```

SanjalKatiyar: why u need to know "has" state and not "is" state ??

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
```diff
@@ -0,0 +1,243 @@
...
+    </>
+  );
+};
+
+const ConfigurePerformanceProfile: React.FC = () => {
+  const { t } = useTranslation();
+  const { resourceName, namespace } = useParams();
+  const navigate = useNavigate();
+
+  const isProviderMode = useFlag(PROVIDER_MODE);
```
SanjalKatiyar: why we need isProviderMode check here ??

if it's ProviderMode this page itself should be blocked...
block the route extension using `PROVIDER_MODE` flag...

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
```diff
@@ -0,0 +1,243 @@
...
+      type: ConfigurePerformanceProfileActionType.SET_INPROGRESS,
+      payload: true,
+    });
+
+    try {
+      if (
+        showCoreStorage &&
+        !isCoreStorageSaveDisabled(state.resourceProfile)
+      ) {
+        await submitCoreStorageProfile({
```
SanjalKatiyar:
```suggestion
        await patchCoreStorageProfile({
```

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
```diff
@@ -0,0 +1,243 @@
...
+          storageCluster,
+          resourceProfile: state.resourceProfile,
+        });
+      }
+      if (showMcgPerformance && !isMcgPerformanceSaveDisabled()) {
+        await submitMcgPerformanceProfile();
```
SanjalKatiyar:
```suggestion
        await patchMcgPerformanceProfile();
```

packages/odf/components/configure-performance-profiles/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+): boolean =>
+  isCoreStorageSectionVisible(visibility) ||
+  isMcgPerformanceSectionVisible(visibility);
+
+export const isCoreStorageSaveDisabled = (
+  resourceProfile: ResourceProfile | null
+): boolean => !resourceProfile;
+
+// Todo: Add logic later
+export const isMcgPerformanceSaveDisabled = (): boolean => true;
```
SanjalKatiyar: why do u even need this util is `!resourceProfile` is the only check u have to do ??

for more than one check/condition, it make sense to have a separate util...

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+  const visibility = {
+    storageCluster,
+    hasExternalMode,
+    isProviderMode,
+    isNoobaaAvailable: !!systemFlags[namespace]?.isNoobaaAvailable,
+  };
+  const showCoreStorage = isCoreStorageSectionVisible(visibility);
+  const showMcgPerformance =
+    isMcgPerformanceSectionVisible(visibility) && false; // will remove false once MCG performance profile is implemented
```
SanjalKatiyar: even todo comments should be bit structured...
```suggestion
// ToDo: remove false once MCG performance profile is implemented
  const showMcgPerformance =
    isMcgPerformanceSectionVisible(visibility) && false;
```

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
```diff
@@ -0,0 +1,243 @@
...
+  const showConfigurePerformance =
+    isConfigurePerformanceProfileVisible(visibility);
```
SanjalKatiyar: how is user landing on this page, if we don't yet know to `showConfigurePerformance` at all or not ??

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
```diff
@@ -0,0 +1,243 @@
...
+  const { clusterNodes, nodesLoaded, nodesLoadError } = useClusterNodes();
+
+  const isLoaded =
+    areFlagsLoaded && storageClusterLoaded && (!showCoreStorage || nodesLoaded);
```
SanjalKatiyar: u need nodes data for mcg as well... so `!showCoreStorage || nodesLoaded` doesn't make sense...

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
```diff
@@ -0,0 +1,243 @@
...
+  const isLoadError =
+    flagsLoadError ||
+    storageClusterLoadError ||
+    (showCoreStorage && !!nodesLoadError);
```
SanjalKatiyar: same here...

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
```diff
@@ -0,0 +1,243 @@
...
+      replace: true,
+    });
+  }, [
+    isLoaded,
+    isLoadError,
+    showConfigurePerformance,
+    navigate,
+    namespace,
+    resourceName,
+  ]);
```
SanjalKatiyar: why ?? there's absolutely no need for adding unnecessary hook... why are user even handing on this page if we don't need to show performance profile at all ??
```suggestion
```

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
```diff
@@ -0,0 +1,243 @@
...
+            label={t('Configure performance profile')}
+          />
+        </div>
+      </>
+    );
+  }
+
+  if (!showConfigurePerformance) {
+    return null;
+  }
```
SanjalKatiyar: unnecessary...
```suggestion
```

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
```diff
@@ -0,0 +1,243 @@
...
+        <div className="pf-v6-u-pb-xl">
+          <StatusBox
+            loaded={isLoaded}
+            loadError={isLoadError}
+            label={t('Configure performance profile')}
+          />
+        </div>
+      </>
+    );
+  }
```
SanjalKatiyar:
```suggestion
```

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
```diff
@@ -0,0 +1,243 @@
...
+          )}
+        </Content>
+      </PageHeading>
+      <ConfigurePerformanceProfileForm
+        storageCluster={storageCluster}
+        clusterNodes={clusterNodes}
+        showCoreStorage={showCoreStorage}
+        showMcgPerformance={showMcgPerformance}
+        onClose={onClose}
+      />
```
SanjalKatiyar:
```suggestion
{(!isLoaded || isLoadError) ?
      <StatusBox
         loaded={isLoaded}
         loadError={isLoadError}
         label={t('Configure performance profile')}
       /> :
      <ConfigurePerformanceProfileForm
        storageCluster={storageCluster}
        clusterNodes={clusterNodes}
        showCoreStorage={showCoreStorage}
        showMcgPerformance={showMcgPerformance}
        onClose={onClose}
      />
```

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/2942/changes#r3576712014

packages/odf/components/configure-performance-profiles/core-storage-section.tsx
```diff
@@ -0,0 +1,121 @@
...
+  resourceProfile: ResourceProfile | null;
+};
+
+export const submitCoreStorageProfile = async ({
+  storageCluster,
+  resourceProfile,
+}: SubmitCoreStorageProfileParams): Promise<void> => {
+  if (isCoreStorageSaveDisabled(resourceProfile)) {
+    return;
+  }
```
SanjalKatiyar: why are we even calling this function if `isCoreStorageSaveDisabled` ??

packages/odf/components/configure-performance-profiles/multicloud-object-gateway-section.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import { useCustomTranslation } from '@odf/shared/useCustomTranslationHook';
+import { Content, ContentVariants } from '@patternfly/react-core';
+
+export const submitMcgPerformanceProfile = async (): Promise<void> => {
+  //ToDo in sprint2: MCG performance profile patch will be added here.
```
SanjalKatiyar:
```suggestion
  // ToDo: add MCG performance profile patch here.
```

packages/odf/components/configure-performance-profiles/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+import { StorageClusterKind } from '@odf/shared';
+import { ConfigurePerformanceProfileFormState } from './state';
+
+export const getOcsLabeledWizardNodes = (
+  nodes: WizardNodeState[],
+  namespace: string
+): WizardNodeState[] =>
+  nodes.filter((node) => isOcsLabeledWizardNode(node, namespace));
+
+export type ConfigurePerformanceProfileVisibility = {
```
SanjalKatiyar: rename accordingly...

packages/odf/components/configure-performance-profiles/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export const checkRequiredValues = (
+  state: ConfigurePerformanceProfileFormState,
+  showCoreStorage: boolean,
+  showMcgPerformance: boolean
+): boolean => {
+  const coreStorageDisabled = showCoreStorage
+    ? isCoreStorageSaveDisabled(state.resourceProfile)
+    : true;
+  const mcgPerformanceDisabled = showMcgPerformance
```
SanjalKatiyar:
```suggestion
  const isCoreDisabled = showCoreStorage
    ? isCoreStorageSaveDisabled(state.resourceProfile)
    : true;
  const isMcgDisabled = showMcgPerformance
```

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/configure-performance.tsx
SanjalKatiyar:
```suggestion
  if (!disableUnsupportedProfiles && availableNodesLoaded && !availableNodesLoadError) {
```

SanjalKatiyar: we don't need these api calls when `disableUnsupportedProfiles` is `true`...

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/inline-resource-requirements-text.tsx
```diff
@@ -0,0 +1,57 @@
...
+};
+
+export const InlineResourceRequirementsText: React.FC<
+  InlineResourceRequirementsTextProps
+> = ({
+  minCpu,
+  minMem,
+  clusterCpu,
+  clusterMemoryGiB,
+  id = 'resource-requirements',
```
SanjalKatiyar: it doesn't need to be a prop...
```suggestion
```

packages/odf/components/Sections/StorageClusterSection.tsx
```diff
@@ -74,12 +78,30 @@ const storageClusterActions =
...
+            getNamespace(storageCluster),
+            getName(storageCluster)
+          ),
         });
       }
     }
+    if (
+      isConfigurePerformanceProfileVisible({
+        storageCluster,
+        hasExternalMode: isExternalMode,
```
SanjalKatiyar:
```suggestion
        isExternalMode,
```

packages/odf/components/Sections/StorageClusterSection.tsx
```diff
@@ -148,7 +174,9 @@ const StorageClusterSection: React.FC = () => {
           currentStorageCluster,
           infrastructure,
           isLSOInstalled,
-          hasExternalMode
+          hasExternalMode,
```
SanjalKatiyar:
```suggestion
          isExternalMode: systemFlags[namespace]?.isExternalMode,
```

plugins/odf/console-extensions.json
```diff
@@ -763,6 +763,14 @@
...
     }
   },
+  {
+    "type": "console.page/route",
+    "properties": {
+      "path": "/odf/system/ns/:namespace/ocs.openshift.io~v1~storagecluster/:resourceName/~configureperformanceprofile",
+      "exact": false,
+      "component": { "$codeRef": "configurePerformanceProfile.default" }
+    }
+  },
```
SanjalKatiyar, line 773: disable it here if it's provider mode (assuming we don't want to support performance profiles for provider mode)...

SanjalKatiyar (changes requested):

NIKHITHAVADDEMPUDI:
@SanjalKatiyar i have addressed all the comments in the pr except a few 

https://github.com/red-hat-storage/odf-console/pull/2942#discussion_r3576958432
https://github.com/red-hat-storage/odf-console/pull/2942#discussion_r3576674058
https://github.com/red-hat-storage/odf-console/pull/2942#discussion_r3576698536
https://github.com/red-hat-storage/odf-console/pull/2942#discussion_r3576701656
https://github.com/red-hat-storage/odf-console/pull/2942#discussion_r3576704988

because of the following reasons:

showConfigurePerformance gates the page because users can hit the direct URL even when the menu action is hidden, and we can’t enforce all eligibility rules in console-extensions.json since route flags are global while section visibility is per-cluster: core storage only when the cluster is not external/provider (shouldShowCoreStorageSection), and MCG only when Noobaa is available for that namespace (shouldShowMcgPerformanceSection) — runtime systemFlags[namespace] state that static route flags can’t evaluate for the specific :namespace/:resourceName in the URL.

SanjalKatiyar:
> showConfigurePerformance gates the page because users can hit the direct URL even when the menu action is hidden, and we can’t enforce all eligibility rules in console-extensions.json since route flags are global while section visibility is per-cluster: core storage only when the cluster is not external/provider (shouldShowCoreStorageSection), and MCG only when Noobaa is available for that namespace (shouldShowMcgPerformanceSection) — runtime systemFlags[namespace] state that static route flags can’t evaluate for the specific :namespace/:resourceName in the URL.

What I am saying is, we should decide `shouldShowConfigurePerformanceProfile` before even allowing user to be able to navigate to ConfigurePerformanceProfile page.
Once you are already on ConfigurePerformanceProfile, then you don't have to decide `shouldShowConfigurePerformanceProfile`, then u need to just decide `shouldShowMcgPerformanceSection` and `shouldShowCoreStorageSection`.

For extensions file, I never said to gate based on internal/external etc, u need to just gate here for provider mode (we already have a global flag for this). Disable the Route here itself for such case.

Which case are you worried about, can you plz elaborate ?? Which case is getting missed as per above suggestion ??

SanjalKatiyar:
https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/system-list/external-systems-list.tsx#L528

This flow is missing the configure profile option. Here as well we first use `shouldShowConfigurePerformanceProfile` to decide whether to add the kebab item, if added, then on the ConfigurePerformanceProfile page we use `shouldShowMcgPerformanceSection` and `shouldShowCoreStorageSection` to decide what to show.

NIKHITHAVADDEMPUDI:
> > showConfigurePerformance gates the page because users can hit the direct URL even when the menu action is hidden, and we can’t enforce all eligibility rules in console-extensions.json since route flags are global while section visibility is per-cluster: core storage only when the cluster is not external/provider (shouldShowCoreStorageSection), and MCG only when Noobaa is available for that namespace (shouldShowMcgPerformanceSection) — runtime systemFlags[namespace] state that static route flags can’t evaluate for the specific :namespace/:resourceName in the URL.
> 
> What I am saying is, we should decide `shouldShowConfigurePerformanceProfile` before even allowing user to be able to navigate to ConfigurePerformanceProfile page. Once you are already on ConfigurePerformanceProfile, then you don't have to decide `shouldShowConfigurePerformanceProfile`, then u need to just decide `shouldShowMcgPerformanceSection` and `shouldShowCoreStorageSection`.
> 
> For extensions file, I never said to gate based on internal/external etc, u need to just gate here for provider mode (we already have a global flag for this). Disable the Route here itself for such case.
> 
> Which case are you worried about, can you plz elaborate ?? Which case is getting missed as per above suggestion ??

@SanjalKatiyar 
core-storage section is shown when it is not external mode and not provider mode 
and 
mcg profiles section is shown only when noobaa is available 

but if we block entire route with provider mode we might hide mcg section as well, though it should not be hiddden in that case

So as of now
gating using provider flag is not done 

and since it is not possible to gate using all three cases (provider, external modes and is noobaa available) how is it possible to gate the route before it is opened is my doubt., though we are gating it in the kebab menus , if we use direct URL we may still reach the page.

NIKHITHAVADDEMPUDI:
> https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/system-list/external-systems-list.tsx#L528
> 
> This flow is missing the configure profile option. Here as well we first use `shouldShowConfigurePerformanceProfile` to decide whether to add the kebab item, if added, then on the ConfigurePerformanceProfile page we use `shouldShowMcgPerformanceSection` and `shouldShowCoreStorageSection` to decide what to show.

Hi @SanjalKatiyar , I included this change in the latest commit

SanjalKatiyar:
> but if we block entire route with provider mode we might hide mcg section as well, though it should not be hiddden in that case

Okay, my understanding was both core and mcg were not supported on provider mode.

That said, this PR has a regression now, I checked, even core profiles are supported for the provider mode, so let's not block them.

NIKHITHAVADDEMPUDI:
> > but if we block entire route with provider mode we might hide mcg section as well, though it should not be hiddden in that case
> 
> Okay, my understanding was both core and mcg were not supported on provider mode.
> 
> That said, this PR has a regression now, I checked, even core profiles are supported for the provider mode, so let's not block them.

Removed isProviderMode from utils.ts and all action/page call sites (StorageClusterSection, csv-actions, system-list actions) in the latest commit

packages/odf/components/configure-performance-profiles/configure-performance-profile-footer.tsx
```diff
@@ -0,0 +1,56 @@
...
+  ButtonType,
+  ButtonVariant,
+} from '@patternfly/react-core';
+import { ConfigurePerformanceProfileFormState } from './state';
+import { checkRequiredValues } from './utils';
+
+export const ConfigurePerformanceProfileFormFooter = (
+  props: ConfigurePerformanceProfileFormFooterProps
+) => {
+  const { state, cancel, onConfirm, showCoreStorage, showMcgPerformance } =
```
bipuladh:
```suggestion
export const ConfigurePerformanceProfileFormFooter: React.FC<ConfigurePerformanceProfileFormFooterProps> = (
{ state, cancel, onConfirm, showCoreStorage, showMcgPerformance }
) => {

```

packages/odf/components/configure-performance-profiles/configure-performance-profile.scss
```diff
@@ -0,0 +1,21 @@
+.configure-performance-profile {
+  display: flex;
+  flex-direction: column;
+  overflow: hidden;
+
+  > .odf-m-nav-title {
+    flex-shrink: 0;
```
bipuladh: Why?

packages/odf/components/configure-performance-profiles/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export const shouldShowConfigurePerformanceProfile = (
+  params: ShouldShowConfigurePerformanceProfileParams
+): boolean =>
+  shouldShowCoreStorageSection(params) ||
+  shouldShowMcgPerformanceSection(params);
+
+export const checkRequiredValues = (
+  state: ConfigurePerformanceProfileFormState,
+  showCoreStorage: boolean,
```
bipuladh, line 39: What is coreStorage?

packages/odf/components/configure-performance-profiles/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+): WizardNodeState[] =>
+  nodes.filter((node) => isOcsLabeledWizardNode(node, namespace));
+
+export type ShouldShowConfigurePerformanceProfileParams = {
+  storageCluster?: StorageClusterKind;
+  isExternalMode: boolean;
+  isNoobaaAvailable: boolean;
+};
+
+export const shouldShowCoreStorageSection = ({
```
bipuladh, line 16:
```suggestion
export const shouldShowCephSection = ({
```
This makes block and file more clearer. Core storage is a confusing term

packages/odf/components/configure-performance-profiles/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+  storageCluster,
+  isExternalMode,
+}: ShouldShowConfigurePerformanceProfileParams): boolean =>
+  !!storageCluster && !isExternalMode;
```
bipuladh: I don't see the need to pass two variables. We can figure out if it's a external mode cluster from the storagecluster resource itself. Use https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/utils/odf.ts#L49

plugins/odf/console-extensions.json
bipuladh: @NIKHITHAVADDEMPUDI did we check if we do support it?

packages/odf/components/system-list/actions.ts
```diff
@@ -24,23 +30,25 @@ const getKindOfExternalSystem = (obj: StorageSystemKind): string => {
...
+  customActions: CustomKebabItem[];
   hiddenActions: ModalKeys[];
 };
 
 export const getActions = (
   obj: StorageSystemKind,
   t: TFunction,
-  isSANSystemDeletable: boolean
+  isSANSystemDeletable: boolean,
+  options: ConfigurePerformanceActionOptions = {}
```
bipuladh:
```suggestion
  configurePerformanceOptions: ConfigurePerformanceActionOptions = {}
```

packages/odf/components/create-storage-system/payloads.ts
```diff
@@ -251,13 +251,20 @@ export const createStorageAutoScaler = (
   return k8sCreate({ model: StorageAutoScalerModel, data });
 };
 
+export const isOcsLabeledWizardNode = (
+  node: WizardNodeState,
+  namespace: string
```
bipuladh: Do we really need to pass the namespace can we do partial matching instead this can be useful for external mode in a later day as well after that? Also move this shared/utils

packages/odf/components/configure-performance-profiles/utils.ts
NIKHITHAVADDEMPUDI: Hi @bipul, used the name coreStoragesection as per the suggestion in the ux designs https://www.figma.com/design/zIHZR2BEoe54aND03T7PZJ/Support-Performance-profiles-for-MCG?node-id=12264-2939&t=lvjA9qkiouOiHgk7-4, but is it fine to keep the title as coreStorage and the variable names related as ceph? if it is ok i will change the names please confirm

packages/odf/components/configure-performance-profiles/utils.ts
NIKHITHAVADDEMPUDI: https://github.com/red-hat-storage/odf-console/pull/2942#discussion_r3664563748

NIKHITHAVADDEMPUDI:
https://github.com/red-hat-storage/odf-console/pull/2942#discussion_r3662604251
@bipul yes i have verified the path by navigating to the configure performance URL loads the page correctly, this follows the same route pattern as ~attachstorage.

Screenshot 2026-07-28 at 5 32 27 PM

plugins/odf/console-extensions.json
NIKHITHAVADDEMPUDI: @bipuladh yes @SanjalKatiyar confirmed that provider mode is supported for core-storage section https://github.com/red-hat-storage/odf-console/pull/2942#issuecomment-5070209996

packages/odf/components/configure-performance-profiles/configure-performance-profile-footer.tsx
NIKHITHAVADDEMPUDI: I included this change in the latest commit

packages/odf/components/configure-performance-profiles/configure-performance-profile.scss
NIKHITHAVADDEMPUDI: removed the unused styles in the latest commit

packages/odf/components/configure-performance-profiles/utils.ts
NIKHITHAVADDEMPUDI: I included this change in the latest commit

packages/odf/components/system-list/actions.ts
NIKHITHAVADDEMPUDI: I included this change in the latest commit

packages/odf/components/create-storage-system/payloads.ts
NIKHITHAVADDEMPUDI: I included this change in the latest commit

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
```diff
@@ -0,0 +1,193 @@
...
+              showCoreStorage={showCoreStorage}
+              showMcgPerformance={showMcgPerformance}
+              cancel={onClose}
+              onConfirm={onConfirm}
+            />
+          </div>
+        </>
+      )}
+    </>
+  );
```
bipuladh:
```suggestion

          <StatusBox
            loaded={nodesLoaded}
            loadError={nodesLoadError}
            label={t('Configure performance profile')}
>
          <div className="odf-m-pane__body odf-m-pane__form configure-performance-profile__content pf-v6-u-mt-md">
            {showCoreStorage && (
              <CoreStorageSection
                state={state}
                dispatch={dispatch}
                storageCluster={storageCluster}
                clusterNodes={clusterNodes}
              />
            )}
            {showMcgPerformance && <McgPerformanceSection />}
          </div>
          <div className="odf-m-pane__body configure-performance-profile__footer">
            <ConfigurePerformanceProfileFormFooter
              state={state}
              showCoreStorage={showCoreStorage}
              showMcgPerformance={showMcgPerformance}
              cancel={onClose}
              onConfirm={onConfirm}
            />
          </div>
</StatusBox>
  );
```

packages/odf/components/configure-performance-profiles/utils.ts
bipuladh: I guess we can live with this

bipuladh (commented):
Statusbox is used incorrectly

packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx
NIKHITHAVADDEMPUDI: @bipuladh, I have included this suggestion in the latest commit, after testing the layout and function once again.

bipuladh (approved):

packages/odf/components/system-list/external-systems-list.tsx
```diff
@@ -466,17 +467,31 @@ const StorageSystemRow: React.FC<RowProps<StorageSystemKind, CustomData>> = ({
...
+          getName(sc) === obj.spec.name &&
+          getNamespace(sc) === obj.spec.namespace
+      )
+    : undefined;
   const { customActions, hiddenActions } = getActions(
     obj,
     t,
     isSANSystemDeletable,
-    isRemoteClusterDeletable
+    isRemoteClusterDeletable,
```
SanjalKatiyar, line 486: is this flow tested or we just assumed that agent's changes are correct ??

SanjalKatiyar: I don't think StorageCluster has spec.name and spec.namespace.

packages/odf/components/system-list/external-systems-list.tsx
```diff
@@ -466,17 +467,31 @@ const StorageSystemRow: React.FC<RowProps<StorageSystemKind, CustomData>> = ({
     isSANSystemDeletable,
     sanClustersByName,
     filesystemsByRemoteCluster,
+    storageClusters,
```
SanjalKatiyar: make it same like other systems, instead of passing entire list...

SanjalKatiyar (changes requested):

packages/odf/components/system-list/external-systems-list.tsx
NIKHITHAVADDEMPUDI: I have includes this suggestion in the latest commit

packages/odf/components/system-list/external-systems-list.tsx
NIKHITHAVADDEMPUDI: @SanjalKatiyar , Without a real external StorageCluster, I tested the External systems kebab → configure-performance wiring by temporarily forcing the same inputs that path reads (**all reverted before merge**):

1. Forced isExternalCluster to true in [packages/odf/utils/odf.ts](https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/utils/odf.ts) so the cluster is treated as external.
2. Forced isNoobaaAvailable to true in [packages/odf/redux/provider-hooks/useODFSystemFlags.ts](https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/redux/provider-hooks/useODFSystemFlags.ts) so flags match the external + NooBaa case.
3. Set useWatchStorageSystems(false) in [ExternalSystemsListPage.tsx](https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/Sections/ExternalSystemsListPage.tsx) and [external-systems-list.tsx](https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/system-list/external-systems-list.tsx) so the internal cluster appears on the External systems list (parent empty-state gate + list data).
4. Confirmed kebab shows Configure performance profiles, redirect uses the real StorageCluster name/namespace via [actions.ts](https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/system-list/actions.ts), and forced the MCG section in [configure-performance-profile.tsx](https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/components/configure-performance-profiles/configure-performance-profile.tsx).
Screenshot 2026-07-29 at 3 32 35 PM

Screenshot 2026-07-29 at 3 32 44 PM
Screenshot 2026-07-29 at 3 32 48 PM

NIKHITHAVADDEMPUDI: external-systems-list.tsx calls useWatchStorageSystems(true) → that hook watches real StorageClusters → maps each one with mapStorageClusterToStorageSystem (which copies the CR’s metadata.name/metadata.namespace into spec.name/spec.namespace) → returns those objects as storageSystems → the table passes each one into StorageSystemRow as obj → so when the row reads obj.spec.name / obj.spec.namespace, those fields are already present on that mapped list object.

packages/odf/components/system-list/external-systems-list.tsx
```diff
@@ -664,6 +675,17 @@ export const StorageSystemListPage: React.FC = () => {
   const [storageClusters] = useK8sWatchResource<StorageClusterKind[]>(
     storageClusterResource
   );
+  const storageClustersByName = React.useMemo(
+    () =>
+      (storageClusters ?? []).reduce<Record<string, StorageClusterKind>>(
+        (acc, cluster) => {
+          acc[getName(cluster)] = cluster;
```
SanjalKatiyar, line 682: u can have multiple systems with same name in this case, use `namespace + % + name` as a key instead...

SanjalKatiyar: plz address this in a follow-up PR...

PR #2947, opened 2026-06-29
react migration

Harshvardhan-handa123:
Jira Link: https://redhat.atlassian.net/browse/RHSTOR-9022, https://redhat.atlassian.net/browse/RHSTOR-9023

https://github.com/user-attachments/assets/19d83901-fa75-408d-9844-03951da1c92e


jest/setup.ts
```diff
@@ -1,8 +1,14 @@
...
+import { TextEncoder, TextDecoder } from 'util';
 import { configure } from '@testing-library/react';
 
+global.TextEncoder = TextEncoder as typeof global.TextEncoder;
+global.TextDecoder = TextDecoder as typeof global.TextDecoder;
+
 configure({ testIdAttribute: 'data-test-id' });
 
+globalThis.IS_REACT_ACT_ENVIRONMENT = true;
+
```
bipuladh, line 11: Why are these changes required?

packages/client/src/components/dashboards/cards/status-card.tsx
```diff
@@ -23,6 +23,8 @@ import {
 } from '@patternfly/react-core';
 import '../../../../style.scss';
 
+const HealthBody = HealthBodyBase as React.FC<React.PropsWithChildren<{}>>;
```
bipuladh, line 26: Why do we require this stub?

packages/mco/components/create-dr-policy/utils/cluster-list-utils.ts
```diff
@@ -15,9 +15,9 @@ import {
 import { ConfigMapKind } from '@odf/shared/types';
 import { sortRows } from '@odf/shared/utils';
 import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
+import { TFunction } from 'i18next';
```
bipuladh, line 18: Has TFunction moved?

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -90,9 +90,9 @@ const SelectNodesText: React.FC<SelectNodesTextProps> = React.memo(
       <Content>
         <Content component="p">{text}</Content>
         <Content component="p">
-          <Trans t={t as any} ns="plugin__odf-console">
+          <Trans t={t as any} ns="plugin__odf-console" values={{ label }}>
```
bipuladh, line 93: For some places ns is used some places it isn't can we remove it if not required? I guess t is already initialized to some namespace so perhaps we can remove it, wdyt?

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
```diff
@@ -375,11 +375,11 @@ const SelectedCapacityAndNodes: React.FC<SelectedCapacityAndNodesProps> = ({
               />
               <Content>
                 <Content component={ContentVariants.small}>
-                  <Trans ns="plugin__odf-console">
+                  <Trans ns="plugin__odf-console" values={{ storageClassName }}>
                     The available capacity is based on all attached disks
                     associated with the selected{' '}
                     {/* eslint-disable-next-line react/no-unescaped-entities */}
-                    StorageClass <b>{{ storageClassName }}</b>
+                    StorageClass <b>{'{{storageClassName}}'}</b>
```
bipuladh, line 382: Why do we need `'`?

packages/shared/src/charts/PrometheusGraph.tsx
```diff
@@ -2,22 +2,23 @@ import * as React from 'react';
...
-    >
-      {title && <h5 className="graph-title">{title}</h5>}
-      {children}
-    </div>
-  )
-);
+export const PrometheusGraph = React.forwardRef<
+  HTMLDivElement,
+  PrometheusGraphProps
+>(({ children, className, title }, ref) => (
```
bipuladh: Why are we dropping the type for ref?

packages/shared/src/dashboards/graphs/prometheus-graph.tsx
```diff
@@ -58,29 +58,32 @@ export const PrometheusGraphLink: React.FC<PrometheusGraphLinkProps> = ({
...
-    >
-      {title && <h5 className="graph-title">{title}</h5>}
-      {children}
-    </div>
-  )
-);
+export const PrometheusGraph = React.forwardRef<
+  HTMLDivElement,
+  PrometheusGraphProps
+>(({ children, className, title }, ref) => (
```
bipuladh: types for ref

packages/shared/src/dashboards/utilization-card/area-chart.tsx
```diff
@@ -104,7 +104,7 @@ export const AreaChart: React.FC<AreaChartProps> = ({
     return (
       <ChartVoronoiContainer
         voronoiDimension="x"
-        labels={getLabel}
+        labels={(point) => getLabel(point)}
```
bipuladh: Why is this required? Has the function type changed?

packages/shared/src/modals/DeleteModal.tsx
```diff
@@ -138,21 +138,21 @@ const DeleteModal: React.FC<CommonModalProps<DeleteModalExtraProps>> = ({
...
-              {{ resourceName: resource.metadata.name }}
-            </strong>{' '}
-            in namespace{' '}
-            <strong>{{ namespace: resource.metadata.namespace }}</strong>?
+            <strong className="co-break-word">{'{{resourceName}}'}</strong> in
+            namespace <strong>{'{{namespace}}'}</strong>?
           </Trans>
         ) : (
-          <Trans t={t}>
+          <Trans t={t} values={{ resourceName: resource.metadata.name }}>
```
bipuladh, line 153:
```suggestion
          <Trans t={t} values={{ resourceName: getName(resource) }}>
```

packages/shared/src/modals/DeleteModal.tsx
```diff
@@ -138,21 +138,21 @@ const DeleteModal: React.FC<CommonModalProps<DeleteModalExtraProps>> = ({
     >
       <ModalBody>
         {isNamespaced ? (
-          <Trans t={t}>
+          <Trans
+            t={t}
+            values={{
+              resourceName: resource.metadata.name,
+              namespace: resource.metadata.namespace,
```
bipuladh:
```suggestion
              resourceName: getName(resource),
              namespace: getNamespace(resource),
```

packages/shared/package.json
```diff
@@ -48,12 +48,12 @@
     "yup": "^1.7.1"
   },
   "peerDependencies": {
-    "@openshift-console/dynamic-plugin-sdk": "4.21.0",
-    "@openshift-console/dynamic-plugin-sdk-internal": "4.21.0",
+    "@openshift-console/dynamic-plugin-sdk": "4.22.0-prerelease.3",
+    "@openshift-console/dynamic-plugin-sdk-internal": "4.22.0-prerelease.3",
```
bipuladh: I think we have stable 4.22 SDK released now. https://www.npmjs.com/package/@openshift-console/dynamic-plugin-sdk/v/4.22.0

package.json
```diff
@@ -65,43 +65,43 @@
...
-    "@openshift-console/dynamic-plugin-sdk-webpack/ajv": "^6.15.0",
     "@eslint/eslintrc/ajv": "^6.15.0",
-    "eslint/ajv": "^6.15.0",
-    "schema-utils/ajv": "^8.20.0",
-    "table/ajv": "^8.20.0",
-    "ajv-formats/ajv": "^8.20.0",
-    "ajv-keywords/ajv": "^8.20.0",
-    "@types/react": "^17",
-    "@types/react-dom": "^17",
+    "@openshift-console/dynamic-plugin-sdk-webpack/ajv": "^6.15.0",
```
bipuladh, line 69: Are these resolutions still required? Aren't they fixed in the SDK packages?

package.json
```diff
@@ -65,43 +65,43 @@
...
-    "@patternfly/react-charts": "8.4.0",
-    "@patternfly/react-code-editor": "6.4.0",
-    "@patternfly/react-core": "6.4.0",
-    "@patternfly/react-icons": "6.4.0",
-    "@patternfly/react-table": "6.4.0",
-    "@patternfly/react-tokens": "6.4.0",
-    "@patternfly/react-topology": "6.4.0",
-    "@types/react": "^17",
-    "@types/react-dom": "^17",
+    "@openshift-console/dynamic-plugin-sdk-webpack": "4.22.0-prerelease.3",
```
bipuladh: https://www.npmjs.com/package/@openshift-console/dynamic-plugin-sdk/v/4.22.0

jest/setup.ts
Harshvardhan-handa123: https://react.dev/blog/2022/03/08/react-18-upgrade-guide#configuring-your-testing-environment -> For line 10

Harshvardhan-handa123: For line 5,6 textEncoder and textDecoder: 
React Router v7 internally uses the textEncoder API for URL parsing and encoding/decoding from binary to string and visa versa, but the jest environment (jsdom) does not have it available by default.
So when the jest imported React router v7, it showed the following error: 
" ReferenceError: TextEncoder is not defined ". 
To tackle this issue I added these lines.

packages/mco/components/create-dr-policy/utils/cluster-list-utils.ts
Harshvardhan-handa123: Yes, it has moved in react 18

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
Harshvardhan-handa123: So the Trans component has changed a bit. We can't pass a JS Object as a child. 
Instead we pass it as a string literal and which fetches the value from values(line 378).

packages/odf/components/create-storage-system/create-storage-system-steps/capacity-and-nodes-step/capacity-and-nodes-step.tsx
Harshvardhan-handa123: I guess for useCustomTranslation (t)  the namespace Is already defined.. idts Trans is initialised to some namespace.. I'll check into this more

packages/client/src/components/dashboards/cards/status-card.tsx
Harshvardhan-handa123: export declare const HealthBody: FC;
This is the Type definition of HealthBody..
React 17 used to automatically add children to every component's props. But React 18 doesn't do that.. In react 18 we need to explicitly declare that the component must accept children. That is why I added this change.

packages/shared/src/charts/PrometheusGraph.tsx
Harshvardhan-handa123: Changed it back to ref: React.RefObject<HTMLDivElement>

packages/shared/src/dashboards/graphs/prometheus-graph.tsx
Harshvardhan-handa123: changed it back to ref: React.RefObject<HTMLDivElement>

packages/shared/src/dashboards/utilization-card/area-chart.tsx
Harshvardhan-handa123: fixed it now. The function structure defined for ChartVoronoiContainer.labels do not match the getLabel function. 
Earlier in React 17  getLabel was getting accepted as a value for the prop labels. 
But in react 18 it is not. 
Now I added the line  labels= {getLabels as unknown as  ( point: any, index: number, points: any[] ) => string}
This fixes the problem. I checked the UI, the ChartVoronoiContainer component for the charts is working fine.

packages/shared/src/modals/DeleteModal.tsx
Harshvardhan-handa123: done

packages/shared/src/modals/DeleteModal.tsx
Harshvardhan-handa123: done

packages/shared/package.json
Harshvardhan-handa123: Will change it back to 4.22.0.

package.json
Harshvardhan-handa123: Done

package.json
Harshvardhan-handa123: @openshift-console/dynamic-plugin-sdk-webpack still requests for 6.12.3 ajv.. which is vulnerable. 
We need this resolution.

bipuladh (approved):

webpack.config.ts
```diff
@@ -57,6 +57,8 @@ const config: webpack.Configuration & DevServerConfiguration = {
     extensions: ['.ts', '.tsx', '.js', '.jsx'],
     alias: {
       '@odf/shared': path.resolve(__dirname, './packages/shared/src/'),
+      'react/jsx-runtime.js': require.resolve('react/jsx-runtime'),
+      'react/jsx-dev-runtime.js': require.resolve('react/jsx-dev-runtime'),
```
SanjalKatiyar: why is this needed ??

we don't import/use these directly...

Harshvardhan-handa123: Hey.. these lines were added by BOB for the fix for the error which appears on yarn build command: 

ERROR in ./node_modules/react-dnd/dist/esm/core/DndProvider.mjs 1:0-51
Module not found: Error: "./jsx-runtime.js" is not exported under the conditions ["import","module","webpack","production","browser"] from package /Users/harshvardhanhanda/odf-console/node_modules/react (see exports field in /Users/harshvardhanhanda/odf-console/node_modules/react/package.json)
 @ ./node_modules/react-dnd/dist/esm/core/index.mjs
 @ ./node_modules/react-dnd/dist/esm/index.mjs 2:0-33 2:0-33
 @ ./packages/shared/src/utils/NameValueEditor.tsx 4:0-40 17:32-43
 @ ./packages/odf/components/s3-iam/create-user/AddKeyValuePair.tsx 4:0-72 37:32-51
 @ ./packages/odf/components/s3-iam/create-user/CreateUserForm.tsx 14:0-53 152:40-56
 @ container entry

Harshvardhan-handa123: I have removed line 61 now, it is not needed

SanjalKatiyar: it's still a hack which should only be used if all other options fails...

try `grep -R "react/jsx-runtime.js" node_modules` and see which package imports this (react-dnd highly likely)...
upgrade the package which is supported with react 18 version...
test PR once, make sure nothing breaks and `NameValueEditor` is working as expected...

Harshvardhan-handa123: removed it, also migrated react-dnd to v16

SanjalKatiyar: @Harshvardhan-handa123 did u test the PR after making this change ??

PR #2948, opened 2026-06-30
RHSTOR-6406: Add multi-select checkboxes and bulk selector to protected applications list page

vbnrh:
## Design

[Figma](https://www.figma.com/design/aWVx8rBXjiqqswaTqBTTyl/Provide-DR-failover-and-Relocation-operations-on-a-group-of-Applications?node-id=4666-68124&p=f&t=fnouEneFFze9ewoz-0)

## Summary
- Adds per-row checkboxes to the protected applications list page with disabled state for mid-operation apps (FailingOver, Relocating, Deleting)
- Adds a bulk selector dropdown (Select none / Select page / Select all) with eligible-only counting
- Adds a "Failover/Relocate" toolbar button that enables when rows are selected (click handler to be wired in [RHSTOR-6407](https://redhat.atlassian.net/browse/RHSTOR-6407))
- Extends shared `ComposableTable` and `PaginatedListPage` with opt-in selection support (no breaking changes to existing consumers)


vbnrh:
https://github.com/user-attachments/assets/b106aadc-4152-48a7-838f-3f6a13240b33

packages/shared/src/list-page/paginated-list-page.tsx
```diff
@@ -59,42 +61,49 @@ export const PaginatedListPage: React.FC<PaginatedListPageProps> = ({
...
-                onPerPageSelect={(_event, newPerPage, newPage) => {
-                  setPerPage(newPerPage);
-                  setPage(newPage);
-                }}
-              />
-            </GridItem>
-          </Grid>
+          <div className="pf-v6-u-display-flex pf-v6-u-justify-content-space-between pf-v6-u-align-items-center pf-v6-u-flex-wrap pf-v6-u-mt-md">
+            <div
+              className="pf-v6-u-display-flex pf-v6-u-align-items-flex-start pf-v6-u-flex-wrap"
```
SanjalKatiyar, line 74: is shift from PF "grid" to direct "div" with custom classes really needed ??

grid automatically adds these spacing, center, etc... out of the box... why we want to not use that...

also did u test any other page that is already using this component ?? shifting to div could cause styling issue at other existing places...

packages/shared/src/list-page/paginated-list-page.tsx
```diff
@@ -59,42 +61,49 @@ export const PaginatedListPage: React.FC<PaginatedListPageProps> = ({
...
+              {toolbarActions}
+              {!hideFilter && (
+                <ListPageFilterWrapper
+                  {...listPageFilterProps}
+                  data={getValidFilteredData(listPageFilterProps.data)}
+                />
+              )}
+              {!!CreateButton && <CreateButton />}
+            </div>
+            <Pagination
```
SanjalKatiyar, line 90: grid was added so that filter and pagination are always aligned together row-wise... I don't see any such guarantee now ??

now it purely depends on user's screen and will cause inconsistent behaviour between diff systems + affects other places which uses this shared component ??

packages/mco/components/protected-applications/list-page.tsx
```diff
@@ -52,21 +51,71 @@ import {
...
+        isAllSelected={isAllPageSelected}
+        onSelectNone={onSelectNone}
+        onSelectPage={onSelectPage}
+        onSelectAll={onSelectAll}
+      />
+      <Button
+        variant={ButtonVariant.secondary}
+        isDisabled={selectedCount === 0}
+      >
+        {t('Failover/relocate')}
```
SanjalKatiyar: nits:
```suggestion
        {t('Failover/Relocate')}
```

packages/mco/components/protected-applications/list-page.tsx
```diff
@@ -109,6 +160,15 @@ const ProtectedAppsTableRow: React.FC<
             expandId: 'expandable-table',
           }}
         />
+        <Td
+          select={{
+            rowIndex: rowIndex ?? 0,
+            onSelect: selectProps.onRowSelect,
+            isSelected: selectProps.isSelected(pav),
+            isDisabled: selectProps.isDisabled(pav),
+            props: { id: pav.metadata?.uid },
```
SanjalKatiyar: nits:
```suggestion
            props: { id: getUID(pav) },
```

packages/mco/components/protected-applications/list-page.tsx
```diff
@@ -177,12 +236,53 @@ export const ProtectedApplicationsListPage: React.FC = () => {
...
+  const [pagePavs, setPagePavs] = React.useState<
+    ProtectedApplicationViewKind[]
+  >([]);
+
+  const onPaginatedDataChange = React.useCallback(
+    (paginatedData: K8sResourceCommon[]) => {
+      setPagePavs(paginatedData as ProtectedApplicationViewKind[]);
+    },
+    []
+  );
```
SanjalKatiyar: unnecessary useCallback...
```suggestion
```

SanjalKatiyar: use `setPagePavs` directly anywhere needed...

packages/mco/components/protected-applications/list-page.tsx
```diff
@@ -177,12 +236,53 @@ export const ProtectedApplicationsListPage: React.FC = () => {
...
+
+  const rowSelectProps: RowSelectProps = React.useMemo(
+    () => ({
+      onRowSelect: selection.onRowSelect,
+      isSelected: selection.isSelected,
+      isDisabled: selection.isDisabled,
+    }),
+    [selection.onRowSelect, selection.isSelected, selection.isDisabled]
+  );
```
SanjalKatiyar: nits: useMemo not needed here...

packages/mco/components/protected-applications/use-selection.ts
```diff
@@ -0,0 +1,176 @@
...
+    onSelectAllPage,
+    selectedCount: selectedUids.size,
+    eligiblePageCount: eligiblePage.length,
+    eligibleTotalCount: eligibleAll.length,
+    isPartiallySelected,
+    onSelectNone,
+    onSelectPage,
+    onSelectAll,
+    getSelectedPavs,
+  };
```
SanjalKatiyar, line 168: we are returning lot of things here but not using them on the list page... will those be needed in a followup PR ??

else, let's have only what's needed for now... it's not a `shared` hook...

SanjalKatiyar (changes requested):

packages/shared/src/list-page/paginated-list-page.tsx
vbnrh: image
image

This is what it looks like with grid on. It is a pre existing issue which was worsened by adding the bulk failover button as well.

Since the child items were not in the same grid container together, it was producing lot of issues with aligment which is why i made the switch simpler flexbox so that these items wrap themselves automaticaly

packages/shared/src/list-page/paginated-list-page.tsx
vbnrh: All 3 other consumers of this shared component (S3 browser pages) were verifie.  they use hideFilter and don't pass toolbarActions or CreateButton, so the flex container renders identically to the old Grid for them.

packages/mco/components/protected-applications/list-page.tsx
vbnrh: done

packages/mco/components/protected-applications/list-page.tsx
vbnrh: done

packages/mco/components/protected-applications/list-page.tsx
vbnrh: Done, removed.

packages/mco/components/protected-applications/use-selection.ts
vbnrh: Removed getSelectedPavs. it was the only unused return value. Everything else is consumed on the list page.

packages/mco/components/protected-applications/list-page.tsx
vbnrh: Inlined at the call site. passing (paginatedData) => setPagePavs(paginatedData as ProtectedApplicationViewKind[]) directly. Can't pass setPagePavs bare due to the type cast needed.

vbnrh: removed useCallback

PR #2952, opened 2026-07-07
Reduce s390x minimum CPU requirements to 12/18/24

malayparida2000:
## Summary
- Lowers IBM Z (s390x) resource profile minimum CPU requirements from 15/21/30 to 12/18/24 for Lean, Balanced, and Performance profiles
- Reflects reduced overhead from Noobaa and CSI pod resource optimizations
- Updates configure-performance unit tests to match the new values

Assisted by AI

Ref-https://redhat.atlassian.net/browse/RHSTOR-9418

### DIFF
--- a/packages/odf/constants/common.ts
+++ b/packages/odf/constants/common.ts
@@ -73,15 +73,15 @@ export const ARCHITECTURE_S390X = 's390x';
  */
 export const S390X_CPU_ADJUSTMENTS = {
   [ResourceProfile.Lean]: {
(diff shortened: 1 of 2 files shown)


packages/odf/constants/common.ts
SanjalKatiyar, line 100: what's the impact here ??

malayparida2000: None, this has not been touched. IBM Z team is not worried about NFS right now

PR #2959, opened 2026-07-13
Add Submariner/Globalnet pre-pair checks to Create DRPolicy

TimothyAsirJeyasing:
https://redhat.atlassian.net/browse/RHSTOR-9077

## Summary
Add pre-pair network validation to the existing Create DRPolicy form for the Data Foundation replication path.
Validate Submariner availability before allowing DRPolicy creation.
Detect both ACM-managed and upstream (subctl) Submariner installations during pre-pair validation. ACM-managed installations continue to use health validation, while upstream installations are detected and reported as an informational state because equivalent health signals are not currently available.
Block Create when ACM-managed Submariner is unhealthy or progressing, or when Submariner is detected on only one of the selected clusters.
Treat Globalnet validation as warning-only (never blocks Create).
Skip pre-pair validation for Third Party / Flash replication.

### User flow
```
Select 2 clusters
    → Existing ODF / cluster validation (unchanged)
    → Pre-pair network validation (Data Foundation only)
         → Submariner validation fails?   BLOCK Create
         → Globalnet issue?               WARN only
    → Policy fields / Create
```
Pre-pair validation runs only when:

Exactly two clusters are selected.
Existing cluster validation has passed.
Data Foundation replication is selected.

Behavior:
No Submariner detected on either cluster (for example, HCI) → Skip Submariner validation.
Upstream (subctl) Submariner detected → Show an informational alert and continue.
ACM-managed and upstream installations across the selected clusters → Informational only; continue.
Submariner detected on one selected cluster but missing on the other → Block Create.
Third Party / Flash → Skip the entire pre-pair validation.

Pre-pair runs only when: exactly 2 clusters + cluster selection valid + backend is **not** Third Party.  
Both Submariner addons missing (e.g. HCI) → skip Submariner gate. Third Party / Flash → skip entire pre-pair UI.

### File map
| File | Role |
|------|------|
| `utils/submariner-health.ts` | Pure logic: Submariner health, Globalnet enabled + required (CIDR overlap) |
| `utils/use-pre-pair-network-validation.ts` | Hook: hub watches → evaluation state |
| `pre-pair-network-validation.tsx` | Inline alerts; dispatches `isPrePairNetworkValid` |
| `CreateDRPolicyForm.tsx` / `reducer.ts` | Mount UI; gate fields/Create on `isPrePairNetworkValid` |
| `utils/map-pre-pair-prerequisites.ts` | Maps hook state → checklist items (for wizard PR) |

## Details
Validation sources:
- ManagedClusterAddOn/submariner for ACM-managed Submariner installations.
- ManagedClusterView fallback to read the spoke submariner.io/v1 Cluster resource for upstream (subctl) installations.
- Submariner Broker resources (spec.globalnetEnabled) for Globalnet validation.
- ManagedCluster network claims, with Submariner Cluster CIDRs as a fallback, to determine whether Globalnet is required.

Validation behavior:-
ACM-managed installations use ACM health conditions.
Upstream installations are detected and reported as informational because equivalent health signals are not currently available.
Globalnet is considered required only when Pod or Service CIDRs overlap.
If Globalnet is required but not enabled, a warning is displayed.
Globalnet validation never blocks Create in this PR.

## Out of scope
Advanced health validation for upstream (subctl) installations (Gateway/Endpoint-based validation).
Blocking Create when Globalnet is required but disabled (warning only in this PR).


packages/mco/components/create-dr-policy/utils/reducer.ts
```diff
@@ -118,6 +121,7 @@ export const drPolicyInitialState: DRPolicyState = {
   selectedClustersHaveODF: false,
   enableRBDImageFlatten: false,
   isClusterSelectionValid: false,
+  isPrePairNetworkValid: true,
```
weirdwiz: ideally should be false, so that the button is disabled when the page loads

packages/mco/components/create-dr-policy/utils/map-pre-pair-prerequisites.ts
weirdwiz, line 1: this file is not imported anywhere, just the tests, is it required?

packages/mco/components/create-dr-policy/pre-pair-network-validation.tsx
```diff
@@ -0,0 +1,194 @@
...
+                : AlertVariant.warning
+            }
+            isInline
+            customIcon={
+              globalnetStatus === GlobalnetCheckStatus.Enabled &&
+              !globalnetRequiredButNotEnabled ? (
+                <CheckCircleIcon />
+              ) : (
+                <ExclamationCircleIcon />
+              )
```
weirdwiz: if these are already mapped in the enum we should use that, currently this exists in map-pre-prerequisites as well has this file

packages/mco/components/create-dr-policy/utils/use-pre-pair-network-validation.ts
```diff
@@ -0,0 +1,199 @@
...
+    managedClusterA,
+    managedClusterALoaded,
+    managedClusterAError,
+    managedClusterB,
+    managedClusterBLoaded,
+    managedClusterBError,
+    submarinerClusters,
+    submarinerClustersLoaded,
+    submarinerClustersError,
+  ]);
```
weirdwiz: we can drop the usememo, also if we want to watch all of thes resources, i'd recommend we create a hook for now it's too verbose

weirdwiz (changes requested):

packages/mco/components/create-dr-policy/utils/use-pre-pair-network-validation.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  const [addonA, loadedA, errorA] = useK8sWatchResource<SubmarinerAddOnKind>(
+    shouldWatch ? submarinerAddonResource(clusterA) : null
+  );
+  const [addonB, loadedB, errorB] = useK8sWatchResource<SubmarinerAddOnKind>(
+    shouldWatch ? submarinerAddonResource(clusterB) : null
+  );
+  const [brokers, brokersLoaded, brokersError] = useK8sWatchResource<
+    SubmarinerBrokerKind[]
+  >(shouldWatch ? brokerListResource : null);
```
weirdwiz: if it's a namespaced resource, not sure if we pass it no namespace will it always return empty?

packages/mco/components/create-dr-policy/utils/submariner-health.spec.ts
```diff
@@ -0,0 +1,369 @@
...
+  it('returns skipped when addon is not found', () => {
+    expect(
+      getSubmarinerClusterHealth(undefined, true, {
+        response: { status: 404 },
+        message: 'NotFound',
+      })
+    ).toBe(SubmarinerClusterHealth.Skipped);
+  });
+
+  it('returns healthy for Aswin sample conditions (SubMariner spelling)', () => {
```
weirdwiz: looks like some testing strings slipped in

packages/mco/components/create-dr-policy/utils/reducer.ts
TimothyAsirJeyasing: Agreed - I will update

packages/mco/components/create-dr-policy/utils/map-pre-pair-prerequisites.ts
TimothyAsirJeyasing: it was for the upcoming Configure checklist and wasn’t used by the current form. will remove from this PR; will add it with the wizard work.

packages/mco/components/create-dr-policy/utils/submariner-health.ts
```diff
@@ -0,0 +1,534 @@
...
+  const ipInt = ipv4ToInt(ip);
+  if (ipInt === null || Number.isNaN(prefix) || prefix < 0 || prefix > 32) {
+    return null;
+  }
+  const blockSize = 2 ** (32 - prefix);
+  const start = Math.floor(ipInt / blockSize) * blockSize;
+  const end = start + blockSize - 1;
+  return { start, end };
+};
+
```
weirdwiz: use `packages/odf/components/utils/cidr-utils.ts`

packages/mco/components/create-dr-policy/pre-pair-network-validation.tsx
TimothyAsirJeyasing: Agree. The status→title mapping was duplicated between the form alerts and the mapper. Enums alone didn’t remove that duplication.

Will fix by:
Removing the unused mapper from this PR (it wasn’t wired into the form).
Centralizing alert title keys in submariner-health.ts via getSubmarinerAlertTitleKey / getGlobalnetAlertTitleKey, and using those from the UI.
The Configure checklist mapper can reuse the same helpers when we add that in a follow-up.

packages/mco/components/create-dr-policy/utils/use-pre-pair-network-validation.ts
TimothyAsirJeyasing: Good catch, will add ns submariner-k8s-broker

packages/mco/components/create-dr-policy/utils/submariner-health.ts
```diff
@@ -0,0 +1,534 @@
...
+      ...asStringArray(parsed.podCIDR),
+      ...asStringArray(parsed.podCidr),
+    ];
+    const serviceCidrs = [
+      ...asStringArray(parsed.serviceNetwork),
+      ...asStringArray(parsed.ServiceNetwork),
+      ...asStringArray(parsed.service_cidr),
+      ...asStringArray(parsed.serviceCidr),
+      ...asStringArray(parsed.serviceCIDR),
+    ];
```
weirdwiz: we should parse what the api defines, this parsing is too defensive
we should refer to the api and parse accordingly imo

packages/mco/components/create-dr-policy/utils/submariner-health.ts
```diff
@@ -0,0 +1,534 @@
...
+  'network.openshift.io',
+  'network.cluster.open-cluster-management.io',
+] as const;
+
+export type ManagedClusterAddOnCondition = {
+  type: string;
+  status: string;
+  message?: string;
+};
+
```
weirdwiz: use `K8sResourceCondition`

SanjalKatiyar:
plz test and add screenshots...

packages/mco/components/create-dr-policy/utils/submariner-health.ts
```diff
@@ -0,0 +1,513 @@
...
+  isConditionStatus,
+} from '@odf/shared/selectors';
+import { K8sResourceCondition } from '@odf/shared/types';
+import { doCidrListsOverlap, isNotFoundError } from '@odf/shared/utils';
+import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
+
+export const SUBMARINER_ADDON_NAME = 'submariner';
+
+export const SUBMARINER_MIXED_INSTALL_MESSAGE =
+  'Submariner must be installed on both selected clusters for this deployment.';
```
SanjalKatiyar: externalise...

packages/mco/components/create-dr-policy/utils/submariner-health.ts
```diff
@@ -0,0 +1,513 @@
...
+  Checking = 'checking',
+  Unknown = 'unknown',
+  Required = 'required',
+  NotRequired = 'notRequired',
+}
+
+export const NETWORK_CLUSTER_CLAIM_NAMES = [
+  'network.openshift.io',
+  'network.cluster.open-cluster-management.io',
+] as const;
```
SanjalKatiyar: move to constants under mco... separate file for submariner...

packages/mco/components/create-dr-policy/utils/submariner-health.ts
```diff
@@ -0,0 +1,513 @@
...
+  clusterName: string;
+  clusterClaims?: ClusterClaim[];
+  loaded: boolean;
+  loadError?: unknown;
+};
+
+export type ClusterNetworkCidrs = {
+  clusterCidrs: string[];
+  serviceCidrs: string[];
+};
```
SanjalKatiyar: move to types under mco... separate file for submariner...

packages/mco/utils/submariner-health.ts
```diff
@@ -0,0 +1,513 @@
...
+export const getSubmarinerClusterHealth = (
+  addon: SubmarinerAddOnKind | undefined,
+  loaded: boolean,
+  loadError: unknown
+): SubmarinerClusterHealth => {
+  if (!loaded) {
+    return SubmarinerClusterHealth.Checking;
+  }
+
+  if (loadError && isNotFoundError(loadError)) {
```
SanjalKatiyar, line 41: how come "not found" is classified as "skipped" ??

skipping means u r not checking the health for whatever reason... not found is completely unrelated...

packages/mco/components/create-dr-policy/utils/submariner-health.ts
```diff
@@ -0,0 +1,513 @@
...
+    return SubmarinerClusterHealth.Degraded;
+  }
+
+  if (!agentDegraded || !connectionDegraded) {
+    return SubmarinerClusterHealth.Progressing;
+  }
+
+  if (
+    agentDegraded.status &&
+    connectionDegraded.status &&
```
SanjalKatiyar: how come just checking if "status" exists can guarantee that resource is "healthy" only ??

packages/mco/components/create-dr-policy/utils/submariner-health.ts
```diff
@@ -0,0 +1,513 @@
...
+      isConditionStatus(condition, 'True')
+  );
+  return degradedCondition?.message;
+};
+
+export type ClusterSubmarinerStatus = {
+  clusterName: string;
+  health: SubmarinerClusterHealth;
+  message?: string;
+};
```
SanjalKatiyar: under types...

packages/mco/components/create-dr-policy/utils/submariner-health.ts
```diff
@@ -0,0 +1,513 @@
...
+  if (!loaded) {
+    return GlobalnetCheckStatus.Checking;
+  }
+
+  if (loadError || !brokers?.length) {
+    return GlobalnetCheckStatus.NotFound;
+  }
+
+  const broker =
+    brokers.find((item) => getName(item) === 'submariner-broker') ?? brokers[0];
```
SanjalKatiyar: import from constants...

packages/mco/components/create-dr-policy/utils/submariner-health.ts
```diff
@@ -0,0 +1,513 @@
...
+      return 'Cluster network (Submariner): Not healthy';
+  }
+};
+
+export const getGlobalnetAlertTitleKey = (
+  status: GlobalnetCheckStatus,
+  requirement: GlobalnetRequirement
+): string => {
+  if (isGlobalnetRequiredButNotEnabled(status, requirement)) {
+    return 'Globalnet is required because the selected clusters have overlapping network CIDRs, but it is not enabled.';
```
SanjalKatiyar: externalise...

packages/mco/components/create-dr-policy/utils/submariner-health.ts
```diff
@@ -0,0 +1,513 @@
...
+  switch (status) {
+    case GlobalnetCheckStatus.Enabled:
+      return requirement === GlobalnetRequirement.Required
+        ? 'Globalnet: Enabled (required for overlapping cluster networks)'
+        : 'Globalnet: Enabled';
+    case GlobalnetCheckStatus.Disabled:
+      return 'Globalnet: Not enabled';
+    case GlobalnetCheckStatus.NotFound:
+    default:
+      return 'Globalnet: Broker configuration not found';
```
SanjalKatiyar: externalise...

packages/mco/components/create-dr-policy/utils/use-pre-pair-network-validation.ts
```diff
@@ -0,0 +1,210 @@
...
+  SubmarinerClusterHealth,
+  SubmarinerClusterKind,
+} from './submariner-health';
+
+const SUBMARINER_ADDON_KIND =
+  'addon.open-cluster-management.io~v1alpha1~ManagedClusterAddOn';
+const SUBMARINER_BROKER_KIND = 'submariner.io~v1alpha1~Broker';
+const SUBMARINER_CLUSTER_KIND = 'submariner.io~v1~Cluster';
+// Default ACM Submariner broker namespace (namespaced CR; list needs a namespace).
+const SUBMARINER_BROKER_NAMESPACE = 'submariner-k8s-broker';
```
SanjalKatiyar: move to constants...

packages/mco/components/create-dr-policy/utils/use-pre-pair-network-validation.ts
```diff
@@ -0,0 +1,210 @@
...
+};
+
+const submarinerClusterListResource = {
+  cluster: HUB_CLUSTER_NAME,
+  kind: SUBMARINER_CLUSTER_KIND,
+  isList: true,
+  namespace: SUBMARINER_BROKER_NAMESPACE,
+  namespaced: true,
+  optional: true,
+};
```
SanjalKatiyar: there should be a common place where we keep resources...

packages/mco/components/create-dr-policy/utils/use-pre-pair-network-validation.ts
```diff
@@ -0,0 +1,210 @@
...
+  loaded: true,
+  loadError: null,
+  canProceed: true,
+  submarinerOverallHealth: SubmarinerClusterHealth.Skipped,
+  globalnetStatus: GlobalnetCheckStatus.Skipped,
+  globalnetRequirement: GlobalnetRequirement.Skipped,
+  clusterStatuses: [],
+};
+
+const usePrePairNetworkResources = (
```
SanjalKatiyar: there should be a common place for hooks under mco... move it there...

packages/mco/components/create-dr-policy/utils/use-pre-pair-network-validation.ts
```diff
@@ -0,0 +1,210 @@
...
+        ? getManagedClusterResourceObj({ name: clusterB })
+        : null
+    );
+  const [
+    submarinerClusters,
+    submarinerClustersLoaded,
+    submarinerClustersError,
+  ] = useK8sWatchResource<SubmarinerClusterKind[]>(
+    shouldWatch ? submarinerClusterListResource : null
+  );
```
SanjalKatiyar: use `getValidWatchK8sResourceObj`... plz always refer existing implementation and use wrappers, instead of custom...

packages/mco/components/create-dr-policy/utils/use-pre-pair-network-validation.ts
```diff
@@ -0,0 +1,210 @@
...
+  };
+};
+
+export const usePrePairNetworkValidation = (
+  clusterNames: string[],
+  enabled: boolean
+): PrePairNetworkValidationState => {
+  const memoizedClusterNames = useDeepCompareMemoize(clusterNames);
+  const shouldWatch = enabled && memoizedClusterNames.length === 2;
+  const [clusterA, clusterB] = memoizedClusterNames;
```
SanjalKatiyar: what purpose is using useDeepCompareMemoize even serving here ?????

packages/mco/components/create-dr-policy/pre-pair-network-validation.tsx
```diff
@@ -0,0 +1,177 @@
...
+  const { t } = useCustomTranslation();
+  const clusterNames = React.useMemo(
+    () => selectedClusters.map(getName),
+    [selectedClusters]
+  );
+  const enabled = shouldRunPrePairValidation(
+    selectedClusters,
+    isClusterSelectionValid,
+    replicationBackend
+  );
```
SanjalKatiyar: what is the current UX ??

since selecting a cluster pair sends multiple API calls now, we shouldn't trigger them immediately on every checkbox change... users may quickly select and unselect clusters, which would result in unnecessary API calls, repeatedly closing and recreating watches.

either user need to confirm that they want to validate the current cluster selection before sending any API requests and while validation is in progress, disable the clusters table and provide a "Cancel" option...

or, add debouncing, everything remains same and user will see a loading state on UI immediately, but actual API call should happen with a 1-2 sec delay...

cc @vbnrh FYI.

packages/mco/components/create-dr-policy/pre-pair-network-validation.tsx
```diff
@@ -0,0 +1,177 @@
...
+import { usePrePairNetworkValidation } from './utils/use-pre-pair-network-validation';
+
+const shouldRunPrePairValidation = (
+  selectedClusters: ManagedClusterInfoType[],
+  isClusterSelectionValid: boolean,
+  replicationBackend: BackendType
+): boolean =>
+  selectedClusters.length === 2 &&
+  isClusterSelectionValid &&
+  replicationBackend !== BackendType.ThirdParty;
```
SanjalKatiyar: this is not sufficient...

SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/2959/changes#r3578461899

packages/mco/components/create-dr-policy/pre-pair-network-validation.tsx
```diff
@@ -0,0 +1,177 @@
...
+  selectedClusters,
+  isClusterSelectionValid,
+  replicationBackend,
+  dispatch,
+}) => {
+  const { t } = useCustomTranslation();
+  const clusterNames = React.useMemo(
+    () => selectedClusters.map(getName),
+    [selectedClusters]
+  );
```
SanjalKatiyar: why do u need useMemo here ??

packages/mco/components/create-dr-policy/pre-pair-network-validation.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+  const validation = usePrePairNetworkValidation(clusterNames, enabled);
+
+  React.useEffect(() => {
+    dispatch({
+      type: DRPolicyActionType.SET_PRE_PAIR_NETWORK_VALIDATION,
+      payload: enabled ? validation.canProceed : true,
+    });
+  }, [enabled, validation.canProceed, dispatch]);
```
SanjalKatiyar: why we need extra state and useEffect to set that state when `shouldRunPrePairValidation` can be called else where itself ??

why extra hooks and states ??

why not calculate `shouldRunPrePairValidation` early and pass that as a prop to this hook ??

packages/mco/components/create-dr-policy/pre-pair-network-validation.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+  const submarinerFailureMessage = React.useMemo(
+    () => getSubmarinerPrePairMessage(validation.clusterStatuses),
+    [validation.clusterStatuses]
+  );
```
SanjalKatiyar: why u need useMemo here ??

packages/odf/components/utils/cidr-utils.ts
```diff
@@ -1,54 +1,14 @@
...
-export const isValidCIDRFormat = (value: string): boolean => {
-  if (!value?.trim()) return false;
-  const trimmed = value.trim();
-  const cidrRegex = /^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\/\d{1,2}$/;
-  if (!cidrRegex.test(trimmed)) return false;
-  const [base, prefixStr] = trimmed.split('/');
-  const mask = parseInt(prefixStr, 10);
-  if (isNaN(mask) || mask < 0 || mask > 32) return false;
-  return ipToNumber(base) !== null;
-};
```
SanjalKatiyar, line 51: do these already exists ??

SanjalKatiyar (changes requested):

packages/mco/utils/submariner-health.ts
TimothyAsirJeyasing: right, the current name can be confusing.
In this case, "Not found" means the Submariner addon is not installed on that managed cluster. If it is not found on both clusters, we currently treat it as Submariner is not being used, so the Create action is allowed.

I'll update the status name to something clearer, such as NotInstalled, so it better reflects the actual meaning.

packages/mco/components/create-dr-policy/utils/submariner-health.ts
TimothyAsirJeyasing: those lines only check that status is present, not that the addon is healthy. Healthy is already determined by Available=True and the degraded conditions being False (and we already return Progressing if those conditions are missing). I’ll remove the bare .status checks.

packages/odf/components/utils/cidr-utils.ts
TimothyAsirJeyasing: Yes, that's correct. I moved the common IPv4/CIDR helper functions to @odf/shared/utils so both MCO and ODF can use the same implementation instead of maintaining duplicate code. The ODF cidr-utils now re-exports those shared helpers, while getMonIp remains ODF-specific since it's only used there.

packages/mco/components/create-dr-policy/pre-pair-network-validation.tsx
TimothyAsirJeyasing: agree, I'll move shouldRunPrePairValidation and usePrePairNetworkValidation up to CreateDRPolicyForm so the parent owns the validation result, and simplify the flow by avoiding the extra state synchronization. The child will just render the validation status.

TimothyAsirJeyasing:
image

image

https://github.com/user-attachments/assets/eeb183d5-c32a-4411-b77d-e5ba46e8093d

TimothyAsirJeyasing:
Closing this PR due to a recent design change in the Create DRPolicy pre-pair validation flow.

We are refining how ACM-managed vs upstream Submariner is presented, how Create gating works, and how Globalnet warnings are handled. I will open a new PR with the updated design and carry forward the review feedback from this one.

Thank you for the reviews.

PR #2984, opened 2026-07-21
CVEs fixes to master

Soundarya189r:
## Description

Based on the  **CVE-2026-59869** upgraded  "js-yaml": "^4.3.0"

https://redhat.atlassian.net/browse/DFBUGS-8827 

---

## Change Type

- [ ] Feature
- [ ✅] Bug Fix
- [ ] Improvement
- [ ] Refactor
- [ ] Tests

---

## Component / Area Impacted

- [✅ ] ODF
- [ ] FDF
- [ ] Client
- [ ] MCO
- [ ] CNSA (Remote Mount)
- [ ] E2E Test

---

### Recordings / Demo Videos

https://github.com/user-attachments/assets/a9a344e9-0d93-4429-bf7d-735c94621b78

---

## Testing

- [ ] Unit Tests
- [ ] E2E Tests
- [ ✅] No Tests Required (with justification below)


package.json
```diff
@@ -73,16 +73,16 @@
     "ajv-formats/ajv": "^8.20.0",
     "ajv-keywords": "^5.1.0",
     "ajv-keywords/ajv": "^8.20.0",
-    "brace-expansion": "^1.1.13",
+    "brace-expansion@npm:^1.1.7": "1.1.16",
```
NIKHITHAVADDEMPUDI: @Soundarya189r  please avoid hardcoding the versions on left #2907 (comment)

try to use single line overrides (except in cases of errors, or when global override fail to replace all versions correctly)

Soundarya189r: okay got it, I have updated the resolutions, please check

SanjalKatiyar: @Soundarya189r did you check the dependency tree ?? Does this actually work ??

I remember suggesting something similar to @NIKHITHAVADDEMPUDI in the PR mentioned above, but IIRC the response was that combining all the entries into a single one doesn't work with yarn v4.

Soundarya189r: image

NIKHITHAVADDEMPUDI: > @Soundarya189r did you check the dependency tree ?? Does this actually work ??
> 
> I remember suggesting something similar to @NIKHITHAVADDEMPUDI in the PR mentioned above, but IIRC the response was that combining all the entries into a single one doesn't work with yarn v4.

Hi, @SanjalKatiyar previously ajv caused yarn build errors on using single global override for all versions, so i used multiple path overrides, 
using single global override along with local path overrides is not working to replace tree with globally one version and another version at the particular path , in that case I used only local path overrides.

but in brace-expansion case , it is not throwing errors for single global overrides for all versions

PR #3027, opened 2026-07-29
[MCG] IAM - Bug Fixes

sowmithri2001:
## Description
Jira link : https://redhat.atlassian.net/browse/DFBUGS-9022

---

## Change Type

- [ ] Bug Fix

---

## Component / Area Impacted

- [ ] ODF

---

### Recordings / Demo Videos

https://github.com/user-attachments/assets/6d12d990-7665-4e25-b270-7dff983885af

https://github.com/user-attachments/assets/0015525d-50de-4913-ac32-baa141d8f236

https://github.com/user-attachments/assets/7079cc91-8b4c-43df-a88c-62d9ea19fbeb

https://github.com/user-attachments/assets/d8f5abb8-5620-4196-9892-03a96acd952a

https://github.com/user-attachments/assets/85b2046a-728d-4f0e-a7e3-5a587c9c63ab

---

## Testing

- [ ] Unit Tests
- [ ] E2E Tests

---

## Summary by CodeRabbit

* **New Features**
  * Added version-aware IAM documentation links to the empty user page.
  * Improved access key secret visibility and copying with accessible controls.

* **Bug Fixes**
  * Improved IAM tag handling when creating and deleting users and access keys.
  * Added validation for the maximum number of user tags, including descriptions.
  * Updated the users table layout for clearer column sizing.

* **Accessibility**
  * Added accessible labels to secret-key visibility and copy controls.


packages/odf/modals/s3-iam/AccessKeySecretKeyDisplayModal.tsx
```diff
@@ -116,31 +116,35 @@ export const AccessKeySecretKeyDisplayModal: React.FC<
...
-              <FlexItem>
+            <InputGroup>
+              <InputGroupItem>
                 <TextInput
                   readOnly
                   readOnlyVariant="plain"
-                  value={!hide ? SecretKey : '*'.repeat(SecretKey.length)}
+                  type={hide ? 'password' : 'text'}
+                  value={SecretKey}
+                  style={{ width: `${Math.max(SecretKey.length, 1)}ch` }}
```
Soundarya189r: No hardcoded sizing, try using `InputGroupItem isFill`or  SCSS with PF tokens

packages/odf/components/s3-iam/create-user/AddKeyValuePair.tsx
Soundarya189r, line 95: I think Add tag is not disabled when you hit 50. 
We might have to include `isAddDisabled={remainingTags <= 0}`
@NIKHITHAVADDEMPUDI  what do you think ?

NIKHITHAVADDEMPUDI: yes i agree, Add tag is not getting disabled ,isAddDisabled={remainingTags <= 0} will fix it

sowmithri2001: @Soundarya189r @NIKHITHAVADDEMPUDI  added isAddDisabled={remainingTags <= 0} so that Add tag is disabled when you hit 50

packages/odf/modals/s3-iam/AccessKeySecretKeyDisplayModal.tsx
sowmithri2001: Removed the ch width, The secret field now uses InputGroupItem isFill

packages/shared/src/constants/doc.ts
```diff
@@ -48,3 +48,7 @@ export const tnfHomePage = (odfDocVersion: string): string =>
 
 export const scalingStorageDoc = (odfDocVersion: string): string =>
   `${odfDocBasePath(odfDocVersion)}/scaling_storage/index`;
+export const iamUserDoc = (odfDocVersion: string): string =>
+  `${odfDocBasePath(
+    odfDocVersion
+  )}/managing_hybrid_and_multicloud_resources/index#creating_and_managing_iam_user`;
```
Harshvardhan-handa123, line 54: Has the doc team confirmed whether "Learn more about IAM" should point to this page,  I can see in the video it points to managing hybrid and multicloud resources.

packages/odf/components/s3-iam/create-user/AddKeyValuePair.tsx
```diff
@@ -61,7 +63,9 @@ export const AddKeyValuePairs: React.FC<AddKeyValuePairsProps> = ({
       >
         <Content component="p" className="pf-v6-u-mb-sm">
           <BlueInfoCircleIcon className="pf-v6-u-mr-sm" />
-          {t(`You can add ${remainingTags} more tags`)}
+          {t('You can add {{remaining}} more tags', {
+            remaining: remainingTags,
```
Harshvardhan-handa123: No need to change this part

packages/odf/components/s3-iam/create-user/CreateUserForm.tsx
```diff
@@ -272,7 +276,11 @@ export const CreateUserFormContent = () => {
               }}
             />
 
-            <AddKeyValuePairs pairs={pairs} setPairs={setPairs} />
+            <AddKeyValuePairs
+              pairs={pairs}
+              setPairs={setPairs}
+              descriptionTagValue={descriptionTagValue}
```
Harshvardhan-handa123, line 294: Current procedure is user enters "abc@#$%" -> descriptionTagValue state is updated -> No validation is performed -> User clicks "Create" -> save() is called -> getValueValidations(descriptionTagValue.trim(), t) -> If invalid → setTagErrors → form submission stops. 

Is live validation a better approach for form entries where we use Yup validations.
For reference, We use yup validation in the above part only for the UserName in createIAMUser.

Harshvardhan-handa123 (changes requested):
Also, in the create user page I feel the gap b/w the create user header and Alert info component is bit large. Hence, the gap b/w the AlertInfo and UserName becomes minimal.

packages/shared/src/constants/doc.ts
sowmithri2001: https://docs.redhat.com/en/documentation/red_hat_openshift_data_foundation/4.22/html-single/managing_hybrid_and_multicloud_resources/index#creating_and_managing_iam_user

This is the link given for "Learn more about IAM" but odfDocBasePath uses access.redhat.com, which 301-redirects to docs.redhat.com. That redirect drops #creating_and_managing_iam_user, so we land on the top of the page instead of the IAM section.

packages/odf/components/s3-iam/create-user/CreateUserForm.tsx
sowmithri2001: Both approaches are fine with me.

The current approach validates on submit, which keeps the implementation simple and prevents form submission if the input is invalid. Live validation, on the other hand, provides immediate feedback to the user while they're typing, resulting in a smoother user experience.

packages/shared/src/constants/doc.ts
Soundarya189r: Please check if there is any url mismatch, #creating_and_managing_iam_user is relevant than the broader managing hybrid and multicloud resources

packages/odf/components/s3-iam/create-user/CreateUserForm.tsx
Soundarya189r: Description tag is optional, so I don't think we need live validations
@NIKHITHAVADDEMPUDI  please let us know your thoughts on this

packages/shared/src/constants/doc.ts
sowmithri2001: https://github.com/user-attachments/assets/2b104b2d-218d-4fa9-94d5-ec9b210f01c5

It's working fine now @Harshvardhan-handa123 @Soundarya189r

Harshvardhan-handa123: But I can still see clicking the link drops you to Managing hybrid and multicloud resources and then user scrolls to #creating_and_managing_iam_user section.

Harshvardhan-handa123: I think you should create a new base path in packages/shared/src/constants/doc.ts and use that. Like this -> 
export const odfDocsRedHatBasePath = (odfDocVersion: string): string =>
  `https://docs.redhat.com/en/documentation/red_hat_openshift_data_foundation/${odfDocVersion}/html-single`;

Harshvardhan-handa123: Since we do not have any base path for https://docs.redhat.com, it should be created.

sowmithri2001: No, the user is not manually scrolling to the `#creating_and_managing_iam_user`  section.

What's happening is that the link points to the -Managing hybrid and multicloud resources documentation page, and because it includes the `#creating_and_managing_iam_user` anchor, the browser automatically navigates to that specific section on the page.


https://docs.redhat.com/en/documentation/red_hat_openshift_data_foundation/4.22/html-single/managing_hybrid_and_multicloud_resources/index#creating_and_managing_iam_user

Harshvardhan-handa123: Then it's fine!

SanjalKatiyar:
@coderabbitai review

packages/odf/components/s3-iam/create-user/CreateUserForm.tsx
```diff
@@ -95,14 +95,18 @@ export const CreateUserFormContent = () => {
...
+          Key: accessKeyResponse.AccessKey.AccessKeyId,
+          Value: descriptionTagValueArg.trim(),
+        });
+      }
+      if (allTags.length > 0) {
+        await iamClient.tagUser({
+          UserName: userNameArg,
+          Tags: allTags,
+        });
+      }
```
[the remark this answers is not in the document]
sowmithri2001: Good catch. Agree — I’ll enforce the combined limit (pairs + description) before createIamUser so we don’t leave a half-created user if tagUser fails at 51.
Will also align the remaining-tags UI (Math.max(0, …) + isAddDisabled) with that same capacity, and add a boundary test for 50 pairs + description.

packages/odf/modals/s3-iam/DeleteIamUserModal.tsx
```diff
@@ -168,6 +168,17 @@ export const DeleteIamUserModal: React.FC<
...
+      const tagKeys = accessKeys
+        .map((key) => key.AccessKeyId)
+        .filter(Boolean);
+      if (tagKeys.length > 0) {
+        await iamClient.untagUser({
+          UserName: userName,
+          TagKeys: tagKeys,
+        });
+      }
+
```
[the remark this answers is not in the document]
sowmithri2001: 1. Atomic intent matches the UX
“Delete access keys” in the Delete user modal is a single prerequisite step before user deletion. Treating the batch as all-or-nothing (Promise.all) is intentional: either keys are cleared enough to proceed, or we surface an error and the user retries. We don’t want a half-success state that looks like “done” while some keys remain.

2. Orphan description tags are low-impact here
Description tags are keyed by AccessKeyId and are only used for display on access-key cards. If a key was deleted but its tag remains, that tag is orphaned metadata on the user — it does not restore the key, block user deletion, or affect auth. When the user is deleted next, those tags go away with the user.

3. Retry path is still safe enough
A failed deleteAccessKey on an already-deleted key is typically a not-found / idempotent failure; the user can click “Delete access keys” again. Stale SWR data doesn’t leave the cluster in an unsafe auth state; worst case is a noisy retry until the list refreshes or the user closes/reopens the modal.

packages/odf/components/s3-iam/create-user/CreateUserForm.tsx
Harshvardhan-handa123: LGTM

PR #892, opened 2023-06-06
Manage data policy list view

GowthamShanmugam:
1. Find all assigned policies to the app and display them in a list view.
2. A dummy action is added for bulk policy un assign the policy.
3. A dummy action is added for individual policy-wise un assign the policy.
4. A dummy action is added for to view policy configuration info
5. Dummy primary action button to switch the assign policy view.

PR #3048, opened 2026-08-10
Switch CI image runtime from http-server to nginx

vbnrh:
## Description

The odf-operator now configures TLS settings (`ssl_protocols`, `ssl_ciphers`, `ssl_conf_command`) via nginx directives in a ConfigMap mounted into the odf-console pod. The current CI image uses Node.js `http-server`, which doesn't understand nginx config, causing CI e2e failures when the CSV Deployment spec tries to run `nginx`.

This PR replaces the Node.js `http-server` runtime stage in `Dockerfile.ci` with `ubi9/nginx-120`, matching the ocs-client-console prod image. The existing `default.conf` is baked into the image at `/opt/app-root/etc/nginx.d/default.conf` so it works standalone; the operator can still overlay via ConfigMap mount.

Unblocks odf-operator PR #892 (TLS profile support).

---

## Change Type

- [ ] Feature
- [ ] Bug Fix
- [x] Improvement
- [ ] Refactor
- [ ] Tests

---

## Component / Area Impacted

- [x] ODF
- [ ] FDF
- [x] Client
- [x] MCO
- [ ] CNSA (Remote Mount)
- [ ] E2E Test

---

## Screenshots / Recordings

N/A — infrastructure change, no UI impact.

---

## Testing

- [ ] Unit Tests
- [ ] E2E Tests
- [x] No Tests Required (with justification below)

This is a Dockerfile-only change. Verification:

```bash
docker build -t odf-console:test -f Dockerfile.ci . --build-arg OPENSHIFT_CI=false
docker run --rm odf-console:test which nginx
docker run --rm odf-console:test nginx -v
docker run --rm odf-console:test ls /opt/app-root/src/
docker run --rm odf-console:test cat /opt/app-root/etc/nginx.d/default.conf


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added production container builds for selected plugin distributions.
  * Published both tagged and latest container images through the release workflow.

* **Refactor**
  * Updated application containers to use NGINX for serving built assets.
  * Improved startup and deployment compatibility across supported environments.
  * Removed the previous Node-based static server and script-based startup process.
  * Application functionality remains unchanged for end users.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/.github/workflows/quay-image.yml
+++ b/.github/workflows/quay-image.yml
@@ -16,10 +16,10 @@ jobs:
         id: vars
         run: echo ::set-output name=tag::$(echo ${GITHUB_SHA:0:8})
       - name: Build the tagged image
-        run: docker build . --file Dockerfile.ci --build-arg OPENSHIFT_CI=false --tag quay.io/ocs-dev/odf-console:${{steps.vars.outputs.tag}}
+        run: docker build . --file Dockerfile.prod --build-arg OPENSHIFT_CI=false --tag quay.io/ocs-dev/odf-console:${{steps.vars.outputs.tag}}
       - name: Push the tagged image
         run: docker push quay.io/ocs-dev/odf-console:${{steps.vars.outputs.tag}}
       - name: Build the latest image
-        run: docker build . --file Dockerfile.ci --build-arg OPENSHIFT_CI=false --tag quay.io/ocs-dev/odf-console:latest
+        run: docker build . --file Dockerfile.prod --build-arg OPENSHIFT_CI=false --tag quay.io/ocs-dev/odf-console:latest
       - name: Push the latest image to ocs-dev
         run: docker push quay.io/ocs-dev/odf-console:latest
--- a/Dockerfile.prod
+++ b/Dockerfile.prod
@@ -23,4 +23,5 @@ FROM registry.access.redhat.com/ubi9/nginx-120
 
 COPY --from=builder /dist /opt/app-root/src
 COPY --from=builder /dist /opt/app-root/src/compatibility
+RUN ln -sf /opt/app-root/etc/nginx.d/nginx.conf /etc/nginx/nginx.conf
 CMD ["/usr/libexec/s2i/run"]

### CODE REVIEW

File: Dockerfile.ci [hunk 1]
```diff
@@ -19,10 +19,10 @@ RUN BUILD_SCRIPT=build; \
...
+FROM registry.access.redhat.com/ubi9/nginx-120
 
-RUN npm install -g http-server
-COPY --from=builder /dist /app
-COPY --from=builder /dist /app/compatibility
-COPY --from=builder /app/http-server.sh .
-ENTRYPOINT [ "./http-server.sh", "./app", "-S", "true", "-C", "/var/serving-cert/tls.crt", "-K", "/var/serving-cert/tls.key" ]
+COPY --from=builder /dist /opt/app-root/src
+COPY --from=builder /dist /opt/app-root/src/compatibility
+COPY default.conf /opt/app-root/etc/nginx.d/default.conf
```
SanjalKatiyar: we don't need this, this should be picked from configmap

SanjalKatiyar: u need to add symlink instead... so that our custom config is picked up...

SanjalKatiyar: no harm in keeping it as well, but in that case plz add a comment explaining when this `default.conf` will be needed...

SanjalKatiyar:
we are making changes in odf-operator in 4.23
but CI will use ODF 4.22, also currently we are using ODF 4.19 in the CI... 

what is the impact, will this PR be compatible with all versions ??

Dockerfile.ci
SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/3048#issuecomment-5236984125

SanjalKatiyar: and it will only be helpful as a fallback, if added to `/etc/nginx/nginx.conf` and name should be `nginx.conf`, not `default.conf`

Dockerfile.ci
```diff
@@ -19,10 +19,11 @@ RUN BUILD_SCRIPT=build; \
...
+FROM registry.access.redhat.com/ubi9/nginx-120
 
-RUN npm install -g http-server
-COPY --from=builder /dist /app
-COPY --from=builder /dist /app/compatibility
-COPY --from=builder /app/http-server.sh .
-ENTRYPOINT [ "./http-server.sh", "./app", "-S", "true", "-C", "/var/serving-cert/tls.crt", "-K", "/var/serving-cert/tls.key" ]
+COPY --from=builder /dist /opt/app-root/src
+COPY --from=builder /dist /opt/app-root/src/compatibility
+# Fallback nginx config for local dev and CI; in production the operator overlays this path via ConfigMap.
```
SanjalKatiyar: check: https://github.com/red-hat-storage/odf-console/pull/3048#discussion_r3747375468

SanjalKatiyar: right now it's not even a fallback...

Dockerfile.ci
SanjalKatiyar: even for ` /opt/app-root/etc/nginx.d/`, name should be `nginx.conf`, not `default.conf`

vbnrh: will remove the baked-in COPY default.conf line. The operator always provides the full nginx.conf via ConfigMap, so adding in the partial server block at /opt/app-root/etc/nginx.d/ just creates a duplicate

vbnrh:
> we are making changes in odf-operator in 4.23 but CI will use ODF 4.22, also currently we are using ODF 4.19 in the CI...
> 
> what is the impact, will this PR be compatible with all versions ??

Checked the odf-operator release-4.19 and release-4.22 branches to verify compatibility.

The CSV Deployment spec for odf-console has no command or args, it just relies on whatever the image's ENTRYPOINT/CMD is. And the operator has actually been setting up everything nginx needs since 4.19:

ConfigMap odf-console-nginx-conf created by ensureConsolePlugin() in clusterversion_controller.go, contains the full nginx.conf from nginx_conf.go (worker processes, http block, server block with TLS on port 9001)
Volume mount at /etc/nginx/nginx.conf (subPath: nginx.conf) that overwrites the base image's default config
TLS secret at /var/serving-cert/
emptyDirs at /var/log/nginx and /var/lib/nginx/tmp for the readOnlyRootFilesystem: true constraint
The old http-server image was just ignoring all these nginx mounts the whole time. The new ubi9/nginx-120 image actually uses them. CMD runs /usr/libexec/s2i/run, starts nginx, reads the ConfigMap-provided /etc/nginx/nginx.conf, and it just works.

CI proof: This PR's e2e run ([build 2086712751798882304](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/red-hat-storage_odf-console/3048/pull-ci-red-hat-storage-odf-console-master-odf-console-e2e-aws/2086712751798882304)) ran against ODF 4.19. The odf-console pod (odf-console-577c5c7cd5-nxds4) started fine with nginx, served all plugin assets with 200s. The only test failure was an unrelated Cypress thing (persistent-dashboard-check.spec.ts, SVG visibility timeout).

Dockerfile.ci
vbnrh: The symlink point is also resolved by removing the baked-in config

vbnrh:
@coderabbitai review

Dockerfile.ci
```diff
@@ -19,10 +19,8 @@ RUN BUILD_SCRIPT=build; \
...
+FROM registry.access.redhat.com/ubi9/nginx-120
 
-RUN npm install -g http-server
-COPY --from=builder /dist /app
-COPY --from=builder /dist /app/compatibility
-COPY --from=builder /app/http-server.sh .
-ENTRYPOINT [ "./http-server.sh", "./app", "-S", "true", "-C", "/var/serving-cert/tls.crt", "-K", "/var/serving-cert/tls.key" ]
+COPY --from=builder /dist /opt/app-root/src
+COPY --from=builder /dist /opt/app-root/src/compatibility
+CMD ["/usr/libexec/s2i/run"]
```
SanjalKatiyar, line 26: Just FYI, once we update ODF to 4.23, we will need to add symlink here... till then, I think we are good !

Dockerfile.ci
SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/3048#discussion_r3749073084

Dockerfile.ci
SanjalKatiyar: better to create a GH issue for it...

SanjalKatiyar (approved):

PR #3006, opened 2026-07-27
RHSTOR-9047: Enable Day-2 encryption for local IBM Scale cluster

weirdwiz:
## Summary

- add singleton Day-2 encryption enablement from the local StorageCluster card
- share encryption resource creation between Day-1 and Day-2
- create the CA certificate, credentials, and EncryptionConfig in dependency order
- roll back created dependencies when setup fails
- enforce EncryptionConfig CRD field limits before submission

## Out of scope

- editing or disabling enabled encryption

Split from #2961.

image

PR #3050, opened 2026-08-11
Fix IBM Scale encryption configuration

weirdwiz:
## Summary

- align encryption identifiers and credentials with the IBM Scale contract
- make the CA certificate optional while preserving CA configuration support
- create or update singleton encryption resources for Day-2 retries
- keep form validation and submission fully typed
- show encryption state from resource presence while allowing failed configurations to retry
- delete the cluster-scoped local Scale cluster with the last remote connection so no stale SAN Storage entry remains

## Verification

- `yarn tsc --noEmit`
- targeted ESLint
- 105 focused Jest tests
- manual onboarding, metrics, Day-2, filesystem removal, last-connection offboarding, and restoration

Refs #3006

## Summary by CodeRabbit

* **New Features**
  * CA certificates are now optional when configuring IBM Scale CNSA encryption.
  * Added validation for Remote RKM, server, tenant ID, and resource names, including trimming and format checks.
  * Certificate contents are preserved as entered when submitted.
  * Local and SAN storage clusters can now be discovered without being restricted to a specific namespace.

* **Bug Fixes**
  * Improved encryption configuration submission and handling of configurations without certificates.
  * Corrected local cluster lookup and resource watch behavior.


weirdwiz:
ptal @bipuladh @SanjalKatiyar

packages/odf/components/scale-encryption/enableScaleEncryption.ts
```diff
@@ -31,53 +27,56 @@ export const enableScaleEncryption = async (
...
-      await k8sDelete({
-        model: SecretModel,
-        resource: createdSecret,
-        requestInit: null,
-        json: null,
-      }).catch(() => undefined);
-    }
-    if (createdConfigMap) {
-      await k8sDelete({
-        model: ConfigMapModel,
```
SanjalKatiyar, line 75: this deletion logic on incomplete create is not needed ??

packages/shared/src/text-inputs/password-input.tsx
```diff
@@ -105,25 +115,35 @@ const getVariant = (status: string) => {
...
-}) => {
+}: ValidatedPasswordInputProps<T, TName>) => {
   const { t } = useCustomTranslation();
   const [showPassword, setShowPassword] = React.useState(false);
   const {
     field: { name, value, onChange, onBlur },
     fieldState: { error, isDirty, isTouched },
     formState: { isSubmitted },
-  } = useController({
-    name: textInputProps.name || 'name',
```
SanjalKatiyar, line 126: is fallback not needed and everywhere we use this component we pass `name` prop ??

packages/shared/src/input-with-requirements/TextInputWithFieldRequirements.tsx
```diff
@@ -67,137 +77,143 @@ const getVariant = (status: string) => {
   }
 };
 
-const TextInputWithFieldRequirements: React.FC<
-  TextInputWithFieldRequirementsProps
-> = ({
-  fieldRequirements,
-  control,
-  formGroupProps,
-  textInputProps,
```
SanjalKatiyar, line 76: changes in this files are mostly type updates, right ??
was it required ??

packages/odf/components/scale-encryption/enableScaleEncryption.ts
weirdwiz: mostly not needed because we do createorupdate now instead of create, so collisions won't happen

packages/shared/src/input-with-requirements/TextInputWithFieldRequirements.tsx
weirdwiz: good catch i think this got left from the original commit, which had a larger refactor which i reverted

packages/shared/src/text-inputs/password-input.tsx
weirdwiz: yes, both the consumers pass name prop, but this change was also leftover
let me clean that up

packages/odf/components/scale-encryption/enableScaleEncryption.ts
SanjalKatiyar: I don't think this was about a collision. An API call can fail for any reason. So, if resource A gets created but the creation of resource B fails, IIUC, we were previously making sure to delete resource A as well, so that the cluster was reverted to its original state.
Now, whether we still need that rollback behaviour or not, I will leave that decision up to you and @bipuladh.

packages/odf/components/scale-encryption/enableScaleEncryption.ts
```diff
@@ -31,53 +27,56 @@ export const enableScaleEncryption = async (
 ): Promise<void> => {
   const configName = ENCRYPTION_CONFIG_NAME;
   const secretName = 'encryption-secret';
-  const createConfigMap = createConfigMapPayload(configName, {
```
bipuladh, line 34: Why are we changing the usage of this?

packages/odf/components/scale-encryption/enableScaleEncryption.ts
bipuladh: I think we can live with the new logic

packages/odf/components/scale-encryption/enableScaleEncryption.ts
weirdwiz: w.r.t to IBM docs, CA certificate is optional and it needs to be in plaintext not base64 encoded

https://www.ibm.com/docs/en/scalecontainernative/6.0.1?topic=resources-encryption

bipuladh: Can we update the function itself then?

PR #3071, opened 2026-08-22
DFBUGS-9581: CVE Fixes to release-4.14

sowmithri2001:
## Description

Jira Link :https://redhat.atlassian.net/browse/DFBUGS-9581
https://redhat.atlassian.net/browse/DFBUGS-9245
https://redhat.atlassian.net/browse/DFBUGS-9207
https://redhat.atlassian.net/browse/DFBUGS-9195
https://redhat.atlassian.net/browse/DFBUGS-9960
https://redhat.atlassian.net/browse/DFBUGS-9786

The following packages were added/updated in the resolutions block of package.json to address CVE vulnerabilities:

Tested locally ocp release-4.14 + odf release-4.14

Updated:
Brace expansion: ^1.1.18

Added:
immutable: ^4.3.9
tar: ^7.5.21
browserslist: 4.28.7

- [ ] Bug Fix

---

## Component / Area Impacted

- [ ] ODF

---

### Recordings / Demo Videos

https://github.com/user-attachments/assets/173ad5d0-cfc1-47f5-a3d5-1c1419f7b95d

## Testing

- [ ] Unit Tests
- [ ] E2E Tests

---


package.json
```diff
@@ -168,7 +168,7 @@
     "**/@typescript-eslint/typescript-estree/semver": "^7.7.1",
     "**/cypress/semver": "^7.7.1",
     "async": "^3.2.6",
-    "brace-expansion": "1.1.12",
+    "brace-expansion": "^5.0.8",
```
Harshvardhan-handa123: V1 to v5 is a major jump. Do we need any refactoring in the code for this?

package.json
```diff
@@ -190,7 +190,11 @@
     "ws": "^8.18.0",
     "serialize-javascript": "^6.0.2",
     "minimatch": "^3.0.5",
-    "underscore.string": "^3.3.6"
+    "underscore.string": "^3.3.6",
+    "@openshift-console/dynamic-plugin-sdk-internal/immutable": "^4.3.9",
+    "@openshift-console/dynamic-plugin-sdk/immutable": "^4.3.9",
```
Harshvardhan-handa123, line 196: Can we upgrade the dynamic-plugin-sdk versions in this branch?  Like: https://github.com/red-hat-storage/odf-console/blob/5f8dcdf3a5e7381b66507f1587cde77f1955536b/package.json#L106

Harshvardhan-handa123 (changes requested):

package.json
sowmithri2001: This is release-4.14, so we can’t bump @openshift-console/dynamic-plugin-sdk* to 4.22.0. That pin is for Console 4.22; 4.14 still needs 0.0.21 / 0.0.12 to stay compatible with OCP 4.14 (React 17 / PF4).

immutable is a transitive dependency of those SDK packages (and sass). The CVE is addressed by resolving it to 4.3.9, which is the same approach we used on 4.16.

Harshvardhan-handa123: Yeah, obv I didn’t mean to bump it to 4.22.0 directly, it wouldn’t be compatible with ODF 4.14. (I added a wrong reference prolly, my bad). I meant to check if we can upgrade the SDK deps to a higher compatible version instead of adding a specific resolution for dynamic-plugin-sdk/immutable.
The last time I fixed the CVEs, I added specific resolutions for dynamic-plugin-sdk/ajv. Bipul had suggested that we first check if we can upgrade the SDK packages to a higher compatible version before changing a specific resolution from the SDK (I can’t find the exact GitHub comment or Slack thread now :)). 
Something similar to: 
https://github.com/red-hat-storage/odf-console/commit/e4bcae5185958d87cf91aa76ace09652b4128880. 
https://github.com/red-hat-storage/odf-console/commit/e8181527da28cf69b2a97d9f3c40955e55cc5edd
https://github.com/red-hat-storage/odf-console/pull/2788

package.json
sowmithri2001: Yes, I changed the vulnerable version to the fixed version within V1 itself.

package.json
```diff
@@ -190,7 +191,12 @@
...
+    "immutable": "^4.3.9",
+    "tar": "7.5.21",
```
Soundarya189r:
```suggestion
    "tar": "^7.5.21",
```

package.json
sowmithri2001: Checked this — we can’t drop the resolution by bumping the SDK on 4.16.

We’re already on dynamic-plugin-sdk@1.3.0 / internal@1.0.0 / webpack@1.1.1. The latest 4.16-compatible core SDK is 1.4.0, and it still depends on immutable: "3.x" (same as 1.3.0). The 1.8.0 / 4.21.0 bumps from e818152 / e4bcae5 / #2788 are for newer Console versions and aren’t compatible here.

immutable is only a transitive dep of those SDK packages, so a resolution is still the way to get a patched version (CVE-2026-29063). I’ll drop the two nested @openshift-console/.../immutable keys though — the existing "immutable": "^4.3.9" resolution already covers them.

Soundarya189r: >  I’ll drop the two nested @openshift-console/.../immutable keys though — the existing "immutable": "^4.3.9" resolution already covers them.

does it mean you are going to drop ?
`"@openshift-console/dynamic-plugin-sdk-internal/immutable": "^4.3.9",`
 ` "@openshift-console/dynamic-plugin-sdk/immutable": "^4.3.9",`

Harshvardhan-handa123: please drop all the immutable resolutions. Just keep "immutable": "^4.3.9"

PR #3072, opened 2026-08-22
DFBUGS-9580: CVE Fixes to release-4.16

sowmithri2001:
## Description

jira Link: https://redhat.atlassian.net/browse/DFBUGS-9580
https://redhat.atlassian.net/browse/DFBUGS-9235
https://redhat.atlassian.net/browse/DFBUGS-9214
https://redhat.atlassian.net/browse/DFBUGS-9201
https://redhat.atlassian.net/browse/DFBUGS-9789

The following packages were added/updated in the resolutions block of package.json to address CVE vulnerabilities:

Tested locally ocp release-4.16 + odf release-4.16

Updated:
Brace expansion: ^1.1.8
"postcss": "^8.5.19"
Added: 
"immutable": "^4.3.9",
browserslist: 4.28.7

- [ ] Bug Fix

---

## Component / Area Impacted

- [ ] ODF

---

### Recordings / Demo Videos

https://github.com/user-attachments/assets/40eb9e6a-9a7d-4237-ac4a-a22eb53324ad

## Testing

- [ ] Unit Tests
- [ ] E2E Tests

---


package.json
```diff
@@ -176,7 +176,7 @@
     "**/@typescript-eslint/typescript-estree/semver": "^7.7.1",
     "**/cypress/semver": "^7.7.1",
     "async": "^3.2.6",
-    "brace-expansion": "1.1.12",
+    "brace-expansion": "^5.0.8",
```
Harshvardhan-handa123: V1 to v5 is a major jump. Do we need any refactoring in the code for this?

package.json
```diff
@@ -200,9 +200,10 @@
     "underscore.string": "^3.3.6",
     "node-forge": "^1.3.2",
     "qs": "^6.14.1",
-    "@openshift-console/dynamic-plugin-sdk-internal/immutable": "^3.8.3",
-    "@openshift-console/dynamic-plugin-sdk/immutable": "^3.8.3",
-    "sass/immutable": "^4.3.7",
+    "@openshift-console/dynamic-plugin-sdk-internal/immutable": "^4.3.9",
+    "@openshift-console/dynamic-plugin-sdk/immutable": "^4.3.9",
```
Harshvardhan-handa123, line 205: Can we upgrade the sdk version instead of adding a specific resolution for immutable?

Harshvardhan-handa123 (changes requested):

package.json
sowmithri2001: This is release-4.16, so we shouldn’t bump @openshift-console/dynamic-plugin-sdk* to a 4.22 SDK. 4.16 needs 1.3.0 / 1.0.0 to stay compatible with OCP 4.16.

Those SDK packages still depend on immutable@3.x, and sass also pulls immutable. The CVE is fixed by resolving all of those to 4.3.9 rather than changing the Console SDK this plugin is built against.

Harshvardhan-handa123: https://github.com/red-hat-storage/odf-console/pull/3071#discussion_r3869331534

package.json
sowmithri2001: Yes, I changed the vulnerable version to the fixed version within V1 itself.

package.json
sowmithri2001: Checked this — we can’t drop the resolution by bumping the SDK on 4.16.

We’re already on dynamic-plugin-sdk@1.3.0 / internal@1.0.0 / webpack@1.1.1. The latest 4.16-compatible core SDK is 1.4.0, and it still depends on immutable: "3.x" (same as 1.3.0). The 1.8.0 / 4.21.0 bumps from e818152 / e4bcae5 / #2788 are for newer Console versions and aren’t compatible here.

immutable is only a transitive dep of those SDK packages, so a resolution is still the way to get a patched version (CVE-2026-29063). I’ll drop the two nested @openshift-console/.../immutable keys though — the existing "immutable": "^4.3.9" resolution already covers them.

Harshvardhan-handa123: Can you remove the lines 
 "@openshift-console/dynamic-plugin-sdk-internal/immutable": "^4.3.9",
"@openshift-console/dynamic-plugin-sdk/immutable": "^4.3.9",
It is still there in 4.16, 4.14.

package.json
```diff
@@ -200,9 +201,10 @@
...
+    "immutable": "^4.3.9",
+    "sass/immutable": "^4.3.9",
```
Harshvardhan-handa123: No need.. we can remove "sass/immutable"

Soundarya189r (approved):

PR #3093, opened 2026-08-27
Added Alert for MCG setup on Day 1

Soundarya189r:
## Description

Issue: On day-1, Object Storage → Buckets shows 404: Not Found until NooBaa is up and the noobaa-admin Secret (S3 endpoint credentials) exists. That is normal setup, usually 5–10 minutes after the StorageCluster CR is created, not a real failure.

Fix: If the watch for noobaa-admin returns 404, show an in-progress alert instead of the 404 page. When the Secret is created, the watch succeeds and Buckets loads. Other errors (403, 500, and so on) still use StatusBox.

---

## Change Type

- [ ] Feature
- [✅ ] Bug Fix
- [ ] Improvement
- [ ] Refactor
- [ ] Tests

---

## Component / Area Impacted

- [ ✅] ODF
- [ ] FDF
- [ ] Client
- [ ] MCO
- [ ] CNSA (Remote Mount)
- [ ] E2E Test

---

### Screenshots

Screenshot 2026-08-31 at 13 40 49

---

## Testing

- [ ] Unit Tests
- [ ] E2E Tests

---

## Summary by CodeRabbit

* **New Features**
  * Added localized messaging while the Multicloud Object Gateway initializes.
  * Added an inline progress alert explaining that object storage typically becomes available within 5–10 minutes after StorageCluster creation.
  * Applied initialization guidance across object storage, IAM, S3 browser, and S3 vectors views.

* **Bug Fixes**
  * Replaced generic error states with a clear readiness message when object storage is still initializing.


SanjalKatiyar:
/retitle Added Alert for MCG setup on Day 1

packages/odf/components/s3-browser/s3-context.tsx
```diff
@@ -113,5 +117,24 @@ export const S3Provider: React.FC<S3ProviderProps> = ({
...
+    return (
+      <Alert
+        className="pf-v6-u-m-md"
+        variant="custom"
+        isInline
+        customIcon={<Spinner size="md" />}
+        title={t('Buckets are not available yet')}
+      >
+        {t(
+          'Object Storage setup is still in progress. The Buckets view will be available after MCG is deployed and the S3 endpoint configuration is created. This usually takes 5-10 minutes after StorageCluster creation.'
```
SanjalKatiyar:
```suggestion
          'Object storage setup is still in progress. The Buckets view will be available after MCG is deployed and the S3 endpoint configuration is created. This usually takes 5-10 minutes after StorageCluster creation.'
```

SanjalKatiyar: If we are capitalising "Buckets", we might want to display in bold to reflect the fact that we are talking about a tab.
Plz confirm with UX if this text is suggested by them.

SanjalKatiyar: I am not against this message, but what if Secret is never created due to any bug ??

It's an edge case so I am not too much worried about this, still I would prefer (if possible) that we can show some generic message here rather than always saying "still in progress".

SanjalKatiyar (changes requested):
Few things:
1. Checking if this design has already been discussed/approved with/by the UX team ?? We want to make sure we have their input before adding something like this.
2. What about Vectors tab, RGW tab or IAM tab ?? User can navigate to those pages as well.

SanjalKatiyar:
> 2. What about Vectors tab, RGW tab or IAM tab ?? User can navigate to those pages as well.

NamespaceStore, OBC etc tabs are hidden until MCG is deployed, whereas IAM and Buckets tabs are always visible.
Buckets tab is dependent on both NooBaa and RGW.
We have Client and Non-admin user flows as well and need to check there is no impact there.

This needs quite some UX inputs on how to address this consistently.

PR #3095, opened 2026-08-27
Fix Protected Applications name search for partial matches

TimothyAsirJeyasing:
Refs [DFBUGS-9554](https://redhat.atlassian.net/browse/DFBUGS-9554)

Before this change, the Protected Applications list used the default useListPageFilter name filter, which matches Kubernetes metadata.name (the ProtectedApplicationView / DRPC resource name). The Name column, however, renders getApplicationName(pav) from
status.applicationInfo.applicationRef.name. When those strings differ, searching contiguous parts of the displayed application name failed even though shorter prefixes sometimes still matched metadata.name.

Example: for displayed name appset-workload-eea5e47a9-rbd, queries like "appset" could appear to work while "appset-wo" or "appset-workload" returned no results.

The fix overrides the console name filter (type: 'name'), matching the existing CORS/Lifecycle list-page pattern, and filters only on the displayed application name with case-insensitive substring matching (includes). Policy, cluster, namespace, and DR status are intentionally not searched.

## Description

---

## Change Type

- [ ] Feature
- [x] Bug Fix
- [ ] Improvement
- [ ] Refactor
- [ ] Tests

---

## Component / Area Impacted

- [x] ODF
- [ ] FDF
- [ ] Client
- [ ] MCO
- [ ] CNSA (Remote Mount)
- [ ] E2E Test

---

### Recordings / Demo Videos

https://github.com/user-attachments/assets/824c02f6-94c4-4229-9520-389bf5b86ead

---

## Testing

- [ ] Unit Tests
- [ ] E2E Tests

---

## Summary by CodeRabbit

- **Enhancements**
  - Improved protected application search to support case-insensitive fuzzy matching.
  - Updated filtering behavior for protected application lists.
  - Hid the label filter from the list interface.

### DIFF
--- a/packages/mco/components/protected-applications/list-page.tsx
+++ b/packages/mco/components/protected-applications/list-page.tsx
@@ -18,6 +18,7 @@ import {
   K8sResourceCommon,
   useK8sWatchResource,
   useListPageFilter,
+  RowFilter,
 } from '@openshift-console/dynamic-plugin-sdk';
 import { LaunchModal } from '@openshift-console/dynamic-plugin-sdk/lib/app/modal-support/ModalProvider';
 import { TFunction } from 'i18next';
@@ -399,6 +400,24 @@ const ProtectedAppsTableRow: React.FC<
   );
 };
 
+// Dedicated filter key so Name search is not tied to the default
+// type: 'name' filter handling, which can leave the filter stuck after clear.
+const PAV_NAME_FILTER = 'protected-application';
+
+const nameFilter: RowFilter<ProtectedApplicationViewKind>[] = [
+  {
+    type: PAV_NAME_FILTER,
+    filterGroupName: '',
+    reducer: () => undefined,
+    items: [],
+    filter: (filterValue, pav) =>
+      fuzzyCaseInsensitive(
+        filterValue.selected?.[0],
+        getApplicationName(pav) || ''
+      ),
+  },
+];
+
 export const ProtectedApplicationsListPage: React.FC = () => {
   const { t } = useCustomTranslation();
   const launcher: LaunchModal = useModalWrapper();
@@ -428,7 +447,10 @@ export const ProtectedApplicationsListPage: React.FC = () => {
   const isAllLoadedWOAnyError =
     pavsLoaded && drpcsLoaded && !pavsError && !drpcsError;
 
-  const [data, filteredData, onFilterChange] = useListPageFilter(pavs || []);
+  const [data, filteredData, onFilterChange] = useListPageFilter(
+    pavs || [],
+    nameFilter
+  );
 
   const [pagePavs, setPagePavs] = React.useState<
     ProtectedApplicationViewKind[]
@@ -547,6 +569,8 @@ export const ProtectedApplicationsListPage: React.FC = () => {
           data: data,
           loaded: drpcsLoaded && pavsLoaded,
           onFilterChange: onFilterChange,
+          nameFilter: PAV_NAME_FILTER,
+          hideLabelFilter: true,
         }}
         composableTableProps={{
           columns: getHeaderColumns(t),


packages/mco/components/protected-applications/utils.tsx
```diff
@@ -343,3 +344,13 @@ export const getEnrollDropdownItems = (
...
 ];
+
+export const pavMatchesSearch = (
+  searchText: string,
+  pav: ProtectedApplicationViewKind
+): boolean => {
+  if (!searchText) return true;
+
+  const appName = getApplicationName(pav) || '';
+  return appName.toLowerCase().includes(searchText.toLowerCase());
```
SanjalKatiyar: we already should be having util for filter...

packages/mco/components/protected-applications/list-page.tsx
```diff
@@ -547,6 +565,7 @@ export const ProtectedApplicationsListPage: React.FC = () => {
           data: data,
           loaded: drpcsLoaded && pavsLoaded,
           onFilterChange: onFilterChange,
+          nameFilter: PAV_NAME_FILTER,
```
SanjalKatiyar, line 572:
```suggestion
```

packages/mco/components/protected-applications/list-page.tsx
```diff
@@ -399,6 +401,19 @@ const ProtectedAppsTableRow: React.FC<
...
+const PAV_NAME_FILTER = 'protected-application';
+
+const nameFilterOverride: RowFilter<ProtectedApplicationViewKind>[] = [
+  {
+    type: PAV_NAME_FILTER,
+    filterGroupName: '',
+    reducer: () => undefined,
+    items: [],
+    filter: (filterValue, pav) =>
+      pavMatchesSearch(filterValue.selected?.[0] || '', pav),
```
SanjalKatiyar: how exactly is it an "override" when u are providing an unique identity "PAV_NAME_FILTER" to the filter ??

what exactly is it supposed to override ??

packages/mco/components/protected-applications/list-page.tsx
SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/3095/changes#r3958835896

packages/mco/components/protected-applications/list-page.tsx
TimothyAsirJeyasing: Agreed, removed pavMatchesSearch.
and also, you are right that this is not really an override of type: 'name', so I renamed it to nameFilter.

I first tried using type: 'name', similar to CORS/Lifecycle, and removed nameFilter: PAV_NAME_FILTER. However, on this page, clearing the Name search does not work correctly with the default name filter handling. When the search text is removed using Backspace, the search field becomes empty, but the table continues to show the previously filtered rows.

So I kept a separate filter key (PAV_NAME_FILTER = 'protected-application') for this page. This keeps the Name search tied to our getApplicationName() filtering with fuzzyCaseInsensitive, while avoiding the clear-search issue with the default name filter.

packages/mco/components/protected-applications/list-page.tsx
```diff
@@ -399,6 +400,24 @@ const ProtectedAppsTableRow: React.FC<
   );
 };
 
+// Dedicated filter key so Name search is not tied to the default
+// type: 'name' filter handling, which can leave the filter stuck after clear.
+const PAV_NAME_FILTER = 'protected-application';
+
+const nameFilter: RowFilter<ProtectedApplicationViewKind>[] = [
```
SanjalKatiyar: https://github.com/red-hat-storage/odf-console/pull/3095#discussion_r3958835896

SanjalKatiyar: as I mentioned previously, I doubt u are overriding existing filter, instead adding a new one ?

SanjalKatiyar: what is the intention here ? to change the filtering behaviour of existing "Name" filter added by console, right ?

TimothyAsirJeyasing: Thanks for the review. I have updated to override the Name filter with type: 'name' (cors/lifecycle pattern) and removed the custom filter key / nameFilter prop.

packages/mco/components/protected-applications/list-page.tsx
```diff
@@ -399,6 +400,24 @@ const ProtectedAppsTableRow: React.FC<
...
+  {
+    type: 'name',
+    filterGroupName: '',
+    reducer: () => undefined,
+    items: [],
+    filter: (filterValue, pav) => {
+      const searchText = filterValue.selected?.[0];
+      // When filter is cleared, show all items
+      if (!searchText) return true;
+      return fuzzyCaseInsensitive(searchText, getApplicationName(pav) || '');
```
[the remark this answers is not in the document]
TimothyAsirJeyasing: @coderabbit I'm keeping `fuzzyCaseInsensitive` to maintain consistency with CORS/Lifecycle list pages (both use the same fuzzy matching approach). This follows the established pattern in the codebase.

PR #3105, opened 2026-09-08
fix batch failover/relocate sending empty failoverCluster

vbnrh:
## Summary
- **Root cause:** Batch modal resolved target cluster from DRPC spec fields (`spec.preferredCluster`, `spec.failoverCluster`), which are empty on fresh deployments that have never been failed over. This sent `failoverCluster: ""` in the patch, silently rejected by ramen.
- **Fix:** Resolve cluster pair from PAV status (`drInfo.drClusters` + `primaryCluster`), populated by MCO controller from DRPolicy regardless of prior DR operations — matching how the single-app modals resolve clusters.
- **Pre-validation:** Added PeerReady/Available condition checks from PAV status before patching. Ineligible apps are skipped with clear error messages instead of firing blind patches.
- **UX fixes:** Progress bar now counts only eligible apps (not ineligible), selection is preserved on total failure, and retry no longer includes DRPCs whose PAV has disappeared.

## Test plan
- [x] All 18 existing tests pass
- [ ] Fresh deployment (never failed over): batch failover sets correct `failoverCluster` value
- [ ] Previously failed-over app: batch relocate sets correct `preferredCluster` value
- [ ] Mixed selection (eligible + ineligible): ineligible apps shown in warning alert, skipped during patch
- [ ] Total failure scenario: user selection preserved (not cleared)
- [ ] Retry after partial failure: only retries items with valid PAV data

## Summary by CodeRabbit

- **New Features**
  - Batch failover and relocate actions now validate application and cluster readiness before starting.
  - Ineligible applications are skipped and reported with specific failure details.
  - Progress and completion counts reflect only eligible applications.
  - Inline warnings identify skipped applications, including missing cluster information and peer-readiness issues.
  - Completion is withheld when no applications are eligible.
  - Retrying failed applications preserves failure details when none are eligible.
  - Added localized messages explaining skipped applications and their reported failures.


vbnrh:
https://github.com/user-attachments/assets/3c2f91ea-2e08-49d4-a5a9-3ce46186ef64

vbnrh:
Batch relocate 

https://github.com/user-attachments/assets/eee6315c-ef88-475f-8b80-0d8fe5dac290

packages/mco/components/modals/app-failover-relocate/batch-failover-relocate-modal.tsx
```diff
@@ -72,31 +86,69 @@ const buildDRPCPatch = (drpc: DRPlacementControlKind, action: DRActionType) => {
...
+    eligibleTotal > 0 ? Math.round((completedCount / eligibleTotal) * 100) : 0;
+
+  const ineligibleCount = React.useMemo(() => {
+    if (!selectedAction) return 0;
+    return selectedDRPCs.filter((drpc) => {
+      const key = `${getNamespace(drpc)}/${getName(drpc)}`;
+      const clusterInfo = clusterInfoMap.get(key);
+      return !clusterInfo || !isDRActionReady(clusterInfo, selectedAction);
+    }).length;
+  }, [selectedAction, selectedDRPCs, clusterInfoMap]);
```
SanjalKatiyar, line 106: why can't we just `ineligibleCount = totalCount-eligibleTotal` ??

packages/mco/components/protected-applications/list-page.tsx
```diff
@@ -61,13 +63,34 @@ import {
...
+  conditions: K8sResourceCondition[],
+  type: string
+): boolean =>
+  !!conditions?.some((c) => c?.type === type && c?.status === 'True');
+
+const buildClusterInfo = (
+  pav: ProtectedApplicationViewKind
+): DRPCClusterInfo => {
+  const primaryCluster = pav.status?.drInfo?.primaryCluster || '';
+  const drClusters = pav.status?.drInfo?.drClusters || [];
```
SanjalKatiyar: I vaguely remember,  we had this bug earlier where we used to read primary/secondary cluster from DRPC, but later shifted to PAV. We created some util which first checks PAV and then DRPC as default (if PAV is missing info somehow)...

I could be wrong here, but can u plz check ?? If we already have a util better to use that...

SanjalKatiyar:
> * **Root cause:** Batch modal resolved target cluster from DRPC spec fields (`spec.preferredCluster`, `spec.failoverCluster`), which are empty on fresh deployments that have never been failed over. This sent `failoverCluster: ""` in the patch, silently rejected by ramen.

We don't use DRPC for single failover/relocate action ?? Only did it for batch ??

vbnrh:
> > * **Root cause:** Batch modal resolved target cluster from DRPC spec fields (`spec.preferredCluster`, `spec.failoverCluster`), which are empty on fresh deployments that have never been failed over. This sent `failoverCluster: ""` in the patch, silently rejected by ramen.
> 
> We don't use DRPC for single failover/relocate action ?? Only did it for batch ??

Right, single-app modal fetches DRClusters, ManagedClusters etc. and resolves clusters via findDeploymentClusters() + findCluster(). It never reads from spec.preferredCluster/spec.failoverCluster. Only the batch modal was using those DRPC spec fields, which are empty on fresh deployments. Fixed it to read from PAV drInfo.drClusters + primaryCluster instead, same source of truth.

packages/mco/components/protected-applications/list-page.tsx
vbnrh: Checked the codebase. getPrimaryCluster(pav) in pav.ts only returns the primary cluster string, and getPrimaryClusterName(drpc) reads from DRPC not PAV. Neither covers what we need here (primary + target + PeerReady/Available conditions). Moved buildClusterInfo and DRPCClusterInfo to pav.ts as a shared util alongside getPrimaryCluster.

packages/mco/components/modals/app-failover-relocate/batch-failover-relocate-modal.tsx
vbnrh: Good catch, simplified. Removed eligibleTotal state and now derive eligibleCount = totalCount - ineligibleCount directly. Did it this way round (not ineligibleCount = totalCount - eligibleTotal) because ineligibleCount is needed as soon as the user picks an action to show the warning alert, while eligibleTotal was only set after clicking Initiate.

SanjalKatiyar (approved):

